PR #1598, opened 2013-01-11
[Alt-Tab] Show windows from all workspaces

autarkper:
This makes Alt-Tab show windows from all workspaces, so you can use it to switch to any open window, regardless of which workspace it is on.

This behavior should probably be optional, but I have not added a settings key for this yet.

Current status: It works fine, as far as I can see.


halfer:
This is now an option in Cinnamon core - I presume it was not before.

In Settings, go to the Windows section, click on the Alt-Tab tab, then tick "Show windows from all workspaces".

Issue #1692, opened 2013-02-08
Workspace Names


Cobinja:
If the schema key "org.cinnamon.workspace-names-overrides" is changed from outside cinnamon (e.g. through dconf-editor), the change is not noticed in cinnamon.

PR #1993, opened 2013-05-09
Make cinnamon aware of workspace name gsetting changes

Cobinja:
Addresses #1692


BeatLink:
Allowing cinnamon to recognize gsettings workspace name changes, in concert with the num-workspace schema would allow external applications to be able to manage workspaces. I myself am seeking to build an app to do just that, however, i am hindered by cinnamon's inability to update the names until it is restarted. See #8508

PR #2140, opened 2013-06-06
[Design] Easier connection management (recommit)

Lusito:
@autarkper recently closed #1737, which added connector.js, a concept to ease the connection management.

I had a similar idea for a while now, especially for applets, since currently most don't even bother to disconnect from their signals.
So, my approach has 3 different ways you can choose to use:

##### Connector.createConnection(object, 'signal', callback, ....);
- You can essentially replace any object.connect('signal', callback, ...) with this one.
- Returns a new Connection object, which has methods disconnect, forget and getTarget
- This can be used if you only have a single connection to manage.

##### let connector = new Connector.Connector();
- This is a connection manager which keeps a list of connections and supplies these methods:
- createConnection(..): same parameters as the manual function above
 - Also returns the connection object, so you can choose to forget or disconnect a single connection separate from all others.
 - The connection will be stored internally
- destroyConnections()
 - Calls disconnect on all connections and then clears the connections array.
- addMaster(master, signal): default value for signal is 'destroy'
 - When a master has been added, the manager will watch that object for the specified destruction signal.
 - When master emits signal, destroyConnections() will be called as well as removeMasters()
- removeMasters()
 - Calls disconnect on all master connections and then clears the masters array.

##### Connector.addConnectorMethods(MyClass.prototype);
- This adds createConnection, destroyConnections, addConnectionMaster and removeConnectionMasters to the class prototype.
- This uses a Connector instance internally, so you don't have to create a connector manually and can skip the connector variable.

### More work required
This should be used widely across cinnamon, especially on the applets. But I'm not gonna start working on that until I know this gets accepted.

### DIFF
--- a/js/Makefile.am
+++ b/js/Makefile.am
@@ -3,6 +3,7 @@ jsdir = $(pkgdatadir)/js
 
 nobase_dist_js_DATA = \
 	misc/config.js		\
+	misc/config.js          \
 	misc/docInfo.js		\
 	misc/fileUtils.js	\
 	misc/format.js		\
--- a/js/misc/connector.js
+++ b/js/misc/connector.js
@@ -1,0 +1,149 @@
+// -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
+
+const Lang = imports.lang;
+
+function _createConnection() {
+    if (!('_connectionManager' in this))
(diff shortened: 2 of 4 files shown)


clefebvre:
Closing due to lack of activity.

PR #3037, opened 2014-04-12
[2.4] Modified to work better with multiple batteries e.g. on Lenovo T440s

michaeldjcox:
Modified to:

a) show the total time remaining in the panel rather than that of one of possible several batteries
b) where individual batteries in the menu are listed show them with time or percent or both as selected from the menu
c) fix the issue where you select time remaining and the panel does not update and remaining at percent (this one is a hack that seems to work)

All the above designed to improve the use of the power applet on machines with multiple batteries like the Lenovo thinkpad T440s

NOTE: That I made the changes on a earlier version supplied with Linux Mint 16 and tested there. This commit is the same change applied to the current master but I am not really set up to test with the latest and also to test with a variety of machines. I wanted to make the changes available though and they can be picked up and evaluated and made available to all in future releases of Cinnamon

Some background and images on my blog:  http://thinkpadmint.blogspot.co.uk/2014/04/fixing-cinnamon-power-applet-for.html


Kabouik:
An update for Cinnamon 17 would be awesome!

letronje:
+1

michaeldjcox:
Closed this pull request - going to have another crack at this starting from the latest code. Back with another pull request later....

PR #3553, opened 2014-09-30
[WIP] Added flickr support for slideshow

KOLEGA007:
Added flickr icon and changed the makefile.am

### DIFF
--- a/data/icons/Makefile.am
+++ b/data/icons/Makefile.am
@@ -55,6 +55,7 @@ public_icons = \
 	hicolor_emblems_scalable_cs-xlet-running.svg \
 	hicolor_emblems_scalable_cs-xlet-system.svg \
 	hicolor_emblems_scalable_cs-xlet-update.svg \
+	hicolor_places_scalable_flickr.svg \
 	$(NULL)
 
 EXTRA_DIST = \
--- a/data/icons/hicolor_places_scalable_flickr.svg
+++ b/data/icons/hicolor_places_scalable_flickr.svg
@@ -1,0 +1,320 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   height="96"
+   width="96"
+   version="1.1"
+   id="svg2"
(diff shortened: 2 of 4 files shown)


clefebvre:
A couple of small issues easy to fix:

Popen -> subprocess.Popen
time.sleep() -> missing import time
chmod a+rx files/usr/lib/cinnamon-settings/bin/flickrapi.py (else we get a permission denied)

A couple of other issues:

That API key is a real show stopper... we could ask the user to provide a key in the preferences and expose flickr support only then, but that's tedious. I don't expect users to be able to go and get a key themselves... they might not even understand what it means.

As bgs are downloaded/converted, the BG view in the settings doesn't get updated.

KOLEGA007:
@GLolol Got your point, made it like this because I want to make it independent (async) from cs_backgrounds.py...

If you have any advices ... Teach me, thanks :)

KOLEGA007:
@clefebvre Clem, could somebody of team (probably you) contant  Flickr and offer something like "collaboration"? I think, that huge number of Linux User could convince them, we could point this in backgrounds ... :)

dalcde:
I'm not sure how likely that would be. Since Cinnamon is open-source and the key we use will be publicly available, if we have something like an unlimited key, anyone would be able to take it and exploit it to do whatever they want.

KOLEGA007:
@dalcde some binary from Flickr aka License server? ... (just idea, do not hate me, I am liberal)

Or it does not have to be a binary ...

KOLEGA007:
Some OAuth site? On first run with Captcha? Crackers can although take a key from flickr site ... I think, that main problem is here, that we would not convince them ...

KOLEGA007:
UPDATE: Trying to contant flickr ... Found some form for contacting them... Got small solution, like users should've log into flickr (f.e. using OAuth) and after that, Flickr could provide us unique api key for each user, which we will save localy ... Anyway I think it will fail, from the Flickr side ...

KOLEGA007:
@clefebvre Clem, please read this topic: https://www.flickr.com/groups/51035612836@N01/discuss/72157652534378861/72157650245110204/

In another topic I've read, that mainly IP's, not keys are banned. 

Posibly solution: We will release it with our (my) api key and we will give users notice, that they should get their own (with link to flickr api garden).

Anyway still there is the problem with no updating of downloaded pictures :/

lestcape:
@KOLEGA007 sorry for my stupid suggestion, because i really don't understand all what you are doing. Just a question. As this is an online Api, will not help redirect the request to an intermediate server, who can include the key and redirect the request to the real server with the correct data. Possible the server could automatically request a new key when is needed. In that way, the user not need to know about the internal implementation and less about keys and complex things.

KOLEGA007:
I mentioned that the second way, but, we will rather force user to put his own API key into CS

lestcape:
Inded i know that have speed on settings is important, but there are a possibility to create an spices session or something to add settings modules? This could be a user settings module(a module that could be loaded from the user home folder). 

I have two modules that could be a user settings modules. One of then allow to create more user settings modules using a, desktop file or a custtom command.

clefebvre:
Hi,

I'm going to close this PR.

It's a fantastic feature and it works quite well but it relies on an API key and so we couldn't/shouldn't ship it by default. It's probably better as a cinnamon spice with the user entering his own flickr key.

Issue #3603, opened 2014-10-17
Feature request for Cinnamon settings: Setting type for Tween functions


pixunil:
In the effects section of cinnamon settings, and by some xlet devs, there is a nasty long combobox used for Tween functions (e.g. `easeInQuad`). These names aren't very informative, many users can't understand them. My idea: a dialog to choose a function, a table with animated time - value graphs like in the image:
![cinnamon](https://cloud.githubusercontent.com/assets/7093655/4682012/5ed8a1aa-5618-11e4-8741-fb3dff5a8979.png)

Issue #1682, opened 2013-02-08
More and Better Cinnamon Effects


Brahimsalem:
I suggest more compiz-like effects for Cinnamon; like burn, glide and beam effects. The fire in the burn effect can be replaced with water, stars, flowers or magic potion for instance. Please bring back old-time memories to Cinnamon.

PR #3662, opened 2014-11-07
several window effects changes

pixunil:
Depends on linuxmint/cjs#18

## setting widgets
new super class `BaseChooserButton`
### EffectChooserButton
- For cs_effects
- displays static previews (can be improved)
- starts an animation on hover
shot

### TweenChooserButton
#3603
- For cs_effects and Xlet settings
- displays static graphs
- starts an animation (a little arc moving up)
shot

## window effects
#1682

### New effects
Effect | Mapping | Closing and Minimizing
-----|-------|------
Blend | Window fades in and scales from 1.5 down to 1 | Window fades out and scales to 1.5
Move | starts at the current mouse pointer position and then scales and moves to the destination. | scales down to 0 and moves to the current mouse pointer position
Fly up | starts at the bottom edge and moves to the destination | moves to the top edge
Fly down | starts at the top edge and moves to the destination | moves to the bottom edge

All effects are now in a new file `windowEffects.js`

## other changes
- scale tweens can't go under 0 anymore (like in the elastic and back transitions)
- opacity tweens can't go under 0 or above 255 anymore

### DIFF
--- a/data/org.cinnamon.gschema.xml.in
+++ b/data/org.cinnamon.gschema.xml.in
@@ -156,7 +156,15 @@
         Whether to enable desktop effects on dialog boxes.
       </_description>
     </key>
-    
+
+    <key type="s" name="desktop-effects-style">
+      <default>"cinnamon"</default>
+      <_summary>The style of desktop effects</_summary>
+      <_description>
+        An overall style used for desktop effects
+      </_description>
+    </key>
+
     <key type="s" name="desktop-effects-close-effect">
       <default>"none"</default>
       <_summary>Effect used when closing windows</_summary>
--- a/files/usr/lib/cinnamon-settings/bin/SettingsWidgets.py
+++ b/files/usr/lib/cinnamon-settings/bin/SettingsWidgets.py
@@ -23,6 +23,7 @@ try:
     import tempfile
     import math
     import subprocess
+    import tweenEquations
 
 except Exception, detail:
     print detail
@@ -90,15 +91,9 @@ class EditableEntry (Gtk.Notebook):
     def get_text(self):
         return self.entry.get_text()
 
-class PictureChooserButton (Gtk.Button):
-
-    def __init__ (self, num_cols=4, button_picture_size=None, menu_pictures_size=None, has_button_label=False):        
-        super(PictureChooserButton, self).__init__()
-        self.num_cols = num_cols
-        self.button_picture_size = button_picture_size
-        self.menu_pictures_size = menu_pictures_size
-        self.row = 0
-        self.col = 0
+class BaseChooserButton(Gtk.Button):
+    def __init__ (self, has_button_label=False):
+        super(BaseChooserButton, self).__init__()
         self.menu = Gtk.Menu()
         self.button_box = Gtk.Box(orientation=Gtk.Orientation.VERTICAL, spacing=2)
         self.button_image = Gtk.Image()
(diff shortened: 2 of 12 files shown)


mtwebster:
Can you try to rebase?  I'll try to take a good look this weekend.

mtwebster:
Did you forget to git add a file?

```
Loading Effects module
  File "/usr/lib/cinnamon-settings/cinnamon-settings.py", line 490, in button_press
    self.side_view_nav(widget, None, category)
  File "/usr/lib/cinnamon-settings/cinnamon-settings.py", line 108, in side_view_nav
    self.go_to_sidepage(cat, filtered_path)
  File "/usr/lib/cinnamon-settings/cinnamon-settings.py", line 117, in go_to_sidepage
    sidePage.build()
  File "/usr/lib/cinnamon-settings/bin/SettingsWidgets.py", line 275, in build
  File "/usr/lib/cinnamon-settings/modules/cs_effects.py", line 99, in on_module_selected
    self.make_effect_group(_("Mapping windows:"), "map", effects, 1)
  File "/usr/lib/cinnamon-settings/modules/cs_effects.py", line 134, in make_effect_group
    tween = GSettingsTweenChooserButton(root, template % (key, "transition"), path)
  File "/usr/lib/cinnamon-settings/bin/SettingsWidgets.py", line 1098, in __init__
    super(TweenChooserButton, self).__init__()
NameError: global name 'TweenChooserButton' is not defined
```

mtwebster:
Nevermind, easy fix to test.. 

At first I assumed the "Cinnamon" style was the current (pre-this-pr) defaults, like "Legacy effects" or something, but I see you don't actually have a gsettings key for the effect style, so you're using "Cinnamon" when the effects have been customized.

I'd rather the current style be stored in a setting - so I can customize, close settings, open them again, and it still says 'Flydown, up' or 'Fade' or whatever the user explicitly set it to.  Right now, once I close cinnamon-settings, then re-open, it will say "Cinnamon", which is confusing.  We can just remove the "Cinnamon" option.

A couple more things:

- Is one of the effect styles equivalent to the old default settings for effects?  I'd like there to be one (maybe called "Default") and make that the default, at least to start out - I'd rather users not face a jolt after upgrading.  Maybe we will tweak what those defaults are later, to take advantage of some of the new effects, but clem would want the final decision on it.

- Many of the presets feel slow - I think 400ms is way too long for anything.

JosephMcc:
That's what I had done in the tweaked version I made. I created an "effects-style" key and set "cinnamon" as the default. It used a set of animations that are very close to the current ones with a couple of tweaks.

I also agree with the speeding some things up but I think some slow overally dramatic ones are alright as part of some of the sets. Some people like that sort of thing. I think in the default one "cinnmaon" I didn't set anything over 200.

pixunil:
@JosephMcc I removed this as I didn't understand it, but I think this is okay.

@mtwebster Yes, I've forgotten to commit one line change…

The storage method could be improved and (as always) the effect sets.

I reduced the times of the (un-) maximize / tile effects to 100ms, except in the Default set (which is like the gsettings schema now)
the default gsettings values could be set to e.g. the cinnamon set

mtwebster:
Ok I think we're ready to squash :)

- Go ahead and enable those remaining effects sets (fancy scale, move) - why not let the user see them.
- When you disable effects entirely, make the "Customize Effects" control go insensitive along with everything else.

I'm wondering about different/better names for some of these (Should "Cinnamon" be called "Traditional"?) But these are things that can be tweaked later.  Let's get it in there.  I'll merge once you squash.

Issue #3655, opened 2014-11-05
Can't find how to change "Sunday" as first day of the week.


DrOptix:
The title explains it all. I can't find a way to change the damn ugly Sunday as the first day of the week. I want a normal, Monday starting calendar, even if I use en-us locale for other aspects of the desktop.

I'm sure there must be some arcane way to do it by editing some config file, but the calendar plugin should be able to set this from the GUI.

Thank you!

dalcde:
Fixed in git master

PR #3665, opened 2014-11-10
Configure first day of week in calendar

AlbertJP:
Default setting is still the locale's default.
Sunday and Monday can be set in Cinnamon Settings, other days can be easily added in the python code if necessary.

#3655 
Requires linuxmint/cinnamon-desktop#28

### DIFF
--- a/src/cinnamon-util.c
+++ b/src/cinnamon-util.c
@@ -10,6 +10,9 @@
 #include <langinfo.h>
 #endif
 
+#define DESKTOP_SCHEMA "org.cinnamon.desktop.interface"
+#define FIRST_WEEKDAY_KEY "first-day-of-week"
+
 #include <libxml/parser.h>
 #include <libxml/tree.h>
 #include <libxml/xmlmemory.h>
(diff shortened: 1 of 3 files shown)


dalcde:
For some reason, this hangs Cinnamon when the first-day-of-week key is set to, say, 8. Perhaps you can let Cinnamon interpret all values not in 0 - 6 as locale default, instead of only 7.

AlbertJP:
I don't expect end-users to change gsettings keys manually, but as this would be easily fixed I've put it on my to-do list anyway.

src/cinnamon-util.c
```diff
@@ -540,7 +543,12 @@ cinnamon_util_format_date (const char *format,
 int
 cinnamon_util_get_week_start ()
 {
-  int week_start;
+  /* Try to get first weekday from gsettings */
+  /* If 7, continue to get the locale's first weekday */
+  GSettings *settings = g_settings_new (DESKTOP_SCHEMA);
+  int week_start = g_settings_get_int (settings, FIRST_WEEKDAY_KEY);
+  if (week_start != 7) return week_start;
```
dalcde: Changing this to

    if (0 <= week_start && week_start <= 7) return week_start;

should solve the problem

AlbertJP:
I have just changed that, however I used < 7 instead of <= 7, as 7 is not a value that we want to return.

dalcde:
The changes in Cinnamon settings probably broke the pr

PR #3733, opened 2014-12-05
List icon themes with cursor theme

City-busz:
Some icon theme (e.g. the new Adwaita icon theme) contains both icon and cursor theme, so don't filter out.


clefebvre:
I agree with the issue but the fix is not good enough, it results in showing cursor themes which don't provide icons..

City-busz:
No, because they don't have a "folder" icon, therefore not visible in the chooser.

Issue #130, opened 2012-01-11
Taskbar on second monitor


tbekolay:
Really loving Mint 12 with Cinnamon. One feature that would be great for the future would be to allow a second taskbar on a second monitor, listing the windows currently showing on that monitor, as was possible in Gnome 2.

mdsitton:
So just got back, and im playing with it a bit more. Another issue that im seeing quite a lot are ghost panels. Panels with nothing on them that with varying levels on interactivity. Some of them i can right click on and get that context menu. Others don't do anything. On the ones that do give me a context menu being in edit mode, and selecting remove panel does nothing.

dalcde:
You're getting weird issues that neither I nor JosephM has experienced. Just to make things clear, this is how add/move panels work:

When you click the add/move panels button, dummy panels will appear in possible locations of the new panel (they can be seen if and only if you are using the default theme). You click on one of theme to select where you want the panel to be. If you change your mind, you can press Esc to cancel the action.

mdsitton:
That would explain why It was hard to figure out how to move panels. I wasn't using the default theme.

Still getting those ghost panels though. However i have figured out when they start happening.

Seems that moving any blank panel works perfectly. However when moving panels with the default applet configuration on it things, they create ghost panels.

Haven't tested with alternative panel configurations to see if there is any specific applet that is causing it though. 

Also here is a bit of clarification with what I'm calling ghost panels, and how they behave.

They basically seem to reference the panel that left it behind. So if i select move on a ghost panel it moves the original panel that created it.

Selecting remove panel on a ghost panel removes neither the ghost panel nor the original panel. However it does break the original panel. When the original panel gets broken like that it cannot be moved or  removed, and copying the applets on it seems to crash everything.

Also another issue i came across when trying to track this down, removing a panel doesn't remove the applets from that panel. So whenever you add a new panel it recycles previous panel configurations. Also because of that applets will show as still on a panel, but they have an error status(in the applet settings menu).

All my previous posts are basically invalid :P

dalcde:
Do you get errors in .xsession-errors/lg?

The retaining of applets is intentional (and was the default behaviour with the old hardcoded 2-panels iirc). It would be annoying if you accidentally removed a panel and have to add all the applets back. However, I also see cases where the user would expect the applets to disappear. It wouldn't be difficult to change this behaviour, so I guess we can sort all bugs out before going into these design decisions.

mdsitton:
[...]

PR #3746, opened 2014-12-10
Panels on Multiple Monitors

dalcde:
Known issues:
- Possibly breaks collapsible systray (3rd party applet) because Main is no longer starting statusicondispatcher

#130


JosephMcc:
Been using this quite a bit so I thought I would post a couple issues I've run into. Some of these I think I mentioned in IRC but I'll drop them here just so they don't get lost.

Using a box-shadow with the #panel css selector in the theme no longer works.

Sometimes opening the panel settings in cinnamon-settings just gives an empty window. Haven't found a reliable way to reproduce this but it's happened a couple times.

If you open "cinnamon-settings panel" from a terminal then use the panel context menu to play around with the new panel settings a lot of errors are thrown. Maybe related to the window occasionally opening empty?

When changing themes the font size of items on the panel don't properly update. Not sure if this is new in your branch or exists in Master as well.

Workflow: okay not a bug, just a thought. I would rather have all the new panel settings in the context menu available in a new "Panel..." submenu. It doesn't feel like there is any reason to make me use panel edit mode to access them. I can add panels from cinnamon settings without doing so.
Also maybe when using the functions to add or move panels show a popup in much the same way the snapping feature does to give the user a bit more of a clue as to how to use them. Something like "Select the panel location, Esc to cancel". Just something to give a bit of a hint.

dalcde:
Thanks. The theme thing is an old issue reproducable on master

Issue #3817, opened 2015-01-14
[Sound applet] Removing a player does not remove it from the list


gauravjuvekar:
1. `sudo apt-get install rhythmbox`
2. Open sound-applet > Launch player
  - Rhythmbox is shown
3. `sudo apt-get remove rhythmbox`
4. Open sound-applet > Launch player
  - **Rhythmbox is still shown**
  - Clicking it gives a failed to launch notification

pixunil:
https://github.com/pixunil/Cinnamon/commit/e78fe896ce41d115cbefe3413147924514ebb6fb

mtwebster:
fixed as of 05f65cf and 6b7d342

Issue #3750, opened 2014-12-11
[Sound applet] slider position goes backwards when using VLC


aukevo:
When a media player seeks and send updates via MPRIS over DBus, the sound applet can get its displayed position out of sync by upto a second.  Play any media with VLC AND seek 1/2 way through, the sound applet slider position will jump backwards and forwards in time as vlc plays. 

The cause is the sound applet runs its own clock to display playback position progress, when a media player sends a MPRIS Seeked over DBus, sound applet leaves its clock running while updating and may get a timer tick immediatly after the update or up to one second after the update. the timing is unpredictable. 

Solution is to restart the sound applets timer when playback position is updated via MPRIS/DBus so the 1 second tick actually happens 1 second after the update.

Here is a patch against the ubuntu utopic nightlies from thursday the 11th december. The patch stops and restarts the timer when a MPRIS/DBus position update happens.

BTW I am not a .js programmer!!  so check I'm not doing something naughty in the .js world.
my version of the _setPosition function:

    _setPosition: function(value) {
        if (value == null && this._playerStatus != 'Stopped') {
            this._updatePositionSlider(false);
        }
        else {
            this._currentTime = value / 1000000;
            if (this._playerStatus == 'Playing') {
                if (this._timeoutId != 0) {
                    Mainloop.source_remove(this._timeoutId);
                    this._timeoutId = 0;
                }
                this._timeoutId = Mainloop.timeout_add_seconds(1, Lang.bind(this, this._runTimerCallback));
	        }
            this._updateTimer();
        }
    },

pixunil:
I don't really notice this bug on my machine with normal playing speed, but it could be a little issue, as the difference will only be some seconds. The code looks fine to me, I can include it in my PR #3858.

Regarding playing media on faster / slower speed: this is a bigger issue imo, and could be fixed with the [rate property](http://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Property:Rate).

mtwebster:
fixed as of 05f65cf

Issue #3385, opened 2014-07-20
[feature request] sound applet: control more applications


who-me:
By default, the sound applet in Cinnamon allows the user to control the last application that is started. For example, if one has Banshee running and then opens VLC, they will only be able to control VLC through the sound applet. If Banshee is minimized (hidden) then, the only way to get it back is to close VLC and then click the button in the applet to bring up the window.

I think that the sound applet should be extended to support controlling more than one application at a time.

mtwebster:
Implemented as of 05f65cf

Issue #2888, opened 2014-02-13
[sound applet] missing output devices


gregsheremeta:
In the Output device submenu of the sound applet, only two devices show up, but I have 3. When I click the Sound Settings submenu, I can then choose from all 3.

For some reason, get_sinks() must only return the two devices. Maybe it's that my digital and analog outputs are both "Built-in Audio"?

cinnamon.x86_64  2.0.14-4.fc20
Fedora 20
greg@dauntless:~$ yum list installed alsa*
Loaded plugins: langpacks, refresh-packagekit
Installed Packages
alsa-firmware.noarch                                                 1.0.27-2.fc20                                         installed
alsa-lib.x86_64                                                      1.0.27.2-2.fc20                                       installed
alsa-plugins-pulseaudio.x86_64                                       1.0.27-2.fc20                                         installed
alsa-tools-firmware.x86_64                                           1.0.27-3.fc20                                         installed
alsa-utils.x86_64                                                    1.0.27.2-3.fc20                                       installed
greg@dauntless:~$ yum list installed pulseaudio*
Loaded plugins: langpacks, refresh-packagekit
Installed Packages
pulseaudio.x86_64                                                    4.0-9.gitf81e3.fc20                                   installed
pulseaudio-gdm-hooks.x86_64                                          4.0-9.gitf81e3.fc20                                   installed
pulseaudio-libs.x86_64                                               4.0-9.gitf81e3.fc20                                   installed
pulseaudio-libs-glib2.x86_64                                         4.0-9.gitf81e3.fc20                                   installed
pulseaudio-module-bluetooth.x86_64                                   4.0-9.gitf81e3.fc20                                   installed
pulseaudio-module-x11.x86_64                                         4.0-9.gitf81e3.fc20                                   installed
pulseaudio-utils.x86_64                                              4.0-9.gitf81e3.fc20                                   installed

greg@dauntless:~$ aplay -l
**** List of PLAYBACK Hardware Devices ****
card 0: PCH [HDA Intel PCH], device 0: ALC892 Analog [ALC892 Analog]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 0: PCH [HDA Intel PCH], device 1: ALC892 Digital [ALC892 Digital]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 1: NVidia [HDA NVidia], device 3: HDMI 0 [HDMI 0]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
[...]

PR #3858, opened 2015-01-28
Sound applet

pixunil:
Depends on linuxmint/cinnamon-settings-daemon#79

## New features
- application mixer to control the volume of each app #1874
- input device selector #1512
- if multiple player are launched, the player info "_Name_ - playing" is working as a submenu to select other players; no player gets "forgotten" #3385 #3882 
- middle click action can be configured to (un-) mute output or to toggle pause / play for the active player
- new players, got by DBus, are accessible via the player launcher

## Little enhancements
- can slide to 1%
- seek slider respects mouse scroll
- output device list should update whenever it is necessary #464 [lp:1285912](https://bugs.launchpad.net/linuxmint/+bug/1285912) [lp:1326590](https://bugs.launchpad.net/linuxmint/+bug/1326590)
- output device list is like the one in settings #2888 
- player launcher checks if the players still exist when software is uninstalled #3817
- input change is hearable
- respect `Rate` property #3750
- `soundManager.js`: handle gsettings update properly #3941

## Theming
- input and output sliders have got an icon left to them
- "Volume: n%" and "Microphone: n%" now are in a tooltip
- dropped class `sound-volume-menu-item`
- dropped SVGs `player-paused`, `-playing` and `-stopped`

sound-applet


mtwebster:
Some feedback (Testing on Debian Jessie)

- No players appear by default (that's fine) - if I launch a player, it's recognized by the applet, also good.  But if I close the program, it won't let me re-launch it until I restart Cinnamon.
-- Launched Gnome Music, closed - result-> Launch Player now appears, but there is no expansion under it - it's empty.  Restart cinnamon, all is well.
-- Launch rhythmbox, close - result-> Still only showing gnome music, until I restart.

- If I have multiple players running, I switch between one to the other within the applet, I don't get track length info (and possibly other mpris info) - just the current time in the track.  (I could only test with a radio station, not sure what else might be missing)

- Should quit player be next to the player name at the top?  (just asking for discussion, not saying it should be so)

- This seems a bit confused:
image

-- I think it may be the fact that the slider doesn't align with the text above it - lose the icon and center the text?  or align the text with the slider?  Not sure - try some things?  Also, maybe there should be a separator between the main volume and the app volumes?  Seems a little crowded.

- Weird things happen to the main volume if you adjust the player volume - expected?

pixunil:
1. _No players appear by default_ - fixed
2. _I don't get track length info_ - Can't reproduce atm
3. _Should quit player be next to the player name at the top?_ - Nice thought, but then Raise should be also there
4. _This seems a bit confused_ - Theme issue, with Mint-X and other 3rd party the submenus are colored something other. I think I talked a bit with @josephmcc on IRC
5. _Weird things happen_ - Can't reproduce atm

JosephMcc:
3. If you do this you should probably add them with a style class of their own. Otherwise the buttons will probably be much larger than what you want in most cases. Not sure if I like the idea or not.

4. This is only partially a theme issue. I think a separator between the main volume slider and the "Applications..." label may not be a bad idea. Also kind of like the idea of centering the icon/label for the individiual applications in the submenu. Would have to see what it actually looks like.

pixunil:
3) Yep, that's true, but as we aren't sure whether we implement it, we shouldn't discuss the details
4)
> I think a separator between the main volume slider and the "Applications..." label may not be a bad idea.

Personally, I thought this for a little moment in development but discarded it, because I thought that we have "Player | Output | (Input) | Settings" (| = separator) and I would dived the Output section with it. But if you think it has got better readability, I'll write this line of course.

> I think a separator between the main volume slider and the "Applications..." label may not be a bad idea.

A bit unconventional, right? (as everything is left-aligned).
The fact that the icon - label menuitem is not aligned like the icon - slider is true, I'm using a `St.BoxLayout` for the first and a `_allocate` hack (which I would like to dump) which respects the `_spacing` property for the slider.
I think that the easiest (and laziest) way for me to solve this is to wait for #3902.

pixunil:
3) Yep, that's true, but as we aren't sure whether we implement it, we shouldn't discuss the details
4)
> I think a separator between the main volume slider and the "Applications..." label may not be a bad idea.

Personally, I thought this for a little moment in development but discarded it, because I thought that we have "Player | Output | (Input) | Settings" (| = separator) and I would dived the Output section with it. But if you think it has got better readability, I'll write this line of course.

> I think a separator between the main volume slider and the "Applications..." label may not be a bad idea.

A bit unconventional, right? (as everything is left-aligned).
The fact that the icon - label menuitem is not aligned like the icon - slider is true, I'm using a `St.BoxLayout` for the first and a `_allocate` hack (which I would like to dump) which respects the `_spacing` property for the slider.
I think that the easiest (and laziest) way for me to solve this is to wait for #3902.

mtwebster:
Reminder - we can discuss anything we want, even if it's out of the scope of the current PR - maybe we want it *in* the scope :) - not saying we do, but we only know by discussion.

anandrkris:
I remember suggesting this earlier, how about publishing this as a new applet in spices and seek feedback (assuming it works on 2.4)....

pixunil:
@mtwebster, regarding 2), this shouldn't happen as the player section exists and can be modified. It is not added though, but this shouldn't have an effect an this. Do you have some more information for 5?

@anandrkris it works on 2.4, but you don't have got the audio feedback when changing sound, because I modified two Cinnamon js files and cinnamon-settings-daemon, which can't be installed over spices.

mtwebster:
Sorry, got pulled away at work - 

I actually like this: https://trello-attachments.s3.amazonaws.com/5482ec374223fe02944e9598/279x360/b124c1e4f883830583edd14b71753b3e/sound_applet.png because the theme makes things not stand out...

I don't imagine our popup menu api would let you indent the section under "Applications..." ?

mtwebster:
I think we're good here - the volume issue I was talking about, I think it's intentional, where if you raise an application's volume, it increases the main volume.

Can you squash?

JosephMcc:
The changes to the applet broke compatibility with Pithos. No album art, song info, progress or anything is shown. It all worked before. The player can be tested by installing from the repos and just using a free account on Pandora.

pixunil:
Sorry, I couldn't create an account for pandora, so I couldn't try to reproduce.
Do you have got something in your logs?

mtwebster:
I did a little bit of debugging on it, it seems like the program doesn't set the 'rate' property, and everything falls apart from there.. defaulting to rate of 1 or rate of 1000 upon initializing the player interface allows the ui to finish loading, but we get no progress info - you can't seek pandora, but it should be able to show/update the current position.

I can make a dummy account for you

pixunil:
okay, I blamed the desktopentry first, but this sounds more logical…

and according to http://www.pandora.com/restricted:

> We are deeply, deeply sorry to say that due to licensing constraints, we can no longer allow access to Pandora for listeners located outside of the U.S., Australia and New Zealand. We will continue to work diligently to realize the vision of a truly global Pandora, but for the time being we are required to restrict its use. We are very sad to have to do this, but there is no other alternative.

JosephMcc:
@pixunil These are the errors I get in .xsession-errors whenever a song changes with pithos
```
(cinnamon:3348): Cjs-WARNING **: JS ERROR: Exception in callback for signal: PropertiesChanged: Error: value is out of range for Argument 'interval' (type guint32)
timeout_source@resource:///org/gnome/gjs/modules/mainloop.js:60
timeout_add@resource:///org/gnome/gjs/modules/mainloop.js:72
Player.prototype._runTimer@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:690
Player.prototype._setStatus@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:624
@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:444
_emit@resource:///org/gnome/gjs/modules/signals.js:124
_convertToNativeSignal@resource:///org/gnome/gjs/modules/overrides/Gio.js:129
```

PR #3863, opened 2015-01-30
Fix DST related weeknumber calculation errors

netzvieh:
In locales with DST diffDate is missing 1 hour which leads to dayNumber
being one off. This leads to wrong week numbers for the summer time
period, if Jan 1st is a Thursday.

For CET CW 13 was double every few years, e.g. 2009 and 2015, and
CW 43 is missing.

Using 12 noon instead of midnight for the next Thursday date works
around this, since the 1h DST difference has no influence and gets
floored either way in the dayNumber calculation.


mtwebster:
Hi - it seems like https://github.com/linuxmint/Cinnamon/commit/6032c918037333352133bc19bf5bba514a8ae3a4 is intended to fix the same issue - can you confirm if this does indeed fix it?  If so I'll close this one.  Thanks

netzvieh:
Hi,
I couldn't test the code, but it definitely looks like it does. toLocaleFormat uses C strftime formatting and %V is the ISO Week number. Wasn't sure about using that, since it's a non standard function.

PR #3867, opened 2015-01-31
T440s / Linux mint 17.1 / Cinnamon 2.4.6 - missing battery summary and tooltip

michaeldjcox:
This is a one liner.

The device details passed to GetPrimaryDeviceRemote seem to be an array of devices rather than a single device.

This worked for me on a two battery T440s.


dalcde:
Does this work if you have a single battery only?

Also, if it is a one-liner (which it is), it would be much appreciated if you could re-do the pull request to have a clean commit instead of a lot of things done and then reverted.

michaeldjcox:
Unfortunately I do not have a single battery machine. 

I notice if I take out one battery I am left with just the top summary and no single battery detail in the menu.

If then revert the one line change I loose the summary but gain the single battery detail.

There might be a bit more to do to cater for all machines but I don't have a truly single battery machine to test with.

Are you able to try the fix?

Also I will try and rework the pull request

Closing this one for now

PR #3870, opened 2015-01-31
Fix missing battery summary in applet menu and missing "time left" tooltip

michaeldjcox:
On a T440s which has two batteries this one liner restored the tool tip and menu summary indicating the battery time left across both batteries.

It has not been tried on a standard laptop with one battery as I don't have one - someone please test


dalcde:
Doesn't work with laptops with single battery

dalcde:
If you change line 165 to

    this._proxy.GetDevicesRemote(Lang.bind(this, function(device, error){

together with your change, it should work on all devices.

michaeldjcox:
Hmm not does not work for me.  I'll see if I can borrow a one battery laptop.

PR #3877, opened 2015-02-02
Bug fixes and improvements for multiple battery laptops

michaeldjcox:
On T440s which has multiple batteries there were a number of issues:

1) The panel applet tooltip just says "Laptop battery" rather that "x hours and y minutes remaining"
2) The applet menu is missing the top item which says "x hours and y minutes remaining   x%"
3) The data feeding both those entries was being driven by the state of one of the two batteries rather than the primary device which provides the summary state for two batteries.

I could have extended to show the time remaining on each battery listed but it turns out that the battery not currently being discharged cannot provide accurate seconds remaining. So instead the individual laptop battery items in the popup menu show only the percent.

I am able to remove one battery and see that the applet still makes sense.

I don't currently have access to a single battery laptop and would appreciate some feedback on how this applet looks on a more standard laptop.

### DIFF
--- a/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
@@ -169,7 +169,7 @@ MyApplet.prototype = {
                 this._batteryItem.actor.hide();                
                 return;
             }
-            let [device_id, device_type, icon, percentage, state, seconds] = device;
+            let [device_id, device_type, icon, percentage, state, seconds] = device[0];
             if (device_type == UPDeviceType.BATTERY) {
                 this._hasPrimary = true;
                 let time = Math.round(seconds / 60);
--- a/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
@@ -181,7 +181,7 @@ MyApplet.prototype = {
             if (error) {
                 this._hasPrimary = false;
                 this._primaryDeviceId = null;
-                this._batteryItem.actor.hide();                
+                //this._batteryItem.actor.hide();
                 return;
             }
             let [device_id, device_type, icon, percentage, state, time] = device[0];
@@ -215,7 +215,7 @@ MyApplet.prototype = {
                 this._batteryItem.actor.show();
             } else {
                 this._hasPrimary = false;
-                this._batteryItem.actor.hide();
+                //this._batteryItem.actor.hide();
             }
 
             this._primaryDeviceId = device_id;
@@ -232,8 +232,9 @@ MyApplet.prototype = {
             }
             let devices = result[0];
             let position = 0;
+            let timeToCharged = 0;
             for (let i = 0; i < devices.length; i++) {
-                let [device_id, device_type] = devices[i];
+                let [device_id, device_type, icon, percentage, state, time] = devices[i];
 
                 if (this._hasPrimary == false) {
                 	if (device_type == UPDeviceType.AC_POWER) {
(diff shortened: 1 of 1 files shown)


dalcde:
Just some nit-picking - you can ignore me if you wish :)
I think using a date object for determining the time is overkill. A simple Math.round(time/60) or something like that would do.

michaeldjcox:
Nit-pick away - glad to have some help with this. 

I went through similar attempts to fix this with mint 16 but the changes never got pulled despite numerous "+1"s.  I'm very keen to do everything possible this time to make sure the fix is acceptable, works for everyone, and gets pulled into the next release of cinnamon.

Regarding the time handling. I lost some faith in some of the existing calculations - saw some strange results, couldn't explain the code that was there, and saw the opportunity to consolidate similar calculations made in several places into one function call.

I can have a look to amending the function I created.

clefebvre:
Hi Michael, thanks for persisting with this. I remember the original PR. I hope we find a solution that works for all devices out there.

I can help testing this in a few days if nobody else has a single battery device (I'm using the laptop to work on betsy atm).

dalcde:
It works well on my laptop. Curiously the original version does not. (Lenovo E440, single battery, for the record)

However, now I have two labels in the popup, one saying "... hours ... minutes remaining" and the other showing "Laptop Battery", both with the remaining percent label. I think the idea is to show the time remaining for your primary battery, and then percentage remaining for the other batteries below. But would it make a bit more sense if the "Laptop Battery" item is not shown for the primary battery in use?

michaeldjcox:
If I have two batteries I am interested knowing the total time remaining.

But I take your point if there is only one battery then the % remaining is there twice.

So we could either:

1) remove the percentage from the top line if there is only one battery...
2) remove the laptop battery / % entry

I prefer the former really because it seems to offer a more consistent experience across laptop types and if I remove one of my batteries altogether.

dalcde:
Just spotted another issue: The percentage remaining does not show up when I am charging. Presumably because it thinks the AC is the primary battery, and is unable to retrieve a percentage reading from it.

michaeldjcox:
ideally it would show time left until fully charged. Need to look into this. I think the information is available.

clefebvre:
Hi,

I tested the applet on a macbook pro. On the master branch the battery doesn't show the remaining time, only the percentage. This is due to "device" (returned in GetPrimaryDeviceRemote) being an array.

This PR fixes the issue.

Remaining issues: 
-------------------------

- As mentioned above, is the duplication of the percentage. In my example I see "3 hours remaining ... 70%" and just below I see "Battery ... 70%".

- When plugged on AC power, the overall information disappears and all I see then is "Battery .. 70%" (This time with a charging icon). The problem though, is that the percentage is no longer visible in the panel itself.

clefebvre:
In the cinnamon power settings, you can see the remaining time until full charge. The source code for this is in cinnamon-control-center (https://github.com/linuxmint/cinnamon-control-center/blob/master/panels/power/cc-power-panel.c#L230)

michaeldjcox:
I will try and fix the remaining issues.

I'm also separately addressing the cinnamon-control-center now.

There I only see the text "using Battery Power" (no time remaining) and the T440s internal battery is shown. The status of the 2nd external battery is only shown if its full discharged or fully discharged 

Theres some kind over overarching device that reports the overall time across the two batteries and the overall percentage. 

The applet has access to that but the cinnamon control panel does not. I'm not sure how to get access from the C++ code of the cinnamon control panel.

But thats an issue for another repo, fork and pull request....

michaeldjcox:
So I found that the "time" available with the device shows the time remaining until charged when the AC power is plugged. This is good: I can add the time across the batteries and display that in the menu top summary line. 

I can also show this on the panel which I will do next time I'm near a power source.

What I don't have any more is an overall percentage of battery charge across the two batteries.

My fixes are also a bit "hacky".

The reason for both these things is that the picture of the world I get when the battery is discharging is:

Composite (96% 11hrs 21 min DISCHARGING)
    |
    +--Battery 0 (99% unknown UNKNOWN)
    |
    +- Battery 1 (96% 9hrs 7min DISCHARGING) 
  
(This reflects what can be seen with acpitool -B)

But when the AC Adapter is plugged in I get:

    +--Battery 0 (99% 0hrs 5min CHARGING)
    |
    +- Battery 1 (96% 9hrs 7min CHARGING) 

(This does not reflect what is in acpitool -B)

The GetPrimaryDeviceRemote call seems to return with error when the AC power is plugged in.

Theres good information available in upower and acpitool but I'm not sure this applet is getting a consistent picture from whatever is feeding this information.

Work continues...

clefebvre:
Hi Michael,

A quick follow up on this. I used your feedback and redesigned the applet to make the code much simpler than before. I also changed the rationale so it lists all devices and information for each of them. Finally I took the opportunity to revamp the UI as well. 

Most issues should be solved now. The guys also merged brightness and keyboard backlight into it (prior to my changes so I had to continue on master and close this PR). The remaining thing I didn't take care of is for cases with multiple laptop batteries with none of them defined as primary... in this case the panel shows the info for the first one rather than an average of the values.

michaeldjcox:
Thank you.

Finally upgraded to 17.2 - without a hitch.

The power applet and control panel still does not really work well for 
me on reporting status of multiple batteries on this particular laptop.

I miss seeing the time remaining on the panel and a cumulative % 
indication.

I'm not sure whether % time remaining above the brightness I see in the 
applet is for one battery or cumulative.



If you restart cinnamon the % charge remaining in the panel disappears 
and you cannot get it back

The battery info in the control panel does not list all the batteries 
and their status




I will take a look again as for some reason knowing what my batteries 
are up to seems to be very important to me!

Unless you are aware of any work already in hand here.

Hopefully I can make a contribution suitable for adoption.

Regards,

Michael

Issue #3801, opened 2015-01-05
System Info applet crashes with mounted Samba Shares


Apostle7:
Linux Mint 17.1 Cinnamon 64-bit

System Info applet crashed when launched from cinnamon-settings.

I have extensive programming experience. However, I have never written in Python
or submitted a code fix. I hope submitting it as a bug is the proper procedure. 
I've already tracked the problem and corrected it.

The function:

getDiskSize()

In file:

/usr/lib/cinnamon-settings/modules/cs_info.py

This function gets the output from the 'df' command and adds up the total amount
of disk space for the machine. The problem is that I mount Samba shares at
boot and this command ends up returning a string that is being converted to a
float: error!

If I use: "df -l" in terminal, it only returns local info. However, if I pass
that argument in python, it does not like it. So here is my amateur fix...

Original code:

def getDiskSize():
    disksize = 0
    for line in getProcessOut("df"):
        if line.startswith("/"):
            disksize += float(line.split()[1])
            
    return disksize

My code:

def getDiskSize():
    disksize = 0
    for line in getProcessOut("df"):
        if line.startswith("//"):
            pass
        elif line.startswith("/"):
            disksize += float(line.split()[1])
            
    return disksize

Python looks very clean, is very picky and I think I like it. However, I felt it
better to let a pro take a look...

GLolol:
Try this patch:

```
diff --git a/files/usr/lib/cinnamon-settings/modules/cs_info.py b/files/usr/lib/cinnamon-settings/modules/cs_info.py
index d50be27..e5aba69 100644
--- a/files/usr/lib/cinnamon-settings/modules/cs_info.py
+++ b/files/usr/lib/cinnamon-settings/modules/cs_info.py
@@ -50,7 +50,7 @@ def getGraphicsInfos():
 
 def getDiskSize():
     disksize = 0
-    for line in getProcessOut("df"):
+    for line in getProcessOut(("df", "-l")):
         if line.startswith("/"):
             disksize += float(line.split()[1])
```

It's probably better to grab a copy of the source and build it yourself, as editing the files of an installed copy of Cinnamon might cause upgrade issues later on if you choose to do so.

Issue #3569, opened 2014-10-04
System Info - Copy to Clipboard


anandrkris:
This would be much useful in reporting the system details while asking for help in forums, etc.
Moderators often ask people to provide the output of commands like 'inxi - Fx' from terminal,  but copy option using GUI could be handy instead avoiding the terminal route for new comers.

This system information details also could be enhanced perhaps to include other details.

anandrkris:
Looks like now we have a xapps common library that allows uploading all relevant details still would like to see full output instead of limited info.

https://github.com/linuxmint/Cinnamon/commit/e7d7a9cafcb72b0bbb3cce5d52ef5685cd44c6f2

clefebvre:
I'll close this for now and it's almost implemented.

I tried to show the inxi output in a textview, but it's very messy.

anandrkris:
You mean **System Info** shows detailed output like `inxi -F` ?
> System:    Host: anand-Intense-PC Kernel: 4.4.0-21-generic x86_64 (64 bit)
           Desktop: Cinnamon 3.0.7  Distro: Linux Mint 18 Sarah
Machine:   Mobo: CompuLab model: Intense-PC v: 1.x
           Bios: Phoenix v: CR_2.2.0.377 X64 date: 04/10/2013
CPU:       Dual core Intel Core i5-3337U (-HT-MCP-) cache: 3072 KB 
           clock speeds: max: 2700 MHz 1: 801 MHz 2: 802 MHz 3: 847 MHz
           4: 802 MHz
Graphics:  Card: Intel 3rd Gen Core processor Graphics Controller
           Display Server: X.Org 1.18.3 drivers: intel (unloaded: fbdev,vesa)
           Resolution: 1920x1080@60.00hz
           GLX Renderer: Mesa DRI Intel Ivybridge Mobile
           GLX Version: 3.0 Mesa 11.2.0
Audio:     Card Intel 7 Series/C210 Series Family High Definition Audio Controller
           driver: snd_hda_intel
           Sound: Advanced Linux Sound Architecture v: k4.4.0-21-generic
Network:   Card-1: Intel 82579LM Gigabit Network Connection driver: e1000e
           IF: enp0s25 state: down mac: 00:01:c0:13:68:87
           Card-2: Realtek RTL8723AE PCIe Wireless Network Adapter
           driver: rtl8723ae
           IF: wlp4s0 state: up mac: 00:0d:f0:ad:6b:c4
[...]

PR #3885, opened 2015-02-07
Minor improvements to System Info applet

GLolol:
* Use `df -l` to find system hard disk size (Possible fix for #3801?)
* Use a × (U+00D7) instead of an x to count CPU cores.
* Reformat Operating System checking, preventing empty fields such as `debian 8.0 '' (x86-64)` from showing up.
  `platform.linux_distribution()` doesn't always have all 3 fields filled out: e.g. Arch Linux gives `('arch', '', '')`

### DIFF
--- a/files/usr/lib/cinnamon-settings/modules/cs_info.py
+++ b/files/usr/lib/cinnamon-settings/modules/cs_info.py
@@ -50,7 +50,7 @@ def getGraphicsInfos():
 
 def getDiskSize():
     disksize = 0
-    for line in getProcessOut("df"):
+    for line in getProcessOut(("df", "-l")):
         if line.startswith("/"):
             disksize += float(line.split()[1])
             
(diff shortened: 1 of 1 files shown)


anandrkris:
#3569 > For your interest :smile:

GLolol:
Cinnamon doesn't directly depend on inxi, but it might be interesting to have that and use the current behavior as a fallback.

I'm not sure what the optimal formatting of output would be though; just using a text box to catch STDOUT would look really awkward. Instead of tediously having to parse inxi's command output, why not implement it into the applet itself? After all, inxi itself is a bash script that wraps around commands to fetch information.

anandrkris:
Dependency should not be a issue, at least for Mint distro. And, we could have existing as a fallback mechanism as you pointed out.
Hmm, not a developer here but have seen inxi output in color mode as well. Perhaps it supports formatted output? Also, instead of fetching full specifications we could also employ filters such as `inxi -D` `inxi -G` `inxi -M` etc to extract relevant info? http://manpages.ubuntu.com/manpages/trusty/man1/inxi.1.html

My original idea was to enhance existing info and add copy to clipboard button to maintain existing legibility.

GLolol:
Cinnamon isn't exclusive to Linux Mint, other distros have it packaged too. Copy to clipboard doesn't sound like a bad idea; it would be indeed useful if the applet was more verbose.

anandrkris:
@GLolol - Yes, agree. I was only saying it won't be an issue on Mint.

PR #3894, opened 2015-02-09
Changing the default icon for the file types 

karakartala:
Right now, changing the default icon for the file types is some troublesome. But this option add in theme settings, will be easier than right now.

Note: Sorry. My English is bad.


dalcde:
You might want to clean up your pull request

AlbertJP:
@karakartala Pull requests are used, if you already have programmed the fix to the problem and want your code to be included. If you didn't code the fix yet, please open an issue instead.

PR #3903, opened 2015-02-13
Don't close menu every time a favorite is removed

RavetcoFX:


RavetcoFX:
Causing freezes, as it basically reverts https://github.com/linuxmint/Cinnamon/commit/8d7984bd5abbaf56da585f623c7658553fedeb1d

PR #3936, opened 2015-02-26
Use -1 for infinite max xlet instance (semi-backwards compatibility)

dalcde:


collinss:
This makes a lot more sense. :+1: 

On a related note, I have been thinking that it might be nice to have another setting (perhaps using max_instances=0) which only allows one instance per monitor, and mirrors settings across all instances. It will probably take quite a bit of additional work, but it might be a nice feature to have in the future.

lestcape:
I don't see yet a reason to have for example 3 instances... Or you want to have 1, or 1 per monitor, or infinite, or one per another thing, but not n arbitrary instance. Another thing is that the cinnamon setting think that i want to have a new settings instance for all my extension instances and this is not necessary a true.

Issue #542, opened 2012-03-16
Window Manager: Placement mode


Generator:
Like in Compiz: Place Windows, users can select the windows placement mode,  Cascate, Centered, Smart, Maximized...
One workaround is devilspie, but if implemented would be nice

Anachron:
This has been implemented as of
https://github.com/leigh123linux/muffin/commit/43d5e6fe4ec489fef6acc08479691b6eeb5230d1

But I think muffins version included in Cinnamon is older than that. 
@mtwebster when do you expect a new Cinnamon-Release?

mtwebster:
It will be *around* May (no fixed date)

Anachron:
Thanks for the info! 
Will this be selectable with a dropdown menu in the "Windows"-Settings or only via `gsettings`?

mtwebster:
We'll add a setting in Windows

maninak:
I've looked everywhere for a workaround/hacked implementation of Compiz's "smart placement" of windows, where it will put them where there's open space etc. Sorely needed on a laptop mainly handled with the trackpad and an awesome feature to boot.

Would love to see it officially integrated, but until then if someone has found a way to bolt that feature in manually, please let me know.

adam900710:
+1 for compiz/xfce4/openbox "smart placement".
Current muffin "auto" is only smart when there is enough open space.
And if there is, it just stack winodows into left up corner.

But compiz/xfce4/openbox's "smart placement" is *MUCH MUCH MUCH MUCH* smarter than that.
It will place window to ensure least coverage with others.

For me that's almost a killer feature.

adam900710:
Any update?

JosephMcc:
Not as far as I know

PR #3945, opened 2015-03-04
add a setting to control initial window placement

JosephMcc:


anandrkris:
https://github.com/linuxmint/Cinnamon/issues/542 > Linking for easy reference.

dreamhunt:
And option to remember position and size would be nice. It's very annoying popups from FF for example to popup in the middle of the screen. Like smiles in phpbb3 for example.

PR #3961, opened 2015-03-14
Add background icon for lmde

RavetcoFX:
icon


lestcape:
If cinnamon is used on debian is used the debian icon and if used on ubuntu the Mint icon? I really don't see the logic... what is the difference between Mint base on Ubuntu and Mint base on Debian? I don't need the answer, just to be taking on account.

PR #3976, opened 2015-03-19
tweak the layout of background settings a bit. Looks a bit off without t...

JosephMcc:
...he sectionbg


dalcde:
I think just removing the indenthboxes looks better. I like the two white boxes to align. (also the second item doesn't exactly need an hbox, but it doesn't hurt to have one for the sake of consistency in code). But again I usually lack proper aesthetic senses...

JosephMcc:
After looking at it longer I have to agree. Re-did this here: https://github.com/linuxmint/Cinnamon/pull/3977

PR #3978, opened 2015-03-19
testing: convert cs_mouse to use a list box style of layout

JosephMcc:


dalcde:
I'd say it probably looks better, but it makes the window uncomfortably tall. Also, does it work well when a row has multiple widgets? eg. cs_background

PR #3980, opened 2015-03-21
Fix broken panel settings. Fixes #3922

### LINKED ISSUE #3922: Wrong selection of the panel id
I use only one bottom panel.

This is the gsettings output:

lestcape@lestcape:~$ gsettings get org.cinnamon panels-resizable
['1:false', '-1:true', '2:false']
lestcape@lestcape:~$ gsettings get org.cinnamon panel-edit-mode
false
lestcape@lestcape:~$ gsettings get org.cinnamon panels-height
['1:25', '-1:32', '2:25']

here i modify on cinnamon-settings the panel height (one panel) and then I repeat again the command, to see the id that changed:

lestcape@lestcape:~$ gsettings get org.cinnamon panels-height
['1:25', '-1:42', '2:25']

To see what happend on cjs, I insert this line:

´´´
log("PANEL_HEIGHT_KEY " + panelHeight + " PANEL_RESIZABLE_KEY " + this._getProperty(PANEL_RESIZABLE_KEY, "b") + " PANEL_SCALE_TEXT_ICONS_KEY " + this._getProperty(PANEL_SCALE_TEXT_ICONS_KEY, "b") + " panelId " +this.panelId);
´´´

before this line https://github.com/linuxmint/Cinnamon/blob/master/js/ui/panel.js#L1644

This is the output:
Cjs-Message: JS LOG: PANEL_HEIGHT_KEY 25 PANEL_RESIZABLE_KEY false PANEL_SCALE_TEXT_ICONS_KEY false panelId 1
Cjs-Message: JS LOG: PANEL_HEIGHT_KEY 25 PANEL_RESIZABLE_KEY false PANEL_SCALE_TEXT_ICONS_KEY false panelId 1
Cjs-Message: JS LOG: PANEL_HEIGHT_KEY 25 PANEL_RESIZABLE_KEY false PANEL_SCALE_TEXT_ICONS_KEY false panelId 1

So cinnamon-settings modify the panel id: -1 and my panel have the id 1 on cjs.

dalcde:
This is actually three fixes
 1. #3922 Don't set the panel_id as -1 when not supplied (since it is never supplied if accessed through navigating cinnamon settings as opposed to right clicking panel)
 2. Update the panel_id property of the widgets when panels-enabled is changed
 3. Don't cast int(self.panel_id) since self.panel_id is always an int

### DIFF
--- a/files/usr/lib/cinnamon-settings/modules/cs_panel.py
+++ b/files/usr/lib/cinnamon-settings/modules/cs_panel.py
@@ -24,10 +24,12 @@ class Module:
 
         self.widgets = []
         self.panel_id = None
-        if len(sys.argv) > 2:
-            if sys.argv[1] == "panel":
+        try:
+            if len(sys.argv) > 2 and sys.argv[1] == "panel":
                 self.panel_id = int(sys.argv[2])
-        else:
+            else:
+                self.panel_id = int(self.settings.get_strv("panels-enabled")[0].split(":")[0])
+        except:
             self.panel_id = -1
 
     def on_module_selected(self):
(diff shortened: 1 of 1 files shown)


lestcape:
+1

PR #3994, opened 2015-03-24
UX improvments in cs-backgrounds

RavetcoFX:
-Made icons colourless
-Tweaked thumbnail drop-shadow

screenshot from 2015-03-24 11 59 05


clefebvre:
I don't think that looks better:

- folders are still fullcolor, so that's inconsistent
- mint logo look ugly in symbolic
- can't see what's better about the shadows?

Issue #3849, opened 2015-01-24
[improvement/feature request] centralized applets on panel


gilbertohasnofb:
The panel in my Linux 17.1 64-bit Cinnamon has three areas for adding applets (in edit mode, they are color coded: a red section on the left, a green section in the centre and a blue section on the right). Dragging applets to each one of them will move their icons to the left, centre and right of the panel, respectively.

The problem is the central section panel is not adding the applets to the centre of the panel itself (that is, exaclty in the middle of the screen), but rather to the centre of the green section. So if for instance an user has a lot of applets in the blue section and only the menu applet in the red section, it will make the green section be quite shifted to the left, and thus the centralized applets will look bad and not centralized to the screen (one of the typical uses of centralized applets nowadays is, for instance, to put the clock in that position). Also, when using applets such as "show/hide applets", it will move anything located in the green section since it dynamically changes the size of the blue section. 

Here are some images of what I am describing:

![a-0](https://cloud.githubusercontent.com/assets/5700795/5885695/6c79f166-a374-11e4-95b1-6653d3a620b3.png)
![a-2](https://cloud.githubusercontent.com/assets/5700795/5885696/6c915284-a374-11e4-8057-5396b12fa3aa.png)

Please let me know if I can be of any help with this.

JosephMcc:
Thanks for the great explanation of what you're after. Not sure you can really do this though. Think about it this way. If I have a panel on the bottom with a window list applet in the left area of the panel and a large number of windows open, the applet takes up more than half the panel. Now I can't have another applet centered in the area already used by the window list.
![screenshot-1](https://cloud.githubusercontent.com/assets/2231759/5886093/69ed60a4-a33d-11e4-92f9-eb49ae3e1b42.png)

daviday:
I don't think you're defining the issues/attributes re the applets properly. You should be able to simply reassign to fit your needs. Justification is ledger dependent, obviously, but if you're unable to add/remove applets from the tool bar you may be in for more work than you bargained for. 
This is a non issue for me on 15.01 btw. 

Sorry if this is not helpful.

JosephMcc:
@daviday Sorry but what are you talking about? No one said anything about not being able to add or remove applets from the panel. The original poster brought up a legitimate issue, just not sure it can be changed. And what is 15.01?

gilbertohasnofb:
@daviday thanks for the reply, but I also do not understand your message - I am sorry :(

[...]

Issue #3100, opened 2014-04-27
Setting to center/freely place applets in bars


t3mux:
Well i think the title says everthing. I just want my clock in the center of a bar but i cannot center it because its position is based on the applets on the right and the left.

I would like to see a center settings to avoid this or that applets could placed freely in the bar. The latter could be the better option.

JosephMcc:
Applets placed in the center are now placed at screen center and ignore the size of the left and right boxes and it's been this way for awhile.

PR #4041, opened 2015-04-12
Add proper panel center region. Applets are actually at the center if possible

dalcde:
If there is enough space, centered applets will be placed at the very center of the applet. Otherwise, the centered applets will appear as close to the center as possible. See centered menu applet:

screenshot from 2015-04-12 14-40-09

#3849, #3100

### DIFF
--- a/js/ui/appletManager.js
+++ b/js/ui/appletManager.js
@@ -294,11 +294,7 @@ function addAppletToPanels(extension, appletDefinition) {
             appletDefinition.location.remove_actor(appletsToMove[i]);
         }
 
-        if (appletDefinition.center) {
-            appletDefinition.location.add(applet.actor, {x_align: St.Align.CENTER_SPECIAL});
-        } else {
-            appletDefinition.location.add(applet.actor);
-        }
+        appletDefinition.location.add(applet.actor);
 
         applet._panelLocation = appletDefinition.location;
         for (let i=0; i<appletsToMove.length; i++) {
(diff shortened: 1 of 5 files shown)


brownsr:
Looks interesting.  Not quite clear on what is being changed.  Are you saying that the icons are currently offset just to left or right of the centre if dragged into the centre panel, and this corrects that ?

dalcde:
See https://github.com/linuxmint/Cinnamon/issues/3849 for how the center region works without this pr

mohammad-sn:
hi, personaly, i prefer this behavior to the default... i did the same by an extension... (http://cinnamon-spices.linuxmint.com/extensions/view/33) but I think this should be configurable by user, because in some cases panel looks better at the current state, with (equal) free space in both sides of centered applets. presence of these free spaces is important when there is not enough space on the panel, if you align applets to the left (or right) in that case, center panel's applets will not be seperated from left (or right) panel and will look (and feel) like they are attached to them. It's true that center panel should try to keep its contents as centered as possible but i think it also has to avoid touching the side panels!

dalcde:
I tried to make it use the old behaviour when the left/right panel is too long, but wasn't happy with the discontinuity caused - when the left panel gets longer and starts to cover the center region, the center applets suddenly "jump" to the new position. If you want a padding between the different regions, it can and should be done on the theme level.

mohammad-sn:
you're right! :) padding can do that, the only thing is that they are static.
couldn't we have a center panel that acts like that its corners is connected to side panels by springs, and its center to panel's center by stronger spring??! so the padding will dynamically, smoothly and wisely change its size, and adopt to available free space.

PR #4045, opened 2015-04-13
[WIP] Fix some errors in Panel Manager and make it extensible.

lestcape:


dalcde:
I'm not sure how naming the signals are useful. If you want to modify its behaviour in an extension, either you replace the onPanelsEnabledChanged etc functions, and those get called instead, or you replace the whole Panel object, in which what gets connected doesn't matter. Anyway, the id names should start with an _, eg this._panelsEnabledId instead of this.panelsEnabledId

lestcape:
@dalcde i cannot understand your idea, but let me explain why this is for me the best solution and the only one.

1- I can forget complete the cinnamon panel implementation and make a new way to do what i want, but this could be really confused to user that can create the panels on differents ways, so i discard this idea.

2- The second idea will be replace the current panelmanager behavior and extends the functionality to add vertical panels and this is the idea that i use.

If i replace the cinnamon panel manager functionality, i need to also replace the current panels when the extension start for a new one create for me, or if the panel is shared with the cinnamon implementation, this will be acessible on two different ways, creating then a big miss. If i replace the panel without disconnect the function, will be two versions of the same panel at the same time one for my extension and another for the cinnamon current implementation.

How i can prevent that the cinnamon panelmanager could capture the signals and then create or remove panels? I can not destroy the global.settings and global.screen objects. I can replace onPanelsEnabledChanged etc functions for the new connection that occurs after the replace, but not for the internally original version that this objects actually have, so this is why my replace will not be called. Will called the original one, the original function that was refer when the connection occurs.

You can only replace the function for a connection before the connection occurs, after is impossible and panel manager start before my extension, so there are not a possible solution here. If you found one way to do that, please i will be happy to used it...
Thanks for your suggestion but i don't know how do that.

dalcde:
Okay you're right. There isn't any easy way of changing it. But still make those private variables instead of public variables.

dalcde:
This looks good to me, but the pull request needs some rebasing (for some reason I cannot explain - nothing much has changed in panel.js). Also would be great if you could squash the relevant commits together.

lestcape:
Why you don't forget this commit and then make one that can be easy merged? Not need to have the "save id", if some one not want this... but throw a lot of event that can crash cinnamon, not have reason to not be merged i think... I don't need my name on this, I only want fix the problem and if is possible add also the possibility to modify the panel, if some one need that to create a dock bar, as when the vertical panel will be include this will not have any other usage, i think...

clefebvre:
@lestcape can you join us on IRC (irc.spotchat.org #linuxmint-dev)?

lestcape:
Hi, @clefebvre. I will try more, but my current internet connection won't let me... I'm not angry or something, just i have not internet connection to do all that i want or all that is necessary do.

Regards.

PR #4047, opened 2015-04-14
Add file chooser dialog to Cinnamon javascript

collinss:
This introduces a file dialog to Cinnamon. The dialog itself is Gtk, but uses the Util.spawn_async method to open the dialog and return the results. It supports 
* open, save, and open folder types
* single or multiple selection
* setting default directory, path and/or name
* custom file filters

Typical usage might be something like:

    const FileDialog = imports.misc.fileDialog;
    
    let params = {};
    params.directory = "~/Pictures/";
    params.name = "untitled";
    params.selectMultiple = true;
    
    let anyFilter = new FileDialog.Filter("any");
    anyFilter.addPattern("*");
    
    let imageFilter = new FileDialog.Filter("image");
    imageFilter.addMimeType("image/jpeg");
    imageFilter.addMimeType("image/png");
    imageFilter.addMimeType("image/svg+xml");
    
    params.filters = [anyFilter, imageFilter];
    
    FileDialog.open(function(path){
        //do something with the path
    }, params);

### DIFF
--- a/files/usr/bin/cinnamon-file-dialog
+++ b/files/usr/bin/cinnamon-file-dialog
@@ -1,0 +1,60 @@
+#!/usr/bin/env python3
+
+import sys
+from gi.repository import Gio, Gtk
+
+cancelButton = (Gtk.STOCK_CANCEL, Gtk.ResponseType.CANCEL)
+typeArg = int(sys.argv[1]) 
+fdType = typeArg % 3
+if fdType == 0:
+    msg = "Open"
+    action = Gtk.FileChooserAction.OPEN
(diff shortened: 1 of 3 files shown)


clefebvre:
@collinss shouldn't the js file be added to Makefile.am? (and possibly configure.ac?)

collinss:
I don't really know anything about makefiles and building and all of that except that I've heard of them and that I there is a lot I don't know about them. I've never had to worry about that before. I built it like normal on my machine (using dpkg-buildpackage) and it seemed to work. What does adding them to those files do?

collinss:
@clefebvre I added it to Makefile.am. Let me know if I did something wrong...

PR #4051, opened 2015-04-15
Change icon properly with PopupIconMenuitem

dalcde:
Should we do the same for IconApplet as well?


mtwebster:
We can try, I imagine it'd have to be quicker.

PR #4058, opened 2015-04-16
cs_backgrounds: use both cinnamon-background-properties and gnome-backgr...

dalcde:
...ound-properties


clefebvre:
No, that would result in Cinnamon showing GNOME backgrounds. That's undesirable at distribution level.

PR #4063, opened 2015-04-18
Fix regression in multi-instance applets

collinss:
There was a regression introduced by https://github.com/linuxmint/Cinnamon/commit/b1a6bce1dd879a05b480cc78f3496cab9d0ef3e2 which was causing multi-instance applets to not be recognized as multi-instance by certain Cinnamon components, which was causing it to not recognize the settings for the first instance of the applet.


dalcde:
Ah thanks. I was experiencing that issue as well.

collinss:
:)

PR #4067, opened 2015-04-21
new file

shilpa12345:


dalcde:
The Cinnamon team is very grateful for your input. However, after careful discussions among the current developers, we have decided that this feature is not very suitable for Cinnamon, as we aim for a more traditional user experience. You can consider submitting your patch to GNOME Shell - they are more on the cutting-edge and are willing to innovate on brand new features.

PR #4068, opened 2015-04-21
Make sure the button in the window list keeps the "focus" style class wh...

JosephMcc:
...en moving between monitors and window lists


JosephMcc:
@dalcde has a PR for a more complete rework of the window list.

Issue #2912, opened 2014-02-27
Remember which Workspace each window belongs to across crashes/reboots


logannc:
This is a feature request, I guess.

It would be nice to have Cinnamon remember which windows belonged where. Occasionally I'll have Cinnamon crash and all of my windows get shoved in Workspace 1 and I have to resort them to the other workspaces. It's not a problem, nor is it frequent, but I still feel like it'd be a good feature to have.

Optionally, this could also be persistent across reboots/OS crashes. I imagine solving this optional problem would also yield the ability to automatically restart the programs you had open on shutdown. I, personally, feel that it is more likely you want the same applications open again when you reboot more often than you want nothing to be there.

Incidentally, regarding the Cinnamon crash, I *think* it was caused by an overflow in a notification? But I can't be sure. Would that be an issue in Cinnamon or a different project?

Vahan86:
Issue https://github.com/linuxmint/Cinnamon/issues/821

Issue #1824, opened 2013-03-29
Autostart applications on specific workspaces


balliet:
Provide an option for applications to be started on specific workspaces following boot-up of the computer so the user does not have to manually place windows onto the workspaces one at a time.

dalcde:
Closing in favour of more comprehensive #4161

iamfedoroff:
+1

PR #4069, opened 2015-04-21
auto assign applications to workspaces feature

megz:
KDE like application feature ( but simpler and through the menu system ) where by applications can be manually assigned to workspaces on launch. 

Why is this useful ? example: 

launching the web browser ( firefox ) it will open on workspace 1 of 4 workspaces
launching the web browser again it will open on workspace 1 or 4 workspaces
launching banshee it will open on workspace 1 of 4 workspaces
launching nemo it will open on workspace 1 of 4 workspaces

All the same, every boot or from every resume from suspend. It seems too static a configuration for the heavier user  ( aren’t we all now ! ) 

Of course, If I know that Firefox is on 1 and Dolphin is always on 2 then I know that I always need to go to 2 ..  this increases switching speed between workspaces, it also largely removes the need for constant access to the "show all workspaces" function and drag and drop .. every session, every program.

In fact, the idea of a dynamic workspace is nice, but without at least some permanency you cannot design a workflow around them, making workspaces in their current form almost non effective vs windows quick list.

its a fairly essential addon i would hope to find in cinnamon settings, much like when you select your favourite application for opening a specific task, we can do the same for a specific workspace.

thanks

* I would add, being able to somehow do this per monitor too would be amazingly useful and open up many possibilities.


dalcde:
Duplicate of #1824, #2912 etc. Also you want to open a *issue*, not a *pull request*. A pull request is for submitting patches for things *you have implemented*.

Issue #3812, opened 2015-01-11
Skype new message notification in Window list applet


nbasov:
Hello developers!

Is it possible to implement into Window applet following fix for Skype new message notification?
https://superuser.com/questions/726444/highlight-skype-window-when-i-receive-a-message

Missing this solution is very inconvenient to chat in Skype.

anandrkris:
I understand this is for Window list. Just checking, it's been sometime since I used Skype, do notifications appear on System tray for incoming messages?

nbasov:
No, notifications still doesn't appear on System tray for incoming messages

dalcde:
Should be fixed by new window-list applet. Please confirm

nbasov:
When new skype chat appears - system creates new window list button that is flashing. It flashes about 4 times and then stops even if i don't open this chat window. If there is new message in this chat - window list button never flash.

Is this right behavior? I assume that window list chat button should always flash until it opened.

PR #4075, opened 2015-04-24
Rework window-list applet

### LINKED ISSUE #3574: when drag&dropping windows on the window list, not all positions can be reached
Hi.

When one moves a window on the window list via drag&drop, then sometimes[0] on cannot drop the window at all positions, since some of the other windows "stick together".

The following three screenshots triy to visualise what I mean:
screen
(The screenshots go chronologically from up to down.)

The window that I drag here from right to left, is the one titled "FRITZ!Box 7490-Hilfe...". As I move it from right to left, I move over the middle window ("Screenshot from 20...") to just a bit left of the middle from the leftmost window ("FRITZ!Box - Iceweasel").
The problem: During the whole continuous move, there is no position in which the moved window would be dropped between "FRITZ!Box - Iceweasel" and "Screenshot from 20..." - at a certain point, both of these two jump to the other side, as if they'd stick together.

This is subtly related to #3573.

Cheers,
Chris.

[0] Sometimes it works however, and I couldn't yet find out, when it does and when not.

dalcde:
- Refactor code to make more sense - this applet was hack on top of
   hack and was a huge mess. A lot of things were done in convoluted
   ways.
 - Add option to scroll window list in the other direction with mouse
   scroll wheel
 - Disconnect signals properly when destroyed
 - Listen to windows demanding attention in addition to urgent-windows
   (might solve #3812)
 - Changed window-list dnd to firefox-tab style
 - Fix broken applet style when changing panel orientation (in the past,
   after changing panel orientation, it will simultaneously have top and
   bottom styles).
 - Added a brief overview of the code structure

Might fix #3574

### DIFF
--- a/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
@@ -1,420 +1,244 @@
-const Applet = imports.ui.applet;
-const Clutter = imports.gi.Clutter;
-const Lang = imports.lang;
+/* Window-list applet
+ *
+ * The applet code consists of four main object. The SignalManager,
+ * AppMenuButton, AppMenuButtonRightClickMenu and the main applet code.
+ *
+ * The SignalManager is a helper object that manages the signals the applet
+ * connects to, providing handy functions to connect and disconnect to signals.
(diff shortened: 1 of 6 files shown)


clefebvre:
When clicking a window list item for a window that's already focused, the window fails to minimize.

PR #4087, opened 2015-04-27
window-list applet: reasonable AppMenuButton preferred width

pawel-wiejacha:
Currently, window-list applet buttons have width limited to 150px. This value is too small to distinguish different windows having the same icon or the same window title prefix.

This patch takes window title length into account when computing natural width.


clefebvre:
Thanks. I couldn't merge it due to dalcde's redesign. I implemented it as an option with https://github.com/linuxmint/Cinnamon/commit/3d4fa41384b82c49db6cdb5f9cb68beea6c6f9ac

pawel-wiejacha:
Cool. Thanks!

PR #4120, opened 2015-05-06
Display cinnamon-menu-editor in menu applications

GLolol:
This is a useful tool for configuring menu items (similar to Alacarte and Menulibre), and only being able to access it via deep, nested Applet options isn't very convenient.


clefebvre:
I disagree. That tool is specifically related to the configuration of that menu, as such it's part of its configuration screen. The way it's implemented it can also be used by other menu applets, that's not an issue.

PR #4199, opened 2015-05-25
Support keybindings to toggle horizontal/vertical maximized state

szeder:
Some graphical applications are ment to be controlled mainly though the
keyboard, think terminal windows.  For these it's extremely handy to have
keybindings to toggle horizontal or vertical maximized state, so one
doesn't have to reach out to the mouse.

Support setting keybindings for 'toggle-maximized-horizontally' and
'toggle-maximized-vertically' in Cinnamon Settings.

See my related pull requests in the [cinnamon-desktop](https://github.com/linuxmint/cinnamon-desktop/pull/47) and [muffin](https://github.com/linuxmint/muffin/pull/192) repositories.


mtwebster:
This is already implemented (but with no default bindings) - pardon my gimp skills:

screenshot from 2015-05-25 12 10 01

I suppose they could say 'Toggle' but that's what they do (and for now I'm not going to lose our translations for it by changing it before a release)

szeder:
Oh, great.
Yeah, the naming is misleading, they should definitely say "Toggle", both in the UI and in code, because that's why I never even bothered to try it while looking for this feature.

PR #4229, opened 2015-05-29
signalManager.js: fix iteration of the object map

collinss:
This was causing errors when the systray applet was moved around in the panel, either directly, or by moving other applets near it.

### DIFF
--- a/js/misc/signalManager.js
+++ b/js/misc/signalManager.js
@@ -136,7 +136,7 @@ SignalManager.prototype = {
         if (!this._storage.has(sigName))
             return false;
 
-        for (let signal of this._storage.get(sigName))
+        for (let [key, signal] of this._storage) 
             if ((!obj || signal[0] == obj) &&
                 (!callback || signal[2] == callback) &&
                 signal[0] &&


dalcde:
I'm not convinced this is the right thing to do. `this._storage` is a map that maps a signal name to an array, and each element of the array is an `[object, signalId, callback]` tuple. Here you are making it loop through the `[key, signal]`, and `signal` will give you the array of all signals with the name `key`, but what we want is the individual signals.

What are the errors you've got?

collinss:
You're right. I misread line https://github.com/linuxmint/Cinnamon/blob/master/js/misc/signalManager.js#L116, and then I forgot that String(array) removes the brackets in Javascript, so it looked like it was passing me the correct values. I guess I shouldn't try to code when I'm tired...
The error I'm getting is http://pastebin.com/M3AGMmtc.

collinss:
I think I figured out the real problem. The systray applet is trying to connect to ```Main.statusIconDispatcher```, which is a javascript object. It then gets passed as the first argument to https://github.com/linuxmint/Cinnamon/blob/master/js/misc/signalManager.js#L143, but ```GObject.signal_handler_is_connected``` requires a gobject, which ```Main.statusIconDispatcher``` is not. It looks like it will need a different test if it is to work properly with javascript objects.

dalcde:
Does this fix the issue for you: https://github.com/linuxmint/Cinnamon/pull/4231

collinss:
Now I'm getting http://pastebin.com/JwfAkjHp

dalcde:
Force-pushed. Try again

collinss:
Yeah, that seems to fix it. :)

lestcape:
I found the same, 2 hours trying to find the reason of the fail...  yumm..
Cjs-WARNING **: JS ERROR: Exception in callback for signal: destroy: TypeError: Object 0x7f7926d0d340 is not a subclass of GObject_Object, it's a Object

disconnectAllSignals not need the same?

Anyway, with the change:
(cinnamon:3134): Cjs-WARNING **: JS ERROR: Exception in callback for signal: destroy: TypeError: signal[0].signalHandlerIsConnected is not a function

I'm try to use this on the indicators, but i thing i need to revert all... There are a solution?
Thanks.

collinss:
@lestcape Did you also build cjs with https://github.com/linuxmint/cjs/pull/24?

lestcape:
ups, not... Thanks @collinss

collinss:
No problem :)

lestcape:
Do you know why this is not merged? I don't know what i can do now. This is really useful, i have a lot of signals.

collinss:
No idea. Maybe @clefebvre forgot about it, or maybe he hasn't got around to testing it yet. I imagine he's been really busy with getting ready for 17.2 right now.

lestcape:
Ok, thanks again i will wait for the @dalcde recommendation and then decide if i will do a regression or not. As is code for cinnamon i don't know the right decision here.

PR #4231, opened 2015-05-30
signalManager.js: fix errors when disconnecting JS objects

dalcde:
Requires https://github.com/linuxmint/cjs/pull/24


lestcape:
I know that this depend of a change in cjs and a change in cjs need to be made with carefully because in the future cinnamon can take the gjs branch again, and if the patch is not set in gjs then this will be a thing to take in account. I will recommend also provide a patch in gjs, with the current cinnamon patch (this is a help to gnome devs and also help cinnamon). I will also recommend revert then the current problems that was the cause of this change if this will not be merged now.

PR #4248, opened 2015-06-03
overrides.js: override GSettings connect function since this doesn't …

dalcde:
…actually connect

### DIFF
--- a/js/ui/overrides.js
+++ b/js/ui/overrides.js
@@ -70,6 +70,7 @@ function overrideGio() {
     Gio._real_set_enum     = Gio.Settings.prototype.set_enum;
     Gio._real_get_flags    = Gio.Settings.prototype.get_flags;
     Gio._real_set_flags    = Gio.Settings.prototype.set_flags;
+    Gio._real_connect      = Gio.Settings.prototype.connect;
 
     Gio.Settings.prototype._init        = function(params)   { check_schema_and_init(this, Gio._real_init, params); }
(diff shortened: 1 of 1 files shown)


clefebvre:
Well ok.. I fully agree with the need for connect to work.. but:

- What about the C layer?
- What about other Cinnamon projects (muffin etc..)?
- What about everything else? Comes Mint 18 and LMDE 3 we'll have no choice but to patch glib to revert their change anyway. There's no way we're going to wait for 60,000 packages to feature random() gsettings regressions... so with that in mind, why fix the JS layer of Cinnamon in isolation? We'd be much more likely to simply reject bug reports in Cinnamon and ask other distributions to patch glib itself since that's where the regression actually is.

You restarted the discussion on bugzilla with mtwebster, let's wait and give Ryan the opportunity to revert this himself. There's plenty of time before the next cycle and we should start seeing regressions in Fedora/Arch before then. So with all that in mind it's not really up to us to fix this, especially not partially like that.

clefebvre:
https://bugzilla.gnome.org/show_bug.cgi?id=733791

leigh123linux:
@clefebvre 

What happens to your patched glib idea when the other 60,000 packages start to switch to the new way?

clefebvre:
Nothing. You will be able to read and/or connect independently from each others, like it's been done so far until they introduced that regression.

If some apps adapt their code to work around that regression, that will have no impact whatsoever with a patched version of GLib.

Issue #4267, opened 2015-06-05
Option to disable tooltips in Window list Applet


simonwiles:
The mouseover tooltips in the window list applet frequently get in my way (most noticeably if I use the mouse to activate my terminal emulator, and then I get a popup which covers the active line I'm typing on), and they strike me as almost entirely superfluous anyway (esp. if using the option to allow the window buttons to use lots of space).  Would it be possible to have an option to disable them?

dalcde:
No (#4269)

PR #4269, opened 2015-06-05
add option to disable tooltips on window-list applet

simonwiles:
as described; addresses #4267 (that I added earlier today).  don't think there's any need to make this a "live" setting that is implemented/removed as the checkbox is set/cleared?


clefebvre:
I'm sorry, I don't like saying NO, but this is really pushing it in terms of configuration. If we start adding options for tooltips now, where do we stop? That's really overkill imo.

simonwiles:
fair enough in terms of configuration options; can the tooltips just be turned off, then?  They don't serve any useful purpose that I can see, and they do get in the way...  (I just removed the code on my own version, and then added the config option back in for the sake of a pull request!)  Just a suggestion, won't be upset if it's declined :)

mondalaci:
This might seem overkill in most cases, but in this case the fact that 2 people (@simonwiles and me in #4980) independently opened this issue proves that there's a reasonable need. (And generally only 1 person proactively acts in such situations out of 100.)

PR #4282, opened 2015-06-07
[2.8] cs_themes.py: remove redundant theme-dir check as it is preventing Ad…

dalcde:
…waita from showing up


clefebvre:
  File "/usr/lib/cinnamon-settings/cinnamon-settings.py", line 38, in <module>
    modules = map(__import__, mod_files)
  File "/usr/lib/cinnamon-settings/modules/cs_themes.py", line 264
    valid = walk_directories(dirs, lambda d: os.path.isdir(d)) and os.path.exists(os.path.join(d, "index.theme")))
                                                                                                                 ^
SyntaxError: invalid syntax

dalcde:
Fixed

clefebvre:
Closing this as discussed on IRC. That test isn't enough... it would bring false-positives such as the DMZ cursor themes.

PR #4298, opened 2015-06-10
[2.8] Add support for indicators.

lestcape:
This will add support for indicators. There are some things that need to be done yet, but as @dalcde recommended i will share what i have. Try to find the FIXME in the code, to find the critical areas. The Api is not binding with Main yet, and this will remove the watchdog class.

### DIFF
--- a/files/usr/share/cinnamon/applets/systray@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/systray@cinnamon.org/applet.js
@@ -5,6 +5,7 @@ const Clutter = imports.gi.Clutter;
 const Applet = imports.ui.applet;
 const PopupMenu = imports.ui.popupMenu;
 const Main = imports.ui.main;
+const IndicatorManager = imports.ui.indicatorManager;
 const Mainloop = imports.mainloop;
 const SignalManager = imports.misc.signalManager;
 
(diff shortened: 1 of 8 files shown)


dalcde:
Don't have the opportunity to test or really look into it yet, but there already exists RadioButton and CheckBox objects in js/ui/checkBox.js and js/ui/radioButton.js respectively, so we don't have to implement it again in popupMenu.js

lestcape:
Yes, the problem is that, they are constructed to be real buttons (with labels ...) In the case of the radiobuttons is more problematics because they react as a group. In the DbusMenu all those things are getting separate, then this is different as how they are design in the cinnamon api, so you need to rewrite the cinnamon api for this class if you want reuse it. If someone want to do this is ok, but if i try to do it, i will never update the applet.

dalcde:
It is highly desirable that we don't have two classes that do the same thing. For the radio buttons, we can create a radio button group that only has one radio button. It would be great if we can aim for something in the middle ground. For example, if your checkbox doesn't have labels, we can make it such that if you pass an empty label to the CheckBox constructor, a label is not created. Then we can use it for both the indicator menu and the current use case.

A quick grep through the existing xlets on cinnamon-spices shows that no xlet actually uses those classes, so it doesn't hurt if how they work slightly changes (apart from @collinss who uses the CheckBox on some WIP code that is not yet published).

All these can be sorted out later on, though. I'm currently more interested in making sure we have a working applet.

lestcape:
I try to not affect the current behavior. When you merge this you can take the decision that you want... Any way both need a theme fix. The most important reason is not the label, is Cinnamon.GenericContainer, that have a lot of code.

lestcape:
I suppose that you can do what you want here right now. You are a collaborator of this branch.

PR #4303, opened 2015-06-11
[2.8] cs-backgrounds: add revealer to gradient widgets

pixunil:


clefebvre:
Color options aren't shown when "no image" is chosen.

PR #4338, opened 2015-06-19
WIP · 2.8 · Automatically reload xlets when updating

pixunil:
Settings informs Cinnamon via DBus, when an update was successful, about the uuids of the xlets.
Errors should also be reported like when adding them.

### DIFF
--- a/files/usr/lib/cinnamon-settings/bin/ExtensionCore.py
+++ b/files/usr/lib/cinnamon-settings/bin/ExtensionCore.py
@@ -896,8 +896,10 @@ class ExtensionSidePage (SidePage):
         self.install_button.set_sensitive(False)
         self.install_list = []
         self.load_extensions()
-        if need_restart:
-            self.show_info(_("Please restart Cinnamon for the changes to take effect"))
+
+        for uuid in need_restart:
+            self.connect_proxy("XletAddedComplete", self.xlet_added_callback)
+            self._proxy.ReloadXlet("(s)", uuid)
 
     def on_spice_load(self, spicesData):
         # print "total spices loaded: %d" % len(spicesData)
--- a/files/usr/lib/cinnamon-settings/bin/Spices.py
+++ b/files/usr/lib/cinnamon-settings/bin/Spices.py
@@ -422,11 +422,14 @@ class Spice_Harvester:
         return basename.replace("jpg", "png").replace("JPG", "png").replace("PNG", "png")
 
     def install_all(self, install_list=[], onFinished=None):
-        need_restart = False
+        need_restart = []
         success = False
         for uuid, is_update, is_active in install_list:
             success = self.install(uuid, is_update, is_active)
-            need_restart = need_restart or (is_update and is_active and success)
+
+            if is_update and is_active and success:
+                need_restart.append(uuid)
+
         ui_thread_do(self.progress_window.hide)
         self.abort_download = False
 
--- a/js/ui/appletManager.js
+++ b/js/ui/appletManager.js
@@ -65,10 +65,10 @@ function finishExtensionLoad(extension) {
 }
 
 // Callback for extension.js
-function prepareExtensionUnload(extension) {
+function prepareExtensionUnload(extension, deleteConfig) {
     // Remove all applet instances for this extension
     for(let applet_id in extension._loadedDefinitions) {
-        removeAppletFromPanels(extension._loadedDefinitions[applet_id]);
+        removeAppletFromPanels(extension._loadedDefinitions[applet_id], deleteConfig);
     }
 }
 
@@ -214,7 +214,7 @@ function onEnabledAppletsChanged() {
     Main.statusIconDispatcher.redisplay();
 }
 
-function removeAppletFromPanels(appletDefinition) {
+function removeAppletFromPanels(appletDefinition, deleteConfig) {
     let applet = appletObj[appletDefinition.applet_id];
     if (applet) {
         try {
@@ -231,7 +231,8 @@ function removeAppletFromPanels(appletDefinition) {
         delete applet._extension._loadedDefinitions[appletDefinition.applet_id];
         delete appletObj[appletDefinition.applet_id];
 
-        _removeAppletConfigFile(appletDefinition.uuid, appletDefinition.applet_id);
+        if (deleteConfig)
+            _removeAppletConfigFile(appletDefinition.uuid, appletDefinition.applet_id);
 
         /* normal occurs during _onAppletRemovedFromPanel, but when a panel is removed,
          * appletObj hasn't had the instance removed yet, so let's run it one more time
@@ -555,7 +556,7 @@ function updateAppletsOnPanel (panel) {
 function unloadAppletsOnPanel (panel) {
     for (let applet_id in enabledAppletDefinitions.idMap){
         if(enabledAppletDefinitions.idMap[applet_id].panel == panel) {
-            removeAppletFromPanels(enabledAppletDefinitions.idMap[applet_id]);
+            removeAppletFromPanels(enabledAppletDefinitions.idMap[applet_id], true);
         }
     }
 }
--- a/js/ui/cinnamonDBus.js
+++ b/js/ui/cinnamonDBus.js
@@ -4,6 +4,7 @@ const Lang = imports.lang;
 const Gio = imports.gi.Gio;
 const GLib = imports.gi.GLib;
 const Config = imports.misc.config;
+const Extension = imports.ui.extension;
 const Flashspot = imports.ui.flashspot;
 const Main = imports.ui.main;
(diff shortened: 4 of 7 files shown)


lestcape:
It's really necessary back to cjs? This will preserve the position of the applets inside the panel (i.e. the applet and desklet definition)?

collinss:
There does seem to be a problem with this. If the applet is multi-instance (which is most of them now) the settings file gets destroyed when ```Extension.unloadExtension``` is called, so the applet loses it's configuration, which is not ideal. I think you'll have to add a function to ```extension.js``` to handle reloading cleanly.

lestcape:
@collinss why we need to do this? All this can be do in the python side (cinnamon-settings).
Read all instance that we update from gsettings "enabled-desklets", "enabled-applets"... whatever the case(extension type) and store it. Then update the available extension (enabled- applets/desklets) gsettings value, without add the update extensions (this will close all that we update currently). Again update the enabled- but now with all using the save values (the definition).

lestcape:
I can help in this, but i don't know how create more than one pull, and the indicator is a big pull to remove it.

lestcape:
This is what i propose:
```
    def install_finished(self, need_restart):
        for row in self.gm_model:
            self.gm_model.set_value(row.iter, 2, 0)
        self.install_button.set_sensitive(False)
        self.install_list = []

        not_update_extensions = [];
        all_current_extensions = [];
        for uuid in need_restart:
            for enabled_extension in self.enabled_extensions:
                if uuid not in enabled_extension:
                    not_update_extensions.append(enabled_extension)
                all_current_extensions.append(enabled_extension)

        self.settings.set_strv(("enabled-%ss") % (self.collection_type), not_update_extensions)
        self.settings.set_strv(("enabled-%ss") % (self.collection_type), all_current_extensions)

        self.load_extensions()
```
Is possible that we need to add a timeout, waiting for the change, between the update of not_update_extensions and all_current_extensions. Need to be tested, if is enough, or is needed the timeout.

lestcape:
I tested it with the last change, apparently is working...

collinss:
@lestcape You can do multiple pull requests by using branches.

https://git-scm.com/book/en/v2/Git-Basics-Working-with-Remotes
https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging
https://git-scm.com/book/en/v2/Git-Branching-Remote-Branches

1. Create a branch on your fork
2. Switch to the new branch
3. Make the changes
4. Commit the changes
5. Push your branch to github

When you've done that, github will pop up a notification (you might need to go to the cinnamon repo or your fork for it to do that) that asks if you want to submit a pull to upstream.

(Also, you probably want to make sure your master is up to date with linuxmint/master before you create the branch so you don't have to worry about merge conflicts later.)

lestcape:
@collinss thanks, but I did not create a branch on indicator. The last part is the problem to me. How update the repo when i have different pending branches and i want create a new one?

pixunil:
@lestcape, you can use

```
git add remote upstream git@github.com:linuxmint/Cinnamon.git
git checkout upstream/master
git checkout -b newbranch
```

**********

Back to topic, I don't know what you want to achieve by updating it via gsettings, when I'm taking it correctly, it updates all xlets? Also those which aren't updated?

lestcape:
@pixunil when a change in gsettings occurs, then cinnamon (cjs), read this value and only process the value that have a change, so, only will be processed the difference 
between the old value and the current value.
There are two cases:
Case 1: 
self.settings.set_strv(("enabled-%ss") % (self.collection_type), not_update_extensions)
Case 2:
self.settings.set_strv(("enabled-%ss") % (self.collection_type), all_current_extensions)

In the case 1:
old value is: all_current_extensions
current value is:  not_update_extensions

So the difference:
(all_current_extensions - not_update_extensions) as a set difference. 
This difference is the extensions that need a restart. 

In the case 2:
old value is: (all_current_extensions - not_update_extensions)
current value is: all_current_extensions

Then, cjs will process the current difference again, this actually is:  
((all_current_extensions - not_update_extensions) - all_current_extensions) 
So is the same as:
(all_current_extensions - not_update_extensions) 
So again will be process only the extension that need a restart.

In both calls,  only will be processed the extension that need a restart, and will not touched any other extension. In the first called, process means unload the extensions as now are not present, and in the second, means load the extensions with an specific definition (the old one) because is added.

You will understand easy this after read: https://github.com/linuxmint/Cinnamon/blob/master/js/ui/extensionSystem.js#L62-L73

Please see:
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/extensionSystem.js#L66
The condition: enabledExtensions.indexOf(uuid) == -1 will satisfy the case 1, then only will be unload the  extensions that currently are not present...

Case 2 apparently is not satisfy here:
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/extensionSystem.js#L71
but it is satisfied here:
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/extension.js#L461

We need to be checked the gsettings next-applet-id or next-desklet-id values, is possible that we need to update this value, but i think is internally updated by cinnamon.

collinss:
I don't think setting the gsettings key is the way to do it. Not only does it feel hasckish to me, but it calls all of the same functions that @pixunil's method does, which means that the problem I mentioned above is still relevant. The proper solution IMHO is to implement a separate 'reload' function as I also mentioned above. This would have several advantages:
1. It would ensure that everything is done right.
2. It would (theoretically at least) be more efficient, as there is probably some code that wouldn't have to be run again
3. Such a function would enable applet developers to easily reload their code, without having to restart all of Cinnamon every time they made a minor change to their applet.

lestcape:
Same function for applets: https://github.com/linuxmint/Cinnamon/blob/master/js/ui/appletManager.js#L173-L215
Same function for desklets: https://github.com/linuxmint/Cinnamon/blob/master/js/ui/deskletManager.js#L168-L210

@collinss 
About what you say, yes is probably that what i say not preserved the configuration, as i don't take on account this. I have not tried.

lestcape:
@collinss 
1- "It would ensure that everything is done right", not really, as this is more complex than that, because involve two sides, and will be more easy maybe for cjs, but what is relate to the gsettings definitions, is not clear anyway... Also this not involve a properly call to destroy the desklets and applets, some signals to the cinnamon object will keeping connected. some of them with catastrophic result as for example the global menu...
2- What you propose, how is over Dbus, will involve be less efficient (practical at least), also a more complex mechanism and will require add more code.
3- "Such a function would enable applet developers to easily reload their code, without having to restart all of Cinnamon every time they made a minor change to their applet." This is now possible, what you say is with less code? I don't find the utility anyway.

lestcape:
What i propose will not remove the configuration for the non multi-instances extensions (applet, desklets), but will be removed for the multi-instances. I think, will be possible do what @pixunil do and in the change event send a parameter deleteConfig with false as is what occurs for only one instance (will be preserved the configuration).

lestcape:
I tested, this current proposition and what i say about not properly destroy the objects is not true, as this is done in prepareExtensionUnload. Apparently the definition (in gsettings) is not touch, so will not create problems.

Maybe, this line is not necessary: https://github.com/pixunil/Cinnamon/blob/b2263a25bc621705f885fa261f7c1254a74980aa/files/usr/lib/cinnamon-settings/bin/ExtensionCore.py#L901 is here the connection: https://github.com/pixunil/Cinnamon/blob/b2263a25bc621705f885fa261f7c1254a74980aa/files/usr/lib/cinnamon-settings/bin/ExtensionCore.py#L968

Both ways apparently will work ok...

collinss:
@lestcape ```prepareExtensionUnload``` is called by ```unloadExtension``` (unless the extension object doesn't exist), and ```unloadExtension``` is called by ```onEnabledAppletsChanged```, which in turn is called when the gsettings key is changed:

https://github.com/linuxmint/Cinnamon/blob/master/js/ui/appletManager.js#L51
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/appletManager.js#L187
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/extension.js#L510

Thus all multi-instance settings will be lost by this method.

lestcape:
@collinss in both methods, will be needed the same thing to not lost the settings of multi-instance (and can be done in both), just is a different way to do the same, but please see if this line is necessary: https://github.com/pixunil/Cinnamon/blob/b2263a25bc621705f885fa261f7c1254a74980aa/files/usr/lib/cinnamon-settings/bin/ExtensionCore.py#L901

collinss:
@lestcape 
https://github.com/linuxmint/Cinnamon/blob/master/files/usr/lib/cinnamon-settings/bin/ExtensionCore.py#L1001
It looks like it is needed because it gets disconnected every time the callback gets called...

lestcape:
That's, ok then, only is strange...

PR #4348, opened 2015-06-21
[2.8] alter mouse motion values to reflect whats in c-s-d

leigh123linux:
The defaults in cinnamon-settings-daemon are -1 for both these settings, also it make no sense to invert motion-threshold.


dalcde:
I think it makes sense to invert motion threshold, since it is labelled "sensitivity". High sensitivity would correspond to low threshold

leigh123linux:
Having it inverted makes it look out of place compared to the other sliders.

dalcde:
How?

dalcde:
ok I see. I think we should fix that by impoving our inverted implementation instead.

leigh123linux:
I agree.

PR #4349, opened 2015-06-21
[2.8] FHS fixes

Fantu:


clefebvre:
Hi Fantu, you should make separate PRs.. this already needs a rebase. Some commits are trivial, others arent..

Fantu:
all changes are need or partially are not working, separate them for each folder or file is a big waste of time
I taken all debian patches (already tested by many users that uses debian packages) and I added files rename instead actually workaround in debian .install files.
Git already see the file rename showing only effective code changes (see file changes in this PR)
When needed for test or before accept it at 2.8 develop. start I'll do rebase but I don't have time to do rebase any upstream commit pushed

Fantu:
Or do you think is better separe it in 2 PR where in the second the 2 commits files rename that require rebase any minimal changes?
The second can be done easy and faster updated doing commits where needed with fast mv or git mv

Fantu:
I did a git rebase without conflict to solves manually, strange that can't be the same with github.
I can't other PR with dependency on other FWIK, on this I keep all changes except the file rename that can cause high probable merge conflicts.
The other 2 commit can be done easy and fast with a "git mv" after apply this PR, are here:
https://github.com/Fantu/Cinnamon/commits/fhs-fixes2
Probably the best thing for avoid too many conflicts will be apply this PR and others in 2.8 start and after do the 2 commits of file rename.

Issue #4243, opened 2015-06-03
Panel edge and hot corner issue


bastimeyer:
Hi,
I'm not sure if this is a design choice or not, but the hot corners are not accessible if the cursor is hovering a panel and the pointer passthrough switch is disabled in the new panel settings (2.6.7). You have to move the cursor to the other edge of the screen first and then move it into the corner, then the hot corner can be triggered.
I've read that the panels have been reworked, but shouldn't this pointer passthrough setting only affect edges between two screens?

dalcde:
Not reproduced, but I might have mis-interpreted you. What do you mean by "the other edge"?

bastimeyer:
I'm sorry if my description wasn't clear enough...

Do this to reproduce the issue:
Enable all four hot corners and turn off the pointer pass through switch in the panel settings. Now try to trigger one of righthand corners - the corners on the left do not seem to be affected by this issue.
If the cursor is positioned at the top/bottom edge of the screen (if there is no panel set) or is hovering a panel first, you're unable to trigger the hot corner. You need to enable the pointer pass through setting to make this work.

> What do you mean by "the other edge"?

What I meant was, that if the cursor is at the right edge of the screen and is then being moved into the top or bottom corner, it will trigger regardless of the pointer pass through setting (even if the panel is always visible).

JosephMcc:
I can't reproduce this on my dual monitor setup. The pointer pass through setting is there for use with multiple monitors. There is no reason to disable it when using only one.

bastimeyer:
> The pointer pass through setting is there for use with multiple monitors

Who was talking about using only one monitor?

See this video I've just uploaded if you're unable to reproduce this issue...
https://youtu.be/latsHDXQY74 (no sound)

But it's actually a good point: Why is this option available if you disable all other monitors?

JosephMcc:
Ok, I was finally able to reproduce this. In my case I have the primary on the left with a panel at the top and a second monitor on the right. If I disable pointer pass through I can't trigger the hot corner if my mouse is on the top panel. I can trigger it if I move my cursor from the right monitor on to the panel on the left. It's almost like the barrier is off by 1px and stops you from hitting the hotspot.

PR #4358, opened 2015-06-23
Panel-barrier: move the right side panel barrier all the way to the m…

JosephMcc:
…onitor edge. The current placement makes hitting the right side hot corners difficult.
Closes: https://github.com/linuxmint/Cinnamon/issues/4243


dalcde:
I vaguely recall this causing panel edges to not work when we have adjacent panels, which is why they are made to be off by one pixel

JosephMcc:
@dalcde Trying out some different configurations, it seems you're right. The cursor does want to slip through in some cases.

PR #4361, opened 2015-06-24
[2.8] Calendar applet: Avoid unnecessary wakeups

phillipberndt:
The calendar applet currently updates once per second, even if a date format that does not include seconds is used. This patch changes this behaviour such that unless a format specifier that includes seconds is used, the timeout is instead set such that the update is performed at the next full minute (one second later actually, to account for possible inaccuracy).

### DIFF
--- a/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
@@ -114,7 +114,10 @@ MyApplet.prototype = {
     },
 
     on_settings_changed: function() {        
-        this._updateClockAndDate();
+        if (this._periodicTimeoutId){
+            Mainloop.source_remove(this._periodicTimeoutId);
+        }
+        this._updateClockAndDatePeriodic();
     },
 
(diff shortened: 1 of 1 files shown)


clefebvre:
This is brilliant. I worked on this so I really appreciate reducing CPU wakeups even more.

That said, I won't risk a regression that close to release. We might push that in 2.8 or as an update to 2.6, but let's give ourselves time (no pun intended) to run that for a while.

phillipberndt:
Sure thing. I'm running Mint's stable version of cinnamon, so I'll be using a local fork of the applet in the foreseeable feature anyway…

Hanmac:
thanks for breaking it, now on my system it does only update each minute, even if the clock does should show seconds.

https://github.com/linuxmint/Cinnamon/issues/4588

Vahan86:
In my opinion the way how the calendar applet is updated should be based on current settings. If you set in Date&Time settings to show seconds as well, then calendar applet should be updated once per second.

phillipberndt:
Of course it should be. This obviously is a bug, I'll look into it and push
a new pr..

PR #4397, opened 2015-07-05
Shortened the text a bit

lcafaro:


clefebvre:
Hi, 

Can you provide a screenshot showing why it's too large?

lcafaro:
Sure, here's a before/after.

Looks a bit more in proportion with the rest of the applet, in my opinion of course.

before
after

tomoe-musashi:
i don't agree this commit.
"until full" and "Discharging" are pretty weird in this case.
also, i think it's better to show actual words like "hours" and "minutes".
it's confuse if it just show a string...

clefebvre:
Hi,

I left this PR opened for a while. I have to admit I've been hesitating.. there's pros and cons.. it does make things more concise on the one hand, but also harder to translate and slightly less explicit on the other hand.

The width of the applet, although large, isn't problematic as such, at least not to the point where it can only be considered a bug.

In the end I decided to leave things as they were then, and I'm closing this PR. I'd like to thank @lcafaro for his work on this. This didn't make it in, but we do appreciate every bit of help we can get.

Issue #4245, opened 2015-06-03
2.6.7 - Regression ? Battery applet does not let you show time remaining


Myu-Unix:
On Cinnamon 2.4 series, the battery applet used to be able to display the time remaining, on 2.6.7, it doesn't seems possible anymore (option is not here).

Could this be a regression ? It was a convenient feature.

Cheers and thanks for your work on Cinnamon !

leoluk:
Yes, would be nice to get it back!

Vahan86:
I agree, it was really useful feature...

numian:
This was really useful, please put it back, or at least as an option.

In the meanwhile, I've changed a little bit the applet to get the time remaining.
It's here https://github.com/numian/cinnamon-power-applet if anyone need it too.
This version ignores the "show percentage" option.

Vahan86:
This issue is fixed in Cinnamon 3.0.7. 

![untitled](https://cloud.githubusercontent.com/assets/12819058/18698981/3cf96824-7fdf-11e6-9346-b747f9d055cb.png)

Issue #4329, opened 2015-06-17
Battery icon changes strangely with percentage?


lf-:
The battery icon in the power applet is definitely bugged. At 24%, it was showing a third full. Now at 23%, it's showing full:

![image](https://cloud.githubusercontent.com/assets/6652840/8212051/d9b30a0a-14d6-11e5-84f7-854f7a63b982.png)

Now it's showing something somewhat indicative of percentage:

![image](https://cloud.githubusercontent.com/assets/6652840/8212111/17a1c860-14d7-11e5-98ff-d36e20db4436.png)

Does this correspond with half percentage points? This is highly unintuitive behaviour.

Expected behaviour: battery icon "fullness" should correspond linearly with battery percentage.

rpbaptist:
I experience the same issue since upgrading Cinnamon to 2.6.11 along with Mint 17.2. (Coming from Mint 17.1)

sfilipov:
I experience a similar issue when charging. Every 2-3 percent change while charging, the icon will add or remove the little lightning sign even though the laptop is plugged in and charging all the time. It is confusing whether the laptop is plugged in or not. Clicking the applet will show the true information as shown in the screenshot below.

![power_applet2](https://cloud.githubusercontent.com/assets/2607726/8633477/1c9e54dc-27d3-11e5-9a92-75e8a50611d6.png)

pixunil:
@sfilipov, Yep, that's true - for a little explanation: after each update (cable (un-) plugged, charge change) the icon for the status is extracted and if the icon is the same than before, it sets to the full one. It's also fixed.

rpbaptist:
I locally merged both the pull requests which references this, but found neither of them worked.

Which version was used in Mint 17.1? I will have to rever the power applet back to that one.

lf-:
I don't know. I thought (90% sure) it was fixed when I reverted the applet but I'm not reinstalling a Linux version with Cinnamon on my laptop at this time. (this decision has nothing to do with this bug. I switched to bspwm ;P)

rpbaptist:
I copied the power applet from Cinnamon 2.4.0 and it's working as I expected it to be. I believe the problem has been introduced in 2.4.1 with these two commits:

https://github.com/linuxmint/Cinnamon/commit/bad763dfd3363a31a336fea8cf0a022d97d9c294
https://github.com/linuxmint/Cinnamon/commit/035e2fca3a2570300d7592ec22ec3e2085c4e128

[...]

PR #4419, opened 2015-07-08
Restore time remaining option for power applet.

mpetroff:
Cinnamon 2.4 had an option for displaying the estimated battery time remaining in the power applet, but the feature was removed in Cinnamon 2.6. This pull request fixes that regression, fixing #4245.

### DIFF
--- a/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
@@ -374,18 +374,18 @@ MyApplet.prototype = {
                                 let time = Math.round(seconds / 60);
                                 let minutes = Math.floor(time % 60);
                                 let hours = Math.floor(time / 60);
-                                labelText = _("time of battery remaining", "%d:%02d").format(hours,minutes);
+                                labelText = C_("time of battery remaining", "%d:%02d").format(hours,minutes);
                             }
                             else if (this.labelinfo == "percentage" ||
                                     (this.labelinfo == "percentage_time" && seconds == 0)) {
-                                labelText = _("percent of battery remaining", "%d%%").format(Math.round(percentage));
+                                labelText = C_("percent of battery remaining", "%d%%").format(Math.round(percentage));
                             }
                             else if (this.labelinfo == "percentage_time") {
                                 let time = Math.round(seconds / 60);
                                 let minutes = Math.floor(time % 60);
                                 let hours = Math.floor(time / 60);
-                                labelText = _("percent of battery remaining", "%d%%").format(Math.round(percentage)) + " (" +
-                                            _("time of battery remaining", "%d:%02d").format(hours,minutes) + ")";
+                                labelText = C_("percent of battery remaining", "%d%%").format(Math.round(percentage)) + " (" +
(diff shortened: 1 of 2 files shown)


muzena:
@mpetroff 
I think this patch is in conflict with this https://github.com/linuxmint/Cinnamon/pull/4427

mpetroff:
This modifies the code immediately before what #4427 modifies.

pixunil:
@muzena, yes they're conflicting (it buts my changes in the middle of @mpetroff's) - but the conflict can only be resolved when one of the PRs is merged.

files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
```diff
@@ -366,12 +366,29 @@ MyApplet.prototype = {
...
+                            if (this.labelinfo == "nothing") {
+                                ;
                             }
-                            else {
-                                this.set_applet_label("");
+                            else if (this.labelinfo == "time" && seconds != 0) {
+                                let time = Math.round(seconds / 60);
+                                let minutes = Math.floor(time % 60);
+                                let hours = Math.floor(time / 60);
+                                labelText = _("time of battery remaining", "%d:%02d").format(hours,minutes);
```
pixunil: You're wrong with the use of `_(msgid)`, you want to use `C_(msgctx, msgid)` (it point to `Gettext.pgettext`)

mpetroff: Thanks. I originally had `C_` but then saw it only had one other use. I didn't notice the `p` when I glanced at `js/ui/environment.js`, so I thought they were the same function and figured the `C_` form must have been deprecated, so I changed it to `_` (and apparently didn't test it again after making the change).

muzena:
@pixunil I don't understand, I compile and install you repository.
I have wirelles mouse Logitech G700s.
I make patch to upower for G700s/G700: http://cgit.freedesktop.org/upower/commit/?id=305f62adf052aa972523d083ca44d3050f659ec9

Now I look this bug https://github.com/linuxmint/Cinnamon/issues/4329
When I check in applet options Show percentage
In tray icon is only icon without percentage.
Is this work only on laptop battery, or must work on other devices such as a mouse.

pixunil:
@muzena, this isn't a conflict, it is the fault of this [if clause](https://github.com/linuxmint/Cinnamon/pull/4419/files#diff-132ff4b31f6edff30eb6b243f205a330L365), as it only processes Laptop batteries.
Don't know whether this is indented or not.

btw: you can also just swap the files instead of compiling

rpbaptist:
This is actually unrelated to the referenced issue: https://github.com/linuxmint/Cinnamon/issues/4329

mpetroff:
I think I successfully merged in the recent changes to `master`. However, the substantial changes in 2c5b9d5aa3381a1c2e31bdf5f1ee11a7e3574179 break compatibility with the stable 2.6 branch that I'm running, so the post-merge version is untested. As #4427 has been merged, there is no longer a conflict with it.

clefebvre:
Thanks for rebasing this. Successfully tested and about to merge.

PR #4427, opened 2015-07-10
Power applet: Fixed a logical bug

### LINKED ISSUE #4329: Battery icon changes strangely with percentage?

pixunil:
What was wrong:
When the primary device does not change its icon string after an update, `battery-full` will be used.

Fixes #4329


rpbaptist:
I merged this locally but did not find it changed the behaviour.

PR #4432, opened 2015-07-11
popupMenu: Don't always ignore Submenu children in width requests. St…

JosephMcc:
…ops the menu size from jumping around when opening submenus.

### DIFF
--- a/js/ui/popupMenu.js
+++ b/js/ui/popupMenu.js
@@ -1208,7 +1208,8 @@ PopupMenuBase.prototype = {
         let columnWidths = [];
         let items = this.box.get_children();
         for (let i = 0; i < items.length; i++) {
-            if (!items[i].visible)
+            if (!items[i].visible &&
+                !(items[i]._delegate instanceof PopupSubMenu && items[i-1].visible))
                 continue;
             if (items[i]._delegate instanceof PopupBaseMenuItem || items[i]._delegate instanceof PopupMenuBase) {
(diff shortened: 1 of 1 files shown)


clefebvre:
Hmm... I don't know, there's pros and cons on that one.

Pros: The size remains consistent
Cons: Well... we could end up with a menu that is much larger than what it contains (and until the user actually expands the revealer(s), he/she wouldn't necessarily have any visual clues as to why the menu is so large).

If a menu was much larger than its visual content, that could also be considered a bug, couldn't it? So I don't know.. I'm not really convinced by this PR :)

JosephMcc:
The jumping around of the menu size just looks bad and I'm not sure the con is really a con. Currently in the situation your talking about where a submenu item is much wider than a normal menu item, opening the submenu can actually push the sides of the menu completely off the screen. Thats even worse.

pixunil:
Just throwing two ideas in this discussion: I'm fine with changing sizes (`cinnamon-settings` does it also), but I think using a tween could make it feeling more correct.
Another thing is applet design, for instance, the sound applet has got those launchers, but there is no reason I see to have the icon taking a second column.

JosephMcc:
As I pointed out before the menu resizing is an issue when it can push the the menu content completely off the screen. Having a tween to make it smoother doesn't help that. And like I said before, imho, it just looks poor and unpolished.

clefebvre:
Closing this one. We should take the width into consideration in the tween animation instead.

PR #4434, opened 2015-07-11
Cleanup: instead of attempting to clean up deprecated functionality i…

JosephMcc:
…n cinnamon-arrow, just remove it since it seems to be unused.


dalcde:
Again no external xlets use this

PR #4436, opened 2015-07-12
Cleanup: Remove st-tooltip. It's the source of a lot of build warning…

JosephMcc:
…s and we no longer use it.


dalcde:
Currently no xlets on cinnamon-spcies use st-tooltip. So lgtm

Issue #4411, opened 2015-07-07
Cinnamon-preload preloading for all users in LDAP


Craigacp:
I'm using Cinnamon 2.6 on a machine using an LDAP server for user authentication & automounted home directories. The cinnamon-preload daemon is iterating through the users in the LDAP directory and trying to preload information for them (which causes their home directory to be mounted). Is there an option to turn off this behaviour as it makes cinnamon unusable in my environment. Also, defaulting to su-ing from root to a user in a network environment is unfriendly.

I'm running Ubuntu 15.04 and the systemd journal is full of lines like:
Jul 07 08:11:12 `<hostname>` su[11353]: Successful su for `<username>` by root
Jul 07 08:11:12 `<hostname>` su[11353]: + ??? root:`<username>`
Jul 07 08:11:12 `<hostname>` su[11353]: pam_unix(su:session): session opened for user `<username>` by (uid=0)
Jul 07 08:11:12 `<hostname>` cinnamon[839]: gdk_mir_display_open
Jul 07 08:11:12 `<hostname>` cinnamon[839]: Failed to connect to Mir: Failed to connect to server socket: Permission denied
Jul 07 08:11:12 `<hostname>` cinnamon[839]: Unable to init server: Could not connect: Connection refused
Jul 07 08:11:12 `<hostname>` cinnamon[839]: (cinnamon-preload:11367): dconf-CRITICAL **: unable to create directory '/run/user/0/dconf': Permission denied.  dconf will not work properly.
Jul 07 08:11:12 `<hostname>` cinnamon-preload[11367]: Loading theme gnome for user ID `<userid>`
Jul 07 08:11:13 `<hostname>` cinnamon-preload[11367]: Loading gnome took 0.77 seconds
Jul 07 08:11:13 `<hostname>` su[11353]: pam_unix(su:session): session closed for user `<username>`

mtwebster:
I'm doing some research on ldap ... do you guys know of a sure way to detect ldap is being used - would save me some trouble.

Thanks

dalcde:
I don't think detecting ldap would be an effective solution. It is also possible that a computer has many non-ldap accounts, and preloading all of them would still take quite a lot of time.

Craigacp:
[...]

PR #4462, opened 2015-07-22
Only get local users

Aviator45003:
Avoid hitting the network if LDAP users present by only looking at users with home folders in `/home`.

This is a problem on one network I am aware of which has about 800 users via LDAP, and is additionally the source of #4411.

This has the extra bonus of not including system users.

### DIFF
--- a/files/usr/bin/cinnamon-preload
+++ b/files/usr/bin/cinnamon-preload
@@ -26,7 +26,7 @@ if __name__ == "__main__":
     if os.getuid() == 0:
         start = time.time()
         has_encrypted_home = False
-        for p in pwd.getpwall():
+        for p in [pwd.getpwnam(name) for name in os.listdir('/home')]:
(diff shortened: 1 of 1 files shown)


Craigacp:
In my LDAP environment all local users mount their home directory in /export/home or /local/home. Network users have their home directories auto-mounted in /home, and on my machine the os.listdir check does return my network mounted home directory and not the local users directories in /export/home.

Aviator45003:
This still solves the problem of every single user being polled. In this case it means your local users don't actually get their themes pre-loaded, but that's a matter of where your setup put home directories.

glebihan:
Listing the contents of /home is not a reliable way to list the users (additional files or directories could very well exist there).
If we're going to filter on the home folder being inside /home, I think the original loop on pwd.getpwall() should be kept, with an additional test.

PR #4463, opened 2015-07-23
settings.js: allow for dynamic getting and setting of options for sup…

collinss:
…ported key types (eg comboboxes and radio groups)

This will allow applets to add more options to a combobox after installation, without having to modify the ```settings-schema.json``` file. For example, say you want to to select between options that vary from system to system, such as software, etc.

It will also make it easier for applet developers as they can just pull options from here rather than maintaining a duplicate list on the js side and having to check it every time the list is updated to ensure compatibility.


lestcape:
+1

PR #4465, opened 2015-07-23
Make more cursor types available to Cinnamon

collinss:
Adds 
  CINNAMON_CURSOR_RESIZE_BOTTOM,
  CINNAMON_CURSOR_RESIZE_TOP,
  CINNAMON_CURSOR_RESIZE_BOTTOM_RIGHT,
  CINNAMON_CURSOR_RESIZE_BOTTOM_LEFT,
  CINNAMON_CURSOR_RESIZE_TOP_RIGHT,
  CINNAMON_CURSOR_RESIZE_TOP_LEFT,
  CINNAMON_CURSOR_CROSSHAIR,
  CINNAMON_CURSOR_TEXT


clefebvre:
Hi,

Can you describe what it fixes, how to test/verify it?

collinss:
It's more a feature than a fix I suppose. For my sticky notes applet, for example, I would like to be able to at least use the text one to make it feel more natural when hovering over the text, and the resize ones to indicate when the note can be resized. Neither of these is currently available to Cinnamon using ```global.set_cursor```, and (as far as I can tell at least) there is no other way to change the mouse cursor in javascript, as Gdk cursor functionality seems to be very lacking in terms of introspection.

It seems to me that these could also be useful to other applet developers in the future as well. I know for example, that the sticky notes Desklet and Configurable menu also have resizing implemented, but it seems awkward with only the drag related icons available. The crosshair cursor would be useful in, for example, the color chooser applet or my devtools desklet.

lestcape:
+1 I agree, that this is helpfully, there are another way to change the cursor, and this is hide the real cursor and then add an icon as a cursor, but require more things, and is a fake. See for example: https://github.com/lestcape/Color-Chooser/blob/master/colorChooser%40lestcape/colorChooser.js#L1186

PR #4473, opened 2015-07-28
removable-drives applet: don't hardcode Nemo

City-busz:
Use Gio to open the default file manager at 'computer:///'.


clefebvre:
That could trigger the launch of Caja, Nautilus etc.. on systems where other users installed/use other DEs. I'd rather leave that hardcoded to Nemo, or add the file manager as yet another option in "favorite applications".

City-busz:
But if you open any places from the main menu, it opened with the default file manager anyway. So if you hardcode nemo here, it makes it inconsistent.

The correct way to set the default file manager is to add inode/directory= line to .config/mimeapps.list

PR #4505, opened 2015-08-12
Fix #3223 - Update applet.js delay resize of bugged icons

### LINKED ISSUE #3223: Pidgin status icon.
Pidgin status icon is not showing even if in the settings it’s set to always show. Workaround: open Settings - set "never show" and "show always" again.
Using: Mint 17 cinnamon (32bit).

brouillon:
Add shutter in bugged icons list and add a delay before applying resize.
Fix for  #3223

### DIFF
--- a/files/usr/share/cinnamon/applets/systray@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/systray@cinnamon.org/applet.js
@@ -118,7 +118,7 @@ MyApplet.prototype = {
 
             // Don't try to scale buggy icons, give them predefined sizes
             // This, in the case of pidgin, fixes the icon being cropped in the systray
-            if (["pidgin", "thunderbird"].indexOf(role) != -1) {
+            if (["pidgin", "thunderbird", "shutter"].indexOf(role) != -1) {
                 if (disp_size < 22) {
                     size = 16;
                 }
@@ -132,16 +132,30 @@ MyApplet.prototype = {
                     size = 48;
                 }
             }
-
-            icon.set_size(size, size);
-
-            global.log("Resized " + role + " (" + icon.get_width() + "x" + icon.get_height() + "px)");
+            let timerId = 0;
+            let i = 0;
+            timerId = Mainloop.timeout_add(500, Lang.bind(this, function() { 
+                icon.set_size(size, size);
+                global.log("Resized " + role + " (" + icon.get_width() + "x" + icon.get_height() + "px)");
+                i++;
+                if (i == 2) {
+                    Mainloop.source_remove(timerId);
+                }
+            }));
         }
         else {
             // Force buggy icon size when not in scale mode
-            if (["pidgin", "thunderbird"].indexOf(role) != -1) {
-                icon.set_size(16, 16);
-                global.log("Resized " + role + " (" + icon.get_width() + "x" + icon.get_height() + "px)");
+            if (["pidgin", "thunderbird", "shutter"].indexOf(role) != -1) {
+                let timerId = 0;
+                let i = 0;
+                timerId = Mainloop.timeout_add(500, Lang.bind(this, function() { 
+                    icon.set_size(16, 16);
+                    global.log("Resized " + role + " (" + icon.get_width() + "x" + icon.get_height() + "px)");
+                    i++;
+                    if (i == 2) {
+                        Mainloop.source_remove(timerId);
+                    }
+                }));
             }
         }
     },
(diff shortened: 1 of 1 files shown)


Emmanux:
Can't believe this bug is still alive. If there are workarounds, there is no reason to not apply them. This bug is CRITICAL, it is nowadays the only show-stopper for mass production deployments.

clefebvre:
Hi brouillon,

Can you explain the results of your analysis/troubleshooting, what you think the cause of the bug is and how the timer helps to fix it?

Also, does this fix all cases? or just makes the issue happen less often?

brouillon:
Hi clefebvre,

The main troubleshooting, is blank space in systray at first launch of shutter or pidgin.

If you inspect the icon object when it's a blank space in the cinnamon debugger and change it's size with eg. ```r(0).set_height(20);``` and ```r(0).set_height(16);``` the icon is displayed.

So my workaround was just playing with this fact, delaying two resize operation on this icon, display it.
With that, It's always fixed for these apps, but there is an Edge effect : on second launch (close/open), you will see the two delayed resize operation 20px and 16px.

I don't know the real problem, it's seems these icons objects are not fully loaded/ready when we try to add it to systray. All case aren't fixed because i don't understand why some icon fails displaying, maybe the case was these apps use icon that are not in cinnamon theme ? 
Is there a way to inspect deeply the icon object, where can we find documentation about icon object ?

It's just a dirty workaround, but this bug was so annoying, for my use i accept it, but i hope for a better solution.

Have you experienced this issue ?

Sorry for my english, i can be more clear in french ;)

Emmanux:
My workaround is to launch pidgin, and then, after few seconds, call it again from Alt+F2.Another (uglier) workaround is to launch Skype first, then after few seconds launch pidgin, quit Skype (the icon of pidgin appears), and launch it again (if I want Skype running). 16.08.2015, 10:15, "brouillon" <notifications@github.com>:Hi clefebvre,

The main troubleshooting, is blank space in systray at first launch of shutter or pidgin.

If you inspect the icon object when it's a blank space in the cinnamon debugger and change it's size with eg. r(0).set_height(20); and r(0).set_height(16); the icon is displayed.

So my workaround was just playing with this fact, delaying two resize operation on this icon, display it.
With that, It's always fixed for these apps, but there is an Edge effect : on second launch (close/open), you will see the two delayed resize operation 20px and 16px.

I don't know the real problem, it's seems these icons objects are not fully loaded/ready when we try to add it to systray. All case aren't fixed because i don't understand why some icon fails displaying, maybe the case was these apps use icon that are not in cinnamon theme ? 
Is there a way to inspect deeply the icon object, where can we find documentation about icon object ?

It's just a dirty workaround, but this bug was so annoying, for my use i accept it, but i hope for a better solution.

Have you experienced this issue ?

Sorry for my english, i can be more clear in french ;)

—Reply to this email directly or view it on GitHub.


  
  



Best regards,Emmanuel Arias

brouillon:
@clefebvre 
Have you experienced this issue ? 
Have you tried the workaround ? and do you agree as is ?
Have you some documentation or a better way to tackle this bug ?

clefebvre:
Hi,

I've got good news. I found a reliable way to reproduce the shutter issue, simply by clearing the disk cache:

echo 3 | sudo tee /proc/sys/vm/drop_caches

That result in the first launch of shutter to display a transparent icon (no need to restart the computer or cinnamon anymore).

Your PR seems to fix that but the resulting icon isn't fitting properly (its dimensions are wrong).

Also, it doesn't fix the pidgin issue.

So I'm still looking into that.

brouillon:
@clefebvre yeah good news !
I remove this fix in my version cinnamon, so actually with cinnamon vanilla (2.6.13) : 

 - shutter never display a transparent icon, just a little bit smaller and not centred at the first launch
 - pidgin always display a transparent icon at first launch

Yes, clearing the cache make the icon respond as first launch.

I hope you will catch this !

clefebvre:
Ok good news again. I've rewritten all that to work differently. It delays insertion instead of doing some black magic with multiple resizes/timeouts.

Here's the commit:

https://github.com/linuxmint/Cinnamon/commit/f2134a4225f82a82ba57bd370b2a51312b731d9a

It fixes the following cases:

- pidgin not showing up on empty disk cache (i.e. first launch after reboot for instance)
- shutter not showing up on empty disk cache

It also probably fixes icons which were out of place or badly resized.

It has the following limitations:

- shutter and filezilla always show in their original size (they don't scale). Scaling is disabled for them as they are buggy when scaled.

- dropbox won't scale, although we do scale it. That's a bug on dropbox side, we can't fix that one.

PR #4531, opened 2015-08-19
windowManager: Destroy the tile hud/preview when we are done with it.…

JosephMcc:
… Currently after showing the preview you will get a noticeable slowdown when moving windows that pass off the monitor edges. This should help solve that.


RavetcoFX:
I was able to reproduce the slowdown, this fixed it

PR #4537, opened 2015-08-21
Better xlet management (uniqueness/type, easier imports, autoreload)

pixunil:
Depends on linuxmint/cjs#26
Extends / Remake of #4338 

#### Uniqueness of uuids
Before this, there were problems when two xlets with the same uuid, but different xlet types, should be added / removed, as the code in `extension.js` saves the objects with the uuid, not the type.
It solves this by adding a `maps` object to every `Extension.Type`, the members `meta`, `importObjects` and `objects` which replaces the maps on the top `Extension` layer (see the Possible known regressions section).
#### Auto-reload of xlets when updating them via `cinnamon-settings`
Replaces the dialog with an automated action.
#### Better reload of xlets with Melange
Doesn't remove config files
#### Easier imports for files in xlet directory
Personally, for me an important change.
```
imports.applet.*
imports.desklet.*
imports.extension.*
imports.search_provider.*
```
Like the normal uses of `imports.*`, which includes files under `/usr/share/cinnamon/js/*` or cjs modules, is `imports.xletType.*` (replace `xletType` like above) used for imports in the directory where the xlet lives.
There were two solutions which existed, but they had side-effects:
- Getting `AppletManager.applets[uuid]` - It worked, but _every_ file which wants to make an import has to save the uuid and to request `AppletManager.applets[uuid]`, which is a bit complicated
- Modifying `imports.searchPath` - I used this method as it is nice and easy, but it is **dangerous**. cjs does "seal" the imported objects, resulting that a xlet can spoil properties on `imports`, so it does work for one or two xlets, but if two are using the same name, it will fail.

With `imports.xletType`, it is possible to easily import files in the directory without hardcoding the uuid, doing extra calls nor side effects, as long (and this is the only restriction so far), no import is done in runtime (e.g. as another applet could take `imports.applet`)

##### `Extension.unloadExtension`
Does require now two arguments, `uuid` and `type` (due uniqueness).
##### `Extension.reloadExtension`
New.
##### Code lifting in `extension.js`
Macro function to create `Extension.Type` members.
##### Remove `cinnamon_js_add_extension_importer`
Replaced by `imports.addSubImporter(name, path)`, see cjs PR.

## Possible known regressions
3rd party could fail, in this situations:

Fails | Should be used instead
--------- | ---------
`Extension.removeExtension(uuid)` | `type` argument is needed
`Extension.meta`, `Extension.importObjects` or `Extension.objects` | `Extension.Type[type].maps[mapType]`


lestcape:
This will be a big deal to all my current code, but +1 +1 +1.

clefebvre:
@pixunil please come and chat with us on #linuxmint-dev (irc.spotchat.org) when you get a chance. I'd like to get a clearer picture of the potential regressions involved.

bluedxca93:
Stark menu and tycho places not working on 2.6.13 mint and 2.8 propably cause of the internal changes in cinnamon.

PR #4559, opened 2015-08-29
popupMenu: Add back the expander arrow on PopupSubMenuMenuItem but wi…

JosephMcc:
…th a different layout. Also tweak the layout of a couple of our applet menus to look nicer with this change. The addition of the icons to Cinnamon itself is a workaround for Mint. Newer versions of the gnome icon theme have these as built in icons but not in the version supplied with Mint 17. They can be removed in the future.

### DIFF
--- a/data/icons/Makefile.am
+++ b/data/icons/Makefile.am
@@ -5,6 +5,12 @@ public_icons_themes = \
 	$(NULL)
 
 public_icons = \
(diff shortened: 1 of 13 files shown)


JosephMcc:
Adding a screenshot: http://imgur.com/79GJBYi

PR #4560, opened 2015-08-29
[Glib connect] overrides.js: override GSettings connect function since this doesn't actually connect

dalcde:


dalcde:
Closing since GLib reverted to sanity

PR #4579, opened 2015-09-03
Window list window preview

dalcde:
The theme for preview is hardcoded. Probably want to improve it later. \*hint\* \*hint\*
thing


clefebvre:
To be merged, pending:

- removal of last commit (use a simple checkbox in settings)
- addition of theming (can use WindowListGroup spice as an example)

Cobinja:
Why don't you use the cloning mechanism in js/misc/WindowUtils.js ? That way you can easily show transient child windows as well (which may not have an own window list entry)

PR #4590, opened 2015-09-04
Convert keyboard short name to upper case.

KEIII:
Hello. I think keyboard short name is more readable in upper case.


clefebvre:
Hi;

I know this is very subjective. I disagree that this looks better though.

PR #4606, opened 2015-09-08
[3.0] Theme revamp 2

JosephMcc:
A first attempt at converting some theme elements to generic style classes. The first commit is just rearranging, cleaning up whitespace, and removing some unused style classes. The second commit adds some generic button styling and converts some of our internal elements to use it.


JosephMcc:
Going to close this for now. I think there is a better way to approach this so it has less effect on existing themes.

PR #4626, opened 2015-09-17
[3.0] Signal manager: Timeouts and DBus

pixunil:
Adds following features to `SignalManager.SignalManager`:

- `SignalManager.addTimeout(name, interval, callback)`
 * `interval` is a number → milliseconds
 * `interval` is `null` → idle timeout
- `SignalManager.removeTimeout(name)`
- for `SignalManager.connect`, `obj` can also be a `Gio.DBusProxy`

I didn't convert all the old code, to have more security regarding regressions.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -14,6 +14,7 @@ const Gio = imports.gi.Gio;
 const Signals = imports.signals;
 const GnomeSession = imports.misc.gnomeSession;
 const ScreenSaver = imports.misc.screenSaver;
+const SignalManager = imports.misc.signalManager;
 const FileUtils = imports.misc.fileUtils;
 const Util = imports.misc.util;
 const Tweener = imports.ui.tweener;
@@ -1066,6 +1067,8 @@ MyApplet.prototype = {
         this.menu = new Applet.AppletPopupMenu(this, orientation);
         this.menuManager.addMenu(this.menu);
 
+        this.signals = new SignalManager.SignalManager(this);
+
         this.actor.connect('key-press-event', Lang.bind(this, this._onSourceKeyPress));
 
         this.settings = new Settings.AppletSettings(this, "menu@cinnamon.org", instance_id);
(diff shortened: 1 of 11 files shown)


dalcde:
I think this will conflict horribly with #4612, which simplifies a lot of the code in signalManager. Might be worth doing this on top of that commit instead.

pixunil:
Okay, when #4612 gets merged, I'll rebase again.

clefebvre:
That's not tested enough. I didn't review the code but found two regressions:

- The sound applet no longer reacts to the launch or death of an MPRIS media player.
- Window list thumbs seem to get stuck when hovering fast over many window list items and messing with Alt-Tab at the same time (probably a performance loss in signal manager).

I found a missing } which crashed Cinnamon in the other PR (Extension work) so I'm worried about the lack of testing here. It's a pity cause you're quite active on Cinnamon and these are really good ideas you're implementing. If you were more present on IRC and we could get to the point where I can trust your code blindly we'd do a lot more together.

It's still a really good PR but as it is I can't merge it in just yet. These two regressions need to be addressed and from a more general point of view, I'm getting really worried about potential regressions.

We've another day until feature-freeze. We can also postpone this to 3.0.

dalcde:
I think we can separate out the api part and get that merged, but not actually use it for anything yet. That would be easier to audit and it would be rather helpful to 3rd party xlet developers.

clefebvre:
Every time I add the sound applet to the panel, I get this:

(cinnamon:21219): Cjs-WARNING **: JS ERROR: TypeError: obj.connect is not a function
SignalManager.prototype._connectSignal@/usr/share/cinnamon/js/misc/signalManager.js:126
SignalManager.prototype.connect@/usr/share/cinnamon/js/misc/signalManager.js:116
MyApplet.prototype._init/<@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:910
_makeProxyWrapper/</<@resource:///org/gnome/gjs/modules/overrides/Gio.js:203

The sound applet successfully shows up, but I guess its signals are messed up. After that I start noticing the weird issue I mentioned earlier with the window list thumbs, I guess the signalManager is confused then.

I'm just looking at it from a black box tester point of view.

This WIP is promising but we're too close to 2.8 so I'm postponing it to 3.0.

Issue #4472, opened 2015-07-27
[Menu applet] Icons with same path/filename not reloaded


martinrotter:
*How to reproduce:*
1. Open menu applet settings and select some image with path *x/y/z.png*. Menu now shows the image. Close settings.
2. Overwrite file *x/y/z.png* with another picture.
3. Open menu applet settings again and again navigate and select *x/y/z.png* as an image.

*Expected result:* After 3, new image (although from the same path/filename) is applied.
*Real result:* Nothing happens after 3, new image (with the same path/filename) is not selected.

It seems that if there is image from the same path/filename cached, then cache is not cleared and that image from cache is used, despite the path/filename may now in fact contain new image.

pixunil:
Regression should be caused by commit https://github.com/linuxmint/Cinnamon/commit/51942292a87b585d76402f6d0b6b1bd8d2081d9a. Guess it was performance vs feature.

martinrotter:
@pixunil It may be related to tha commit. But, when I explicitly select new icon through "Select Icon" dialog, I would expect icon to change itself. This is 100% bug. Performance optimizations should not introduce bugs.

JosephMcc:
If this is a good feature vs performance design choice is debatable. However, since it behaves as the code intends for it to behave it is 100% NOT a bug.

martinrotter:
Bug = A software bug is an error, flaw, failure, or fault in a computer program or system that causes it to produce an incorrect or unexpected result.

Thus, this IS a bug. Intentional bug.

martinrotter:
:)

PR #4630, opened 2015-09-19
Add texture monitors as an option for theme developers

### LINKED ISSUE #4610: Cinnamon restart required on image edit
It would be pretty awesome for themers if this nuisance could be solved.

When editing any images in cinnamon themes, you have to restart cinnamon to see changes.

This is quite annoying as I need to make hundreds of little tweaks on particular assets sometimes. Under gnome-shell 3.16 only a theme restart is required to update an image. I even made a gulpfile which uses this for autoreloading on changes (absolutely awesome to work with!):
https://github.com/zagortenay333/ciliora-tertia-shell/blob/dd4f75d127412d2fa6bc91ec5f704ac6ce26e127/Ciliora-Tertia/gnome-shell/gulpfile.js#L39-L43

rjanja:
Cinnamon 2.6.0 disabled the filesystem monitoring of assets used by textures for a performance bump (51942292a87b585d76402f6d0b6b1bd8d2081d9a). While that is a good idea for most users, it makes theme development much more of a challenge. This change set makes it an option in Cinnamon Settings "General" tab and defaults it to off.

Would resolve #4610 and #4472.

screenshot-area-2015-09-19-132804

### DIFF
--- a/data/org.cinnamon.gschema.xml.in
+++ b/data/org.cinnamon.gschema.xml.in
@@ -424,6 +424,14 @@
       </_description>
     </key>
 
+    <key name="enable-texture-monitors" type="b">
+      <default>false</default>
+      <_summary>Enable or disable watching textures for changes</_summary>
+      <_description>
+        Whether image files are monitored for changes in order to reload or not
+      </_description>
+    </key>
+
     <key name="workspace-expo-view-as-grid" type="b">
       <default>false</default>
       <summary>Display the Expo view as a grid</summary>
--- a/files/usr/lib/cinnamon-settings/modules/cs_general.py
+++ b/files/usr/lib/cinnamon-settings/modules/cs_general.py
@@ -52,3 +52,7 @@ class Module:
 
             switch = GSettingsSwitch(_("Log LookingGlass output to ~/.cinnamon/glass.log (Requires Cinnamon restart)"), "org.cinnamon", "enable-looking-glass-logs")
             settings.add_row(switch)
+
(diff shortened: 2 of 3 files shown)


rjanja:
This option could also go into the "Themes" settings window, with a simplified label, e.g. "Reload images when changed on disk". But it affects cache of all texture images, not just those specified in themes.

JosephMcc:
I'm not sure how others feel about this but I don't really know if this needs to be an option. Most users aren't  going to have a clue what that is. If there isn't an obvious/noticeable performance improvement happening here it should probably just be reverted.

rjanja:
I'd love to see some performance stats including number of textures loaded, but I wouldn't know where to begin - maybe @clefebvre or @glebihan can chime in there. But I agree with you, if there's not a noticeable improvement between using monitors and not, the original change could simply be reverted as the most straightforward option. Otherwise, it may be beneficial to move such an option to a "Developer" menu of sorts and document it for theme developers on the wiki/forums etc.

JosephMcc:
Best test: just revert and run cinnamon that way for awhile. The two places I would think you would be most likely to notice a difference really would be startup or the cinnamon menu. Or does it make a noticeable difference in cpu or memory usage?

If the difference is so minute that we have to create some test to capture a couple milliseconds, I'm not sure it's worth the regressions.

zagortenay333:
It does really look a little bit weird as an option in the system settings.

clefebvre:
It was part of a much bigger effort to reduce performance. Overall it worked.. but whether that particular change gave us significant results or not... I can't remember to be honest.

We need to reassess that, measure it, see how often things get called etc.

Once we know more, we can then decide if we want to revert the change or not.

Either way, making this configurable is not a good idea imo.

PR #4631, opened 2015-09-20
Show additional audio device description in sound applet

rjanja:
When there are many devices present, the current list isn't helpful. Example:

screenshot-cinnamon-2015-09-19-203908

This minimal change adds a trailing label with the second line description so devices can more easily be identified. Example:

screenshot-cinnamon-2015-09-19-202819

**Note**: the style applied on the right is `menuitem-detail`, inherited from the desktop-capture applet. Cinnamon *should* have a similar style, and when that style is selected and implemented this pull request can be updated and considered.

### DIFF
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -1420,6 +1420,11 @@ MyApplet.prototype = {
             this._control["change_" + type](device);
         });
 
+        let bin = new St.Bin({ x_align: St.Align.END, style_class: 'menuitem-detail' });
+        let label = new St.Label({ text: device.origin });
+        bin.add_actor(label);
+        item.addActor(bin, { expand: false, span: 1, align: St.Align.END });
+
         let selectItem = this["_select" + type[0].toUpperCase() + type.slice(1) + "DeviceItem"];
         selectItem.menu.addMenuItem(item);
         //show the menu if we have more than two devices


files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
```diff
@@ -1420,6 +1420,11 @@ MyApplet.prototype = {
             this._control["change_" + type](device);
         });
 
+        let bin = new St.Bin({ x_align: St.Align.END, style_class: 'menuitem-detail' });
```
pixunil: I believe that `.popup-inactive-menu-item` is a built-in cinnamon style class which should have the same style.

pixunil: I believe that `.popup-inactive-menu-item` is a built-in cinnamon style class which should have the same style.

rjanja: Thanks for the feedback. I had a look at that style, but I think an inactive menu item should be able to be styled differently than context for an active menu item.

JosephMcc: @rjanja I'm with @pixunil on this on this one. 'popup-inactive-menu-item' is the same style used for the status label on things like switches. You can see these in the network menu. For theming purposes all of these "additional info" labels should probably share the same style class. We already have a bit of a style class mess in Cinnamon and I'm not a huge fan of adding even more if we don't need to.

PR #4643, opened 2015-09-22
split sliders and separators out of popupMenu.js. This should make th…

JosephMcc:
…em easier for desklet and extension devs to use.

The brightness slider in the power applet needs to be double checked since I don't have the available hardware to test it.


dalcde:
Wouldn't this break many third party xlets?

clefebvre:
https://github.com/linuxmint/Cinnamon/pull/4644

PR #4673, opened 2015-10-03
Removed dateFormat from toLocaleTimeString() call

filakhtov:
Notifications are broken due to invalid format supplied. Calling toLocaleTimeString with "%l:%M %p" format gives a RangeError exception with "invalid language tag" message.

### DIFF
--- a/js/ui/messageTray.js
+++ b/js/ui/messageTray.js
@@ -446,7 +446,6 @@ Notification.prototype = {
         this._imageBin = null;
         this._timestamp = new Date();
         this._inNotificationBin = false;
-        this.dateFormat = _("%l:%M %p");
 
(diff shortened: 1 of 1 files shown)


clefebvre:
In what locale? Could that be a translation issue?

filakhtov:
Getting this problem on Gentoo with three locales: de_DE, en_US, ru_RU.

clefebvre:
Well that's weird.. I can't reproduce the issue.

That said, the API doesn't seem to expect a date format as argument here... and testing it here, no matter what format I give it, I end up with the same string so... :)

If it helps you and it doesn't break anything for us, I see no reason not to merge it :)

filakhtov:
Thanks, that helps a lot.

PR #4680, opened 2015-10-05
[WIP] Alter panel allocation logic

Cinnamal:
Corrects that centering behavior of panels with applets in the center box.


dalcde:
Explain and justify. Why is the old one wrong? Words are preferred over screenshots

Cinnamal:
With the old code, the left box and right box are treated as if they require the same minimum space. When one of the side boxes is smaller than the other, it creates an unusable gap between the smaller of the two side boxes and the center box. This gap grows proportionally to the difference in size between the two side boxes.
This code still centers the center box but fills all the available space proportionally to its fair share.

Cinnamal:
I just fixed another bug. Will that also be of this pull request or do I have to make another one?

dalcde:
The old behaviour is desired behaviour. It is so that the center box will be centered all the time, unless this requires allocating to something less than its minimum width.

In any case, I am not convinced that your code is correct. Can you explain in full what the expected behaviour of your allocation algorithm is?

dalcde:
I'll add that the old algorithm emphasizes on centering objects because that the common use case is to have something fixed in size (eg. clock) in the center while the things in the side boxes are expandable (eg. window list)

Cinnamal:
Actually the behavior is exactly as you described it. It's actually based of some of the behavior of the old code. It is perfectly centered. The only change is that it will expand if necessary and there is a fair portion space available. I've added extensive comments to clarify.

The code is as follows:
//Find space wanted beyond minimal size
               "Total Want"  = "Total Nat Size" - "Total Min Size" + "1 to avoid divide by zero"
              "Center Want" = "Center Nat Size" - "Center Min Size"

//Now we want to set center to its Min size + if fair portion of wanted space
//Which I admit will scale it slightly if their is another compressible applet in different box.
//However this is unlikely and won't be an issue once the boxes meet
               "Center Width" = "Center Min Size" + ("Center Want" * ("Center Want"/ "Total Want"));

// Now the left box is set to Half the Total Space subtracting Haft the Center Width
// This perfectly centers the center box. It's half the remaining space.
		(Left Width) = ("Total Space" - "Center Width")/2;
//And check to see if we've run out of space in the left box
		if("Left Width"<"Left Min Width")"Left Width" = "Left Min Width";
//Give the other half of the remaining space to right box.
		"Right Width" ="Total Space" - "Left Width" -  "Center Width";
//And check to see if we've run out of space in the right box
		if("Right Width"<"Right Min Width")"Right Width" = "Right Min Width";

dalcde:
On Tue, Oct 06, 2015 at 07:32:47PM -0700, Cinnamal wrote:
>Actually the behavior is exactly as you described it.

I did not describe the behaviour of my code.

> It's actually based of some of the behavior of the old code. It is perfectly centered. The only change is that it will expand if necessary and there is a fair portion space available. I've added extensive comments to clarify.
>
>The code is as follows:
>//Find space wanted beyond minimal size
>               "Total Want"  = "Total Nat Size" - "Total Min Size" + "1 to avoid divide by zero"
>              "Center Want" = "Center Nat Size" - "Center Min Size"
>
>...
>
>		"Right Width" ="Total Space" - "Left Width" -  "Center Width";
>//And check to see if we've run out of space in the right box
>		if("Right Width"<"Right Min Width")"Right Width" = "Right Min Width";

This does not clarify anything at all. Incidentally, we are able to read
code, and converting from rightWidth to "Right Width" does not in any
wany make the code easier to understand. I was looking for a description
of the expected outcome *without reference to any code*. You have a set
of expected outcomes, and then the code is implemented such that this
set of expected outcomes is always achieved.

For example, a description of the old code could be

> Assuming that the centerBox is filled, the primary objective is to
> center the centerBox whenever possible.  This will be done all the
> time unless doing so requires some box's width to go under its minimum
> width.
>
> If we are centering the centerBox, there are two possible scenarios.
> Firstly, if the centerBox can be perfectly centered while everything
> takes their natural size, then everything will be allocated at least
> their natural size such that the centerBox is centered, leftBox is
> left aligned, rightBox is right aligned.
>
> Otherwise, we first allocate the minWidth to every box, and then
> distribute the remaining space proportional to how much more space
> each box wants. This is done in a way that ensures the leftWidth and
> rightWidth are equal.
>
> If it is not possible to center the centerBox, but there is enough
> space to just allocate the boxes, the centerBox will be made as
> centered as possible without making things go under their minWidth.
> This is achieved by making the shorter box go to their min width, and
> distributing the remaining space among the two other boxes.
>
> Finally, if there isn't even enough space to just put the things, the
> width allocated is just proportional to the minimum width.
>
> In the cases where the centerBox is not occupied, a similar mechanism
> is employed. If there is enough space for everything to get their
> natural width, this will happen. Otherwise, we first allocate the
> minimum width and then distribute the remaining space proportional to
> how much more space each box wants. In the scenario where the isn't
> enough space to just allocate the minimum width, we just allocate
> proportional to the minimum width.

Can you come up with a similar description of what the expected
behaviour of your code is, and hopefully convince yourself that your
code does indeed perform correctly in all scenarios?

dalcde:
I'll come up with another case where your code fails. Suppose the three boxes all have a minWidth of 0 and naturalWidth of 1. In this case, the centerBox will be allocated a width of 1/4, even though there is more than sufficient space to fully allocate the naturalWidth of 1.

clefebvre:
Hi,

I'll close this PR now. @dalcde worked a lot on panels and so it was important to reach consensus. There's also a very important PR from @brownsr coming in 3.2, which brings vertical panels to Cinnamon, so it's better to work on layout after that gets merged as well.

PR #4738, opened 2015-10-23
Fixed code injection bug in screensaver lock dialog (Fixes #4652)

### LINKED ISSUE #4652: Code injection in cinnamon-screensaver-lock-dialog.py
screenshot

Have a look on the screenshot first.
The program xeyes will be started because of the special characters before and behind it.
The screen will not be locked until xeyes is closed.
You can have the same effect with the shell command $(xeyes)

Reason is the os.system call in the python script where the shell command is injected in %s
Please use subprocess , not os.system()

-----
    def lock_screen(self, data):
        message = self.entry.get_text()
        if (message != ""):
            os.system("cinnamon-screensaver-command --lock --away-message \"%s\" &" % self.entry.get_text())

dralley:
Entering ```$( any command )``` into the lock screen dialog box will currently pass that command through to ```os.system()```.

I replaced os.system() calls with subprocess.call() and did some PEP8 formatting

Fixes #4652

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-screensaver-lock-dialog/cinnamon-screensaver-lock-dialog.py
+++ b/files/usr/share/cinnamon/cinnamon-screensaver-lock-dialog/cinnamon-screensaver-lock-dialog.py
@@ -1,22 +1,23 @@
 #!/usr/bin/env python2
 
-import sys
 import os
+import subprocess
 import gettext
 from gi.repository import Gtk
 import pwd
-import socket
 
 # i18n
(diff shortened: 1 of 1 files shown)


clefebvre:
Careful, call() is synchronous... os.system("%s &" % command) is not.
It might be ok if cinnamon-screensaver-command returns immediately.

Otherwise we'd need to use popen instead. (or glib's spawn functions if we're to face zombies ;)).

clefebvre:
Looks good.

Issue #4723, opened 2015-10-19
Make systray-applet icon size themable


A-Shahbazi:
I added systray applet to my panel but the icons seemed a little small. I did a little investigation and realized that systray applet icons do not obey theme icon-size. I have a cinnamon theme which defines several panel properties including panel height and icon-size:

```
...
.system-status-icon {
    icon-size: 16px;
    icon-shadow: black 1px 1px 2px 0px;
}
...
```

The systray icons, however, do not follow the assigned icon-size. Following are 2 screenshots. I have changed icon-size to 21px in one of them so that the issue become more clear.
icon-size: 16px
![screenshot from 2015-10-19 18 04 42](https://cloud.githubusercontent.com/assets/7456141/10580998/49c14bf0-768d-11e5-9351-b0eef22a6146.png)
icon-size: 21px
![screenshot from 2015-10-19 18 06 59](https://cloud.githubusercontent.com/assets/7456141/10581009/5962541e-768d-11e5-8ef8-6cca363dde25.png)

ManIVIctorious:
If I change the panel height all applets, including the system tray icons, rescale properly. Is this still an issue for you in cinnamon 3.0?

JosephMcc:
I would have to double check by I'm pretty sure you still can't set the size of the icons from the theme. I'll mark this as a feature request but wouldn't hold my breath because the hard coded icon sizing helps work around some systray issues. Just do a search for systray issues here to get a hint at how problematic it is to keep the systray working.

lestcape:
Thank to @mtwebster that remove the icon cache for Gio I can use gicon in the indicators, but i can not use system-status-icon in combination with panel-button. I need to use instead applet-box and applet-icon, because indicators can not used the same styles as status icons. I think, because icons are not directly inside the panel-button actor. Also this styles are not working for the status icons. Also indicator icons are not thinking to be integrated with the size property of the theme, In some cases the icons are an array of bytes and they are loaded form a clutter colg texture, with a predefine size in the cinnamon implementation. Now i'm loaded the icon for a gicon, but also with a predefine size, using a GdkPixbuf.Pixbuf that lucky implements gicon and allow me make indicators theme-ables.

The indicators are not homogeneous and will not be loaded into a vertical panel if will not be estimated the size of the icons, base on the panel layout (the orientation of the applet). I changed all size estimation to be possible estimate correctly the height and width of the indicator actors, base on the with and height of the panel and his Layout.

[...]

PR #4739, opened 2015-10-23
make systray icon size and scaling factor customizable

A-Shahbazi:
I have fixed the below issue already reported by me.
systray-applet icon size is hardcoded: linuxmint/Cinnamon#4723


clefebvre:
Thanks.

@dalcde @JosephMcc what's your opinion on this?

It certainly is an improvement, but should it be defined in the theme or in the configuration?

dalcde:
It is a good idea imo, but it makes more sense to make the size a fixed multiple of usual icon sizes. In any case, it should never be a gsetting that is read when the file is loaded and not monitored for change.

A-Shahbazi:
It is checked on every cinnamon start. Do you mean it should take place immediately?

dalcde:
It should be read when the applet load, stored as a setting within the applet, not a global variable. Then the applet to listen to changes and then act accordingly when the settings changes, *just like how every other setting is read and handled in Cinnamon*.

lestcape:
@A-Shahbazi :

Maybe something like this: Won't set this properties (Schema, ICON_SCALE_FACTOR, ICON_SIZE) global, set it as private to the class in the constructor:
```
MyApplet.prototype = {
    __proto__: Applet.Applet.prototype,

    _init: function(orientation, panel_height, instance_id) {
        Applet.Applet.prototype._init.call(this, orientation, panel_height, instance_id);
        ...
        this._settings = new Gio.Settings({ schema: 'org.cinnamon.applets.systray' });
        this._icon_scale_factor =  this._settings.get_double('icon-scaling-factor');
        this._icon_size =  this._settings.get_int('icon-size');
        // Then connect to any changes:
        this._scalingFactorId = this._settings.connect('changed::icon-scaling-factor', Lang.bind(this,  this.on_panel_height_changed));
        this._iconSizeId = this._settings.connect('changed::icon-size', Lang.bind(this,  this.on_panel_height_changed));
        ...    
    },
```

And read the values for the class in all places where they are called.

What i disagree with this solution is about the performance involved to change the standard icons size (not the appindicators), as they need to be rebuilded to only change his sizes... So that’s isn't nice but i think rebuild all on panel change, to just to change the size of the icon is not a good choice currently.

Another will be relate with this:
https://github.com/A-Shahbazi/Cinnamon/blob/145d988e05e00d0f5a06625712b5458b68b293ca/files/usr/share/cinnamon/applets/systray%40cinnamon.org/applet.js#L308
https://github.com/A-Shahbazi/Cinnamon/blob/145d988e05e00d0f5a06625712b5458b68b293ca/files/usr/share/cinnamon/applets/systray%40cinnamon.org/applet.js#L311
https://github.com/A-Shahbazi/Cinnamon/blob/145d988e05e00d0f5a06625712b5458b68b293ca/files/usr/share/cinnamon/applets/systray%40cinnamon.org/applet.js#L314
https://github.com/A-Shahbazi/Cinnamon/blob/145d988e05e00d0f5a06625712b5458b68b293ca/files/usr/share/cinnamon/applets/systray%40cinnamon.org/applet.js#L317

If you want to dynamically set the size for the appindicator, you need to also dynamically set the size of the standard systray-icons, or will be a mixture of things and this will look like ugly in my opinion.

lestcape:
Let me propose a solution to this function:
https://github.com/A-Shahbazi/Cinnamon/blob/145d988e05e00d0f5a06625712b5458b68b293ca/files/usr/share/cinnamon/applets/systray%40cinnamon.org/applet.js#L297

As i know the reason of this function is for handle the buggy icons, and @clefebvre estimated the size of this icons statically.

So, if someone want to set it dinamically, the solution could be find a good function to scale icons, in the domain of the real icon size, to the domain of the cinnamon icon size. As @clefebvre was estimated the value of this function in some points, we can construct a polinomial aproximation to the unknown function that can interpolate the estimate points of @clefebvre and then we can evaluate the polinomial to find the cinnamon icons size values in the range of the function where we don't know actually the best "match".

A-Shahbazi:
@lestcape @clefebvre 

> Let me propose a solution to this function:
https://github.com/A-Shahbazi/Cinnamon/blob/145d988e05e00d0f5a06625712b5458b68b293ca/files/usr/share/cinnamon/applets/systray%40cinnamon.org/applet.js#L297

>As i know the reason of this function is for handle the buggy icons, and @clefebvre estimated the size of this icons statically.

I didn't exactly understand what for is the part you mentioned. What do you mean by buggy icons? (Did you mean "shutter", "filezilla"?) Why don't we just use the following?
```
...
        let disp_size = this._panelHeight * this._icon_scale_factor;
        if (["shutter", "filezilla"].indexOf(role) != -1) {
            global.log("Not resizing " + role + " as it's known to be buggy (" + icon.get_width() + "x" + icon.get_height() + "px)");
        }
        else {
            icon.set_size(disp_size, disp_size);
            global.log("Resized " + role + " with normalized size (" + icon.get_width() + "x" + icon.get_height() + "px)");
            //Note: dropbox doesn't scale, even though we resize it...
```

lestcape:
@A-Shahbazi yes, about the shutter, filezilla (but more icons)... see the history of the applet: 

https://github.com/linuxmint/Cinnamon/commits/master/files/usr/share/cinnamon/applets/systray%40cinnamon.org/applet.js

You will see, your proposition was maybe the first approximation to the problem. Anyway, what i say is that:

1- The standard systray icons need to use the same way to resize icons of the indicator or this will look like ugly.

2- Will be nice create an alternative way to resize the icons when a panel size change or if will be implemented a way to handle the custom icon size (this will be instead of rebuild all again).

For your proposition about how handle the buggy icons, I think you are who need to propose a solid and general solution better than the currently if you want to modify the hard work of @clefebvre and others in this (again, please see the history of the applet). In my personal opinion, this is complex and is better not be modify the current way if you are not pretty sure your solution is at less equivalent or better, but you can try.

lestcape:
In the function _resizeStatusItem, maybe the best solution will be apply the scaling factor, after the case switch, to be really dynamic and then impact the icon size in a dynamic way. So, something like that maybe:
```
 _resizeStatusItem: function(role, icon) {
        if (icon.obsolete == true) {
            return;
        }
        let size;
        if (["shutter", "filezilla"].indexOf(role) != -1) {
            global.log("Not resizing " + role + " as it's known to be buggy (" + icon.get_width() + "x" + icon.get_height() + "px)");
        }
        else {
            if (this._panelHeight < 22) {
                size = 16;
            }
            else if (this._panelHeight < 32) {
                size = 22;
            }
            else if ( this._panelHeight < 48) {
                size = 32;
            }
            else {
                size = 48;
            }
            size = size * ICON_SCALE_FACTOR;
            icon.set_size(size, size);
            global.log("Resized " + role + " with normalized size (" + icon.get_width() + "x" + icon.get_height() + "px)");
            //Note: dropbox doesn't scale, even though we resize it...
        }
    },
```

lestcape:
```
   _getIndicatorSize: function(appIndicator) {
        if (this._scaleMode)
            return this._panelHeight * ICON_SCALE_FACTOR;
        return ICON_SIZE;
    },
```

Need to be then:

```
   _getIndicatorSize: function(appIndicator) {
        if (this._scaleMode) {
            let size;
            if (this._panelHeight < 22) {
                size = 16;
            }
            else if (this._panelHeight < 32) {
                size = 22;
            }
            else if ( this._panelHeight < 48) {
                size = 32;
            }
            else {
                size = 48;
            }
            return  size * ICON_SCALE_FACTOR;
        },
        return ICON_SIZE;
    },
```

lestcape:
In others words if you let the function _resizeStatusItem as it is, then you will get spaces without any modification between the icon size change and this is ok when they are statically, but not when they are dynamically.

PR #4748, opened 2015-10-25
Change generate_desktop_files script to refer to the new location of cinnamon-settings.py

dralley:
And some PEP8 formatting

### DIFF
--- a/files/generate_desktop_files
+++ b/files/generate_desktop_files
@@ -1,6 +1,7 @@
 #!/usr/bin/python2
 
-import os, commands
+import os
+import commands
 
 sourceFile = "usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py"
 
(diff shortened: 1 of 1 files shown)


clefebvre:
hmm... there's nothing wrong with the PR, but this file shouldn't be here anymore.

The cinnamon-settings desktop file is generated by generate_additional_files.py.

dralley:
Is [cinnamon-looking-glass/page_extensions.py](https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/cinnamon-looking-glass/page_extensions.py) also no longer being used?  It isn't imported from anywhere else in the entire project, and some of the code in there is clearly nonfunctional. (Specifically, the variable ```lookingGlassProxy``` is referenced about a dozen times without ever having been defined anywhere)

Issue #2453, opened 2013-10-12
Is OpenCV an unnecessary dependency?


krasnoglaz:
I can't imagine why a desktop environment needs a computer vision library.

show0k:
Nice, and if I want to use opencv3 ?

tgarc:
@show0k Don't know what distribution/package manger you're using but in debian/ubuntu `apt` will install packages in `/usr/`. You can install opencv3 (or any other package for that matter) alongside the apt installation without conflict by specifying `/usr/local` as the install prefix. For opencv use the cmake_install_prefix when doing your build e.g., `cmake -D CMAKE_INSTALL_PREFIX=/usr/local`). Python will automatically try to load from `/usr/local` first so as far as I can tell the `/usr/ ` install of opencv should not cause any issues with your local install.

show0k:
The auto-generated CMakeCache.txt is good, the prefix is already set to /usr/local. There is no conflicts, but the '/usr/lib/python2.7/dist-packages/cv2.so' is loaded instead of '/usr/local/lib/python2.7/dist-packages/cv2.so' when I import cv2, because **it's after in the sys.path list**.
The only solution I found, is to **add**  'usr/local/lib/python2.7/dist-packages' **at the beginning** of sys.path 
``` python
import sys
sys.path.insert(1,'/usr/local/lib/python2.7/dist-packages')
```

Btw, I'm running on Linux Mint 17.

dralley:
Considering the code under question uses the OpenCV 1.x API which [is deprecated](http://answers.opencv.org/question/17546/opencv-will-drop-c-api-support-soon/) and has been removed in OpenCV 3.x entirely, would a Pull Request to use the new API be desirable?

It is not likely that the ```python-opencv``` package will switch to OpenCV 3.x soon, however if and when it does happen this code will need to be changed, and I wouldn't mind doing so preemptively.

clefebvre:
Yes, it would be considered for cinnamon 3.0.

PS: In the absence of bindings we can also use gi if available.

dralley:
@clefebvre Fantastic, I'm testing the patch as we speak.

[...]

PR #4795, opened 2015-11-06
[3.0] Use streamer instead of python-opencv for user profile photos

dralley:
Discussion in Issue #2453

* Removes ```python-opencv``` as a dependency, which should reduce Cinnamon's disk footprint by 20-30 megabytes
* Webcam is no longer temporarily activated when opening "Account details" settings panel
* Adds ```streamer``` as a dependency

Credit to @kernc for the idea and the basic implementation of lines 119-124 via. [PR](https://github.com/linuxmint/live-installer/pull/57)


dralley:
It should also be noted that the current Cinnamon user profile code will break once the package ```python-opencv``` is updated to use OpenCV 3.0.  This transition is currently in the Debian Experimental branch.

Issue #4730, opened 2015-10-21
Chrome/Chromium Webapps fail to use proper icon on panel


ReplicatedGrace:
Issue occurs on Linux Mint 17.2 64-bit, Cinnamon 2.6.13.

Steps to reproduce:
1. Install Chromium or Google Chrome
2. Open chrome://apps/ in Chromium/Chrome.
3. Right-click a webapp (e.g. YouTube) and check as "Open as Window". 
4. Left click the chosen webapp to open it.

Expected Behaviour:
Webapp opens in its own window, and the webapp's icon shows on the panel.

Current Behaviour:
Webapp opens in its own window, but uses the default Chromium/Chrome webapp.

Notes:
It is possible to create webapps in Google Chrome (version 46.0.2490.71 64-bit) and the open-source Chromium browser, and set these webapps to open in their own window without the usual browser chrome visible. These webapps should use their web icon on the panel rather than the typical Chrome or Chromium icon to differentiate them from regular browser windows. This behaviour fails on Cinnamon.

I tested this in all the major Linux Mint desktop flavours and indeed it is only Cinnamon that fails to handle the webapp icons properly.

Screenshots linked below show how all the different desktops handle webapp icons fine, but Cinnamon does not. Cinnamon has this erroneous behaviour no matter what theme is installed, and whether or not it is a fresh install of Mint.

XFCE - 
![xfce](https://cloud.githubusercontent.com/assets/15222985/10624900/23200aca-776b-11e5-9433-9413b7be70e6.png)

MATE - 
![mate](https://cloud.githubusercontent.com/assets/15222985/10624916/3ce1cfde-776b-11e5-95d2-0fe3aa5f29d6.png)

KDE (alternate theme) - 
![kde alt theme](https://cloud.githubusercontent.com/assets/15222985/10624921/4543f8f0-776b-11e5-84dd-3c1fa76b2f9a.png)

Cinnamon (Numix theme) - 
![cinnamon numix theme](https://cloud.githubusercontent.com/assets/15222985/10624930/59d0b7ae-776b-11e5-815d-5664b8607f24.png)

Cinnamon (default theme, fresh install) - 
![cinnamon default](https://cloud.githubusercontent.com/assets/15222985/10624934/6999f7ea-776b-11e5-8f2a-3a5d8ee12bd0.png)

Same problem is reported for open source Chromium as well as Google Chrome. - http://forums.linuxmint.com/viewtopic.php?f=47&t=206590

[...]

PR #4806, opened 2015-11-09
CinnamonWindowTracker: Pull some upstream fixes so google chrome apps…

JosephMcc:
… will use the proper icons in the window list


JosephMcc:
Fixes: https://github.com/linuxmint/Cinnamon/issues/4730

PR #4807, opened 2015-11-09
[3.0] Panel Lock (Fix #4720)

### LINKED ISSUE #4720: add lock/unlock panel setting
It is very nice that cinnamon panel is highly configurable(like the early days of gnome2) but after one finishes the configuration it's better to have a "lock panel" option in panel settings (System Settings > Panel) so that the right clicking on blank space or on applets doesn't bring up context menus that are for configuration purposes.

The idea behind this suggestion is that once you finish the configuration you want the panel to feel SOLID for everyday use. You rarely decide to change panel configuration once you have it set. Using this option you make sure there won't be unintentional pop up of a configuration window or even unintentional change in panel layout. Moreover, this is the default behavior for other desktop environments.

BTW, this bug report is not from a security point of view like what we have at #2719. So this is not a duplicate of that bug.

A-Shahbazi:
This implements lock/unlock feature for panel through gsettings key org.cinnamon.panel.locked. There are some points:
1. The feature is implemented using gsettings so that only power users who really know what they are doing can use it.
2. The default option is false so that every modifying option is available from the very first start.
3. The key description fully warns the user about not being able to configure panel once locked unless they unlock it first.
4. It's just an OPTION and it's really hard to accidentally change because of the way it's implemented(gsettings key)


dalcde:
This is pointless. You have not managed to persuade even a single person that we should do this. Wasting your time to create a pull request does not help.

clefebvre:
No need to be rude or at least so direct :)

A few notes:

- Some people did ask for that, so there's a bit of demand for it.
- This is definitely 3.0 material and we're pre-RC on Mint for 2.8 so it will take time before it gets in, if it does.
- We should discuss it (IRC is best for this, this isn't your only contribution, you could meet with the team and talk about your ideas too. We're on #linuxmint-dev irc.spotchat.org)
- I can see how locking the panel (or even Cinnamon globally) can be useful in some circumstances... a cyber-cafe for instance, a school etc... 
- Should anyone be able to unlock via gsettings?

A-Shahbazi:
> We should discuss it (IRC is best for this, this isn't your only contribution, you could meet with the team and talk about your ideas too. We're on #linuxmint-dev irc.spotchat.org)

I think you have seen linuxmint/Cinnamon/issues/4720. I thought the guys there didn't consider it necessary so I implemented this feature and tried to address their concerns explained there. I think there is a difference between not being willing to implement an OPTION and preventing an OPTION from getting merged. Specially when the option is going along with the mainstream linux desktop environments.
> Should anyone be able to unlock via gsettings?

IMO, each user should have the right to unlock panel. Like any other configuration in desktop environment.

JosephMcc:
> I think there is a difference between not being willing to implement an OPTION and preventing an OPTION from getting merged. 

It's not about preventing an option from getting merged. It's about preventing unneeded options from getting merged. Every additional option adds more complexity and maintenance work in the code. In my personal opinion, which others certainly might not agree with, you haven't provided any real reason to need this.

> Specially when the option is going along with the mainstream linux desktop environments.

That just isn't true. Some desktop do this and some do not. We already have a form of this with Panel Edit mode.

A-Shahbazi:
> That just isn't true. Some desktop do this and some do not. We already have a form of this with Panel Edit mode.

Would you name some of those that provide configuration options through a simple right-click and don't provide a lock/unlock feature?

dalcde:
On Mon, Nov 09, 2015 at 12:56:56AM -0800, Clement Lefebvre wrote:
>- I can see how locking the panel (or even Cinnamon globally) can be useful in some circumstances... a cyber-cafe for instance, a school etc...

This is a completely separate issue. This pull requests adds an option
to hide certain buttons used to configure the panel. Users are still
able to edit it through, say, cinnamon settings or dconf-editor.

Locking the options themselves is something that can be done directly in
dconf iirc, where the options just cannot be changed (you can press the
buttons, but nothing happens)

lestcape:
Whatever... I want an option to hide some uglys menu items, well better, one option per item will be the best, in this way will can hide to prevent an accidental. But well maybe is better another option to hide the option of hide the menu item. Thats will be pretty secure.

A-Shahbazi:
@lestcape 
Would you explain more what you mean? Do you mean kind of "Configure Panel Context Menu" under Cinnamon Settings > Panel through which you can show/hide each item(i.e. "Add applets ...", "Panel settings", "Themes", "All settings", "Troubleshoot", "Modify panel", "Panel edit mode")?

lestcape:
What i say is that is not ok to me, make disappear options, because if you want to do that for only some items, why not for others, this will look like a patch. I don't think as a good one patch to fix something... If this will have sense to me, need to be be a general option, where you can control the items that you want to disappear as occurs more less in gtk, where you can control the Gtk Actions... In any case i see this as an unnecessary overhead for designing wherever be applicable and will only create confusions for novice user. I don't see how this can be useful and is easy to reverse the option at any time if the option was activated by mistake. Which I do not see how this could be occurred anyway, if can be is not common and is like any other things. for example the desktop console can be activated also by mistake, and it is well annoying in the middle of the screen, if you do not know what is and how remove it. Is a dev tools in the middle of the desktop, not useful for novice user. Will convenient an option to disable the option of active it? There are more things like that? Of course  yes. Then we will create an option to disable all? Not of course not.

clefebvre:
Hi,

I'll close this PR as it is. The code changes are disposable, this can be re-implemented easily if needed.

The idea is much more valuable though and it should be discussed. I think the main reason people aren't convinced is because it doesn't address another "much more serious" need, which is to really lock things up.

Say I'm a public library, or a school, and users come and go on a guest account, I want to lock things up and not let them mess the desktop. If I want to do that though, I want my password or admin permissions to unlock it again and I want that lock to be configurable and global (we might need to think about xapps here).

DarianAnjuhal:
Hi, I would need the same option. Is there a solution already. Maybe its done because the discussion was 4 years ago.

Maybe there are other possiblities to customize the drop down menu wich appears on right click on a panel?

PR #4813, opened 2015-11-10
Places and Recent should not show icon if CATEGORY_ICON_SIZE = 0

A-Shahbazi:
I didn't want the category icons in menu applet to show up so I set CATEGORY_ICON_SIZE = 0 but places and recent items where still showing icons.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -790,9 +790,11 @@ PlaceCategoryButton.prototype = {
         this.actor.set_style_class_name('menu-category-button');
         this.actor._delegate = this;
         this.label = new St.Label({ text: _("Places"), style_class: 'menu-category-button-label' });
+        if (CATEGORY_ICON_SIZE > 0){
         this.icon = new St.Icon({icon_name: "folder", icon_size: CATEGORY_ICON_SIZE, icon_type: St.IconType.FULLCOLOR});
         this.addActor(this.icon);
         this.icon.realize();
+        }
         this.addActor(this.label);
         this.label.realize();
     }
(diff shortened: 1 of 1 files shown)


dalcde:
errr... CATEGORY_ICON_SIZE is defined to be 22. If you don't want them to show up by editing your local copy of the code, why don't you just comment that block out directly?

A-Shahbazi:
Errr... I thought adding a "category icon size" or "show category icon" option to applet might be considered unnecessary/unneeded option which adds more complexity and maintenance work in the code ;)

JosephMcc:
> I thought adding a "category icon size" or "show category icon" option to applet might be considered unnecessary/unneeded option which adds more complexity and maintenance work in the code ;)

That sounds like something I would say not @dalcde ;)

I think the point is, if you are editing your own local copy you could just comment out the couple lines that add the icon instead of doing this.

A-Shahbazi:
> I think the point is, if you are editing your own local copy you could just comment out the couple lines that add the icon instead of doing this.

What if one changes their mind about having category icons back? They should find all those commented lines and uncomment them. Moreover, setting CATEGORY_ICON_SIZE to 0 worked for other entries. So one doesn't have to also comment the lines related to them. It seems more efficient this way ;)

dalcde:
On Tue, Nov 10, 2015 at 12:15:57AM -0800, A-Shahbazi wrote:
>> I think the point is, if you are editing your own local copy you could just comment out the couple lines that add the icon instead of doing this.
>
>What if one changes their mind about having category icons back? They should find all those commented lines and uncomment them. Moreover, setting CATEGORY_ICON_SIZE to 0 worked for other entries. So one doesn't have to also comment the lines related to them. It seems more efficient this way ;)

Except that the code will completely make no sense, becasue
CATEGORY_ICON_SIZE is *defined* to be 22, and then you are checking if
it is greater than 0. The code is not intended to be manually changed
that way. If you wish to change it yourself, then it's your job to make
it work, not Cinnamon's.

mtwebster:
I wouldn't be against this being an actual option in the applet - it's substantially faster without icons, which might appeal to some people.

This isn't the way to do it though - use the settings API.

A-Shahbazi:
@JosephMcc 
According to what @mtwebster said is it considered unnecessary/unneeded now?

clefebvre:
Hi,

I'm not against it either.

+1 on using the settings API.

dalcde:
+1 for making it a proper option

Issue #4767, opened 2015-10-30
cinnamon-control-center can not load panel module


kronolio:
Loading Calendar module and Keyboard module
/usr/share/cinnamon/cinnamon-settings/bin/capi.py:53: Warning: instance of invalid non-instantiatable type 'CcDateTimePanel'
  return GObject.new(panel_type)

crash cinnamon-control-center

==================================

Display module
/usr/share/cinnamon/cinnamon-settings/bin/capi.py:53: Warning: invalid cast from 'CcDisplayPanel' to '<invalid>'
  return GObject.new(panel_type)
/usr/share/cinnamon/cinnamon-settings/bin/capi.py:53: Warning: instance of invalid non-instantiatable type 'CcDisplayPanel'
  return GObject.new(panel_type)

====================================

list /usr/lib{64}/cinnamon-control-center-1/panels/
-rwxr-xr-x 1 root root  56K окт 28 23:20 libcolor.so
-rwxr-xr-x 1 root root  72K окт 28 23:20 libdate_time.so
-rwxr-xr-x 1 root root  81K окт 28 23:20 libdisplay.so
-rwxr-xr-x 1 root root 561K окт 28 23:20 libnetwork.so
-rwxr-xr-x 1 root root  73K окт 28 23:20 libregion.so
-rwxr-xr-x 1 root root 193K окт 28 23:20 libsoundnua.so
-rwxr-xr-x 1 root root  36K окт 28 23:20 libuniversal-access.so

os: gentoo
cinnamon: 2.8.2

lisuml:
Similar problems on cinnamon 2.8.2 on Manjaro Linux. Affected settings modules:

* Accessibility
* Date & Time
* Color
* Display
* Graphics Tablet
* Networking

kronolio:
How to fix it? This is upstream bug?

lisuml:
@kronolio: If you are on x86_64 can you check if your `/usr/lib64` is a symlink to `/usr/lib` please?

kronolio:
os:gento64 multilib
in gentoo 
/usr/lib64 is a directory
/usr/lib symlynk to /usr/lib64

ls -alh /usr/lib
lrwxrwxrwx 1 root root 5 июл 24  2014 /usr/lib -> lib64

kronolio:
paths += ["/usr/lib/x86_64", "/usr/lib/x86_64-linux-gnu"]
            if not os.path.islink("/usr/lib"):
[...]

PR #4822, opened 2015-11-13
#4767: Handle situations when /usr/lib64 is a symlink to /usr/lib and…

lisuml:
… vice versa on some distributions causing modules are being loaded multiple times


lisuml:
Not sure if you are OK with the change, but it at least fixes the issue on Gentoo and Manjaro (probably Arch users will be happy too).

clefebvre:
Looks good to me.

I don't see how it could be an issue. I'll just test it just in case but it's very likely to be merged.

dalcde:
I can confirm this fixes#4767

Issue #4357, opened 2015-06-23
cinnamon 2.6.9 keyboard applet cannot use flags to display current layout


petrucci4prez:
Keyboard applet no longer displays a flag for the current keyboard layout (only "en" for english for example) This occurs regardless of the configuration setting. Downgrading to 2.6.8 solves issue

EmergReanimator:
I have made small patch for this which works for me.

$ diff -u  applet.js /usr/share/cinnamon/applets/keyboard\@cinnamon.org/applet.js
--- applet.js	2015-07-11 20:47:40.000000000 +0200
+++ /usr/share/cinnamon/applets/keyboard@cinnamon.org/applet.js	2015-07-11 20:54:05.000000000 +0200
@@ -139,7 +139,7 @@
         for (let i = 0; i < groups.length; i++) {
             let icon_name = this._config.get_group_name(i);
             let actor;
-            if (this._showFlags && this.icon_theme.has_icon(icon_name))
+            if (this._showFlags)
                 actor = new St.Icon({ icon_name: icon_name, icon_type: St.IconType.FULLCOLOR, style_class: 'popup-menu-icon' });
             else
                 actor = new St.Label({ text: short_names[i] });
@@ -171,7 +171,7 @@
         let selectedLabel = this._labelActors[selected];
 
         this.set_applet_tooltip(item._long_name)
-        if (this._showFlags && this.icon_theme.has_icon(item._icon_name)) {
+        if (this._showFlags) {
             this.set_applet_icon_name(item._icon_name);
             this.set_applet_label("");
         } else {

leigh123linux:
This commit breaks it

https://github.com/linuxmint/Cinnamon/commit/e6c3fd9fb247f08e65b316540c2b0e5fd6de65f4

perfect7gentleman:
EmergReanimator, thanx for patch

JosephMcc:
That commit only breaks things under some newer gtk versions. Gtk.IconTheme.has_icon() returns the incorrect value for some reason. Either it's behavior was changed or there is a bug which needs to be fixed in gtk. Works fine under gtk3.10

Rolice:
The patch worked for me as it seems the bug is still present in the current version, thanks @EmergReanimator . I wonder why the release contain such a malfunction (extra logic check within some icon theme)? Does this patch keep the consistency? Why it is not still applied in the master release?

JosephMcc:
[...]

PR #4838, opened 2015-11-18
#4357: Replace has_icon method with lookup_icon one to fix missing fl…

lisuml:
…ags icons in keyboard applet

Not sure why `has_icon` returns false for not built-in icons while `lookup_icon` finds them without a problem.


lisuml:
By the way, I was experiencing the issue when gtk3-3.18.5.

Issue #4830, opened 2015-11-15
Sliders in submenus


zagortenay333:
The `StDrawingArea` inside of submenus has a weird padding on the right. This is an issue if the theme uses a `border-image` property since it is not affected by it, making it look like the handle cannot be scrolled all the way to the right.
![workspace 5_003](https://cloud.githubusercontent.com/assets/11144630/11168667/8bb7bd16-8b99-11e5-9232-6efd817b9311.png)

Even if you don't use a `border-image`, the problem will be that the slider doesn't align properly on the right. The default theme solved this arbitrarily by setting a `padding-right: 0;` to the submenu items, which is not good.

JosephMcc:
There is something weird going on in the sound applet menu. Hopefully we can manage to get it fixed but we haven't figured out what is causing it. Currently that slider has a padding-right that is hard coded.

zagortenay333:
Ah, ok.

JosephMcc:
I'm closing this since a commit was merged that should fix it.

PR #4855, opened 2015-11-21
Fix volume slider in sound applet

Cobinja:
This fixes the volume slider offset in the sound applet, fixing #4830

### DIFF
--- a/js/ui/popupMenu.js
+++ b/js/ui/popupMenu.js
@@ -758,7 +758,7 @@ PopupSliderMenuItem.prototype = {
     },
 
     _moveHandle: function(absX, absY) {
-        let relX, relY, sliderX, sliderY;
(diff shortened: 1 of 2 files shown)


JosephMcc:
I didn't get the chance to test this. I assume it works but it would be nice if we could figure out how to fix the underlying issue that caused this in the first place. To get the proper padding right on these sliders I had to hard code a value here: https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/applets/sound%40cinnamon.org/applet.js#L227

Without that the slider works as it's supposed to but expands all the way to the menu edge. The same thing had to be done for the items in the Device submenu. The main volume slider works correctly so it would be nice to figure out what's going on with the layout of these submenu items.

PR #4868, opened 2015-11-24
window-list: Properly scale the window previews for hidpi

JosephMcc:


leigh123linux:
It works well here, thank you for fixing it.

Issue #4939, opened 2015-12-15
Indent battery status in power applet


philipzae:
Indenting the battery status line of each battery source will make it look better.

![cinnamon_power](https://cloud.githubusercontent.com/assets/4412114/11814791/208f3fa6-a362-11e5-91fd-375d16c3cddb.png)

On a side note, the mouse icon isnt aligned similar to the other battery icons.

![icon alignment](https://cloud.githubusercontent.com/assets/4412114/11814933/dc9f23d2-a362-11e5-8916-8e1e3fd66bdb.png)

philipzae:
Forgot to also mention that right aligning the charge percentage would make it easier to look through the list.

fahmed24:
I was able to indent the battery status but I did not right align the charge percentage.

icarter09:
Many thanks for submitting your feature request for Cinnamon. Your feature request has been documented and will be reviewed by the Linux Mint team.

For more information on our workflow and feature requests, read https://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/faq.html.

PR #4963, opened 2015-12-24
Indented battery status in power applet. 

fahmed24:
More information about the feature request, here.
https://github.com/linuxmint/Cinnamon/issues/4939

### DIFF
--- a/data/theme/cinnamon.css
+++ b/data/theme/cinnamon.css
@@ -202,6 +202,7 @@ StScrollBar StButton#vhandle:hover {
 .popup-inactive-menu-item {
 	font-weight: normal;
 	color: #999;
+	padding-left: 20px;
 }
 .popup-subtitle-menu-item {
 	font-weight: bold;
--- a/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
@@ -297,7 +297,6 @@ MyApplet.prototype = {
     _getDeviceStatus: function(device) {
         let status = ""
         let [device_id, vendor, model, device_type, icon, percentage, state, seconds] = device;
-		  let indent = "      ";
 
         let time = Math.round(seconds / 60);
         let minutes = time % 60;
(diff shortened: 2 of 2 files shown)


fahmed24:
Can you possibly provide some resources to help me modify the style? I am having some trouble finding adequate documentation for St.BoxLayout and St.Label. Thanks.

lestcape:
Ok... I see the problem. Well see, first thing is that: Add an style is pretty easy, but normally the devs of an app not want to add many different styles for several actors, because then will be pretty hard make a theme. So, they evaluate the best way to add the style that minimize the number of actors with an style and allowing the possibility of apply several modification, but of course not for all possible decisions. They can miss some important ones, but is a decision of who are controlling the cinnamon style add one more. So, i will recommended you instead of make your own style, request what you want, then will be possible evaluate if is really necessary. Create a theme is an art so, your personal appreciation, not always is the best for all people and sometime could not be convenient add an style because there are other reason, also if the changed will look better, probably could not be apply.

General help:
http://developer.linuxmint.com/reference/

A direct reference to the style property:
http://developer.linuxmint.com/reference/git/st/StWidget.html#StWidget--style

Inline Styling example:
actor.style = "<Here an style>"; // This is in the css way:

http://www.w3schools.com/css/css3_intro.asp
http://www.w3schools.com/html/html_css.asp
http://www.w3schools.com/css/css_syntax.asp

Real example:
statusLabel.style = '<Here your style>';

But as you can see here:
https://github.com/fahmed24/Cinnamon/blob/ac02b205454535583db4a41d34745c3f0df1ea46/files/usr/share/cinnamon/applets/power%40cinnamon.org/applet.js#L117

You have an style class called: 'popup-inactive-menu-item'

let statusLabel = new St.Label({ text: "%s".format(status), style_class: 'popup-inactive-menu-item' });

popup-inactive-menu-item is apparently a general style, so, what probably is the best is create a more specific one and add this style to the current one:

statusLabel.add_style_class_name('the-new-name');

Then you can set your properties inside the css file of your theme, or if you want to modify the cinnamon default style here:

https://github.com/linuxmint/Cinnamon/blob/4f40fd2aa40daba6273b388d727d7bae44f7a874/data/theme/cinnamon.css

You will need to know how create an style class in the css.

Regards.

fahmed24:
Thanks @lestcape. I appreciate the feedback.

clefebvre:
Hi,

There are two issues with this:

- RTL languages aren't taken into consideration
- The style you're changing isn't specific to the power applet (it's used somewhere else in Cinnamon and it can be used also by 3rd party applets/desklets).

Instead of adding padding here, I would recommend reviewing the layout of the power applet, maybe using a box to make the icon center on the left and have the two lines of text align with it or something.

Closing the PR in the meantime.

PR #4968, opened 2015-12-25
resize was wrong onTrayIconAdded

ikuria:
i added line 217 because of this._resizeStatusItem(role, icon); function was not there, with that line i got the right size on my tray icon

### DIFF
--- a/files/usr/share/cinnamon/applets/systray@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/systray@cinnamon.org/applet.js
@@ -214,6 +214,8 @@ MyApplet.prototype = {
             if (icon.get_parent())
                 icon.get_parent().remove_child(icon);
 
+            this._resizeStatusItem(role, icon); /* resize the icon (dont know if this was missing) */
+
             icon.obsolete = false;
             this._statusItems.push(icon);
 


lestcape:
Hi. This is wrong...

Please see this description: https://github.com/linuxmint/Cinnamon/blob/master/js/ui/applet.js#L122. So, you can not scale the icons if you are not in the "scale" mode. The scale mode is set inside panel property: https://github.com/linuxmint/Cinnamon/blob/master/js/ui/applet.js#L171 in this place:
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/panel.js#L1738 and this is taking from gsettings: https://github.com/linuxmint/Cinnamon/blob/master/js/ui/panel.js#L1538 and you can set this inside Cinnamon-settings (command: cinnamon-settings panel).

If the scale mode is set, then your line (please see is inside _onTrayIconAdded function) https://github.com/ikuria/Cinnamon/blob/2f2c75eaa8d5c8f7ea58386675200683f701d97b/files/usr/share/cinnamon/applets/systray%40cinnamon.org/applet.js#L217
is not needed, because you always call _insertStatusItemLater on any of this points:

- https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/applets/systray%40cinnamon.org/applet.js#L227
- https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/applets/systray%40cinnamon.org/applet.js#L234
- https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/applets/systray%40cinnamon.org/applet.js#L241

Inside _insertStatusItemLater then you always call with a delay time the function _insertStatusItem:
https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/applets/systray%40cinnamon.org/applet.js#L253

Inside _insertStatusItem, if the scale mode is set
https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/applets/systray%40cinnamon.org/applet.js#L287

you also call the _resizeStatusItem with a delay of 500 miliseconds:
https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/applets/systray%40cinnamon.org/applet.js#L289

Then your call of _resizeStatusItem is unnecessary, and is intentional not call this function directly, because there are some buggy status icons and this depend of the panel scale mode.

Regards.

PR #4970, opened 2015-12-27
Make window list less cluttered by removing special characters, like [ ]

mat-b-:
With many minimized windows, window list gets cluttered: http://i.imgur.com/aViOVfN.png (screenshot)

It's hard to read actual titles.

This change removes additional characters from titles, like [ ]

Imho it should be up to themes to style buttons according to window state (special characters disabled by default), but I'm not skilled to make such a change.

Other systems, like Windows and OS X, hide titles altogether. Hiding just special characters] should be a step to more tidyness, clarity and usability on the taskbar.

### DIFF
--- a/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
@@ -6,10 +6,7 @@
  * The main applet object listens to different events and updates the window
  * list accordingly. Since addition/removal of windows is emitted by the
  * workspace, we have to listen to the changes to the number of workspaces and
- * update our signals accordingly as well. It also listens to the change in
- * window state (eg tile/maximize) since the window titles are displayed
- * differently in the AppMenuButton for each different state, eg minimized
- * windows are shown as [Title].
+ * update our signals accordingly as well.
  *
  * For each window the main applet object wants to show, an AppMenuButton is
  * created. This is created for every window in the monitors it is responsible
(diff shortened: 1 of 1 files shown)


brownsr:
Not sure about this ... this will remove the minimised/snapped/tiled information that is put there deliberately, regardless of how cluttered the list is. Perhaps now that Cinnamon can show thumbnails on hover, this minimised/tiled/snapped information should move there ?  Perhaps some more drastic pruning - no title if minimised, maybe ?

JosephMcc:
I never found the special characters all that useful. I used cinnamon for quite some time before I even realized some of them were there. Having something to indicate the different states is probably good though. Just as an idea, I wonder if we could use style classes for the window list items to differentiate the look depending on the state. Something like what we do by coloring the item red when it needs attention.

dalcde:
I agree the special characters aren't that useful, especially the tiled ones

lestcape:
I don't use the cinnamon windows list, but just for that reason i don't see the usability of the special characters, but even more, i don't see the usability of the titles. Is better, show only the icon and make a group by the same application, like unity, windows... At less could be as an option.

Vahan86:
Agreed with @lestcape

mtwebster:
I agree the current implementation is rather hokey - i'd rather see better 'style' changes to depict differences in states, etc.. and if someone doesn't want them, they can turn them off I suppose.  I don't want to remove a feature like this, however, just because of arbitrary opinion - I guarantee someone uses it in their workflow (https://xkcd.com/1172/)... Other missing features etc.. I'm pretty sure already have their own issues.  Bringing them up in every tangentially related thread doesn't make them come true.

lestcape:
This is real and then true, what could be, is that never comes by default: http://cinnamon-spices.linuxmint.com/applets/view/243, http://cinnamon-spices.linuxmint.com/applets/view/16. What user can do then is just express his arbitrary opinions in every tangentially related thread, as clear is not in his hands the decision.

mat-b-:
As making an option or making 'style' changes as @mtwebster said is beyond my skill level, should I open new issue for this?

PR #4994, opened 2016-01-09
St update1

JosephMcc:
Start cleaning up our st/clutter code and bringing it up to date. Remove some old and/or unused functionality. Make st-widget a solid actor. Put a few "compatibility functions" in place to keep existing xlets working.


lestcape:
@JosephMcc let me ask you a question. I want to add a Ctrl+Z to the st.entry, there are a plan to do this inside the st.entry or i need to do this in the cjs side?

PR #5054, opened 2016-02-01
effects: Make several improvements to handling effects

JosephMcc:
* Add an unminimize effect to the meta-plugin instead of using the map effect

* Add more fine grained control of the effect settings internally and expose those
  changes in Effects settings ui

* Enable effects on dialogs and menus by default

* Add new rollup/down and expand/collapse effects

* Fix an issue where maximizing/unmaximizing minimized windows from the window list
  didn't unminimize them

Requires https://github.com/linuxmint/muffin/pull/216

### DIFF
--- a/data/org.cinnamon.gschema.xml.in
+++ b/data/org.cinnamon.gschema.xml.in
@@ -152,7 +152,7 @@
     </key>
     
     <key name="desktop-effects-on-dialogs" type="b">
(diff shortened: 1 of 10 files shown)


JosephMcc:
Closing this temporarily. Found a bug in it and don't want it to be merged until I sort it out.

Issue #3666, opened 2014-11-10
Allow multiple actions for Panel Launcher Applet icons


theSeppi:
Currently, you can only use the Panel Launcher Applet items for one action: To launch a specified program.  
Here, I like Unity's approach more (one of the few things I actually like in Unity): The Panel Launcher items are links to .desktop launcher files, like in /usr/share/applications. Some of these launchers, e.g. Firefox, support multiple actions.  
In the case of Firefox, if you normally click the launcher, it starts as it would in Cinnamon. However, if you right click the launcher icon, the context menu offers some more options, like opening a window in private mode or a new window.

To show what I mean, I found this image on Google:
![image](https://cloud.githubusercontent.com/assets/5564491/4982605/1291fee0-6914-11e4-886c-f375d6fa27ec.png)
The entries "Wallch" and "Unlock from Launcher" are default, everything above is defined in the .desktop file.

This is a really nice feature which, at least for me, would improve workflow, and which I'd like to see in Cinnamon.

anandrkris:
Duplicate of https://github.com/linuxmint/Cinnamon/issues/2826 ?

JosephMcc:
This is now implemented in git master for Cinnamon 3.0

geoffm:
Actually there are officiel additionnal application actions in the freedesktop.org specifications, they just haven't been implemented yet by most DE's.
https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s10.html

PR #5077, opened 2016-02-11
[Applet][Feature request] Allow multiple actions for Panel Launchers Applet icons

trolleyyy:
I've implemented feature request: (Allow multiple actions for Panel Launcher Applet icons) #3666
You may want to see how it works on screenshots available here: https://github.com/trolleyyy/MintPanelLauncher

### DIFF
--- a/files/usr/share/cinnamon/applets/panel-launchers@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/panel-launchers@cinnamon.org/applet.js
@@ -40,10 +40,24 @@ PanelAppLauncherMenu.prototype = {
 
         Applet.AppletPopupMenu.prototype._init.call(this, launcher, orientation);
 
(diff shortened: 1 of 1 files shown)


JosephMcc:
Cool, I haven't actually tried this yet but have a couple of comments. Please use consistent 4 spaces indentation to stay in line with what is already there. Also remove the "..." from the "More" item. We started removing those after adding back the expander arrows.

trolleyyy:
Done, I prefer using "4 spaces length" tab indent in editors rather than space indent so I haven't seen this inconsistency before. Also dots from "More" item removed already.

mtwebster:
Would it make sense to have this functionality in the window list applet also?

Works nice, thanks :)

trolleyyy:
It would be nice to have this functionality in the window list applet too, but after fast look into window list applet code it seems, that it is really not intended to do it. Apparently in windows list applet we have only handle for window, not referenced desktop file. Maybe there are some way to do it, I'll think about it later.

PR #5131, opened 2016-03-15
Move sound settings to python as a native module

collinss:
Requires https://github.com/linuxmint/cinnamon-control-center/pull/134 and https://github.com/linuxmint/cinnamon-settings-daemon/pull/112

### DIFF
--- a/data/icons/Makefile.am
+++ b/data/icons/Makefile.am
@@ -70,6 +70,15 @@ public_icons = \
 	hicolor_emblems_scalable_cs-xlet-running.svg \
 	hicolor_emblems_scalable_cs-xlet-system.svg \
 	hicolor_emblems_scalable_cs-xlet-update.svg \
+	hicolor_devices_scalable_audio-speaker-left.svg \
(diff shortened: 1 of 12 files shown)


JosephMcc:
I know I mentioned in IRC about the missing icons in the "Test sound" dialog. Since those icons aren't part of the standard gnome/adwaita icon theme you'll probably need to move them to cinnamon so they are accessible regardless of a users icon theme. Much like we do with the other settings icons here: https://github.com/linuxmint/Cinnamon/tree/master/data/icons

mtwebster:
Looks nice and clean.

Observations:

- Test sound dialog has no way to close (other than right-clicking  the title bar).  Also, we should maybe shrink the test window depending on what speakers are active, instead of having a mostly-empty dialog.
- I know there's some lengthy historical reason somewhere about max volume not being 100% ;) so I won't ask...
- Nitpicking, but the effects volume maybe should be outside the effects list container.

Overall looks nice though - when can I merge?

PR #5185, opened 2016-04-06
Fix middle click action in the photo frame

### LINKED ISSUE #5184: photo frame fails to open the image on middle-click
Forwarding Debian bug report 761309 (http://bugs.debian.org/71309)
originally reported by Martín Ferrari:

According to the README, middle-clicking on the desklet would open the
image in an external program, but this does not work. It is a
combination of using an unescaped command-line to pass file names (which
fails if there are spaces), and not actually passing the filename, but
an object.

maxyz:
Avoid using unescaped file names by using spawn, store the path in order
to use it. fixes #5184


maxyz:
fixes #5184

Issue #4758, opened 2015-10-28
Backgrounds Setting Applets Slow to Start with >100 images


ion201:
When running the backgrounds settings application, I noticed that it would take a very long time to populate when there were > ~100 images in the target directory. I have one directory with 300 MB of >300 images and it would take at least 10 seconds for the settings menu to generate everything and would be less responsive while doing it. Since Cinnamon uses Python and PIL for the thumbnail generation, I (correctly) assumed that the bottleneck was recreating all of the thumbnails.

I made the following changes to cs_backgrounds.py to cache thumbnails by a sha1 hash of the original's and use a cached version if available. On my weak laptop and desktop, the the gui finishes loading thumbnails substantially faster.

A concise diff for my changes is below, I've made this an issue instead of a pull request because I'm well aware that my code is more of a quick and dirty hack than a proper implementation -_-
```
 files/usr/share/cinnamon/cinnamon-settings/modules/cs_backgrounds.py:

@17
< import hashlib
@476
<                 with open(filename, 'rb') as f:
<                     h = hashlib.sha1(f.read()).hexdigest()
<                 tmp_cache_path = os.environ['HOME'] + '/.cache/backgroundthumb/'
<                 if not os.path.exists(tmp_cache_path):
<                     os.mkdir(tmp_cache_path)
<                 cache_filename = tmp_cache_path + h
<                 if os.path.exists(cache_filename):
<                     img = Image.open(cache_filename)
<                     (width, height) = Image.open(filename).size
<                else:
...                  #Do the normal thumbnail thing
...                  ...
<                    img.save(cache_filename, "png")

```

dralley:
I also noticed this the other day when I was fixing #4658 , so I will try to find a "proper" solution and submit it as a PR

PR #5235, opened 2016-04-25
Persistently Cache Background Thumbnails

ion201:
This pull request is for issue #4758. It uses hashlib.sha1 to uniquely and deterministically identify images with their name and last modified time as below then caches the thumbnail in `$XDG_CACHE_HOME/cs_backgrounds/`

    h = hashlib.sha1(('%f%s' % (os.path.getmtime(filename), filename)).encode()).hexdigest()

Suggestions welcome if there is a better way I could be approaching this.


clefebvre:
Cool thanks @ion201.

PR #5243, opened 2016-04-26
Fix desktop effects on dialogs and menus

Cobinja:
This prevents desktop effects on dialogs and menus from being used if desktop-effects is turned off. It's like already like this in the effects settings dialog.


JosephMcc:
Thanks. Missed that one.

PR #5303, opened 2016-05-16
cinnamon-settings: Remove the hardcoded theming in the main cinnamon …

JosephMcc:
…settings icon view

This has never worked right in some themes since not all themes use the same convention for "selected_bg_color" or "theme_selected_bg_color"
Also fixes issues with how this looks under gtk3.20


clefebvre:
We need a style class and changes in Mint-X and Mint-Y to make it look the same way as before then.

Issue #5214, opened 2016-04-18
'Remove this applet' can easily remove the system tray applet


ACinnamonUserInUK:
**What happens.** Imagine that you have a program running that puts an icon in panel - say, the program TeamViewer. One wants to (completely) exit that program. To that end, one tries to right-click the panel icon; attempting this generates a 'remove this applet' option, which one might - somewhat unthinkingly - click. Result: the system tray applet is removed from the tray. Now, it is easy to put the applet back, and the message 'remove this applet' - rather than, say, 'quit TeamViewer' - would alert the user who is paying attention, i.e. would tell them what is really going on. However, I submit that it is too easy - I do it myself fairly often - to fall into this confusion, i.e. t_o close the tray when one means to close an application_.

**What should happen:** not sure. Perhaps show the message 'Remove this applet ('System Tray [or what have you])' rather than just 'remove this applet'?

Mint 17.3 x64 Cinnamon 2.8.8.

JosephMcc:
If you disable support for indicators in Cinnamon Settings->General, right clicking should bring up the actual menu from the teamviewer tray icon instead of the applet context menu.

ACinnamonUserInUK:
Aha. It is good to have that option. Why not, though, change the message in the way I suggested?

brownsr:
@ACinnamonUserInUK That's a function that is shared between all applets.  It would be a bit awkward to adapt systray to use its own local version, or find some other coding change.  Quite a bit harder then the simple wording change you might imagine it to be without looking at the code.  A better change might be to consider adding the applet name to the message.

ACinnamonUserInUK:
@brownsr: your notion of altering Cinnamon such that its 'remove this applet' option always reads 'remove this applet (<applet name>)' sounds like a good idea to me!

JosephMcc:
Fix merged.

PR #5318, opened 2016-05-19
[3.2] Add applet name to the applet context menu item that deletes the applet

brownsr:
fixes feature request #5214

### DIFF
--- a/js/ui/applet.js
+++ b/js/ui/applet.js
@@ -401,7 +401,7 @@ Applet.prototype = {
         let items = this._applet_context_menu._getMenuItems();
 
         if (this.context_menu_item_remove == null) {
-            this.context_menu_item_remove = new PopupMenu.PopupIconMenuItem(_("Remove this applet"),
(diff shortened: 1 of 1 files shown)


clefebvre:
It's not very elegant.

Remove applet "User Applet"
Remove applet "Sound"
Remove applet "Calendar"

Could we simplify it to "Remove %s" instead?

Remove "User Applet"
Remove "Sound"
Remove "Calendar"

f2404:
I agree.

And maybe also add applet names to other menu items? Or add a kind of title on top of them.
I mean, it would be elegant to show
```
About...
Configure...
Remove "Power Manager"
```
all items should be in the same style.

clefebvre:
You mean repeating the name of the applet in each item? that's a bit too much isn't it?

Maybe we could add a title on top of the context menu, with the name of the applet, and then leave the action items the way they were.

f2404:
Yes, I realized repeating was too much, and suggested to add a title with applet name.

JosephMcc:
Actually that whole "..." thing trailing the text could really go away as well.

brownsr:
will resubmit this, need to clean my git repository up to move this out of my master

PR #5324, opened 2016-05-21
Xlet settings overhaul

collinss:
Makes the following changes:
- gsettings widgets were refactored to separate the gsettings-specific code into it's own backend. This makes it easier to add new widgets, and makes it possible to add other backends. This in turn allows all widgets of a type to use the same code for ui without duplication for easier maintenance and a more unified appearance and behavior.
- A new JSON backend was created for the settings widgets.
- Several existing widget were improved to remove bugs, simplify code and/or add new features. The effects and keybindings in particular received a great deal of attention.
- Several other widgets were ported from the old xlet settings and are now available for use with or without a backend. 
- settings.js was mostly rewritten to simplify the code and remove the use of object.watch(). According to the MDN website, that function is costly in resources and should be used for debugging purposes only. (No tests have actually been performed to verify if there are any significant gains as of yet.)
- xlet settings now open in their own window/process
- xlet settings now matches the newer style of cinnamon settings
- now supports pages and sections in the settings file
- instances are now automatically highlighted in the panel when the instance is selected in the settings window

All changes are backward-compatible, which means that all applets that functioned before the change should still work.


muzena:
@collinss 
I think, it will be usefule if it possible that widgets have auto update, or at least notification to notify user for available widget updates or maybe this all update integrated through update manager.
I have installed a few additional desklet and extension and usually I notice that they have update after 2-4 weeks after new widget release.

collinss:
@muzena This pr doesn't effect installation/upgrade of applets in any way. I suppose it could, but I think it would be better as a separate pull request. It would be a handy feature though.

muzena:
@collinss ok i will then copy this post as a separate pull request.

lestcape:
@collinss I like a lot what you are doing here. 

I will request you if you can, do not call the cinnamon settings directly from the command line as is how actually occurs inside the applet, desklet class... Instead you can use an instance of an object and store it on Main. Then, you can call this object and this object will call the spaw command line. In that way, some cinnamon extensions can override the object, injecting some code, and redirect the settings call to another provider, for example an applet.

collinss:
@lestcape What do you mean by calling from the command line? Are you talking about when you select configure from the context menu? If so, you can already override it from the applet by defining a function called ```configureApplet``` in your applet object. Also, I don't know why you would store it in main unless you wanted to override it from something other than the applet itself, which IMO is a VERY bad idea. Either way, this would probably be best for a separate pr.

lestcape:
@collinss yes. What i want it's override the call of cinnamon setting applet page from an external applet (a setting applet). This is override all calls from applets and desklets  to the cinnamon-settings, NOT my own applet. I don't know why you say it's a bad idea, because anyway this is optional and only if you want to install the setting applet. In fact, what i considered a bad idea it's handle different programming language, different applications for a thing that can be done inside cjs and in the same application. Of course, this is also not perfect and i agree that's a good idea have an only one app to handle all settings, but in my opinion will be possible implement a virtual setting over dbus and then create a client cjs app. That's clear it's more hard, but not a VERY bad idea.

Benefits:

- Speed.
- Native drag and drop.
- Reduce the numbers of programming languages.
- Provide a native api to create/modify the settings on fly.
- Native theme handling using css.

Anyway you probably have right that's could be better from a separate pr.
Thanks.

lestcape:
Also right now i can iterate over all applets and desklet to search the configuration menu option, remove it and then add my own menu option in the same place. That could happend in the initialization of my applet and when another applet or desklet was added. Just this is not a fast way...

collinss:
@lestcape Why would you want to do this? If there is something missing from the settings api, we can add it. I'm just not sure why anyone would ever want to install such an applet. I can't think of a single good reason for it, and a lot of bad ones.

Think about it from a maintenance point of view. Such an applet would have to be updated regularly and would frequently break, whereas the api stays backward compatible, even with an extensive pr such as this one.

Then you have the performance issues - Clutter on js is much slower and resource intensive than Gtk on python. Add to that the fact that Cinnamon and it's applets currently run in a single thread, which makes it much more likely to cause crashes, blocking, and other issues.

From a user perspective, the implementation in this pr matches most of cinnamon settings in terms of style. Doing this in clutter definitely would not. Then you have the widgets which exist in Gtk, but not in Clutter. There is a lot of work that has been put into those widgets over many years to make them functional, stable, and seemless. You would not have any of that in Clutter, so such an implementation would be much more buggy.

From a programming perspective, python/Gtk is much easier to write and maintain than js/Clutter.

You are welcome to open a feature request or discuss it on IRC, but I'm guessing you will get a similar response from @clefebvre and @mtwebster

lestcape:
@collinss If what you say it's true, that i think is only a partially true... I will ask you, why you don't use Mate? Cinnamon API is over cjs, then as your own words is:  "**much more buggy**",  js is "**much slower**", "**programming perspective, python/Gtk is much easier to write and maintain than js/Clutter**",  "**applet would have to be updated regularly and would frequently break**". Are you really say all that? Because all that you are saying can be apply for Cinnamon in general... Anyway, you are not complete wrong, but you are in the wrong place, i think.

collinss:
The Cinnamon applet api is not as buggy because it also has had years to mature. Also, the more I work with Cinnamon, the more I realize that clutter was probably not best choice for the ui, but I don't use Cinnamon for the speed, or else I _would_ use Mate or XFCE or something.

If I were the one making the decision right now, I would probably go with Gtk for the ui. Gtk is a much more mature toolkit, and is much more well developed, used and supported. However, I'm not making that decision right now, and so I use what's there already. I'm not saying that Clutter is horrible, and that Cinnamon is unusably slow, but if we already use a toolkit for it that's better, why bother to switch to something that's not quite as good and ~~could potentially~~ would likely make things slower and more buggy unless there is something to be gained that would outweigh the risks/losses?

As for what I said about it breaking constantly, the problem is that it would be 3rd party, which means that any time someone makes a change to the internal workings, it could potentially break the applet, whereas all Cinnamon components would be updated to support the change. As an example, many of the releases (at least recently), have included some new settings types. When that happens, the js and python sides of the api are both updated in the same pr, but a 3rd party applet wouldn't get the same treatment. In addition, you have people using multiple versions of Cinnamon, and would have to support all of them with a 3rd party applet.

p.s. It's clutter that's slower - not js. I believe this is due to clutter being a 3d toolkit rather than a 2d one like Gtk.

lestcape:
@collinss now i can agree with you, but really i like more clutter than gtk, probably because clutter is completed over GL.

You don't test the speed difference. You don't know how faster could be the settings for applets and desklet inside clutter (because this mean inside the same application and mean a direct call from the same api).

I have warranty, because i tested it. If i take all time from the click to the complete displayed settings, from the configurable menu applet,  on clutter is at less 5 time more faster than the current cinnamon-settings implementation. I don't say your new pr.

Configurable Menu have the bigger setting configuration of all cinnamon extensions.

All this was tested with:
https://github.com/lestcape/Spices-Menu this applet have the ability for render the settings.

Anyway this is not all. Probably, the color chooser is the most difficult Gtk component that the cinnamon settings have. This no mean that can not be create a similar one in clutter:
https://github.com/lestcape/Color-Chooser What can not be created then?

collinss:
I think you're talking about a different speed than I am. Of course a native call is going to be faster than a dbus one, but everything you add to the clutter stage is going to use more resources. It will slow down Cinnamon itself, even if it's just a little. For most users that will not be a problem, but for older hardware, it could be.

Also, I don't think the little extra speed boost in updating settings is worth the potential loss in performance elsewhere. It's already fairly fast right now, and changing settings is something that isn't usually done on a daily basis.

As for the other widgets, I wasn't saying they couldn't be created, only that it would take a lot of work and years of extensive use and testing before they reach the level of maturity that Gtk widgets have.

lestcape:
@collinss yes, i agree now with you again, but again: The cinnamon API is **clutter** is **not** Gtk and all gnome works is for **clutter**, **not** for Gtk. Cinnamo is a fork of Gnome, **not** a fork of Mate and all this is a fact. Cinnamon it's for take risk, because it's: **"...built from modern technology and introducing brand new innovative features..."**, not for traditional GUI as it's Gtk. And as you know the last words are not mine..

lestcape:
@collinss don't worried, i has not plan to create a pr for change the cinnamon settings as before. I desisted of it a long time ago.... In all case, only a pr to provide a way for intercept a setting call and then use it in a 3rd party extension. In this way, we can all be happy. If this is interesting for users or not never was a question in this case, because is not an imposition, because is optional. Users can decide later what they consider better for his day to day usage.

lestcape:
For record propose only, cinnamon-settings is not slow because of DBus, it's slow because not have a bidirectional Dbus connection. The applet to Cinnamon-Settings connections not exist (only exist the settings to applet), but the most time losted it's because need to load all module at once(really a pre-load, because if perform full load would then be impossible the usage). After a pre-load it's when is decide the module to be loaded, a lot of time and resource was lost to open the settings applet page, loading the main page and pre-loading all modules.

But this is not all, python create a context to launch the GTK GUI on the initialization and this is slow, and take a big time. If I'm a cinnamon user and i click some thing i want a faster response, i don't care why is slow and all technical complications, the reason is not important, the fact it's. 

That's without taking in account the time to prepare the spaw context and all that is involved when you request open a new application from another.

lestcape:
@collinss maybe this is interesting to you. You can create a DBus signal in cinnamonDBus: https://github.com/linuxmint/Cinnamon/blob/master/js/ui/cinnamonDBus.js

This signal will be emitted when a xlet property changed, and will provide the property id and the new value (maybe also the old one).

Then, in the python side you can capture the signal instead (or using both mechanism, with a priority) of monitor the json file: https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/cinnamon-settings/bin/XletSettingsWidgets.py#L86

So, in that way, cinnamon-settings will be informed of a settings changed, several time more faster than monitor a file and not need to wait some milliseconds: https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/cinnamon-settings/bin/XletSettingsWidgets.py#L66  

Then you will have a bidirectional Dbus connection. this mean a faster communication between the cjs and python. This also mean, an improve like this: "settings.js was mostly rewritten to simplify the code and remove the use of object.watch(). According to the MDN website, that function is costly in resources and should be used for debugging purposes only. (No tests have actually been performed to verify if there are any significant gains as of yet.)"

collinss:
@lestcape I was aware of the ability to send a signal over dbus to speed up performance, however, I didn't implement it because I don't really see a use case where it is very important. When changing a setting from the settings page, you would expect the xlet to be updated immediately. However, it is unlikely that you would encounter a situation where changing the setting from the applet would require an immediate change in the settings page (why not just change it in the settings page then). Also, such a solution wouldn't address the situation where you have multiple instances of the settings page open, or when you change the file manually. We could certainly handle the first one with yet more bloat, but due to the possibility of a manual file change we would still need the file monitor, and quite frankly, I don't see the point. The delay is not significant enough, and IMO it doesn't serve a useful purpose. If there comes a time where this is actually needed, we can always add the necessary dbus calls later, but it's worked fine without this feature for a long time, and I don't think we need to add it now.

lestcape:
I know you are right....

The situation occurs when you call several bidirectional properties one behind another. Some of then will lost. Thats for example when you have several profiles like configurable menu and you try to change from one configuration to another,  you can not set all values of the new configuration....

But of course anywhere that demand a lot usage of the settings json file.

lestcape:
This is exactly why for me the cinnamon settings is not a useful feature, because what i need to do it's not common in first place, will stock with the settings limitation in the second or at last will be a VERY bad idea. How i need to go in this direction, what i found in cinnamon settings is more like a problem than a solution. Also i create a pr 2 years ago with a path for the bidirectional connection, but was never merged. Really do you think that this is true? "**If there is something missing from the settings api, we can add it.**" Sorry, but not...

lestcape:
For record: https://github.com/linuxmint/Cinnamon/pull/3368

lestcape:
@collinss you can see the more less the speed of cjs rendering the settings here: 

https://youtu.be/beIdxqtEmo0

Please take the start time after close the menu in both cases...

You will see what i say....

JosephMcc:
That may be faster but it sure is ugly. There is no way I would want to redo all of the settings widgets in St/Clutter. Not when we have the convenience of so many useful widgets already in gtk. I would also point out that accessing the settings with this PR should be faster since it no longer goes through the main cinnamon-settings interface to get to the settings. One of the slowest parts of opening cinnamon-settings is the initial icon view and that is completely bypassed in this PR.

All of the cinnamon settings are done in gtk so I don't see how it doesn't make sense for these to be as well. I agree with every point @collinss made. I personally feel like we would be better off in the long run looking at moving pieces out of cinnamons main thread instead of adding even more to it.

lestcape:
@JosephMcc of course it's ugly, because have not an style yet. If you are against clutter, the same it's for you, go then to other place where the api will not be Clutter. Or maybe, I'm is who are in a wrong place and i need to go from gnome shell where the devs at less love what they create. 

Cinnamon is probably one of the most popular desktops because of Clutter, not because of Gtk... But i can understand, that if you are not the Clutter dev, you could forget that clutter is really new and in development and try to go over a more stable Api. That's what people do always. This is why GNU linux is not the most used O.S.

Anyway, as i say, this is not an intention of change the cinnamon-settings, it's an intention of not used in my desktop and my applications other unrelated application, that have a lot of limitations, are slow and also have unresolved bugs. Because what i report here, it's a bug of cinnamon-settings, that occurs only in some context, it's not just slow, it's a bug of more than two years, with also a patch that not was merged, and apparently will continue and will not be fixed.

I have not much more to say... Really thanks for your comments... But i don't care any personal opinions of my applications, is they are made by people that not take on account my opinion about his applications and also are his application the cause that i need to create my applications...

Regards...

JosephMcc:
> @JosephMcc of course it's ugly, because have not an style yet. If you are against clutter, the same it's for you, go then to other place where the api will not be Clutter. Or maybe, I'm is who are in a wrong place and i need to go from gnome shell where the devs at less love what they create.

Sorry if you took the "ugly" comment as rude. It was not meant to be anything against you. St doesn't currently have the support for the widgets and styling you would need to create nice looking settings. That isn't your fault. The point is that it would huge amounts of work to do that which seems a bit of wasted time when Gtk already gives us everything we need.

> Cinnamon is probably one of the most popular desktops because of Clutter, not because of Gtk... But i can understand, that if you are not the Clutter dev, you could forget that clutter is really new and in development and try to go over a more stable Api. That's what people do always. This is why GNU linux is not the most used O.S.

I don't think Cinnamon is popular because it uses Clutter/St for some interface elements. Clutter is not that new either and is based on an even older toolkit. I also have serious doubts about the future of the standalone clutter/cogl libraries. The gnome devs have moved these to directly being a part of shell/mutter and it remains to be seen how much development/improvements these libraries will see outside of their embedded versions. Pair that with the fact that they plan to add clutter like capabilities to Gtk (a proper scene graph, OpenGL rendering) and it makes Gtk look more and more like the better option.

lestcape:
@JosephMcc ok, Clear i know it's hard and I mention that. Of course i know that Gtk also have this components. My applets and desklet's are not in Gtk and the settings of my extension are a part of my extension, more than a part of the settings. If the current cinnamon-settings implementation won't resolve the problems that i have, what i can do? Please give me a better choice, or do you think that cinnamon-settings it's the only solution in the world, and is perfectly?

What you try to said me? I'm wrong trying to create cinnamon extensions, applets or desklets? This because they used Clutter and probably the Clutter core will be merged with Gtk, and the standalone implementation of Clutter could disappear?

If this is what you try to said, please said it...!! Not matter, this is clear a good reason, and i will agree instantaneously with you. What you can not do, it's use the same arguments for your convenience. As said: Create applet, desklets and extensions for cinnamon is good (not matter if they are over clutter), but override our work in cinnamon-settings is not, because we are like gods and there are not a better or a different way to do what we create...

If the reason that you defend it's like you said... Ok then, please alert other cinnamon devs and if you can, also create a post in segment fault. They need to know that create cinnamon extensions it's not useful, because the extensions run in the main thread and because will be better create a cinnamon extensions in Gtk as a part of the settings that all this work for try to imitate Gtk components without styling. In fact, Gtk components can be embedded in a clutter context to be displayed...

In that case and in my opinion, cinnamon will disappear... Who will want to use it in this condition...?

JosephMcc:
@lestcape I'm have to say I'm struggling  a bit due to the obvious language barrier and I sometimes have a hard time deciphering exactly what your points are.

That being said I'm not sure there is much point in keeping on with this line of conversation here. You have already several times told people they should go somewhere else because they disagree with you. You're responses are also overly dramatic and taking things to the extreme. My comments on Clutters future were just thoughts. Things to think about. Right now it means nothing and maybe it will be a non issue moving forward. I don't know and probably no else does either.

The fact that all of the main Cinnamon components run in a single thread is an issue/limitation that has always been there. You know that. It would be silly for us not to consider ways we might improve on that.

As far as cinnamon settings being the only solution or a perfect solution, I never said that either. Stop putting words in my mouth. Right now it is the solution we have and it works decently. Does that mean it couldn't be better? Of course not. I just disagree that writing them in clutter/st is a better solution. My opinion and I'm entitled to it.

In the end, this PR improves things and thats what we want ultimately.

lestcape:
@JosephMcc I only say:

1- The Cinnamon-Settings have a bugs, that can be produced in certain circumstances.
2- I will want to make an applet like any other, as a replace of the Xlet part of the cinnamon settings
3- The Cinnamon-Settings is slow.

The theoretical controversy about Gtk and Clutter, is the problem, and not need to occurs in first place. At less, I don't have a solution to that and i think that don't used it, it's not the solution. That's why, if you don't want to use it, what will be the case of use cinnamon then? Clutter is the Cinnamon api...

I go to the extreme, because of language barrier, is a good way to show the point. Dramatical yes it's, but probably necessary, because in that way you understand me more easy what i say.

Of course, i put words in your mouse that you don't say. And when this happens it's because i can not see any other reason of why a cinnamon dev, won't recommended the usage of the cinnamon api, and the creation of a cinnamon extension... That's absurd to me... There are always problems in all places and that's not a reason to at less try... If you don't want to try let other do what they want.

collinss:
@lestcape Have you tried this pull request yet? As @JosephMcc mentioned above, the xlet settings is no longer part of cinnamon-settings, and I think you will find it to be much faster than before. I'm using a laptop that is over 10 years old and it loads xlet settings in about a second. On a newer computer it would probably be almost instantaneous. I don't think you'll be able to get much faster than that using clutter/st.

lestcape:
@collinss I'm pretty sure your pull it's at less 5 time more faster. For that reason and for the Xlet pagination it's why I say in the first comment: "I like a lot what you are doing here." ... But is clear you don't catch my real problem, maybe because i merge tow things:

1- I want a faster cinnamon-settings, i want help to detected and fixed bugs...
2- As the cinnamon settings it's an external application, i can not modify it to do what i really want in my extension.

You contribute a lot to the first point and i need to said thanks for that!!!

The seconds point have not a solution, because the cinnamon settings it's a different application and it's part of the core and that's ok to me. I can live with that,  but i can not do what i want in my extension in this way and that's NOT a fault of the Cinnamon-Settings conception or any other things related with any problem of the cinnamon-settings...

I have only one solution to do what i want, without break the compatibility with the cinnamon-settings at the same time. This is: Create my own settings without break the rules of the cinnamon-settings (Use the same json format to build the settings). The special modification to the settings can occurs later as i have a native connection with the settings.

How i need to do that in my extension, it's clear that make all this to be used only my extension it's not a good a practical idea. So, will not much more difficult make it general, to be used for all extensions, it's just redirect the cinnamon settings call to my programming logic...

I hope you understand now. Sorry if this go from the wrong way...

PR #5334, opened 2016-05-25
menu applet: added checkbox to show/hide recent files in the menu

NikoKrause:
The comparison between the old and the new code in the applet.js is a little bit confusing.

There are only 3 lines of new code in the applet.js, i.e.
`this.settings.bindProperty(Settings.BindingDirection.IN, "show-recent", "showRecent", this._refreshBelowApps, null);`
and
`if(this.showRecent) {
}`
Everything between the brackets of the if-case is the same old code, just indented.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1093,6 +1093,7 @@ MyApplet.prototype = {
         this.settings = new Settings.AppletSettings(this, "menu@cinnamon.org", instance_id);
 
         this.settings.bindProperty(Settings.BindingDirection.IN, "show-places", "showPlaces", this._refreshBelowApps, null);
+        this.settings.bindProperty(Settings.BindingDirection.IN, "show-recent", "showRecent", this._refreshBelowApps, null);
 
         this.settings.bindProperty(Settings.BindingDirection.IN, "activate-on-hover", "activateOnHover", this._updateActivateOnHover, null);
         this._updateActivateOnHover();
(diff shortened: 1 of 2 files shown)


collinss:
Thanks! This is something that has been bugging me about the menu for a while. A few notes though:
1. These changes should be in the same commit. In the future you can ```git add``` multiple files before you ```git commit```. Or else ```git commit --amend``` if you already committed and forgot something.
2. It's generally best to create a separate branch for a pull request so that your master branch can stay up to date with upstream master, and it's MUCH easier to submit another pull request later (trust me, I know ;) )
3. It looks like there are merge conflicts, which probably means that you made your changes on an older version of Cinnamon, and something has changed in the master branch that conflicts with this change. In the future it will be best to make sure your branch is up-to-date with master before submitting a pull request.
4. Cinnamon is a large project, with a lot of different parts, so It is generally preferable to specify which component is being changed in your commit messages. The title you used for this pr would be a good commit message for this change.
5. With Cinnamon 3.0 already released, there is a good chance this gets pushed back to 3.2, though I'm not the one that makes that call.
You can take these how you will. I'm only sharing these as one who has had to learn them the hard way.

NikoKrause:
I just see, that this checkbox was removed two years ago:
https://github.com/linuxmint/Cinnamon/commit/249c626b81466c4cdf4c5e7d068e3139d750136e

It is now able to hide recent files in the privacy-settings. 

@clefebvre: Maybe it should be mentioned in the settings of the menu, that it is possible to hide recent files that way?

@collinss: Thanks for your advice. My problem was, that I edited it online on github. I don't know how to make a branch only of the menu@cinnamon.org folder. It seems as there is only the option to fork the whole Cinnamon folder. Or is there a way, I don't know about?

collinss:
You will first need to make a fork and then create a branch on your fork. I think you can do it on github, but I'm not sure how. It's generally easier to do it using the command-line git tool on your local machine and then pushing your changes to github - especially if you contribute a lot.

PR #5342, opened 2016-05-28
Re-order and re-word idle-delay and re-word lock-enabled.

### LINKED ISSUE #103: make gnome-shell extensions compatible
Here is the workflow for getting gnome-shell extensions(most) working with cinnamon for now:
- copy gnome-shell extensions to ~/.local/share/cinnamon/extensions/
- Run the following commands:
  cd ~/.local/share/cinnamon/extensions/
  for i in `ls`; do
  sed -i -e "s@imports.gi.Shell@imports.gi.Cinnamon@g" $i/*
  grep -q cinnamon-version $i/metadata.json || sed -i -e '/{/{n;n;n;s/\,/\,\n"cinnamon-version": [ "1.1.3" ],/;}' $i/metadata.json
  done

Attached patch makes browser plugin work partially, i.e. https://extensions.gnome.org/local/ works fully, but installation of plugins does not work. I do not know any programming but getting the above workflow automated via browser plugin should be fairly simple for someone who knows how to code :)

In short, when user selects the extension to install on extensions.gnome.org, the plugin should download  it and make the above changes automatically.

http://paste.opensuse.org/93726837

marga-debian:
This fixes linuxmint/cinnamon-screensaver#103 by making the strings clearer.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_screensaver.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_screensaver.py
@@ -69,25 +69,27 @@ class Module:
         page = SettingsPage()
         self.sidePage.stack.add_titled(page, "settings", _("Settings"))
 
-        settings = page.add_section(_("Lock settings"))
+        settings = page.add_section(_("Screensaver settings"))
 
         size_group = Gtk.SizeGroup.new(Gtk.SizeGroupMode.HORIZONTAL)
+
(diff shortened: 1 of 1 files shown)


marga-debian:
This is how it looks:
reworded_screensaver_settings

It seems to me that with this wording the actual purpose of the settings is notably clearer. idle-delay is not actually used for locking and so it shouldn't be inside the locking settings nor should it talk about locking.

clefebvre:
Thanks.

This should be reviewed for 3.2. Other WIPs will affect the screensaver there and might impact its design. Also, it's late on 3.0.x to introduce new msgids in translations.

anandrkris:
@mtwebster - We were discussing on IRC earlier on the confusing terms at the moment. This is much clearer IMO.

PR #5362, opened 2016-06-07
Resize a thumbnail when the window resizes

AlbertJP:
I tiled a window by keyboard shortcut while the window preview was open, and the preview became stretched. That is fixed with this pull request.

The problem is seen in this image:
mint-y-bugs


medude:
I just tested this out and it works perfectly.

PR #5369, opened 2016-06-11
Window-list applet: make icon boxes wider

efournival:
Boxes in the window list are really narrow and look small on modern screens.
This commit change the default width of 150 pixels to 250.

### DIFF
--- a/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
@@ -70,6 +70,8 @@ const ICON_HEIGHT_FACTOR = .64;
 const MAX_TEXT_LENGTH = 1000;
 const FLASH_INTERVAL = 500;
 
+const ICONBOX_WIDTH = 250;
+
 const WINDOW_PREVIEW_WIDTH = 200;
 const WINDOW_PREVIEW_HEIGHT = 150;
 
@@ -533,10 +535,10 @@ AppMenuButton.prototype = {
 
         if (this._applet.buttonsUseEntireSpace) {
             let [lminSize, lnaturalSize] = this._label.get_preferred_width(forHeight);
-            alloc.natural_size = Math.max(150 * global.ui_scale,
+            alloc.natural_size = Math.max(ICONBOX_WIDTH * global.ui_scale,
                     lnaturalSize + naturalSize + 3 * 3 * global.ui_scale);
         } else {
-            alloc.natural_size = 150 * global.ui_scale;
+            alloc.natural_size = ICONBOX_WIDTH * global.ui_scale;
         }
     },
 


JosephMcc:
So I just tried this and have to say I'm not a big fan. Just a few open windows crowds up the entire taskbar. In addition to that you can quite easily accomplish something like this already. In the window list simply enable "Window buttons can have different sizes and use the entire space available". The end result is very similar and a user isn't stuck with it.

efournival:
Well here is how it looks with this option enabled:
capture du 2016-06-11 23-54-21
For me, it's far from being very similar to wider icon boxes. Having different widths looks quite strange and even break the width of the whole panel (look at how my clock is going too much to the right).

Without (and with the default 150 pixels width):
capture du 2016-06-11 23-57-27

And with my patch applied:
capture du 2016-06-12 00-01-00

I guess it depends a lot on the screen width you have. The screenshots have been taken on my laptop which have a 1920 pixels wide screen, and it's already too small (for me). So you can imagine easily what's going on on my 2560 pixels wide desktop screen...
I can test with other desktop environements to see what policy regarding this they are following but I don't remember having to change this when I was using MATE.

JosephMcc:
I'm using a 1920 wide screen as well and like the width they currently are over how it is with this patch. I am running it on 22 inch monitors though.  When your talking about resolutions above that your looking at getting into the range of HiDPI which gets handled differently. 

On top of that I think less people then we sometimes assume use monitors at this high of a resolution. A lot of people use inexpensive laptops and older where 1366x768 is a super common size.

I'm not a huge fan of an optional setting for every little thing but maybe allowing a configurable size in the settings would be an option? You would need to get some feedback from other devs and I personally would prefer the current size left as the default.

efournival:
Unfortunately my 2560*1600 monitor looks too big with HiDPI enabled but that's another topic...

I think 1366*768 is as common as 1920*1080 but the first should only decrease in the future.

Personnaly, I really don't like the default size so I'm manually changing (and thinking of making a script to do that) the width of the boxes each time Cinnamon is updated, which is quite common under Arch Linux.

I agree with you that a new setting would be a good idea.
I can code it myself so I guess I'm waiting for some other feedbacks before doing anything.

JosephMcc:
At worst case, instead of fixing this with every update you could copy the window list applet and rename it. You could make the changes you want and just use it instead of the built window list. That way it wouldn't get over written by the constant updates on Arch.

efournival:
Thanks for your suggestion, I did not think about that, will do.

Also, I just had a look at my settings and I just saw that I have 1.2 UI text scaling and a custom panel height on both of my computers.

Even without all those extra parameters, icon boxes still look too narrow for me.
However, it seems we can fix this situation while still pleasing everyone by scaling the boxes' width to their height at a constant ratio when 'allow Cinnamon to resize text and icons...' in panel settings is checked.

What do you think? It seems to be a perfect fix to me (even if my setup is quite "special").
What do others think?

clefebvre:
Hi,

Please feel free to set up an issue to discuss this further. I'm closing the PR as it's a bit old and the current solution isn't accepted.

PR #5402, opened 2016-06-24
[3.2] Scale window buttons to fit the screen

medude:
I inserted new code into the window-list applet ([View master](https://github.com/linuxmint/Cinnamon/tree/master/files/usr/share/cinnamon/applets/window-list%40cinnamon.org)) to re-scale buttons for different screen sizes.

| Resolution  | With                                                                      | Without                      |
|----------------|--------------------------------------------------------------|----------------------------|
| 1920x1080 | 1920x1080 | 1920x1080 without |
| 1600x900 | 1600x900 | 1600x900 without |
| 1280x720 | 1280x720 | 1280x720 without |

(Sorry for any strange artifacts in the 1920x1080 and 1280x720 images; my screen's resolution is 1600x900.)


clefebvre:
That makes them lose consistency (i.e. they start having different sizes then). I'm not saying that's bad or anything, but there's already an option which does that in the window-list applet.

PR #5426, opened 2016-07-07
Add button to remove Xlets

HenryJW:
Added Remove button for Xlets (applets, desklets, etc.). Button is disabled if there are no instances.

remove button

remove button disabled


anandrkris:
Should "Uninstall" button be available as well? (Available only in right-click Context menu at the moment)

HenryJW:
I don't think users uninstall often enough to need an Uninstall button. But I could be wrong.

anandrkris:
Hmm. Thinking if these actions  can perhaps come as an icon against each row. Would it look clumsy?
Eg: Display Uninstall icon where it is not Locked.

HenryJW:
I too think it would look clumsy. Unless someone can come up with a really good design for it.

PR #5495, opened 2016-07-29
Create Yubico

gh4snow:


itzexor:
This sounds like a feature request or bug report rather than a code contribution. If that is the case, then open an issue instead. Thanks.

PR #5499, opened 2016-07-30
Ensure a long applet description does not push applet status icons off the settings screen

brownsr:
A long applet description will push the applet status icons off the visible applet settings screen. This could confuse people. This PR wraps the applet description at a sensible point to avoid this happening.

Sorry about the clutter, I rather dumbly put a small change in master which carries it into every other PR.  reverted now, but my git skills are not able to remove all the evidence!  Can just cherry pick 	d7ad25b

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/ExtensionCore.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/ExtensionCore.py
@@ -130,6 +130,9 @@ class ExtensionSidePage (SidePage):
             column3.set_max_width(300)
             cr.set_property('wrap-mode', Pango.WrapMode.WORD_CHAR)
             cr.set_property('wrap-width', 200)
+        if self.collection_type == 'applet':
+            cr.set_property('wrap-mode', Pango.WrapMode.WORD_CHAR)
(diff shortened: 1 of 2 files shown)


JosephMcc:
For such a simple commit why not just redo it? I nearly always use `git rebase -i HEAD~<numberofcommits> before submitting a PR to clean things up. A decent explanation here: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

brownsr:
@JosephMcc Here my git inexperience shows. I've squashed things on branches as you suggest, before.  However in this case the commits I want to squash are on my master, separated over time,  and rebase -i  is showing me all the other people's commits in between them.  That makes me very wary of making things worse.  I actually did entirely redo the change on a new branch, with a single commit, but sending the PR pulls the commits out of my master as well, even though they net out to nothing.  I do not know how to either get rid of them or to suppress them.  I have certainly learnt never, ever to make a change direct on master again, no matter how trivial.

clefebvre:
Hi @brownsr 

Use a separate branch for each PR to avoid this problem.

Can we close this PR and make a new one?

PR #5500, opened 2016-07-30
More informative menu item to remove an applet 

brownsr:
This small change introduces the applet name into the right click applet menu item that removes the applet, so instead of "Remove this applet"  the user will see "Remove 'Systray'"  for example. I hope this will make a user less likely to remove an applet in error.  Sorry about the git clutter caused by moving the original PR which I unfortunately put in master, my limited git skills do not run as far as removing the evidence.  Perhaps you can just cherry pick 43f99d4 ?


clefebvre:
Same thing as previous PR..

PR #5538, opened 2016-08-10
Update cinnamon-settings-startup.desktop

welblade:
Fix wrong name in brazilian portuguese language


clefebvre:
Thanks, but there's no point fixing it here. It would be overwritten again. The translations happen at https://translations.launchpad.net/linuxmint/latest/. Please fix it there instead.

welblade:
I have fixed it in the translation before the "sarah" release but the name didn't change in menu, so the name in menu and in the control center are diffent now.

itzexor:
@clefebvre Someone needs to regenerate those files with the scripts in the repo. They were last generated by @mtwebster in May and it looks like the update on launchpad was in June. 

Is running the python regeneration scripts in the repo root part of the build process for Mint? They aren't currently regenerating during build on Arch with my PKGBUILD, nor the arch stock PKGBUILD, as far as I can tell. The mint-common dependency for the rebuild scripts isn't currently packaged by anyone for Arch, so I'm pretty sure non-English Arch users are getting whatever is in the repo at release. This may hold true for other distros as well, but I don't know. 

Should we regenerate those files before every release? Or maybe make it a mandatory step in the build process so non-English users don't get mismatched translations in cases like this?

PR #5551, opened 2016-08-14
settings.js: fix some settings not updating due to unspecific conditi…

collinss:
…onal


itzexor:
@mtwebster this should be merged, as it is pretty broken without it

Issue #4755, opened 2015-10-27
Allow switching media players in the new sound-applet


ManIVIctorious:
I have to admit that I am only using the sound applet of cinnamon 2.8 with cinnamon 2.6 so excuse me if this doesn't appear on 2.8

When using the soundapplet of 2.6 you could control more than one player simultaneously (you could click the Name of the active player and the applet expanded to switch betweeen running players)
This functionality doesn't seem to be implemented/exposed on the new applet.

Additionally i discovered that there is still a setting if the timer of the seeker bar should count upwards or downwards. Since the seeker doesn't have a timer anymore this setting should be removed or even better the timer should be readded

ManIVIctorious:
Just upgraded to cinnamon-2.8.5. Seems to still be an issue. The handling of multiple applications was a really neat feature, as it is now you only can control the last started application with the soundapplet.

msjenkins1:
I too have had this concern. I'm currently running LM 17.3 with Cinnamon 2.8.6 and the feature is still missing. Don't get me wrong, the applet is sharp as hell and I love it, but hitting pause when getting a call or something similar all too often yields an undesirable result with a podcast AND a loud song playing instead of pausing audio.

marcus-lundgren:
I believe this behavior was introduced in this commit: a85f67c
I have a semi working fix to this issue, that will introduce a menu of applications from which one can choose which application the controls should be based on. I don't remember if it will look identical to how it was before, but I hope it will be decent/intuitive GUI wise. Stay tuned :-)

marcus-lundgren:
This issue should probably be closed since #5556 has been merged.

NikoKrause:
> Additionally i discovered that there is still a setting if the timer of the seeker bar should count upwards or downwards. Since the seeker doesn't have a timer anymore this setting should be removed or even better the timer should be readded

https://github.com/linuxmint/Cinnamon/issues/5848

PR #5556, opened 2016-08-14
sound applet: Add a menu that allows the user to switch between activ…

marcus-lundgren:
…e players

# Introduction

The intention of this change is to fix the regression of not being able to switch between player controls of application in the sound applet. This regression was reported in the following issue: #4755

I based some of the code on what was removed in a85f67c after having done most of the work, but most of it is "original" and works according to my own impression of how it should look like. Please let me know if you have any comments on either the behavior or the GUI.

# How it works

When there is less than two players active, the sound applet will look the same as it does at the moment. As soon as there are two or more players active, a sub menu will appear at the top of the applet. In this list, the user is able to choose which application controls should be displayed. In order to let the user know which controls are showing, a dot will be present on the left of the currently selected application (similar to how the "Output device" menu works).

A screenshot of how it looks with both VLC and Spotify running:
imgur link


HermannDppes:
Spotify currently crashes randomly, leading me to detect an issue with this.

I'm on Cinnamon 3.0.7 and copy-pasted this over my sound applet. (Works beautifully otherwise, so far.)

Steps to reproduce:

* 1. Open two players. (Spotify and VLC in my tests.)
* 2. Kill one (`$killall spotify` or `$killall vlc`)
    *Ideal result:* Killed player vanishes.
    *Actual result:* Killed player still shown in selection.
* 3. Select killed player.
    *Expected result:* Applet detects mistake, takes steps to correct.
    *Actual result:* Killed player gets “selected” dot.
* 4. Select unkilled player.
    *Expected result:* Chosen player selected.
    *Actual result:* Nothing happens.
* 5. Start killed player again.
    *Expected result:* I don't even know anymore. Everything works again?
    *Actual result:* Player shows double, only selected once. Still can't be unselected.

**Note:** This issue appears if player is restarted *before* selecting the dead version, too. With the same end result.

**Current workaround:** Alt+F2, “r”

So far I have not been able to produce a player showing triple, but I could get both VLC and Spotify to show double.

Last, but not least: Thank you for this patch, it has increased my appreciation for this applet immensely.

marcus-lundgren:
Thank you for the feedback!

I was unable to exactly reproduce the behavior you described. But I did however remember something similar happening to me before, which seemed to be due to Cinnamon crashing. By crashing Cinnamon, I was able to reproduce a behavior similar to what you described.

I've made another commit that will remove entries from the list if the associated player doesn't seem to exist anymore. It also includes a fix for an issue that could arise when a "owner change" happened.

Could you try and see if this fixes your issues? And just as an FYI, you can put this applet in **~/.local/share/cinnamon/applets** instead of overwriting the actual player in **/usr/share/cinnamon/applets**. :-)

HermannDppes:
I can't test properly right now, but first impressions: When killed with `killall` players usually vanish immediately, got a dead player only once. That vanished properly when clicked on, focus returned to only other player.

***tl;dr:*** Looks good so far.

Thanks for he FYI, seems a lot saner than my method.

ManIVIctorious:
I hope this gets merged before cinnamon 3.2

mtwebster:
Hi, thanks for this - I've been testing some other things and ran into the fact I couldn't change running players.  A bit of feedback:

I'm getting the following in ~/.xsession-errors when I close a player (either via the applet x or the player itself)  - 

```
(cinnamon:2307): Cjs-WARNING **: JS ERROR: Exception in callback for signal: destroy: Error: No signal connection 4 found
_disconnect@resource:///org/gnome/gjs/modules/signals.js:74
PopupMenuBase.prototype.addMenuItem/<@/usr/share/cinnamon/js/ui/popupMenu.js:1847
_emit@resource:///org/gnome/gjs/modules/signals.js:140
PopupMenuBase.prototype.destroy@/usr/share/cinnamon/js/ui/popupMenu.js:2008
Player.prototype.destroy@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:834
MyApplet.prototype._removePlayer@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:1247
@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:938
_emit@resource:///org/gnome/gjs/modules/signals.js:140
_convertToNativeSignal@resource:///org/gnome/gjs/modules/overrides/Gio.js:129


(cinnamon:2307): Clutter-CRITICAL **: clutter_actor_remove_child: assertion 'child->priv->parent != NULL' failed

(cinnamon:2307): Cjs-WARNING **: JS ERROR: Exception in callback for signal: destroy: Error: No signal connection 7 found
_disconnect@resource:///org/gnome/gjs/modules/signals.js:74
PopupMenuBase.prototype.addMenuItem/<@/usr/share/cinnamon/js/ui/popupMenu.js:1847
_emit@resource:///org/gnome/gjs/modules/signals.js:140
PopupMenuBase.prototype.destroy@/usr/share/cinnamon/js/ui/popupMenu.js:2008
Player.prototype.destroy@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:834
MyApplet.prototype._removePlayer@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:1247
Player.prototype._dbus_acquired/btn<@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:314


(cinnamon:2307): Cjs-WARNING **: JS ERROR: Exception in callback for signal: destroy: Error: No signal connection 7 found
_disconnect@resource:///org/gnome/gjs/modules/signals.js:74
PopupMenuBase.prototype.addMenuItem/<@/usr/share/cinnamon/js/ui/popupMenu.js:1847
_emit@resource:///org/gnome/gjs/modules/signals.js:140
PopupMenuBase.prototype.destroy@/usr/share/cinnamon/js/ui/popupMenu.js:2008
Player.prototype.destroy@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:834
MyApplet.prototype._removePlayer@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:1247
Player.prototype._dbus_acquired/btn<@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:314


(cinnamon:2307): Clutter-CRITICAL **: clutter_actor_remove_child: assertion 'child->priv->parent != NULL' failed
```

It could be you're destroying something twice.  

.xsession-errors will usually contain errors like this - things related to introspected libraries rather than javascript-specific stuff ( which tends to end up in .cinnamon/glass.log)

One other observation, I'm not sure I'm crazy about having an extra line there simply to tell me there are multiple players.  What about having the expander on the currently active player (so during normal use, the player looks the same as without this chooser, except with the addition of the expander arrow).  You expand it, and can select another player, and that one now becomes the visible player, the expander collapses automatically behind you.  Just something to consider - I was gonna merge this anyhow, and worry about it later, but for the errors.

marcus-lundgren:
I am able to reproduce this error with the existing sound applet, which suggests that it was not introduced in the changes I've made. For me, the easiest way to reproduce this was to have two players active, and then closing them one by one with maybe a small number of seconds apart (haven't tried with faster/slower, as time is sparse). Are you also able to reproduce it in the currently released version?

I've thought about having the player be an expander itself. But for me, the list approach seems nicer. Then the items don't jump around whilst selecting a new player, which means that you can click through them all as you wish. It would also mean that the close button wouldn't be able to be next of the player name. The close button could of course be moved, but I wanted to do as small changes as possible to the existing behavior/look of the applet.

marcus-lundgren:
Nevermind, read one of the stack trace types wrong. One of the stack trace types you have suggests that when using the close button of the active player in the applet could be problematic. I was unable to reproduce the issue, but since it seems to works just fine without the call so I've simply removed it. This should fix the stack trace originating from:

    Player.prototype._dbus_acquired/btn<@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:314

Please let me know if you still have issues related to this.

The other stack trace type is the one that I've been able to reproduce using the original version of applet however. It has something to do with how the DBus listener handles owner changes. My knowledge of how this should work is beyond me at the moment, so I don't feel confident enough to attempt at a fix for it in this PR.

clefebvre:
Hi @marcus-lundgren. Sorry it took me so long to get involved in this PR.

I like it but I'm having cold feet. 

I observed two issues when testing this but I'm unable to reproduce them always.

The first one isn't critical, on my first try I already had Spotify opened. The applet showed me the option to switch between two "spotify" players. After switching from spotify to spotify (hehe), the choice disappeared as if the applet understood there was only one player.

The second one is much more serious. I get random Cinnamon crashes when switching controls between Spotify and Banshee. 

`(cinnamon:2645): St-CRITICAL **: st_theme_context_get_for_stage: assertion 'CLUTTER_IS_STAGE (stage)' failed
Window manager warning: Log level 8: g_object_get: assertion 'G_IS_OBJECT (object)' failed

(cinnamon:2645): St-CRITICAL **: st_theme_context_get_for_stage: assertion 'CLUTTER_IS_STAGE (stage)' failed
Window manager warning: Log level 8: g_object_get: assertion 'G_IS_OBJECT (object)' failed

(cinnamon:2645): Cjs-WARNING **: JS ERROR: Error: cairo error on context: "invalid value (typically too big) for the size of the input (surface, pattern, etc.)" (32)
BoxPointer.prototype._drawBorder@/usr/share/cinnamon/js/ui/boxpointer.js:287


(cinnamon:2645): Cogl-ERROR **: Failed to create texture 2d due to size/format constraints
Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged.

** (gnome-panel:27866): WARNING **: Failed to register client: GDBus.Error:org.gnome.SessionManager.AlreadyRegistered: Unable to register client

(cinnamon:27927): Cjs-WARNING **: JS ERROR: Exception in callback for signal: destroy: Error: No signal connection 7 found
_disconnect@resource:///org/gnome/gjs/modules/signals.js:74
PopupMenuBase.prototype.addMenuItem/<@/usr/share/cinnamon/js/ui/popupMenu.js:1847
_emit@resource:///org/gnome/gjs/modules/signals.js:140
PopupMenuBase.prototype.destroy@/usr/share/cinnamon/js/ui/popupMenu.js:2008
Player.prototype.destroy@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:833
MyApplet.prototype._removePlayer@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:1266
@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:940
_emit@resource:///org/gnome/gjs/modules/signals.js:140
_convertToNativeSignal@resource:///org/gnome/gjs/modules/overrides/Gio.js:129
`

marcus-lundgren:
Thanks for the feedback, @clefebvre.

The first issue is one I've experienced mostly when the applet has been readded to the panel or when Cinnamon has been restarted/crashed. A possible fix could be to redo the whole list of available players based on the list of player objects that exists, but I am not really sure when this should be done (unless there is a good signal/method that is called when left clicking the applet). I've used this applet since August and can't recall having those issues outside of those scenarios mentioned earlier, which makes it at least for me not that big of an issue.

I guess that it is only the last stacktrace that you attribute to my applet. Please correct me if I am wrong in this assumption. This stacktrace is identical to the one that @mtwebster referred to, which I've been able to reproduce with the current version of the applet shipped in stable Cinnamon. This would therefore suggest that the issue wasn't introduced in this PR. Would you mind testing out the current stable version of the sound applet and see if you can also recreate the issue there?

clefebvre:
I'm able to confirm this exception on master:

(cinnamon:28585): Cjs-WARNING **: JS ERROR: Exception in callback for signal: destroy: Error: No signal connection 4 found
_disconnect@resource:///org/gnome/gjs/modules/signals.js:74
PopupMenuBase.prototype.addMenuItem/<@/usr/share/cinnamon/js/ui/popupMenu.js:1936
_emit@resource:///org/gnome/gjs/modules/signals.js:140
PopupMenuBase.prototype.destroy@/usr/share/cinnamon/js/ui/popupMenu.js:2097
Player.prototype.destroy@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:833
MyApplet.prototype._removePlayer@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:1230
@/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js:939
_emit@resource:///org/gnome/gjs/modules/signals.js:140
_convertToNativeSignal@resource:///org/gnome/gjs/modules/overrides/Gio.js:129

It happens if I run two players and I terminate the player tracked by the applet. It doesn't crash cinnamon though.

clefebvre:
Getting these two on master.. same thing:

(cinnamon:28585): St-CRITICAL **: st_theme_context_get_for_stage: assertion 'CLUTTER_IS_STAGE (stage)' failed
Window manager warning: Log level 8: g_object_get: assertion 'G_IS_OBJECT (object)' failed

clefebvre:
I can't make it crash anymore :)

The other day I pasted this:

> (cinnamon:2645): Cjs-WARNING **: JS ERROR: Error: cairo error on context: "invalid value (typically too big) for the size of the input (surface, pattern, etc.)" (32)
> BoxPointer.prototype._drawBorder@/usr/share/cinnamon/js/ui/boxpointer.js:287
> 
> (cinnamon:2645): Cogl-ERROR **: Failed to create texture 2d due to size/format constraints
> 

There's no boxpointer in Cinnamon anymore.. maybe that's why it no longer occurs. In any case, it looks good and ready to merge.

PR #5560, opened 2016-08-16
Menu applet: Show 'Run with nVidia GPU' in context menu

rapgenic:
With this commit Cinnamon menu applet shows in the right click context menu a new option: "Run with nVidia GPU", useful in systems using Bumblebee to manage the GPUs. 

I made this change because I found annoying the fact that I had to open a terminal, than type _optirun ..._ and keep the terminal window in the background while working. This solution is much cleaner: right click, run with nVidia, and you're done.

It isn't too much annoying, in fact it only shows up when _optirun_ (part of Bumblebee) is installed.

I could have made a more general mechanism, such as nemo actions, but I don't think it is necessary, because, apart from adding to the favourites, adding to the panel, adding to the desktop, uninstalling and running with dedicated GPU, there aren't many other possible actions for an application.


mainmachine:
Excellent feature, thanks for the good work! :)

anandrkris:
Thanks for this.

rapgenic:
You're all welcome! :)

PR #5627, opened 2016-09-02
xlet settings: Replace old binding function in settings.js with two n…

collinss:
…ew ones

The primary reason for this is that the binding direction is no longer meaningful internally due to the use of javascript properties rather than using a watch function. For that reason, the 'bindProperty' function is now considered deprecated.

In its place are two new functions:
bind works the same as bindProperty, but without the need to pass a binding direction as the first argument.
bindWithObject works identically to bind but allows the caller to specify the object to which the property will be bound.

The bindProperty function still works as a wrapper around bind, but it is recommended that bind or bindWithObject be used in newer code.

The system applets and desklets were updated to reflect this change.


collinss:
I'm going to close this for now as something occurred to me recently - the binding functions currently only support one binding per key, but I can think of a number of situations in which more than one per key is desirable. It should be fairly easy to implement. I'll create a new pr when I do.

PR #5628, opened 2016-09-02
Refactor sound chooser widgets to simplify and avoid unnecessary dupl…

collinss:
…ication of code


lestcape:
@collinss can you see if can be implemented something like this:
https://github.com/linuxmint/Cinnamon/issues/5659

PR #5631, opened 2016-09-02
Fix lens shape options from showing when lens mode is off

RavetcoFX:
This will happen at the startup of the accessibility module when zoom settings is on.


RavetcoFX:
Closed because it magically fixed itself

PR #5677, opened 2016-09-11
Vertical panel improvements

JosephMcc:


LinuxHax00r:
This is not a very sensible change as it reflects the sheeple quality that thou bring for being a mindless Ubuntu user and censorer of the great minds here in the linux community such as thyself with all the great ideas thy bring. So one would expect an apology to be brought thereforth so all communique can thusforth be mended. :)

clefebvre:
"This applet is not suitable for vertical panel" --> doesn't say which one :)

My network (above) and keyboard applets (just below) (you need two keyboard layouts to see that one) touch each others.

clefebvre:
not sure why, but I have one panel at the bottom.. when I created a new (on the right side), it automatically received all the applets I already had.

When creating a third and then fourth panel, these were empty.

JosephMcc:
@clefebvre can you reproduce those issues on git master? This PR shouldn't have any effect on those things. It sounds like you turned up some other issues that need to be addressed.

PR #5679, opened 2016-09-11
Update cinnamon-global.c

LinuxHax00r:


lestcape:
@LinuxHax00r  what you are trying to do?

Please, there are a lot of people here, working to create a product for other people... They are working because they want, not for money.  They listen to the users, but they need to take on account all users, not only one. They have his own preferences, because they are human and because always you need to have a plan of what they want to do and what not.

In github it's possible block all ways to report issues and not allow user to interact with him. This is what you want to happen?

Please, this is not a community forum. It's a development site of a free product, ours personal consideration are always taking in account but never will be the final decision. You will really hinder the workflow here,  If you do not understand this.

PR #5688, opened 2016-09-14
removable-drives applet: don't hardcode Nemo

City-busz:
Use Gio to open the default file manager at 'computer:///'.

Note that nemo is not hardcoded anywhere else on the panel, so please consider to not hardcode it for this menu item too.

### DIFF
--- a/files/usr/share/cinnamon/applets/removable-drives@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/removable-drives@cinnamon.org/applet.js
@@ -1,6 +1,7 @@
 const Lang = imports.lang;
 const St = imports.gi.St;
-const Cinnamon = imports.gi.Cinnamon;
+const GLib = imports.gi.GLib;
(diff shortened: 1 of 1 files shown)


clefebvre:
Hi,

Please point to ~ instead of computer:///

City-busz:
Are you sure? This is a menu item of the removable-drives applet, and I assumed that it would better to open the file manager at location where the devices (including removable drives) are listed.

clefebvre:
Yes, the current behaviour was to launch Nemo and that showed the home directory by default.

I understand your point although I'm not really happy with computer:/// as it is now. It's hidden away in Nemo and that's a good thing because it doesn't really provide anything valuable. It's more ambiguous than the sidebar, doesn't show space being used, shows devices which cannot be browsed etc.. 

We could consider improving/redesigning a better overview/computer screen and making it more prominent within the navigation options, but that's beyond the scope of this PR.

City-busz:
Okay, I updated my PR.

PR #5692, opened 2016-09-14
Remove panel barriers on panel removal

Cobinja:
This was not done before, now it is.


clefebvre:
hi @Cobinja, can you explain how to reproduce the issue? These barriers are what stop the mouse from going from one monitor to another?

Cobinja:
You can reprod it by removing a vertical panel adjacent to a horizontal one. The barrier between those panels stays after removing the panel. Its only in the direction vertical -> horizontal

PR #5703, opened 2016-09-18
[show-desktop applet] added feature "Peek at desktop"

NikoKrause:


NikoKrause:
That's how it looks like:
peek_at_desktop

clefebvre:
The keyboard applet change is already there isn't it? https://github.com/linuxmint/Cinnamon/pull/5346 I know it's been there a while, sorry it's taking so long to review it.

clefebvre:
What's the rationale behind peeking at the desktop? I mean, what do you use it for? In which case is it useful?

NikoKrause:
Sorry, I mixed the pull request up a little bit. Should have created a new branch for keyboard applet back then.
I'll try to fix it.

For example: Sometimes I download or save files to the desktop and have a bunch of windows open. And I want to be sure the file is really saved on the desktop, then I take a glance at the desktop. I now it's also possible to just press on the button. But I'm to lazy to do that ;)

I think there are also more capabilities for that feature. But they don't come to my mind yet.
That's why this feature is optional. By default it's turned off and it's to have it, if you want to use it.

NikoKrause:
I deleted the changes from the keyboard-applet.
Is this okay now? 

I just wanted to propose this feature to you for discussion.
If after discussion it turns out, that the feature is not needed/wanted, you can decline this pull request.

NikoKrause:
This feature is adapted from an applet on cinnamon spices website:
https://cinnamon-spices.linuxmint.com/applets/view/165

That applet is rated in Top 10. So there are quite some people who use it, I think :)

JosephMcc:
@NikoKrause not sure how familiar you are with github but you should squash all of this down into a single commit. It keeps our history on git master much cleaner.

When you squash it you have the opportunity to update the commit message. When you do add a brief one or two sentence description of what this adds in it. Here is an example of what I mean: https://github.com/linuxmint/Cinnamon/commit/ef1cfead7e2bcbd1e63783d15e6def56fcf6f0e5

PR #5705, opened 2016-09-18
[show-desktop applet] new feature added: Peek at desktop

NikoKrause:
This feature allows you to take a peek at the desktop, if you hover with your mouse over the show-desktop applet. (If you want to take a quick look at your files or desklets (for example weather) stored on the desktop and you are to lazy to press the show-desktop applet for that ;)
peek-at-desktop

You are in control of the following settings:
peek_settings

(1) It's up to you using this feature. Turned off by default
(2) Blur effect for the transparent windows, when peeking at desktop
(3) Time you have to wait, till windows are getting transparent, after you hovered over the applet
(4) You decide how transparent you want the windows to be

This commit tidies this previous messy pull request: https://github.com/linuxmint/Cinnamon/pull/5703

### DIFF
--- a/files/usr/share/cinnamon/applets/show-desktop@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/show-desktop@cinnamon.org/applet.js
@@ -1,5 +1,10 @@
 const Applet = imports.ui.applet;
+const Settings = imports.ui.settings;  // Needed for settings API
 const Main = imports.ui.main;
+const Mainloop = imports.mainloop;
+const Tweener = imports.ui.tweener;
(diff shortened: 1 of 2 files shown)


clefebvre:
Cool.

I'll be honest I thought this was completely pointless at first. I had to think about it and try it, but I think it's great now :)

First, it kinda shows what the applet is about... you get a preview of the desktop on hover before getting it completely on click. That's good.

Second, if you're using desklets, that's a quick way to visualize them. Desklets can show information much better than applets but they're often hidden away by windows. This makes them much more useful all of a sudden.

PR #5711, opened 2016-09-20
centre align the centre box in a vertical panel

brownsr:
and also clean up an if sequence, a few related comments, and a misalignment

### DIFF
--- a/js/ui/panel.js
+++ b/js/ui/panel.js
@@ -1931,12 +1931,10 @@ Panel.prototype = {
             //    at top and bottom of the panel respectively
             // 2) have a central box that can take all the space in between
             // 3) turn on central y-alignment for the central box
-            // 4) there empty panel case is worked around with a kludge when setting edit mode to set box sizes explicitly if empty.
+            // 4) the empty panel case is worked around with a kludge when setting edit mode to set box sizes explicitly if empty.
             //    - there is a similar work around in the allocate logic, but allocate may not get called without this kludge 
             //
-            // The appearance of this looks reasonable - the icons in the boxes have sensible positioning
-            // (css permitting). The central box  position will depend on the relative numbers of 
-            // icons in the top and bottom boxes.
+            // The appearance of this looks reasonable - the icons in the boxes have sensible positioning (css permitting).
             //
             // Some workarounds for the side effects of the central alignment are needed.  
             //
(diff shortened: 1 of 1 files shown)


lestcape:
Hi @brownsr, thanks for the vertical panels...

Let me try (because my english it's not ok) to explain you some thing that i considerer important.
When you change the vertical property of an actor you are inverting how this actor works inside of clutter. As a result if you also don't inverted the vertical properties of his children's, you will not get it the same resulting behavior...

As you change the vertical properties of the panels, you are not get it the same behavior of the horizontal panel, because you also not invert the children's vertical properties.

If you also invert the children vertical property. Something like this:
```
if((appletDefinition.panel.panelPosition ==0) || (appletDefinition.panel.panelPosition ==1))
        applet.actor.set_vertical(false);
else
        applet.actor.set_vertical(true);
```
here, in this position probably:
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/appletManager.js#L320

You then will not need this:
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/panel.js#L2723

and as a result the applet actor will fill the panel "height", in the vertical case the width.

This change will need to be performed on deep if you want to have exact behavior (childrens of a childrens), but will not really needed , because normally there are not a list of children's, so you will have localized problems only... As for example the systray applet, where you will need to take more decisions..

**Another things**

I think this is not your fault, but please see the AppletContextMenu, of all applets are not updated properly his orientation when we move an applet from one panel to another. Also if it's explicit set here:
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/applet.js#L69

The reason if because in the boxpointer implementation, in this function:
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/boxpointer.js#L145

We there are a missing set of this variable: https://github.com/linuxmint/Cinnamon/blob/master/js/ui/boxpointer.js#L38 apparently this variable it's used just tow times and it's not important, but well...

This line will resolve this problem:

```
        this._userArrowSide = side;
```

if you write the line here:
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/boxpointer.js#L147

Regard...

JosephMcc:
@lestcape The issues you pointed out are known. I'm actually working on fixing the applet layout issues for vertical panels.

The boxpointers will hopefully be gone for Cinnamon 3.2.

lestcape:
@JosephMcc there are in my opinion two different thing.

1- In vertical panels the applets are not expanded to fill the panel width.
2- The box pointer are not updated properly.

You will fix both cases?

- The first can be fixed stetting the applet actor vertically.
- The seconds probably is not needed as will be removed the box pointers...

lestcape:
anyway, @JosephMcc thanks.

JosephMcc:
> 1- In vertical panels the applets are not expanded to fill the panel width.

I'm working on this one. Already have this fixed locally but many of our own applets that were enabled in vertical panels need a bit of work to layout properly. They were relying on the broken layout to work in a vertical panel.

lestcape:
Yes @JosephMcc  It is not enough just that, i try to find a short solution for systray, but Clutter.BoxLayout reused to give me a good alternative instead of iterate for all items inside the box..
Again thanks.

brownsr:
I haven't had much time this week, but I am planning to take a look this
weekend. As regards applet expansion I was going to see if there was any
Center alignment where it should be Fill.

On 21 Sep 2016 8:10 p.m., "Lester Carballo Pérez" <notifications@github.com>
wrote:
>
> Yes @JosephMcc It is not enough just that, i try to find a short solution
for systray, but Clutter.BoxLayout reused to give me a good alternative
instead of iterate for all items inside the box..

lestcape:
@brownsr I don' t say there are not, but i think there are not. I try a lot of things Clutter.BoxLayout work for specific actors, not general. Thanks any way.

PR #5728, opened 2016-09-26
Move all settings chooser button code to it's own file

collinss:
There are at least 2 benefits to doing so:
1. The effects and tween chooser button code had a great deal of duplication with the BaseChooserButton class in SettingsWidgets.py, but they couldn't be subclasses of BaseChooserButton because SettingsWidgets.py was already importing EffectsWidgets.py. With BaseChooserWidget in the same file, this is no longer an issue, which allows simplification of code and makes it easier to maintain
2. SettingsWidgets.py is getting very large. This change makes it easier to maintain the settings widgets.


clefebvre:
print("hello world!")

clefebvre:
@collinss can you rebase your previous PR also?

collinss:
Good catch! I always review my code to make sure I got all the logging and commented out code, but somehow this slipped through. It's fixed now.

PR #5740, opened 2016-09-30
[menu-applet] hiding favorites box leaves an empty box in the menu, t…

NikoKrause:
…his should fix this

In the image below you can see the empty box, where the favorites box and the shutdown buttons supposed to be:
hidefavbox_bug


itzexor:
It looks much better with this change when the favorites box is disabled.

PR #5742, opened 2016-10-01
Show desktop applet timeout id tracking + cleanup

itzexor:


NikoKrause:
I think I should be more aware of the whitespaces o.O
Is there an easy way in xed to detect them or do I have to go through line by line manually?

I also just saw, that I conjugated leave wrong. It's leave, left, left.
So it should be _onLeft not _onLeaved. It still sounds strange to me. I thought of it as opposite of _onEnteted, so is there another agreement on how to call it?

JosephMcc:
@NikoKrause Either onLeft or onLeave is probably ok. onLeaved is just a bit strange ;)

itzexor:
I rebased that commit. I changed those function names and a bunch of other small things in an attempt to make the style more consistent and better fit the new coding guidelines

itzexor:
Force pushed again to fix some I missed...

clefebvre:
onLeave is funny... it sounds holiday-ish :)

clefebvre:
In Sublime you can install an extension for trailing spaces. The bright-red they appear in is so annoying, you're sure not to leave any behind anymore. That extension also adds a quick menu option to get rid of them :)

In Xed, there's a plugin to automatically remove trailing spaces on file save if I remember right. It's not as convenient... especially if you want do so in a distinct commit.

PR #5747, opened 2016-10-02
[window-list applet] using icons on context menu

NikoKrause:
The following pictures are saying pretty much everything about this commit:
bildschirmfoto vom 2016-10-02 17-46-03
bildschirmfoto vom 2016-10-02 17-46-26

I saw this here https://cinnamon-spices.linuxmint.com/applets/view/261 and liked it.
Maybe you can use/create different icons.

This is especially helpful if you take a close look at the first picture.
In German the words "Restore" and "Unmaximize" are translated with the same word "Wiederherstellen".
Without the icons you can't tell the difference.
Before I saw this applet with icons, I wasn't even aware, that there was a difference xD

### DIFF
--- a/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
@@ -826,7 +826,7 @@ AppMenuButtonRightClickMenu.prototype = {
         this.addMenuItem(new PopupMenu.PopupSeparatorMenuItem());
 
         // Close all/others
-        item = new PopupMenu.PopupMenuItem(_("Close all"));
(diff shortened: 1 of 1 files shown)


JosephMcc:
I like it :)

lestcape:
I like it also... And i add icons to all my popupmenu always, they say more than 1000 words...

Some days i think also to check for the gtk gsettings property menus-have-icons, but i desist to it, because cinnamon in general don't follow this gtk rule.

Now a day it's a little more complicate get this value for GTK, but also posible:
http://unix.stackexchange.com/questions/150236/enabling-menu-icons-in-gnome3

Use this property will help in opinion to have a more unify desktop, base on user preference...

But thats clear it's a big work....

itzexor:
Just updated to the latest cinnamon with this included, and it looks good. Thanks for adding this.

PR #5749, opened 2016-10-02
[menu-applet] a more intuitive way of keyboard-navigation in menu

NikoKrause:
When opening the menu you can see, that the "All Applications" button is highlighted:
bildschirmfoto vom 2016-10-02 22-02-13

Imagine the following two scenarios:
1. You would expect to get the next category by pressing the  down key on your keyboard, because the first category is already highligted. Wrong.
Instead you get the first application in the "All Applications" category.
2. You press the right key on your keyboard and expect to get the first application in the "All Applications" category. Wrong again. You get the first category "All Applications", which is already highlighted.

I think the suggested way is more intuitive, due to the fact, that the "All Applications" button is highlighted, when you open the menu.

The only thing I wasn't sure about was what the `this._scrollToButton(item_actor._delegate);` function is doing. Maybe somebody, who does know what the function is doing, could check that.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1465,26 +1465,32 @@ MyApplet.prototype = {
         index = this._selectedItemIndex;
 
         if (this._activeContainer === null && symbol == Clutter.KEY_Up) {
-            this._activeContainer = this.applicationsBox;
-            item_actor = this.appBoxIter.getLastVisible();
-            index = this.appBoxIter.getAbsoluteIndexOfChild(item_actor);
-            this._scrollToButton(item_actor._delegate);
+            this._activeContainer = this.categoriesBox;
+            item_actor = this.catBoxIter.getLastVisible();
+            index = this.catBoxIter.getAbsoluteIndexOfChild(item_actor);
         } else if (this._activeContainer === null && symbol == Clutter.KEY_Down) {
+            this._activeContainer = this.categoriesBox;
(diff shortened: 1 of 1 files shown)


lestcape:
@NikoKrause  what is doing the _scrollToButton function, is move the scroll position of the applications, to a place where will  be show the selected application inside the visible viewport of the menu.

So, just take care to always pass as a parameter to this function the selected application  bottom and not the selected category bottom, because categories have not an scroll area.

NikoKrause:
So thanks to lestcape comment `scrollToButton` is now used only by the applicationsBox.
Also `previousVisibleIndes` is now only used by applicationsBox.

This fixes a weird behaviour, i.e. when you scrolled to the last application in "All Applications" category and went from there straight to the favoritesBox and scrolled the favoriteButtons.
First the applicationsBox jumped to the top, when you started scrolling in the favoritesBox and second the last application you visited in the applicationsBox was not remembered. 

I fixed that behavior and rebased the branch to one commit only.

JosephMcc:
Minor comment on the code. Remember conditionals should have a space between if  and ().

NikoKrause:
Thanks I'll change that

JosephMcc:
Are you active in our IRC channel? Maybe under a different name? Feel free to join anytime if you want to discuss changes or know what others are working on. It's #linuxmint-dev

NikoKrause:
I have no idea what an IRC channel is.

lestcape:
@NikoKrause https://community.linuxmint.com/tutorial/view/12

NikoKrause:
Ah thanks. I'll give it a try.

NikoKrause:
Added space between if and ().

PR #5753, opened 2016-10-03
applets: Change how applets set the type of panel they are allowed in

JosephMcc:
Currently an applet has to override getDisplayLayout() to allow itself in a
vertical panel which doesn't really make sense. An applet should be setting the
type of allowed layout it can have. Add an _allowedLayout variable to the base
applet class that is set by default to AllowLayout.HORIZONTAL. Then add two
functions that allow getting and setting that variable. Now to allow an
applet in non-horizontal panels an applet should call setAllowedLayout with the
appropriate value.


NikoKrause:
Branch has conflicts

JosephMcc:
Some things were merged since I made the PR. Fixed now.

JosephMcc:
...And rebased again.

PR #5755, opened 2016-10-03
Remove the boxpointer from popup menus

collinss:


clefebvre:
Hi,

A couple of points:

- I'm really excited about this. I don't think it looks good yet, but I'm pretty sure I'll end up loving it when it's ready.
- The animation is way too slow, it's frustrating to wait that long.
- The position of the menu after the animation is finished must be wrong, because the fonts looks blurry/aliased/blueish/pinkish. Make sure you round the coordinates after applying Maths to them.
- The menu applet uses its own trick or way to show its menu... I'm not sure how best to fix that, but whereas it looked somewhat ok before, with this PR it now looks drastically different when you open the menu and when you open other applets.
- do we want a little gap? do we need rounding on the part touching the panel? how about borders? I'd run that with Joseph or artists, I'm not too sure myself but I think it could look better.

clefebvre:
Can we wrap the top widget into a container styled with boxpointer style class? I'm worried we're breaking all themes significantly here otherwise (making all their applets transparent basically).

PR #5758, opened 2016-10-04
Add hostname label to the "user" applet

aleksey-novikov:
The hostname does not appear anywhere in UI except the lock screen. This patch display hostname in the "user" applet.


clefebvre:
Thanks for contributing to Cinnamon.

I don't think this adds to the user experience though. The hostname is mostly irrelevant to most users. It also no longer appears in the new screensaver in 3.2.

Issue #5719, opened 2016-09-23
mintmenu: Arrows no longer work after tabbing out of the search text field


Piiit:
Steps to reproduce:

1. Open mintmenu in Cinnamon
2. press TAB
3. try to choose an application with keyboard arrows

It does no longer work to choose apps with the keyboard. After clicking on the searchfield again, keyboard navigation works again.

JosephMcc:
You don't need to use tab. You can simply start navigating with the arrow keys

Piiit:
@JosephMcc Hi, I suggest to reopen this issue, because...

- It is not only the graphical issue like described in #5333
- It means generally, that if you open the menu and press incidentally the wrong button you end up in a state where keyboard navigation no longer works as intended 
- It does not matter if you need TAB or not, or if you can navigate without pressing TAB... It would have been the same bugreport if you end up in a "undefined/inconsistent navigation state" by pressing any non-ascii button on your keyboard

I suggest to ignore TAB and all other keys that are not needed to navigate with keyboard, or even better (if this is possible) to let keyboard events bubble up to Cinnamon (Xserver, or whatever tool handles those... don't know). For instance, to allow a workspace switch during an open Cinnamon menu.

JosephMcc:
The main reason I closed it is because it is still related. What ends up happening is we have similar/related issues spread around in a bunch of different places. If someone looks into this I would prefer they can see all the issues in one go so it can be fixed properly.

In this case you might be right though so I'll reopen it.

JosephMcc:
On git master, tab actually works but there is no highlight effect being applied to the menu items so you have no idea where you are.

Issue #5333, opened 2016-05-24
menu applet: weird key navigation behaviour, when first pressed TAB


NikoKrause:
Open menu, press TAB key, press RIGHT key two times. You get the following gap:
![weirdnavigationwithtab](https://cloud.githubusercontent.com/assets/8415124/15491106/a9998980-216d-11e6-967e-4e0048a93028.png)

It seems as it is possible to navigate through the menu. You can see see where your are,
if you press right, because then a gap opens. If you press left, the gap vanishes.
![weirdnavigationwithtab2](https://cloud.githubusercontent.com/assets/8415124/15491182/3fcf5786-216e-11e6-9192-da18320316d9.png)

JosephMcc:
Ok. Can reproduce this one.

PR #5770, opened 2016-10-08
[menu-applet] deactivate TAB-key in menu, because of weird behavior

NikoKrause:
I think this commit solves some weird behavior in the menu, when the TAB key is pressed.
See: https://github.com/linuxmint/Cinnamon/issues/5719 and https://github.com/linuxmint/Cinnamon/issues/5333

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1590,6 +1590,8 @@ MyApplet.prototype = {
             }
             return false;
 
+        } else if (symbol == Clutter.Tab) {
+            return true;
         } else {
             return false;
         }


JosephMcc:
@NikoKrause Can I make a suggestion since a number of your PR's are being merged? We've been trying to get improved commit message for commits/PR's. We are looking for something along the lines of:

```
menu-applet: Deactivate TAB-key in menu

A simple explanation of the "weird" behavior. It doesn't need to be a novel but just a bit of an explanation so we know what it was doing when we look at this a year or two from now.
```

NikoKrause:
@JosephMcc
I'm speaking of the behaviour explained in the issues. I thought it was enough to reference to the issues?

JosephMcc:
That won't show in your commit message. What you will see is this: https://github.com/linuxmint/Cinnamon/pull/5770/commits/7591b1d8af33cd9c22de24dfea991eff4469f90c

NikoKrause:
But that's the commit message I made in my repo before I created the pull request o.O

JosephMcc:
Yes. The commit and commit message from your branch are what is being merged.

NikoKrause:
Okay, I try to change the commit message later. 

But first a question about the content of the commit.
An idea popped up in my head.
This commit right now deactivates the TAB key to prevent those issues, reported above.
Maybe it would be better to define a different behavior for the TAB key?


**The current behavior** is as follows:
IF you have reached the last category/application and then pressing the DOWN key, you get the first category/application.
UP key behaves analogously. 
(In my opinion that can be annoying, because sometimes I want to reach the last category really fast by holding the DOWN key. But I end up getting back to the first or second category, because I was to slow releasing the key.)


**My suggestion** would be that, when you reach the last category, the DOWN key should do nothing and stay there. Vice versa the same with the UP key.
And the TAB key should behave like the DOWN key is behaving right now.
The current behavior of the UP button would be taken by SHIFT+TAB.

The suggested key navigation would have the advantage, that it would be more consistent with other applets. E.g. sound-applet, user-applet, etc. are behaving like that.

If this idea meets with approval, I would try implementing that. 
It would just need somebody, who can tell me, how to address SHIFT+TAB. (Is it `if (symbol == Clutter.Tab && symbol == Clutter.Shift)`?)

NikoKrause:
I close this PR, because I implemented the idea I mentioned above.
https://github.com/linuxmint/Cinnamon/pull/5780

Instead of deactivating TAB key, it's now possible to navigate with the TAB key.

Main reason for this solution is, that it's more consistent with the navigation known from the other applets.

Issue #4000, opened 2015-03-25
Remove from favorites in panel menu with right click option 'remove'


hrvoojex:
When you want to remove an item from favorites, you have to find it in a menu and then right click and select 'remove from favorites'.

Would it be possible to remove it right from the favorites menu with a right click or a mouse?

collinss:
You can just drag it from the favorites to the main part of the menu.

hrvoojex:
Didn't know that because it doesn't work if you drag it outside of the menu

raidermax:
Wait, right clicking the apps on my favorites doesn't bring anything up, I have to search the app first and then I can remove it. Is this intended?

JosephMcc:
@raidermax simply drag the item out of favorites over any other part of the menu and it will be removed from the favorites box

anandrkris:
@NikoKrause - In case you want to pick this up - Right click options for Favorties...

NikoKrause:
I think that right click for favorites is problematic.
The favoritesBox has a fixed height, i.e. it is not scrollable like the applicationsBox.
Which means, that if favorites would have a right click menu, the height of the whole menu would change.
That would look ugly.

Also the favorites are only icons. How should the context menu look like?
I think it's easier to drag and drop to remove it.

The only other way I could imagine is to set the DeleteKey to remove a favorites from it's box.

anandrkris:
Hmm right, it would increase height and will make it look unwieldy. Delete shortcut key is a good idea.

As OP suggests, usually dragging outside the menu is considered as removal - perhaps we can show an delete 'x' indicator beneath the dragged icon when you drag outside Favorites are to make it a bit intuitive?
> Didn't know that because it doesn't work if you drag it outside of the menu

hrvoojex:
Would it be possible to remove the icon from favorites by drag and drop outside of menu also. This might be confusing for a new user when they try first to drag and drop outside the menu and then think it is not possible without trying to drag and drop inside the menu

NikoKrause:
@hrvoojex 
I think that would be too mich.
It is now possible to delete a favorite by pressing the delete button.
And if my other pull request will also be merged, then the mouse cursor will look different, when a Favorite is draged over the menu compared to when it is draged outside the menu.

It would make more sense to add a feature, which drops the applet at the desktop, when you hover it outside the menu. But I don't know how to do it. If you want to, you can create a feature request for that. Maybe someone else will look into it.

[...]

PR #5780, opened 2016-10-10
[menu-applet] rewritten keynavigation codeblock

NikoKrause:
**EDIT:**

_What's new?_
- New keys for menu navigation were added (see below)
- Some bugs were fixed (see below)
- The code, which handles the keynavigation was rewritten. I have splited it into two codeblocks.

    1. One codeblock handles the navigation of the menu, without having to worry, which button you have to press to get this action done. It means, you check where you are (`appsBox, catBox or favBox`) and what should happen if you go right/left/up/down (`whichWay`) from your current position.
    2. In the other codeblock you decide, what you want to do, if a certain key is pressed, for example if you press TAB, you want to go right. At this point you don't have to worry, how this will be handled, because that's the problem of the above mentioned codeblock. You can add restrictions in this block, i.e. if you don't want to go to the right, when you press RIGHT-Key and you are already in the appsBox.

_Key behaviour and new keys:_
- **UP/DOWN/RIGHT/LEFT** behave the same as always
- **PAGE-UP** and **PAGE-DOWN**-keys jump to the first resp. the last button
- **TAB-Key** (catBox -> appBox -> favBox -> catBox -> ...)
- **SHIFT+TAB** behaves same as TAB but backwards.
- **Else turn off TAB-Key**, if it is not used for navigation or search completion (Fixes https://github.com/linuxmint/Cinnamon/issues/5719 and fixes https://github.com/linuxmint/Cinnamon/issues/5333)
- The context menu of an application button can be opened with the **MENU-Key** or with **CTRL+ENTER**
- Favorites can be removed from the favorites box by pressing the **DELETE-Key** (Regarding https://github.com/linuxmint/Cinnamon/issues/4000)
- Position of favorites in the favorites box can be changed by pressing **CTRL+UP** or **CTRL+DOWN**

_Bugs fixed:_
- When searching and nothing is found and then pressing left, the favBox was activated/hovered
- The above mentioned TAB-key issues 
- If there are no recent documents and you press right and then down, the keynavigation breaks, i.e. the keynavigator tries to go to the applicationsBox, even though it has no elements.

_What has to be done or what still doesn't work?_
- It is possible to open the contextMenu, but it is not possible to navigate through the contextMenu with the keyboard. I don't know how to do this. So **help concerning this matter would be appreciated**. This is no regression, given that this hasn't worked before either.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -209,9 +209,7 @@ GenericApplicationButton.prototype = {
             this.activate(event);
         }
         if (event.get_button()==3){
-            if (this.withMenu && !this.menu.isOpen)
-                this.appsMenuButton.closeContextMenus(this.app, true);
-            this.toggleMenu();
+            this.activateContextMenus(event);
         }
         return true;
     },
@@ -220,6 +218,12 @@ GenericApplicationButton.prototype = {
         this.unhighlight();
         this.app.open_new_window(-1);
         this.appsMenuButton.menu.close();
+    },
+
+    activateContextMenus: function(event) {
+        if (this.withMenu && !this.menu.isOpen)
+            this.appsMenuButton.closeContextMenus(this.app, true);
+        this.toggleMenu();
     },
 
     closeMenu: function() {
@@ -599,13 +603,10 @@ RecentButton.prototype = {
 
     _onButtonReleaseEvent: function (actor, event) {
         if (event.get_button()==1){
-            this.file.launch();
-            this.appsMenuButton.menu.close();
+            this.activate(event);
         }
         if (event.get_button()==3){
-            if (!this.menu.isOpen)
-                this.appsMenuButton.closeContextMenus(this, true);
-            this.toggleMenu();
+            this.activateContextMenus(event);
         }
         return true;
     },
@@ -613,6 +614,12 @@ RecentButton.prototype = {
     activate: function(event) {
         this.file.launch();
         this.appsMenuButton.menu.close();
+    },
+
+    activateContextMenus: function(event) {
+        if (!this.menu.isOpen)
+            this.appsMenuButton.closeContextMenus(this, true);
+        this.toggleMenu();
     },
 
     closeMenu: function() {
@@ -1161,6 +1168,7 @@ MyApplet.prototype = {
         this._isBumblebeeInstalled = GLib.file_test("/usr/bin/optirun", GLib.FileTest.EXISTS);
         this.RecentManager = new DocInfo.DocManager();
         this.privacy_settings = new Gio.Settings( {schema_id: PRIVACY_SCHEMA} );
+        this.noRecentDocuments = true;
         this._display();
         appsys.connect('installed-changed', Lang.bind(this, this.onAppSysChanged));
         AppFavorites.getAppFavorites().connect('changed', Lang.bind(this, this._refreshFavs));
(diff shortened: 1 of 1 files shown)


clefebvre:
Hi @NikoKrause,

Wouldn't it be better if TAB witched section (category -> apps -> favorites -> etc..), and SHIFT_TAB switched the opposite way?

Page_Down/Up could also be used to quickly get to the bottom/top of a list.

NikoKrause:
@clefebvre 
That's a good point. I haven't thought about the Page_Up/Down keys.

So just to be clear:
TAB: Should it switch (category -> apps -> favorites -> category -> etc..) in a loop or should it only switch (favorites -> category -> apps), i.e. without jumping from apps to favorites.

UP and DOWN key: They stay how they are right now, i.e. passing last category/applet/favorite.

Page_Up/Down key: The same as UP and Down keys, but stop at last category/applet/favorite.

Did I get that right?

Shouldn't take much time to change that, I think.

clefebvre:
imo, everything should loop.

page_down/up shouldn't go down/up though, they should go directly to the last/first item.

NikoKrause:
> imo, everything should loop.

o.O
But no loop with LEFT and RIGHT key, right? I think that would be too much.

Okay, I have an implementation locally on my notebook. 
But I hesitate to upload it right now.

The else if statements look awful. First I need to do some code cleanup.

clefebvre:
Why not loop on left/right?

NikoKrause:
@clefebvre 
Because I think it's too easy to hit the left/right key accidentally twice.
That's what the TAB key is for.
Also if left/right would loop, there would be no difference between (SHIFT+)TAB and LEFT/RIGHT.
But if you want to you can change that easily.

NikoKrause:
So I have rewritten the code a little bit. I think, that way it's easier to maintain this in the future.

First you get the `goUp, goLeft, etc.` variables. That's way it easier to define the behavior of the keys.
And also to add new keys, for example TAB and RIGHT key behave most of the time the same, so you tell them to `goRight`.

In the `if(navigationKey)` part you check all cases of `activeContainer` and define how the navigation should behave.
For example when you are at applicationsBox you just define, how to get the appButton above, by telling the applicationsBoxButton to `goUp`,without worrying which button you have to press, to get this action done.

The rest stayed untouched, except of indentation.

I hope I did everything right and didn't cause some regression, because there are a few lines I changed. I would be glad, if somebody would test it, before merging.

NikoKrause:
@rjanja 
Hey, you added the line `this._previousTreeSelectedActor._delegate.emit('leave-event');` a few years ago (https://github.com/linuxmint/Cinnamon/commit/be9887691e5b58b295231c9288ffb1281d6f1034). What is it good for? It makes no difference to me, if it's there or not.
But I'm not sure, what if it's needed anyway.

NikoKrause:
I'm thinking about getting a key, which works like the right-mouse-click. I mean, when you navigate over applications and want to open the context?/submenu? (the menu, which says "Add to panel", ... , "Uninstall", etc.). 
Any suggestions, which key to use for that?

Odyseus:
Hello, NikoKrause.

How about the [Menu](https://help.gnome.org/users/gnome-help/stable/keyboard-key-menu.html.en) key? Its sole purpose is to open context menus, so it will be fitting. Its *Clutter code* (I don't know if it's right to call it that way) is **Clutter.KEY_Menu**.

JosephMcc:
Not sure how many keyboards have that key. You would probably be better off doing something like Shift or Ctrl + Enter. @NikoKrause This might be a good thing to bring up in linuxmint-dev and get feedback on. Just be patient as it's quiet there at times.

NikoKrause:
@JosephMcc 

> This might be a good thing to bring up in linuxmint-dev and get feedback on. Just be patient as it's quiet there at times.

I haven't figured out how to get messages, even when the PC is turned off.
If I ask a question at IRC and I don't have that much time to let my PC turned on the whole day, I miss the answers, if the PC is down, didn't I?

rjanja:
@NikoKrause I believe it had to do with behavior when closing and re-opening the menu, so the top-level category would appear selected, and on selection of another category would then appear unselected.

clefebvre:
Many keyboards have the menu key. Looking at the 5 computers closest to me here, 4 of them have it.

It's the key which is left of the right-control key. Its purpose is exactly that, to show context menus.

NikoKrause:
Okay, I'm out of ideas and need help getting the contextMenu of apps key-navigable.
For now the contextMenu of apps can be opened with the CONTROL keys (this will be changed later), but it's not possible to navigate through the contextMenu.

I think it the code to navigate through the contextMenu should be added here at the if query:
```
} else if (this._activeContainer === this.applicationsBox) {
...
} else if (goDown) {
                    this._previousSelectedActor = this.applicationsBox.get_child_at_index(index);
                    if (this._previousSelectedActor._delegate.menu.isOpen) {
                        // How to get the contextMenu navigable with keys?
                    } 
...
```
But I don't know how exactly this is done.
If somebody has any idea, please let me know. (JosephMcc: Maybe you can add the "help wanted" label to this PR?)

NikoKrause:
@Odyseus 
Do you know how to detect Ctrl+Enter?
I don't get it, how the ControlMask works.

Odyseus:
I don't really get it either. To detect modifier keys I use the following code that I barely understand:

```javascript
let ctrlKey = Clutter.ModifierType.CONTROL_MASK & global.get_pointer()[2];
let shiftKey = Clutter.ModifierType.SHIFT_MASK & global.get_pointer()[2];
let altKey = Clutter.ModifierType.MOD1_MASK & global.get_pointer()[2];
```

NikoKrause:
@Odyseus 
Thanks. I found this here an hour ago: https://github.com/linuxmint/Cinnamon/blob/master/js/ui/appSwitcher/appSwitcher.js#L241
That worked for me.

The contextMenu opens now with CTRL+Enter or with the Menu-Key.

I can't still figure out, how to navigate through the contextMenu. Due to lack of time, I give up on trying and start hoping that somebody else has a good idea, how to fix that.

If somebody finds other issues or regressions with this pull request, please let me know.

Otherwise:
@clefebvre 
Let me know if I should rebase/squash this PR to 1 commit.
Or if I should close it. I fixed some other minor stuff with this PR, which is combined with the key-navigation, e.g. the keynavigation of the Recent files is broken, if there are no recent files.
If this PR doesn't make it to be merged, I would create for this stuff separate commits.

rjanja:
Generally it's easier on the maintainers if you separate bug fixes from new features, but your changes are all related to keyboard accessibility in the menu so it's probably fine.

My only minor gripe is that you removed an `else` block of code and only commented out the related `if` in https://github.com/linuxmint/Cinnamon/pull/5780/commits/907045401ac33fce7822e80998c7761569048986. It seems like the leftovers may cause confusion in the future.

Aside from that it looks good, though I haven't tested it out yet (I'm away from my Cinnamon desktop for another week or so).

NikoKrause:
@rjanja
> Generally it's easier on the maintainers if you separate bug fixes from new features, but your changes are all related to keyboard accessibility in the menu so it's probably fine.

I didn't separated the bug fixes, because I think this would lead to conflicts with this PR.

> My only minor gripe is that you removed an else block of code and only commented out the related if in 9070454. It seems like the leftovers may cause confusion in the future.

Yes. The if statement, which I commented out, is the place, where I suppose, that the contextMenuNavigation belongs to. I'll delete it, when I rebase this PR.

NikoKrause:
I tried to recap the changings, which were made in this PR. 
Therefor I edited my very first comment on this PR, so see above.

NikoKrause:
Squashed PR to 1 commit

clefebvre:
Thanks

clefebvre:
Very nice!

NikoKrause:
I set this reference for contextMenu keynavigation here, so I'll be able to find it later :)
https://github.com/Odyseus/CinnamonTools/commit/32c6f7b3d887c1ba7207a66c21f03c85a1bf524d

Odyseus:
Hello, everybody.

@NikoKrause
There is a little error in line [1550](https://github.com/linuxmint/Cinnamon/pull/5780/files#diff-9730f14a54367b06bdc5d0af4a9a5706R1550) of this commit (line [1553](https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/applets/menu%40cinnamon.org/applet.js#L1553) on master). The **item_actor** variable is set twice in a row.

NikoKrause:
It's not an error. The first one gets the first element, which is AllAppsCategory.
The second one gets the next element after the first.

The reason behind that:
If you open the menu the first time, the activeContainer is null, which means no button is selected. But the allAppsCategory button is already highlighted.
If I delete the second line and press down the very first time the menu is opened, the menu selects the first item in the categoriesBox, which is allAppsCategory and already highlighted. You wouldn't notice anything had happened. So you would have to press down twice to get the next category.

Odyseus:
I just read the code again (this time with my eyes open (LOL)). That's what happens when one is in front of a computer while sleep deprived.

Very sorry to waste your time (and everybody's), NikoKrause!!!

Issue #5778, opened 2016-10-10
Cinnamon theme chooser shows theme twice, if local theme has same name as global theme


NikoKrause:
If you are adding a theme into `~/.themes` folder, which has the same name as a theme, which already exists in `/usr/share/themes`, then the theme chooser in Cinnamon shows the theme twice. But it shouldn't.

To understand better, why it shouldn't, do the following:

- Copy the directory `Mint-X` from `/usr/share/themes` to `~/.themes`.
- Rename the local `Mint-X` directory to `Mint-Y`

You get the following result. Observe, that Mint-X is shown twice (but with different thumbnails: One thumbnail shows Mint-X and the other Mint-Y). 
![bildschirmfoto vom 2016-10-10 12-23-55](https://cloud.githubusercontent.com/assets/8415124/19233340/3f93b2fa-8ee5-11e6-95de-006963f0e009.png)

Both Mint-X themes are behaving the same, they use both the local Mint-X theme. (As seen in the image: The chosen Mint-X theme thumbnail shows the global Mint-X, but the local Mint-X (which is Mint-Y) is actually used).
![bildschirmfoto vom 2016-10-10 12-30-38](https://cloud.githubusercontent.com/assets/8415124/19233368/6059afe4-8ee5-11e6-808f-022d3de9d0d0.png)

The behavior is as expected. The local theme should cover/overlap the global theme, which it does.
But the theme chooser should only show the local theme (once).

(The "Add applets to panel" settings are doing that right. If you have local and global applets with the same name, it shows only the local one.)

I hope my explanation was clear enough.

P.S.: The same applies to icons, i.e. for `usr/share/icons` and `~/.icons`

JosephMcc:
Ok, I can reproduce this. You don't even need the rename step. Just copying mint-x to the .themes folder is enough for this to show up.

NikoKrause:
Yes. The rename step was just an attempt to explain this issue better. Also it shows, that one Entry is using the global thumbnail and the other is using the local thumbnail. But both are using the local theme.

PR #5784, opened 2016-10-12
cs_themes: Don't show duplicate themes

JosephMcc:
Currently when we have a theme of the same name in both the /usr and /home
directory both themes are shown in the theme chooser buttons. When choosing either
of the themes in the settings the one in /home will be given precedence. When
both exist, only show the one from the users home directory since that is the
one that will be applied regardless of which the user chooses.

Fixes https://github.com/linuxmint/Cinnamon/issues/5778


NikoKrause:
Works great! It also fixes duplicate icons in the theme chooser.
So maybe changing the title to: 

> cs_themes: Don't show duplicate themes and icons

JosephMcc:
Well it is an icon theme :)

NikoKrause:
That's a good point :)

To my defense:
My fallacy is based on the fact that you put the icon themes into `.icons` directory and not into `.themes` directory ;)

PR #5790, opened 2016-10-13
Improving slightly the appindicators.

lestcape:
- Resize the status indicator and the appindicator icons in the same way.
- Dynamically enable and disable the appindicators using the value of the global.settings.
- Dynamically enable and disable the appindicator when an applet with a same role will do the opposite action. This also will unregistered the appindicator Dbus iface.
- Added shortcut support.
- Encapsulate the blacklist functionality inside the manager.
- Reorganize the position of starting, to allow the access to the systray in the initialization.
- Can be decided the behavior for the indicators menu (right click or left click): https://github.com/linuxmint/Cinnamon/pull/5790/commits/4077efe1a02ac647ae6fe550add910bcb44672b8#diff-05db65ea5625e393ffab82cc7a763e9cR961 **Please use gsettings or create a applet setting to be configured this behavior.**
- Can also be decided the behavior for the indicators on PASSIVE MODE: https://github.com/linuxmint/Cinnamon/pull/5790/commits/4077efe1a02ac647ae6fe550add910bcb44672b8#diff-05db65ea5625e393ffab82cc7a763e9cR960 **Please use gsettings or create a applet setting to be configured this behavior.** In Ubuntu the behavior is not show the indicator as now is by default in cinnamon see the free desktop recommendation: "_Passive: The item doesn't convey important information to the user, it can be considered an "idle" status and is likely that visualizations will chose to hide it._" here: https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem/StatusNotifierItem/
- Added support for a label inside the indicator actor (to the right side of the icon) if somehow is possible (as this is a Jayatana invention https://lists.launchpad.net/ayatana-commits/msg02042.html). **Please see this label need to be set horizontal or hide in the vertical panel orientation**. https://github.com/linuxmint/Cinnamon/pull/5790/commits/4077efe1a02ac647ae6fe550add910bcb44672b8#diff-05db65ea5625e393ffab82cc7a763e9cR969 **Also the label need an style class and is NOT SCALING according to the panel size, this will be performed implicit as a label is expanded because the icon change his size. Can be tested with Psensor Indicator**. 
- Added support for tooltip, showing the appindicator "title" property as a tooltip. The indicator "tooltip" property is an array and could be in html format, so will require a webkit to be properly displayed. Anyway, the support for the html format, need to be placed inside the Cinnamon Tooltip implementation (not inside the appindicator).
- Indicators are now complete theme-ables, but this need a fix (applet-box and applet-icon are not the correct style class): 

https://github.com/linuxmint/Cinnamon/pull/5790/commits/4077efe1a02ac647ae6fe550add910bcb44672b8#diff-05db65ea5625e393ffab82cc7a763e9cR946

https://github.com/linuxmint/Cinnamon/pull/5790/commits/4077efe1a02ac647ae6fe550add910bcb44672b8#diff-05db65ea5625e393ffab82cc7a763e9cR1207

https://github.com/linuxmint/Cinnamon/pull/5790/commits/4077efe1a02ac647ae6fe550add910bcb44672b8#diff-05db65ea5625e393ffab82cc7a763e9cR1250

### DIFF
--- a/files/usr/share/cinnamon/applets/systray@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/systray@cinnamon.org/applet.js
@@ -78,20 +78,31 @@ MyApplet.prototype = {
         this._shellIndicators = {};
         this.menuFactory = new IndicatorMenuFactory();
         this.menuManager = new PopupMenu.PopupMenuManager(this);
-        this.signalAdded = 0;
-        this.signalRemoved = 0;
+        this._signalAdded = 0;
+        this._signalRemoved = 0;
     },
 
     _addIndicatorSupport: function() {
-        let currentIndicators = Main.indicatorManager.getIndicatorIds();
+        let manager = Main.indicatorManager;
+
+        // Blacklist some of the icons
+        // quassel: The proper icon in Quassel is "QuasselIRC",
+        // this is a fallback icon which Quassel launches when it fails to detect
+        // our indicator support (i.e. when Cinnamon is restarted for instance)
+        // The problem is.. Quassel doesn't kill that icon when it creates QuasselIRC again..
+        manager.insertInBlackList("quassel");
+
+        let currentIndicators = manager.getIndicatorIds();
         for (let pos in currentIndicators) {
-            let appIndicator = Main.indicatorManager.getIndicatorById(currentIndicators[pos]);
-            this._onIndicatorAdded(Main.indicatorManager, appIndicator);
+            if (!manager.isInBlackList(currentIndicators[pos])) {
+                let appIndicator = manager.getIndicatorById(currentIndicators[pos]);
+                this._onIndicatorAdded(manager, appIndicator);
+            }
         }
-        if (this.signalAdded == 0)
-            this.signalAdded = Main.indicatorManager.connect('indicator-added', Lang.bind(this, this._onIndicatorAdded));
-        if (this.signalRemoved == 0)
-            this.signalRemoved = Main.indicatorManager.connect('indicator-removed', Lang.bind(this, this._onIndicatorRemoved));
+        if (this._signalAdded == 0)
+            this._signalAdded = manager.connect('indicator-added', Lang.bind(this, this._onIndicatorAdded));
+        if (this._signalRemoved == 0)
+            this._signalRemoved = manager.connect('indicator-removed', Lang.bind(this, this._onIndicatorRemoved));
     },
 
     _removeIndicatorSupport: function() {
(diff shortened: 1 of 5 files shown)


lestcape:
Please let me know if this required more things to be merged or will not be merged finally.

lestcape:
@JosephMcc please see:

There are a problem with non homogeneous indicator icons (like indicator-multiload):
alt tag

When the icons are less homogeneous, they are displaying more and more wrong... This is because GDKPixbuf is NOT scaling according to the icon size and the panel size. If we scale it in a good way, they will not be shown good in a vertical panel orientation.

In the image above indicator-multiload are configured to show several sensors at the same time, while the cinnamon indicator receive just one image, with a large with respect to the height (a non homogeneous image). This image will not be displayed good in a vertical panel orientation. One possible idea could be rotate the image 90 degrees when the aspect-radio of the pixbuff image will be different to 1, but of course this will create a confusing representation of the vertical panel, some things with a rotation others not.

Please see this is not only a problem of indicators as it's currently a problem of the status icons, here you have two images of what currently happen:

systray

systray-fixed

As you can see some time the icon is show as it is received (first image) and some time it's scaled to an homogeneous size (second image). This is with one sensor, but if you add more, really you will not see anything in that way...

We can force all image to use an aspect radius = 1 in indicators, but of course, the first image will be displayed really wrong, but better than now i think.

We can decide what we want on this, it's just a think of decide if the GdkPixbuf will be scale or not and if will be scale how we want to be scaled.

This also could be fixed later, as now they are display better than the status icons in my opinion, also with this problem.

brownsr:
I had a go at testing this a couple of days ago, but had real difficulty finding indicators that would load, and the one that worked for me did not have any interesting functionality like context menus that I would expect to be improved, judging by what I remember from the last PR (which I really liked).  Is this to do with all the  gsetting options, because I did not attempt to set these ?  I have a PR in the queue which trims vertical panel applets, so if that is accepted I would imagine it would avoid problems with using your scaling option 3 above (which I like).

lestcape:
@brownsr I really don' t understand you, but if you speak of the: **Please use gsettings or create a applet setting to be configured this behavior.**, They are things that i don' t want decide, if some one tell me how need to be by default and where will be the place to set it, i will do it...

There are really a lot of problems with the tray like you say. I'm think some times that it' s better simplify the tray to use only the status icons like before. When all bugs will be fixed on status icons then again thinking on merged this.

I think there are a lot of changes for the next release, and one change will break the another, so i consider that and i try to not create more incompatible things, not brother with other changes if this will finally merged.

clefebvre:
Hi,

I had a quick go at this today. I wasn't able to properly test it though, because something seems broken for me, even on master... I can't click on indicators at all, nothing happens. I get this in .xsession-errors:

(cinnamon:28585): St-CRITICAL **: st_theme_node_render_resources: assertion 'width > 0 && height > 0' failed

It could be just my environment I don't know.

Other than that, I'm testing this with "XDG_CURRENT_DESKTOP=KDE mintupdate".

The size is definitely better in this PR (it's too small in master), but I also notice the icon is blurry and aliased.. as if the small icon had be chosen (it's set by name, not by filename in the code) but then sized up, instead of the bigger icon size being used.

lestcape:
@clefebvre 

I will see if i can find why the St-CRITICAL...

About the size, there are some different procedure to find the icon:
- On an array of icons in a binary format.
- Directly for a path to an specific icon
- In a virtual Icon Theme.
- In the default Icon Theme.

From an array of icons, we are selecting here the most biggest that it's less bigger than the panel size, but we can select the most bigger one if this is the idea, i have not notice any difference, but well, this procedure it's the same of the master and also it's not the most commons case (probably it's more commons for QT application do the things in that way).
https://github.com/lestcape/Cinnamon/blob/4f7a3b22df2c7dc1d53283f524b716f2ac119868/js/ui/indicatorManager.js#L1236

The another method could be directly from a path to the icon where we can not do anything:
https://github.com/lestcape/Cinnamon/blob/4f7a3b22df2c7dc1d53283f524b716f2ac119868/js/ui/indicatorManager.js#L1148

For a Icon Theme that have tow cases:

Was create virtually:
https://github.com/lestcape/Cinnamon/blob/4f7a3b22df2c7dc1d53283f524b716f2ac119868/js/ui/indicatorManager.js#L1173

Is the default Icon Theme:
https://github.com/lestcape/Cinnamon/blob/4f7a3b22df2c7dc1d53283f524b716f2ac119868/js/ui/indicatorManager.js#L1180

In both cases, we lookup an icon with the panel size:
https://github.com/lestcape/Cinnamon/blob/4f7a3b22df2c7dc1d53283f524b716f2ac119868/js/ui/indicatorManager.js#L1184

or if the panel it's not scale with 16px:
https://github.com/lestcape/Cinnamon/blob/4f7a3b22df2c7dc1d53283f524b716f2ac119868/js/ui/indicatorManager.js#L1143

lestcape:
I update cinnamon to the last version and i have the same problem... st_theme_node_render_resources: assertion 'width > 0 && height > 0' failed and I have not the menu in the indicators...

lestcape:
@clefebvre
Rebase linuxmint/master to update the popup API and fixed the boxpointer problem that cause the St-CRITICAL bug...

clefebvre:
Thanks @lestcape

Issue #5595, opened 2016-08-24
mint menu


oldtomdevel:
Mint menu in LM18 Cinnamon 64 bit seems to have some bugs.

When pointing at a document in "Mint Menu -> Last Documents", Mint menu shows the path of the document in the bottom line.
If the path contains non-standard-ASCII characters (e.g. German Umlaut), the bottom line is not correct.
May-be, the bottom line does not understand unicode characters.

Within Nemo it is possible to bookmark a link to an external server e.g. a link to a ftp-source within the own network.
However, when pointed at those bookmarks within Mint menu, the menu's bottom line again does not work correctly.
Instead of e.g. "ftp://Thomas@192.168.2.100:21/storage" the bottom line only shows "homas@192.168.2.100:21/storage".
Same with smb links, instead of e.g. "smb://idefix/" the bottom line only shows "defix/".

By the way, for the pre-defined bookmarks like "Music", "Downloads" etc. the bottom line works fine.

The bugs are not new i.e. the description is also true for LM17.3 Cinnamon 64bit.

Issue #3620, opened 2014-10-29
weird escaping in Recent Files menu description area when a file path has square brackets


Lx:
*Menu > Recent Files* shows the full path of the selected item in the description area at the bottom of the menu.

For files with square brackets in the path, the brackets in the full path are URL-escaped.  Please see the attached screen shot.  The correct path is `.../Desktop/Folder [with square brackets]/File.txt`.

![screenshot from 2014-10-29 11 31 52](https://cloud.githubusercontent.com/assets/160164/4819476/fa9d97f0-5f03-11e4-866b-75e584d4c0d5.png)

Vahan86:
Tested in Cinnamon 3.0.7. The issue is still there.

Odyseus:
Hello, everybody.

I'm also seeing this on Cinnamon 3.0.7 (stable and nightly). It seems that this is the product of only un-escaping spaces on the following two lines of the applet code:

[Line 1784](https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/applets/menu%40cinnamon.org/applet.js#L1784)

```javascript
this.selectedAppDescription.set_text(button.place.id.slice(16).replace(/%20/g, ' '));
```

[Line 1851](https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/applets/menu%40cinnamon.org/applet.js#L1851)
```javascript
this.selectedAppDescription.set_text(button.file.uri.slice(7).replace(/%20/g, ' '));
```

To fix it, and as a test, I tried to use **decodeURIComponent** like this:

```javascript
this.selectedAppDescription.set_text(decodeURIComponent(button.place.id.slice(16)));
```

```javascript
this.selectedAppDescription.set_text(decodeURIComponent(button.file.uri.slice(7)));
```

With the previous edit, all paths were shown correctly inside the info box. I repeat, this was just a test. I don't know if it is *safe* to use **decodeURIComponent** in that place. Maybe not all characters should be un-escaped inside a label? It would be great to have the opinion of one of the developers.

Issue #3250, opened 2014-06-09
Improve path show in menu description


highwindmx:
![bug1](https://cloud.githubusercontent.com/assets/6413481/3212821/7cff7734-ef76-11e3-8a96-d9eb312005d2.png)

you can see in Nemo(1) the path with Chinese Character show well,
but in menu description shows like 'read' URI with trans-coded ASCII code(3), 
My Opinion is: as menu "description", it is better to show these paths as what nemo shows.

Issue #2872, opened 2014-02-04
[menu applet] Bookmarks on unmounted volume don't work


collinss:
I suspect that the only way to fix this would be to completely change the way bookmarks are handled, which would be a lot of work, and cause a whole host of other problems, so they should probably be hidden when they aren't mounted (as they are in nemo).

PR #5798, opened 2016-10-14
[menu-applet] decodeUriComponent in app description

NikoKrause:
The app description doesn't handle some characters (e.g. german umlaute ö ü ä) properly.
Fixes https://github.com/linuxmint/Cinnamon/issues/5595

Edit: fixes https://github.com/linuxmint/Cinnamon/issues/3620 and fixes https://github.com/linuxmint/Cinnamon/issues/3250

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1796,7 +1796,12 @@ MyApplet.prototype = {
                         this._clearPrevSelection(button.actor);
                         button.actor.style_class = "menu-application-button-selected";
                         this.selectedAppTitle.set_text("");
-                        this.selectedAppDescription.set_text(button.place.id.slice(16).replace(/%20/g, ' '));
+                        let selectedAppId = button.place.idDecoded;
+                        selectedAppId = selectedAppId.substr(selectedAppId.indexOf(':') + 1);
+                        let fileIndex = selectedAppId.indexOf('file:///');
+                        if (fileIndex !== -1)
+                            selectedAppId = selectedAppId.substr(fileIndex + 7);
+                        this.selectedAppDescription.set_text(selectedAppId);
                         }));
                 button.actor.connect('leave-event', Lang.bind(this, function() {
                             this._previousSelectedActor = button.actor;
@@ -1863,7 +1868,11 @@ MyApplet.prototype = {
                             this._clearPrevSelection(button.actor);
                             button.actor.style_class = "menu-application-button-selected";
                             this.selectedAppTitle.set_text("");
-                            this.selectedAppDescription.set_text(button.file.uri.slice(7).replace(/%20/g, ' '));
+                            let selectedAppUri = button.file.uriDecoded;
+                            let fileIndex = selectedAppUri.indexOf("file:///");
+                            if (fileIndex !== -1)
+                                selectedAppUri = selectedAppUri.substr(fileIndex + 7);
+                            this.selectedAppDescription.set_text(selectedAppUri);
                             }));
                     button.actor.connect('leave-event', Lang.bind(this, function() {
                             button.actor.style_class = "menu-application-button";
--- a/js/misc/docInfo.js
+++ b/js/misc/docInfo.js
@@ -27,6 +27,7 @@ DocInfo.prototype = {
         this.name = recentInfo.get_display_name();
         this._lowerName = this.name.toLowerCase();
         this.uri = recentInfo.get_uri();
+        this.uriDecoded = decodeURIComponent(this.uri);
         this.mimeType = recentInfo.get_mime_type();
         //this.mtime = this._fetch_mtime(); // Expensive
     },
--- a/js/ui/placesManager.js
+++ b/js/ui/placesManager.js
@@ -29,6 +29,7 @@ function PlaceInfo(id, name, iconFactory, launch) {
 PlaceInfo.prototype = {
     _init: function(id, name, iconFactory, launch) {
         this.id = id;
+        this.idDecoded = decodeURIComponent(this.id);
         this.name = name;
         this._lowerName = name.toLowerCase();
         this.iconFactory = iconFactory;
@@ -84,6 +85,7 @@ PlaceDeviceInfo.prototype = {
         this.name = mount.get_name();
         this._lowerName = this.name.toLowerCase();
         this.id = 'mount:' + mount.get_root().get_uri();
+        this.idDecoded = decodeURIComponent(this.id);
     },
 
     iconFactory: function(size) {


Odyseus:
Hello, NikoKrause.

If you use `decodeURIComponent` in those strings, then the `replace` isn't needed. This commit also fixes #3620.

NikoKrause:
@Odyseus 
I didn't saw, that you already solved this issue.
Why haven't you created a pull request?

lestcape:
https://github.com/linuxmint/Cinnamon/commit/b2426c5b6f6163cc97d0c6a7b75914042b806799

NikoKrause:
So it seems that decodeURIComponent is known since over 2 years.
Is there a reason why it hasn't been fixed yet?

lestcape:
As i not get any response on this i don't do anything... But that is my position, i don't use the cinnamon menu, because of this and because of more things... So, i really don't care if is not interesting for who need to be intresting ...

lestcape:
@NikoKrause, @Odyseus have right and also it's not because the replace is not needed. In your context, as you try to decode a replace string, you could tray to decode an invalid hexadecimal string (where some characters are replaced). If you first decode the string and then apply the replace, then it's like @Odyseus  say, but you are not doing that...

lestcape:
Also see the problem it's not inside the menu, and will not be fixed in all places where a bookmarks are used. This have a source and if you fixed this in a source you will not need to fixed it in all places where this was used...

Odyseus:
Hello, g@ys.

@NikoKrause: I didn't really fix anything. I just did some tests and posted my findings in that issue. Like I said in there, I'm not sure if it's *safe* to use `decodeURIComponent` in that place because it could decode an *illegal* character for an `St.Label` element. I don't even know if there is such a thing (an *illegal* character for an `St.Label` element).

I already added `decodeURIComponent` to my own fork of the menu without any apparent drawbacks. I saw that **lestcape** also uses it in his configurable menu applet.

> Why haven't you created a pull request?

Basically, because I'm not a developer. Any possible contribution that I could make would be narrowed by my very limited knowledge. I wouldn't want to waste the developers' valuable time reviewing commits that could contain obvious errors (for a professional developer's eye). Plus, I don't feel comfortable using some GitHub advanced features that I can barely understand.

NikoKrause:
I rebased the commit, thanks to @Odyseus.

@Odyseus 
You shouldn't be afraid to create a pull request, if you think, that you solved an issue.
The Linux community is very nice. And I think one purpose of pull requests is, that commits are reviewed by the owner of the project, so he can check if there are errors or if he want to merge it or not.  
And about to GitHub advanced features. If you don't start to try, you don't learn them.
I also made mistakes and still doing them, when using git. But I learned from my mistakes. And I even learned some new features of git, I didn't even know about, because I made mistakes.
So don't be afraid trying out new stuff and don't bad talk your skills.

@lestcape 
I think you are right, when you say, that this problem should be fixed in the source. But I don't know, where the source of this problem is, so I solved this problem at that specific location.
Can you tell a concrete example, where `decodeURIComponent` could go wrong?
I wasn't able to create one:
bildschirmfoto vom 2016-10-15 14-38-51

lestcape:
@NikoKrause See my original comment:
https://github.com/linuxmint/Cinnamon/commit/b2426c5b6f6163cc97d0c6a7b75914042b806799#commitcomment-5958641

I think this need to be in _reloadBookmarks:
that it's located on: https://github.com/linuxmint/Cinnamon/blob/master/js/ui/placesManager.js file.

clefebvre:
hi everyone.

@NikoKrause can you adapt the change to reflect @lestcape's last comment?

@lestcape sorry I didn't spot your comment on my commit earlier. I try to have eyes everywhere... emails, PRs, github notifications, slack internally, IRC etc... but sometimes I don't see everything. I didn't ignore you on this, I just didn't see it.

@Odyseus don't be afraid to get involved :)

clefebvre:
@NikoKrause .. not related to this PR, but I see you're focusing a lot on the menu :) I've got something on my list if you want to have a look. Opening the menu always freezes Cinnamon momentarily.

It only freezes it for a few milliseconds, but it's enough to notice it.

If you play an HTML5 video in Firefox for instance (I use matches on http://fullmatchesandshows.com/ to test) and then repeatedly open/close the menu, you'll see the movement of the football will lag every time you open the menu.

NikoKrause:
> @NikoKrause can you adapt the change to reflect @lestcape's last comment?

I don't know what to change in _reloadBookmarks. @lestcape Could you be more specific what to change their (or even create an own pull request, which solves this problem)? I'm not very familiar with the placesManager.js file (I tried to change the labels, but that had no effect and even froze cinnamon once ;)

@clefebvre 

>  Opening the menu always freezes Cinnamon momentarily. It only freezes it for a few milliseconds, but it's enough to notice it.

I opened a video on the page you mentioned. When I open the menu, the video freezes for one or two seconds (My notebook I code on is very old). But this happens also, when I open other applets.
(With keyboard-applet, sound-applet etc. the freeze is shorter. The longest freeze I get is with the network-applet. It's almost as long as with the menu applet.) 
If I had to guess, I would say this problem is related with the `open` or the `on_paint` function in `popupMenu.js`.

NikoKrause:
With network-applet the freeze gets longer, when you show more wireless networks, i.e. when you hit the `More` button.

lestcape:
@clefebvre no problem, this is also not important to me, so you have nothing for to say sorry.

@NikoKrause can you send me your bookmark file lestcape@gmail.com?

Wharever you do a long process, Cinnamon main thread will freeze for some milliseconds, thats i think have not solution. There are of course some workaround to reduce the lag.

One place where i really considered there are a problem, it's with the updates and the menu. Like when you install a package we trigger the cinnamon menu reload, this is really a problem, when you perform an update of several packages that trigger several times the menu reload.

lestcape:
@clefebvre Probably remove the boxpointer will have a positive impact in the performance, like the boxpointer it's a direct call to paint on gl.

lestcape:
@NikoKrause just send me that's file that have not effect if you try to decode the label in the place manager file. I think need to be the same whatever you decode the file, in the menu or in the placemanager, just is better in the place manager because it's more short fix and because not need to be done several times but well.

NikoKrause:
@lestcape 
I'm not sure, if I understood, what you said.

> So, on my opinion will be good that the fix will be do on the source of the problem. On the function:
> _reloadBookmarks
> that it's located on: https://github.com/linuxmint/Cinnamon/blob/master/js/ui/placesManager.js

I thought you said, instead of using decodeURIComponent in the applet.js file of the menu, the decodeURIComponent should be used in the placesManager.js file? 

I don't know in which line decodeURIComponent should be added in this file.

lestcape:
@NikoKrause i also don't know, this can be added where will work. For example, in this line:

 https://github.com/linuxmint/Cinnamon/blob/master/js/ui/placesManager.js#L339

You can do this:
```
        let bookmarks = decodeURIComponent(bookmarksContent).split('\n');
```
This will decode all file contents at ones, but you will need to check if this continues returning a valid file:
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/placesManager.js#L359

and if this return a valid label:
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/placesManager.js#L361

If this both function required an hexadecimal string, you will need to decode the string after apply this function, and this will required probably more changes to the logic... This is a code, not the recipe of the rotisserie chicken from my grandmother :)

lestcape:
@NikoKrause the other source of this it`s from here:
https://github.com/linuxmint/Cinnamon/blob/master/js/misc/docInfo.js#L26-L30

Odyseus:
Hello, g@us.

@lestcape: Decoding all bookmarks at once will break bookmarks with spaces in their path. See that the immediately next **for** loop extracts the URI by doing [`let components = bookmarkLine.split(' ');`](https://github.com/linuxmint/Cinnamon/blob/master/js/ui/placesManager.js#L345). Even if the path is decoded in the correct place, it could break any of the methods that uses that URI and might even need it to be encoded (**Gio.file_new_for_uri**, **Cinnamon.util_get_label_for_uri**, **Cinnamon.util_get_icon_for_uri**, **Gio.app_info_launch_default_for_uri**, etc.). It seems to much of a risk to fix something that isn't broken. Because, as it is, it works perfectly. It just force us to use `decodeURIComponent` to display the label correctly on the menu.


@NikoKrause: There is another error displaying labels in the applications info box. Mounted devices on the **Places** category will be displayed cut down. It will display their path like this **dia/user_name/MountedDeviceLabel** instead of **/media/user_name/MountedDeviceLabel**. The error is produced by `button.place.id.slice(16)` that assumes that all URIs start with **bookmark:file:///** (it has to cut 16 characters from the start). But some URIs start with **mount:file:///** (it has to cut only 14 characters).

lestcape:
@Odyseus like i say, is not the recipe of the rotisserie chicken from my grandmother and i don't know. I only throw an idea. If this will not work in this way, not need to be in this way and thanks for search more on this...

If you get it the logic after let components = bookmarkLine.split(' '); it used to get the first position later, as an uri it's not the first position in the list (the first position it`s the label), nothing will happen.

All function get_for_uri will work with an encode and/or decode uri as i see for now.

But you probably are right that it's better let the things as they are and only fix the problem that appear, it's more easy and you will ensure that you are not introduced more problems.

NikoKrause:
I rebased the pull request. I hope that this solution is ok.

@Odyseus 
Thanks for your explanation of the `bookmarks:file://` and `mount:file://` problem.
This problem was also mentioned here https://github.com/linuxmint/Cinnamon/issues/5595
But I didn't understand it there.

My solution for this problem only works, when the bookmarks have the structure `something-cool:file:///`. Does anybody know, if it's true for every bookmark even with `ftp` and `smb`, which are mentioned in the referenced issue report.

Odyseus:
Hello, @NikoKrause.

You could do it like this:

```javascript
let selectedAppId = button.place.idDecoded;
if (/file:\/\//.test(selectedAppId)) // If it's an URI to a local file system.
    selectedAppId = selectedAppId.substr(selectedAppId.indexOf(':///') + 3);
else // Any other URI
    selectedAppId = selectedAppId.substr(selectedAppId.indexOf(':') + 1);
this.selectedAppDescription.set_text(selectedAppId);
```

This way, you don't have to worry about the length you have to use with `slice`. In fact, `slice` isn't needed, because it will always give you the exact string that you need/want (hopefully).

NikoKrause:
@Odyseus 
I like the `if` case. But does the `else` case ever appear? Do you know an example?

Odyseus:
If you add an ftp as a bookmark on Nemo (for example "ftp://ftp.wayne.edu/ldp/manpages"). It's id will be **bookmark:ftp://ftp.wayne.edu/ldp/manpages**. But you don't want to get rid of the URI schema (the *ftp://* part), that's when the **else** part removes the **bookmark:** from the string. Leaving the **selectedAppId** variable with a value of **ftp://ftp.wayne.edu/ldp/manpages**.

A bookmarked ftp will also create a *mount point* (**mount:ftp://ftp.wayne.edu/** using the previous example) that will also be listed in the **Places** category on the menu. And in the applications box description will be displayed as **ftp://ftp.wayne.edu/**.

Odyseus:
The previous code is for use with the **Places** category. I didn't test it with the **Recent files** category. For the recent files section the code could be like this:

```javascript
let selectedAppId = decodeURIComponent(button.file.uri);
if (/file:\/\//.test(selectedAppId)) // If it's an URI to a local file system.
	selectedAppId = selectedAppId.substr(selectedAppId.indexOf(':///') + 3);
this.selectedAppDescription.set_text(selectedAppId);
```

In this section, the **selectedAppId** variable seems to always be a *pure URI*.

NikoKrause:
Is this also a problem in the Recent files category? I thought Recent files contains only `files:///`, doesn't it?

Odyseus:
> Is this also a problem in the Recent files category? I thought Recent files contains only files:///, doesn't it?

No, not just `file:///` URIs. An opened file from an ftp will produce an URI like `ftp://ftp.wayne.edu/ldp/manpages/man-pages-2.25.tar.bz2` on the **Recent Files** category. In which case, the URI doesn't need to be modified.

NikoKrause:
Rebased again.

I hope it works now as expected :)

Odyseus:
My approval?! LOL

I just tested it and everything works like a charm. I tested Unicode characters, paths to mounted devices, bookmarks and ftps. Everything displays correctly on the application info box.

lestcape:
I also like your solution. I don't know why remove "file:///" or "ftp://" from the output, but well... Maybe just preferences...

anandrkris:
Not sure this would get fixed - https://github.com/linuxmint/Cinnamon/issues/2872
@NikoKrause  - Linking it for interest :smile:

lestcape:
@anandrkris I think not...

PR #5800, opened 2016-10-14
[menu-applet] clear recent list applet with ENTER-key

NikoKrause:
When pressing the ENTER/RETURN-key on the RecentClearButton nothing happened.
Now it clears the recent list, when you hit ENTER.


anandrkris:
Glad to see keyboard friendly fixes. Thanks!

Issue #5347, opened 2016-06-01
[keyboard applet] uses wrong flag


NikoKrause:
If you add "Russian (phonetic for germans)" as second language, the keyboard applet uses the german flag in the panel. It should use the russian flag, because it is meant to type russian letters.

andyn:
The keyboard applet is currently using "group name" instead of "short name" when it determines the flag icon to be used. Would the correct behaviour be to prioritize the short name and fall back to the group name, should the correct flag not be found?

I've fixed this in my local install but I'm not submitting patches/pull requests until I can ensure it works properly with other layout combinations.

NikoKrause:
This bug still exists as I described here: https://github.com/linuxmint/Cinnamon/commit/b9344bb9829a3a2e8d886e60eda7d6dd64360a53

Is it possible to reopen this issue?

JosephMcc:
Sure thing.

NikoKrause:
See explanation of mtwebster https://github.com/linuxmint/Cinnamon/commit/b9344bb9829a3a2e8d886e60eda7d6dd64360a53

PR #5801, opened 2016-10-14
cs_screensaver.py: expose new cinnamon screensaver settings + bugfix

itzexor:
I have renamed the "Date" page to "Customize" and added the new lock screen settings there. I think the away message stuff should probably be in customize too, but the whole reason I added the new stuff on the "Customize" page is to avoid having one page be too long and require scrolling by default. Let me know what you all think of the wording, layout, etc.... 

By the way I recommend we also merge #5342 as it is still compatible with the rewritten screensaver.

preview:
screenshot-window-2016-10-14-132327

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
@@ -162,17 +162,18 @@ class MainWindow:
 
         # Resize vertically depending on the height requested by the module
         use_height = WIN_HEIGHT
+        total_height = n.height + self.bar_heights + WIN_H_PADDING
         if not sidePage.size:
             # No height requested, resize vertically if the module is taller than the window
-            if n.height > WIN_HEIGHT:
-                use_height = n.height + self.bar_heights + WIN_H_PADDING
+            if total_height > WIN_HEIGHT:
+                use_height = total_height
             #self.window.resize(use_width, n.height + self.bar_heights + WIN_H_PADDING)
         elif sidePage.size > 0:
             # Height hardcoded by the module
             use_height = sidePage.size + self.bar_heights + WIN_H_PADDING
         elif sidePage.size == -1:
             # Module requested the window to fit it (i.e. shrink the window if necessary)
-            use_height = n.height + self.bar_heights + WIN_H_PADDING
+            use_height = total_height
 
         self.window.resize(use_width, use_height)
 
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_screensaver.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_screensaver.py
@@ -104,6 +104,10 @@ class Module:
 
         size_group = Gtk.SizeGroup.new(Gtk.SizeGroupMode.HORIZONTAL)
 
+        widget = GSettingsSwitch(_("Always show clock"), schema, "show-clock")
+        widget.set_tooltip_text(_("Show the clock on the wallpaper instead of just the unlock screen"))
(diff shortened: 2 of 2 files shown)


anandrkris:
Looks good. Perhaps "Away - Font selection" alone can be in Customize option IMO...

NikoKrause:
Looks good. Is the flag/letters setting in the keyboard applet now also handled through this settings?
Because I run Cinnamon nightly builds and I just noticed, that the checkboxes in the keyboard applet don't work.

I like that flags are now have numbers. But this bug https://github.com/linuxmint/Cinnamon/issues/5347 reappeared, which was solved two weeks ago :(

itzexor:
@NikoKrause I'm not sure what's going on with the Keyboard Applet at the moment, I haven't had a look at it since the recent changes. These settings relate only to the lock screen though.

itzexor:
Rebased on master and combined these new options with the away message options.

updated preview:
screenshot-window-2016-10-15-181524

itzexor:
I rebased this again and included a bug fix. With the fix I can now use multiple categories without introducing scrollbars. I think this is ready now unless someone has objections.

Also, I renamed "Customize" to "Customize Lock Screen" because I think that helps make it clear that this isn't related to the screensavers at all just the lock screen.

Updated preview:
screenshot from 2016-10-27 20-25-45

NikoKrause:
My opinion:
(1)
I think the **Settings** tab is too general. Does it set the settings for Screensaver or Lock Screen?
I don't know how your Settings tab look right now.
As I see it, there is only one option, which handles the settings for screensaver:
**Screensaver settings**
_Delay before starting the screensaver_
Maybe it would be better just to seperate screensaver and lock screen settings. 

(2)
And I would consider to rename the header **General** in the Customize Lock Screen section.
When you read **Date and Time** or **Away message** it's clear right away what the settings are for.
The **General** section is too general ;) Maybe something like **Keyboard Layout and Album Art** or just **Layout** or something like that.

(3)
In my opinion the "Screensaver" application should be renamed to **Screensaver and Lock Screen**.
That way it would be more clear, that it doesn't handle only the Screensaver.

itzexor:
This is the current settings page:
screenshot from 2016-10-28 12-37-23

All settings in the "Settings" page affect how the screensaver acts, whether you are using a screensaver or the lock screen only (which is considered a type of screensaver as well if you're not using a 'real' screensaver). So if you have the lock screen selected as your screensaver, then the delay before starting the screensaver applies to the lock screen, if you have another screensaver selected then it applies to that.

I think the "General" section is fine as is, personally. If we were to make it specific it should group only one category of settings together.

Renaming the entire module can go in it's own PR separate from this one. I'm just trying to add the new settings here, I didn't intend to redesign the whole thing.

itzexor:
Rebased yet again. Changed "Customize Lock Screen" back to "Customize"

NikoKrause:
> All settings in the "Settings" page affect how the screensaver acts, whether you are using a screensaver or the lock screen only (which is considered a type of screensaver as well if you're not using a 'real' screensaver). So if you have the lock screen selected as your screensaver, then the delay before starting the screensaver applies to the lock screen, if you have another screensaver selected then it applies to that.

Good point

> I think the "General" section is fine as is, personally. If we were to make it specific it should group only one category of settings together.

I also think it shouldn't be too specific, but a little more specific then "General". **Layouts** or **Appearance**  or something like that. But that's a matter of taste.

> Renaming the entire module can go in it's own PR separate from this one. I'm just trying to add the new settings here, I didn't intend to redesign the whole thing.

I think that before this will be merged, the comments here would be read. That's why I throw this idea in.

PR #5814, opened 2016-10-18
clip the applet at the panel edge in a vertical panel

brownsr:
Avoids any part of the applet appearing outside the panel 
e.g. menu and show desktop highlight in Mint-X themes when hovering.
Also adds a bit of further protection against an applet that might be declared
as suitable for a vertical panel but that still has size issues.

### DIFF
--- a/js/ui/applet.js
+++ b/js/ui/applet.js
@@ -155,6 +155,7 @@ Applet.prototype = {
                                         reactive: true,
                                         track_hover: true });
 
+        this._panelHeight = panel_height ? panel_height : 25;
         this._allowedLayout = AllowedLayout.HORIZONTAL;
         this.setOrientationInternal(orientation);
     
@@ -176,7 +177,6 @@ Applet.prototype = {
         this._applet_enabled = true;    // Whether the applet is enabled or not (if not it hides in the panel as if it wasn't there)
         this._orientation = orientation;  // orientation of the panel the applet is on  St.Side.TOP BOTTOM LEFT RIGHT
 
-        this._panelHeight = panel_height ? panel_height : 25;
         this.instance_id = instance_id; // Needed by appletSettings
         this._uuid = null;      // Defined in gsettings, set by Cinnamon.
         this._hook = null;      // Defined in metadata.json, set by appletManager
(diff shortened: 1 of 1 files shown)


JosephMcc:
I'm not sure we need to do this. This is just hiding when something isn't right in the theme. Our own themes aren't updated for all the vertical panel changes yet. I've been holding off because things are still changing and it's easier to do it once :)

Also, we are talking about adding versioning for themes so the old ones that don't properly support new features won't be shown in spices or settings.

js/ui/applet.js
```diff
@@ -377,9 +377,14 @@ Applet.prototype = {
             this.actor.set_important(true);
             this.actor.set_vertical(true);
             this.actor.set_x_expand(true);
+
+            this.actor.set_clip(0,0,this._panelHeight,5000); // ensure no portion of the applet
```
JosephMcc: This should have spaces between the arguments (0, 0, this._panelHeight, 5000)
If you just put the comment above inline
```
// ensure no portion...
this.actor.set_clip()
```
it would all fit in one line. If you're going to put them after the line they are probably better as one single line and not as a multi-line comment.

brownsr:
I pushed vertical panels theme changes a while back, the series of changes over the last day or so are just me noticing (generally) minor issues.  That's not to say there are not any more of course, and I am sure you will be able to improve what's there :)  Changed so far are the cinnamon default, mint-Y, all the mint-X, and Linux Mint.  I've restructured to reflect your comment.

clefebvre:
Hi guys, I had to chat with Joseph, I wasn't sure whether to close or merge this.

After chatting it looks like the best way to go is to keep broken things apparent and not hide them away, to give a better incentive for artists to fix this in the proper place, i.e. in the theme itself.

brownsr:
Sure clem, joseph, I can understand that.  The only current related theme issue is with mint-x and large colour icons where the highlighting on hover can spill over the panel.  But I reckoned a bit of insurance might not go amiss in case something like one of lester's very wide indicators got turned on in a vertical panel.  The degree of panic this could cause a user could be marked - as the panel centres things the icons get pushed way off to one side, in a left panel they will just appear to vanish, real 'wtf' stuff.

JosephMcc:
Now that all of the menu changes and stuff have been merged into git master I can turn attention to updating our themes for it. I've been holding off just so I don't have to redo things because of other major changes.

3rd party themes are going to need to get updated to work well with vertical panels. The good thing is that the new panel stuff doesn't break them from working with anything they supported before.

Issue #5810, opened 2016-10-16
Huge inhibit applet


germanfr:
The inhibit applet has an entry in the menu that is bigger than the other. It goes back to normal once you click on it. It's not a theme issue because it happens with every theme. Here are some screenshots:

![screenshot from 2016-10-16 20-27-05](https://cloud.githubusercontent.com/assets/10391266/19419806/a26a0ede-93df-11e6-896b-f6cde6d96918.png)![screenshot from 2016-10-16 20-28-32]
(https://cloud.githubusercontent.com/assets/10391266/19419805/a26081c0-93df-11e6-857d-92598c708c1c.png)
_Cinnamon default theme and New-Minty theme_

The only difference my inexperienced eyes can see between the two menu items there, is that the first inherits from `PopupBaseMenuItem` (and does more things) and the second is just an instance. I don't know if that's what makes the difference.

I know this is not a priority, but it may be interesting to be fixed one day :)

Version: Cinnamon 3.0.7

JosephMcc:
Once you click on the menu item it restores itself back to the proper size and then is fine until you resart Cinnamon again. It's something to do with the icon that sometimes show on that menu item. Removing it solves the issue. I'll see if I can figure out a fix for it.

germanfr:
Do you mean this icon? 
![screenshot from 2016-10-17 15-53-09](https://cloud.githubusercontent.com/assets/10391266/19441565/67792dda-9486-11e6-838f-e900ad0ae2d9.png)
When it appears it seems to restore its normal size. I haven't clicked on it in the screenshot.

JosephMcc:
Yep, that one.

JosephMcc:
Bit of follow up. I think I have a fix for this. I'll make a PR a little later.

PR #5823, opened 2016-10-21
inhibit-applet: Fix an oversized menu item

JosephMcc:
On initial show the first menu item is overly large. This seem to be caused by
the way we show and hide _statusIcon object. Switching from using show/hide to
hiding the icon by using it's opacity fixes the issue.

Fixes https://github.com/linuxmint/Cinnamon/issues/5810


clefebvre:
Hmm...  I don't really agree with that fix.

Other applets use hide/show... should we not try and fix that within the applet core code instead?

JosephMcc:
@clefebvre Partly I agree. I did do some digging while working on this to find out the cause. The issue is being caused by the icon that is being added in the middle. I'm not sure if it is somehow a side effect of clutter not knowing the proper sized to allocate until the first time the icon is shown or what.

I've not seen it occur anywhere else though. In this case you are creating a custom menu item that is unique to this applet so it doesn't exist anywhere else.

Not sure that leaving it broken is a better solution :)

clefebvre:
OK..

PR #5824, opened 2016-10-21
add icons to panel launcher context menu

brownsr:
Adds symbolic icons for Launch, Edit, Add, Remove
Needed an additional function in popupMenu to be created


brownsr:
not needed once #5833 goes through, so closing

PR #5826, opened 2016-10-22
correct unreactive top/bottom strip in vertical panels

### LINKED ISSUE #5748: Vertical panels: No menu icon and hovering over an applet at a corner
I know that the vertical panels are still in progress and maybe this is already fixed or known, but I wanted to mention it in case it isn't.

(1) In the menu applet you can choose to have a custom icon. You can also choose to have no icon at all. That's a problem with vertical panels, because then the applet is gone. I know, who does that, huh? I don't even know what behavior I would expect at that situation. Maybe using the first letter of the word "Menu" (in the respective language) or something like that?

I'm not sure if the second one is a themeing issue (I'm using Mint-Y):
(2) Imagine you have a vertical panel, here for example left panel, without any horizontal panels. In the top left or bottom left corner you have an applet, for example the menu applet.
I you hover with your mouse to the very edge of the corner the menu applet isn't selected/hovered/green.
With horizontal panels that's not a problem at all.

brownsr:
fixes #5748.

There was a tiny bit of spacing there in the css.  Removed it.

### DIFF
--- a/data/theme/cinnamon.css
+++ b/data/theme/cinnamon.css
@@ -288,16 +288,10 @@ StScrollBar StButton#vhandle:hover {
     spacing: 0px;
 }
 .panelLeft.vertical {
-    padding-left: 0px;
-    padding-right: 0px;
-    padding-top: 2px;
-    padding-bottom: 2px;
+    padding: 0px;
 }
 .panelRight.vertical {
-    padding-left: 0px;
-    padding-right: 0px;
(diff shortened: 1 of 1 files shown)


JosephMcc:
Nothing wrong with this but couldn't you just set padding: 0; instead of doing all four sides.

brownsr:
will do

Issue #1050, opened 2012-08-24
Add configurable delay to Hot corner


juliao:
I think it would be nice if there were a option to add a delay (some milliseconds) to the 'Hot corner', to avoid accidental activation of it. By default should come as 0 ms and user could configure it.

Luke-Nukem:
Good to see someone has taken this on. It's a definite needed item.

theoar:
+1

gordon-morehouse:
This is by far the biggest annoyance I have with Cinnamon on a day-to-day basis. +1000 for a configurable hover delay.

nyeates:
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/hotCorner.js

My guess is that it would be a change in this file, if anyone wants to have a look. There is some code about jitter that assures if the mouse is slightly moving in and out of a certain area, it doesnt trigger the hot corner multiple times. Not sure if that is related or gives someone a hint. I don't know enough to make this change, but maybe someone else can.

ACinnamonUserInUK:
+1.

Hobo4ssassin:
@nyeates I haven't tried it yet, but I would guess that changing line 12 to a higher value is the ticket here

nakhla:
a better alternative "Activate when pushed" like in KDE, that require a deliberate push to trigger the action of hotcorner
or at least a delay option from terminal like:
`gsettings set org.cinnamon hotcorner-delay 500`

PR #5829, opened 2016-10-22
[Hot Corners] Added hover delay functionality.

Odyseus:
This pull request adds the ability to set a hover delay to activate hot corners. Its main purpose is to 
avoid accidental hot corners activation. I also added tooltips to the hot corners icons that reflect their current action, as suggested by @anandrkris.

The new Hot Corners settings window would look like this:

hotcornershoverdelayimplemented2

### Some technicalities
- File js/ui/hotCorner.js 
    - This PR also fixes some minor white-space *inconsistencies* on this file (some missing spaces and some extra tabs). But, as I use a code formatter plugin, it also expanded some objects. Let me know if I should revert back those objects expansions.
- File files/usr/share/cinnamon/cinnamon-settings/modules/cs_hotcorner.py
    - Here I had to add a timer to the **on_widget_changed** function to avoid repeated triggering from the new **SpinButton** elements. I tried to add that timer to a new callback for the **SpinButton** elements, but I wasn't able to pass the correct arguments to the **on_widget_changed** function. Any advice will be welcome.
    - <del>I used a **Gtk.Grid** instead of a **Gtk.HBox** because of its deprecation.</del> Switched to a **Gtk.Box** as suggested by @collinss.
- <del>This PR also adds a new string that needs to be translated. I will need advice on how to proceed with this as well.</del>

### DIFF
--- a/data/org.cinnamon.gschema.xml.in
+++ b/data/org.cinnamon.gschema.xml.in
@@ -359,7 +359,7 @@
     </key>
 
     <key name="overview-corner" type="as">
-      <default>['expo:false:false', 'scale:false:false', 'scale:false:false', 'desktop:false:false']</default>
+      <default>['expo:false:false:0', 'scale:false:false:0', 'scale:false:false:0', 'desktop:false:false:0']</default>
       <_summary>Properties of overview corners</_summary>
       <_description>Properties of overview corners, in the form functionality:hover:icon. The order in which properties are displayed is top left, top right, bottom left, bottom right.</_description>
     </key>
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_hotcorner.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_hotcorner.py
@@ -2,7 +2,7 @@
 import cairo
 import math
 
-from gi.repository import Gio
+from gi.repository import Gio, GLib
 
 from GSettingsWidgets import *
 
@@ -82,11 +82,16 @@ class Module:
             else:
                 isEnabled = False
 
-            corner.setValues(function, visible, enabled)
+            try:
+                delay = prop[3]
+            except:
+                delay = "0"
+
+            corner.setValues(function, visible, enabled, delay)
             self.cornerDisplay.setCornerEnabled(corner.index, isEnabled)
         self.cornerDisplay.queue_draw()
 
-    def onConfigChanged(self, index, function, visible, enabled):
+    def onConfigChanged(self, index, function, visible, enabled, delay):
         self.cornerDisplay.queue_draw()
 
         props = self.properties[index]
@@ -102,6 +107,11 @@ class Module:
             props[2] = 'true'
         else:
             props[2] = 'false'
+
+        try:
+            props[3] = str(delay)
+        except:
+            props.append("0")
 
         self.write_settings()
 
@@ -207,6 +217,7 @@ class HotCornerConfiguration():
     def __init__(self, index, updateCallback):
         self.updateCallback = updateCallback
         self.index = index
+        self.timer = None
         self.functionStore = Gtk.ListStore(str, str)
         self.functionStore.append(['expo', _("Show all workspaces")]) #Expo
         self.functionStore.append(['scale', _("Show all windows")]) #Scale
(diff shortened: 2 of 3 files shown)


anandrkris:
Hope this gets merged - lot of votes for this :)
https://github.com/linuxmint/Cinnamon/issues/1050

anandrkris:
Also, a suggestion - when icon is visible in the corner, can the action be displayed as a tooltip on hover over icon?
Show All workspaces, windows, custom action, etc

Odyseus:
Hello, @anandrkris.

I like your suggestion about adding a tooltip to the hot corners icons. I have been doing some tests but, sadly, without success. No matter how or at which point I add the tooltip, it never shows on icon hover.

Here is a shortened version of what I did in case someone has any suggestion:

After adding the import for the tooltip module in the **js/ui/hotCorner.js** file...

```javascript
const Tooltips = imports.ui.tooltips;
```

...I simply added the tooltip to the icon inside the **setProperties** method like this...

```javascript
if (this.icon) {
    new Tooltips.Tooltip(this.iconActor, "Tooltip text.");
    this.iconActor.show();
} else
    this.iconActor.hide();
```

Is it even possible for the hot corners icons to have a tooltip?

NikoKrause:
@Odyseus 
Try this.
```js 
this.tooltip = new Tooltips.Tooltip(this.iconActor, "Tooltip text.");
```

Odyseus:
Hello, @NikoKrause.

I already tried that. The very first thing that I tried was to create the tooltip inside **HotCorner._init** and then use the **set_text()** method inside the **setProperties** method to apply the correct tooltip text. It's pretty straight forward and it works in every applet code that I ever used, except here with the hot corners.

NikoKrause:
@Odyseus
Does my suggestion doesn't work?
I tried your code and added `this.tooltip =`, which you forgot before `new`
And it showed the Tooltip on icon hover like expected.

Odyseus:
@NikoKrause.

> Does my suggestion doesn't work?

Like I said, I already tried that. I used the exact code that you posted and it didn't work.

> I tried your code and added this.tooltip =, which you forgot before new

I didn't really forget. It isn't needed to store the tooltip in a new variable because the tooltip is already *added* to the element passed by the first argument of the constructor ([here](https://github.com/Odyseus/CinnamonTools/blob/master/Applets/0dyseus%40QuickMenu/applet.js#L375) is a functional example that I use in one of my applets). Eighter way, like I stated before, I already added `this.tooltip = ` in front of that line and it still doesn't work for me.

I did my tests in all of my 4 Cinnamon systems with the same results. The tooltip isn't shown when hovering over the hot corner icon.

I will give this rest for now. Maybe someone else can create a different pull request that adds the tooltips to the hot corners icons.

NikoKrause:
@Odyseus 
Could you check if my branch works for you https://github.com/NikoKrause/Cinnamon/tree/hot-corner-tooltip? I created a commit with your code in there https://github.com/NikoKrause/Cinnamon/commit/60993434fb7410879a943358c02ce0ff334cb2a7.

Just replace the hotCorner.js file in `/usr/share/cinnamon/js/ui` with the file from my branch.

It looks like this on my machine:
bildschirmfoto vom 2016-10-25 20-25-10

Odyseus:
@NikoKrause

No, it still isn't working on any of the 4 Cinnamon systems I tried before. **But**, I tried it on a fith system (an Archlinux system with Cinnamon) and all tooltips work like a charm here! This is very strange!

I hesitate to add to this pull request something that I can't make work reliable on all my systems. If you want to add this feature yourself and send the pull request, [here is a Gist](https://gist.github.com/Odyseus/084193105d6bd8fc9cc06da4e4f4a237) with the **hotCorner.js** file with icons tooltips fully implemented.

Odyseus:
@NikoKrause and other code reviewers.

Ignore the previous posts discussing problems while setting the tooltips for the hot corners icons. I had an extension enabled that was interfering with the hot corners code. I disabled it and everything started working as it should.

@anandrkris: If this PR *doesn't make it*, NikoKrause just send a [pull request](https://github.com/linuxmint/Cinnamon/pull/5852) adding the tooltips to the hot corner icons.

Apologies for the *noise*.

collinss:
@Odyseus A couple notes/suggestions.

- You don't need to worry about the translations. Putting ```_()``` around the string is enough to insure that it will get translated for the next release.
- ```Gtk.HBox``` and ```Gtk.VBox``` were deprecated in favor of ```Gtk.Box```, which adds an additional orientation property to make it vertical. Unless you need the extra layout features of ```Gtk.Grid``` (which you do not here), I would recommend using ```Gtk.Box```.
- I would recommend adding the tooltip to the actor as soon as you create it. If you need to make sure the tooltip is not shown under certain circumstances, you can set the text to an empty string later (you don't need to do this if your actor is hidden as it wont show anyway). The way you did it above could lead to multiple tooltips on the same actor, all competing with each other. This might also solve your problem with it not showing, but I can't guarantee anything there.

Odyseus:
Hello, @collinss.

- **About translations:** Got it.
- **About `Gtk.Box`s:** I have chosen a `Gtk.Grid` because I couldn't make the `Gtk.Box` *"look good"* (basically, I couldn't understand the arguments passed to **pack_start**). But since you recommended me to use a `Gtk.Box`, I gave it another try and I successfully made it work.
- **About the tooltips:** The code that I posted was just an example. So that someone willing to help me out finding the root of my problem could quickly set up a tooltip with just two lines of code. Thanks to NikoKrause's assistance, I figured that the *problem* was on several of my systems.

Thanks for your notes/suggestions, collinss.

Pull request updated.

anandrkris:
Thanks @Odyseus for listening :smile: and again for your efforts.  :clap:

PR #5833, opened 2016-10-23
panel-launchers: Redesign the launcher context menu

JosephMcc:
Now that we have support for launcher specific actions in the menu, put some of
the less used options in a submenu. Also add an additional submenu to hold the
menu items common to the right click context menu for most applets. This also
gives an easy way to remove a single instance of the panel launchers. Something
that is currently missing.

### DIFF
--- a/files/usr/share/cinnamon/applets/panel-launchers@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/panel-launchers@cinnamon.org/applet.js
@@ -1,4 +1,5 @@
 const Applet = imports.ui.applet;
+const AppletManager = imports.ui.appletManager;
 
 const Clutter = imports.gi.Clutter;
 const St = imports.gi.St;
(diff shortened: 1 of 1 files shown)


NikoKrause:
@JosephMcc 
This commit causes a regression:
The options "Add" and "Edit" do not work anymore.

Setting `event.get_time()` to `null` seems to solve this problem. 
```js
     _onAddActivate: function(event) {
         this._launcher.launchersBox.showAddLauncherDialog(null);
     },
 
     _onEditActivate: function(event) {
         this._launcher.launchersBox.showAddLauncherDialog(null, this._launcher);
     },
```
I'm not sure for what `event.get_time()` is needed though and if this would cause some other problems. So I think it's better if you check that.

JosephMcc:
Your right. Thanks. PR'd a fix here https://github.com/linuxmint/Cinnamon/pull/5898

PR #5852, opened 2016-10-25
added tooltip to hot corner icons

NikoKrause:
@Odyseus asked me to create this pull request
It seems, that this works only on 1 out of 5 of his machines.
I didn't had problems on my machine ;)


NikoKrause:
@Odyseus
I'm not very happy wit the getIconTooltip function.
Isn't it possible to get the Strings directly from the hotCorners module python file?

PR #5849, opened 2016-10-25
[sound-applet] add option: show percentage next to volume slider

NikoKrause:
Closes https://github.com/linuxmint/Cinnamon/issues/5809

Thanks to @itzexor and @mtwebster, who helped me preventing the popupMenu to change width.

Please feel free to rephrase description and rename variables/functions.

PR #5854, opened 2016-10-26
[sound-applet] add option: show percentage next to volume slider

### LINKED ISSUE #5809: New feature proposal: Percentage label in sound applet
Currently while you are adjusting the volume you can't see the exact value, only when you finished can you hover over it, so the popup label appears. It'd be nice to have a label between the sound icon and the bar indicating the volume in percentage. If the devs nod to this, i'd be interested in implementing it.

NikoKrause:
Closes #5809

Thanks to @itzexor and @mtwebster, who helped me preventing the popupMenu to change width.

rebased version of https://github.com/linuxmint/Cinnamon/pull/5849


Vahan86:
Would be great to have this feature for brightness slider as well.

NikoKrause:
Yeah, I thought about that. But before doing that I'll wait, if this one will be merged or not.

PR #5859, opened 2016-10-27
[menu-applet & panel-launchers] added new mouse cursors while draggin…

NikoKrause:
…g apps
- dragging an app from appsBox into favsBox shows "dnd-copy" cursor
- dragging an app from favsBox into appsBox/catsBox shows (pointing)"hand" cursor (so you can see that favs can be dropped/deleted there)
- dragging an app from menu into panel-launchers shows "dnd-copy" cursor

Closes https://github.com/linuxmint/Cinnamon/issues/4000

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -961,6 +961,13 @@ CategoriesApplicationsBox.prototype = {
             return true;
         }
         return false;
+    },
+
+    handleDragOver : function(source, actor, x, y, time) {
+        if (source instanceof FavoritesButton)
+            return  global.set_cursor(Cinnamon.Cursor.POINTING_HAND);
+
+        return DND.DragMotionResult.CONTINUE;
     }
 }
 
@@ -1052,6 +1059,13 @@ FavoritesBox.prototype = {
             if (fadeIn)
                 this._dragPlaceholder.animateIn();
         }
+
+        let id = app.get_id();
+        let favoritesMap = AppFavorites.getAppFavorites().getFavoriteMap();
+        let srcIsFavorite = (id in favoritesMap);
+
+        if (!srcIsFavorite)
+            return DND.DragMotionResult.COPY_DROP;
 
         return DND.DragMotionResult.MOVE_DROP;
     },
--- a/files/usr/share/cinnamon/applets/panel-launchers@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/panel-launchers@cinnamon.org/applet.js
@@ -630,7 +630,10 @@ MyApplet.prototype = {
             if (fadeIn) this._dragPlaceholder.animateIn();
         }
 
-        return DND.DragMotionResult.MOVE_DROP;
+        if (source instanceof DND.LauncherDraggable && source.launchersBox == this)
+            return DND.DragMotionResult.MOVE_DROP;
+
+        return DND.DragMotionResult.COPY_DROP;
     },
 
     acceptDrop: function(source, actor, x, y, time) {


NikoKrause:
"dnd-copy" is a hand with a plus symbol.

For the second option (the one with the (pointing)"hand" cursor) I would have liked to have another cursor. In my opinion it should be a hand with a minus symbol, called "dnd-remove".

But there was no such cursor. Maybe it is possible to create such a cursor?

Odyseus:
Hello, @NikoKrause.

The `cinnamon-global.c` file is a source file and it's located only in this repository ([**src/cinnamon-global.c**](https://github.com/linuxmint/Cinnamon/blob/master/src/cinnamon-global.c)). You will not see it in an installed system. In that file are only 15 different cursors defined. As I understand it (I know nothing about C language) the cursors are extracted(?) from a module(?) called [**GdkCursor**](https://developer.gnome.org/gdk3/stable/gdk3-Cursors.html#gdk-cursor-new-from-name)? There are available 30 different cursors. But I thing that the way to go would be to add all the cursors from the the [CSS specification](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor).

Either way, what you want to achieve *"isn't possible"*. The cursor that you want to implement (a hand with a minus symbol) doesn't exist in any standard implementation (not in the CSS specification linked above nor in the [freedesktop cursor conventions specification](https://freedesktop.org/wiki/Specifications/cursor-spec/)) nor in any cursor theme (that I know about).

NikoKrause:
I created such a cursor for the DMZ-White cursor theme.
It looks like this:
dnd-remove

Here is the link to the file: 
https://github.com/NikoKrause/arc-theme/blob/master/dnd-remove

Isn't it possible to include this to the DMZ-White cursor theme and use it in the menu-applet instead of the pointing hand cursor?

NikoKrause:
@Odyseus 
I think Cinnamon uses the cursor theme from Herr
https://packages.qa.debian.org/d/dmz-cursor-theme.html

NikoKrause:
I got in touch with the Debian developers. Here is the response I got:

> Debian is only maintaining the packaging of the Debian
> dmz-cursor-theme package, not the development of the cursor theme nor
> the creation of new icons.
> 
> According to the Debian packaging, the upstream git repository is on
> gitorious, which is now defunct and read-only:
> 
> http://metadata.ftp-master.debian.org/changelogs/main/d/dmz-cursor-theme/unstable_copyright
> https://gitorious.org/
> https://gitorious.org/opensuse/art
> 
> Here is the history of the upstream git repository:
> 
> https://gitorious.org/opensuse/art?p=opensuse:art.git;a=history;f=cursors/dmz;hb=HEAD
> 
> It appears that the right person to contact about adding new icons to
> the DMZ cursor theme is this person:
> 
> Jakub Steiner <jimmac@gmail.com>
> 
> Here is his website:
> 
> http://jimmac.musichall.cz/


Does this

> the upstream git repository is on gitorious, which is now defunct and read-only:

mean, that even when a new icon will be added, I won't be able to use it, because the repository is defunct?

collinss:
@NikoKrause the problem with adding custom cursors is that most cursor themes will not support them, and not everyone uses the same cursor theme. As for adding new cursors, I did that a while back in #4465. You could see what I did there.

NikoKrause:
@collinss 
Thanks for your reply.

> the problem with adding custom cursors is that most cursor themes will not support them, and not everyone uses the same cursor theme

I used some custom cursor themes,, which don't have for example dnd-copy cursor. The cursor falls back and uses the DMZ-White cursor instead. What I was wondering about: Is Cinnamon shipped with the DMZ cursor theme by default for other OSes as well (e.g. in ArchLinux?)

I see that you changed the global file. Where did you put the file of the new cursor? I don't see that in the commit. Thanks for your hell.

Odyseus:
@NikoKrause
Cinnamon doesn't come with any cursor theme (we could say that Cinnamon is cursor theme agnostic). The DMZ-White/Black cursor themes are shipped with Linux Mint. Cinnamon on Arch Linux doesn't install the DMZ cursor theme (although, it can be installed from its package manager). What **collinss** did in that commit was to add to one of Cinnamon's modules the ability to programmatically set more cursor types. It has nothing to do with any cursor theme in particular.

You should abandon the idea of changing the cursor theme itself. In the commit linked by **collinss**, **lestcape** [mentions a method](https://github.com/linuxmint/Cinnamon/pull/4465#issuecomment-124745026) that could be just what you are looking for (although, I didn't understood half of what the code does (LOL)).

NikoKrause:
> What collinss did in that commit was to add to one of Cinnamon's modules the ability to programmatically set more cursor types.

He creates new cursor types, but he doesn't use new cursor icons for that (he uses only existing cursor icons), does he?

So I guess Cinnamon needs his own Cursor Icon Theme ;) I think the absent possibility to add new icons is a restriction you have to live with.

I think it would be better to use an existing icon, isn't it?

Any proposals, which icon I should use for this feature instead? I'm not very happy with the pointing hand. Maybe the dnd-link icon?
dnd-link
Or the dnd-ask icon?
dnd-ask

NikoKrause:
> It appears that the right person to contact about adding new icons to
> the DMZ cursor theme is this person:
> 
> Jakub Steiner

@jimmac
I would like to hear your opinion on this thread, i.e. about how good/bad it is to add new cursor icons.

NikoKrause:
@collinss 
Right now I had an idea, but I don't know if it is possible to do.
I would linke to create a new cursor type like you did. I want to say that cursor type, use the custom cursor (handhand with minus Symbol). If this cursor doesn't existexist in your cursor then, then use the cursor with the pointing Hand, which exists for sure. 
Do you know if this is possible?

jimmac:
DMZ is bitrotten and no longer maintained. gtk+ itself provides core cursors, including improved DND ones.

NikoKrause:
@jimmac 
WhereWhere can I find them? Can you post a link?

jimmac:
Sorry, it's not gtk+ itself, but adwaita-icon-theme. The 'sources' live in https://git.gnome.org/browse/adwaita-icon-theme/tree/src/cursors

clefebvre:
Closing this for now.

NikoKrause:
Why?

clefebvre:
Sorry I assumed this was dead and going nowhere.

NikoKrause:
No, it worked fine. The follow up discussion was only asking if a new cursor icon could be added. Since this is hard to accomplish, I just usw the pointy finger cursor instead.

You can try it, if you like the chosen cursors.

NikoKrause:
So is this closed as "Do not want to" or "Do not like it"?

PR #5861, opened 2016-10-27
menu-applet: Re-enable the animation on menu open and close

JosephMcc:
With the removal of the boxpointer this now looks really out of place. It's still
possible to disable this globally in cinnamon-settings effects.


clefebvre:
I feel different about this. I agree it's inconsistent, but then again so are people's expectations on that particular applet. They want it to show up as fast as possible. The way our animation is at the moment, it's taking quite a lot of time to poup.

Assuming we can't fine-tune this to be fast enough for the menu applet and slow enough for the animation to look good everywhere else... I'd suggest a menu setting, so that no effects would apply to the menu by default, and checking that setting it would then behave like the rest of the applets.

JosephMcc:
The one thing I would disagree with is turning off the animation by default. I can add the setting to the menu applet itself but would prefer to have people turn the animation off if they don't like it. It looks really out of place.

Ultimately this would be a simple change if we get a lot of feedback one way or the other during the beta.

PR #5873, opened 2016-10-29
Window-list applet - ensure that the edge of the screen reacts to mouse clicks

brownsr:
Previously the outside couple of pixels were unreactive.  This was done to avoid recursion errors
but these seem no longer to be there, so this resets to the desired behaviour

### DIFF
--- a/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
@@ -591,8 +591,7 @@ AppMenuButton.prototype = {
                 alloc.natural_size = 150 * global.ui_scale;
             }
         } else {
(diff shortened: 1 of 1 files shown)


files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
```diff
@@ -1033,10 +1029,8 @@ MyApplet.prototype = {
             this.actor.set_important(true);
         }
 
-        //
-        // For horizontal panels any padding/margin is removed on one side
-        // so that the AppMenuButton boxes butt up against the edge of the screen
-        //
+// Any padding/margin is removed on one side so that the AppMenuButton boxes butt up against the edge of the screen
```
JosephMcc: Please indent comments to the same level as the rest of the code.

PR #5878, opened 2016-10-30
[Tooltips] Fixed positioning of tooltips.

Odyseus:
Currently, tooltips created with the **new Tooltips.Tooltip()** constructor are aligned to the top-left corner of the mouse pointer. This leads to having tooltips overlapped by the mouse pointer. And the bigger the mouse pointer, more of the tooltip is covered by it.

This commit aligns the tooltip to the bottom-right corner of the mouse pointer (approximately), reducing the possibility of the mouse pointer to overlap the tooltip.

#### Before
Tooltips broken

#### After
Tooltips fixed
### Some technicalities
- This change only affects tooltips created with the **new Tooltips.Tooltip()** constructor.
- As I didn't want to use a fixed size, I used the global preference _cursor-size_.

### DIFF
--- a/js/ui/tooltips.js
+++ b/js/ui/tooltips.js
@@ -6,7 +6,9 @@ const Applet = imports.ui.applet;
 const Main = imports.ui.main;
 const SignalManager = imports.misc.signalManager;
 const Tweener = imports.ui.tweener;
-
+const Gio = imports.gi.Gio;
+const DESKTOP_SCHEMA = 'org.cinnamon.desktop.interface';
+const CURSOR_SIZE_KEY = 'cursor-size';
 /**
  * #TooltipBase
  * @item (Clutter.Actor): The object owning the tooltip.
@@ -95,7 +97,7 @@ TooltipBase.prototype = {
         }
     },
 
-    _onTimerComplete: function(){
+    _onTimerComplete: function() {
         this._showTimer = null;
 
         if (!this.preventShow)
@@ -125,11 +127,11 @@ TooltipBase.prototype = {
         this.signals.disconnectAllSignals();
         this._destroy();
     }
-}
+};
 
 /**
  * #Tooltip:
- * 
+ *
  * This is a tooltip item that displays some text. The tooltip will be
  * displayed such that the top left corner of the label is at the mouse
  * position.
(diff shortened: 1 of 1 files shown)


collinss:
This looks out of place to me, and could potentially be confusing to the average user in some situations. It might work better if you place it above or to the right instead. That's just a suggestion, though. I'm not exactly a UI expert.

Odyseus:
Hello, @collins.

I'm not an UI expert either, but I have chosen the current positioning by observing how the tooltips of several applications work. And what I observed was that all applications have completely different approaches.

The tooltips of the Gtk2 applications that I checked (Gimp and InkScape) look like this...

ttgimp1 ttgimp2

...and they don't consider the mouse cursor size.

Qt applications are a completely messed up universe (LOL). Most of the ones that I tried have custom tooltips. Others have the same tooltips as the ones used by Gtk2 applications.

And finally, Gtk3 applications. Almost all Gtk3 applications position their tooltips related to the UI elements that they belong and center aligned. Like this on Nemo's UI...

ttnemo2

But inside Nemo views, the tooltips mostly look like this...

ttnemo1

Seeing that all application handle tooltips completely different:

- Gtk2 tooltips are too close to the cursor and a big mouse cursor overlaps the tooltip.
- Qt tooltips are a mess or the obey Gtk2 rules.
- Gtk3 tooltips are too complex to implement (aligned to the element).
- Tooltips on Nemo's internal views are to far away from the mouse cursor.

I opted to use the one implementation that I liked the most and made more sense to me. Which is the tooltips used by Firefox...

ttfirefox1

Although, seeing that the *trend* is for the tooltip to be aligned diagonally with respect to the cursor, I could make them look like Nemo tooltips, like this...

tttweaked

Placing the tooltip above the cursor wouldn't make much sense. All applications across all toolkits display their tooltips bellow the cursor, unless there is no room, then they are displayed above the cursor. And placing them to the right of the cursor would present the original problem. When there is no room to the right to fit the tooltip, the tooltip is moved to the left and the cursor overlaps the tooltip.

I know that this entire commit is more of a workaround than a solution. Mainly because the *cursor-size* setting isn't the actual size of the cursor. But it's an easy fix that would improve the readability of Cinnamon's tooltips in most use cases.

Odyseus:
I updated this pull request.
- I changed the code so the tooltips are now aligned diagonally with respect to the mouse pointer.
- I added a more detailed commit message.
- I edited my first comment to reflect all of this changes.

PR #5887, opened 2016-10-31
[menu-applet] hide recent files, which are no longer available 

NikoKrause:
Recent files, which are no longer available, will be hidden in the recent files category after next call of refreshRecent.
If the file is available again, because it was on a CD or USB, it will appear again.
If a file is no longer available, but the recent files are not refreshed yet, the description of the file shows a warning "**This file is no longer available**".

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -2092,6 +2092,10 @@ MyApplet.prototype = {
                             if (fileIndex !== -1)
                                 selectedAppUri = selectedAppUri.substr(fileIndex + 7);
                             this.selectedAppDescription.set_text(selectedAppUri);
+
+                            let file = Gio.file_new_for_uri(button.file.uriDecoded);
+                            if (!file.query_exists(null))
+                                this.selectedAppTitle.set_text(_("This file is no longer available"));
                             }));
                     button.actor.connect('leave-event', Lang.bind(this, function() {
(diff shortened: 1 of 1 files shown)


DirkHaar:
A plus would be to check file availability when opening the menu - this shouldn't take half a second.
A minus: changing mentiones USB-device in and out when searching something manually ("mmh... which version was the correct one?") wouldn't allow you to click those in the menu afterwards, but I think this can be diregarded.

NikoKrause:
@DirkHaar
I don't understand your comment. Could you explain this more detailed, what you mean or what the issue is and how to reproduce it.

DirkHaar:
Why couldn't the menu check if the files in the "recently used" list still exist *before* listing them?

NikoKrause:
@DirkHaar 
So you mean, files, which aren't no longer available, should be removed from the recent files list?

DirkHaar:
Yes.
(Despite the mentioned disadvantage when you juggle with several external media swapping them in and out searching for something - which would make files temporarily "not existing" if you ask for them. I could live with that and would prefer a "cleaner" list, but if others mind about external files as described it shouldn't be changed. There should be a sentence as comment in the help file to clarify how the recent file list will work, or maybe an entry in the release notes.

NikoKrause:
I rebased the PR and edited the title and description.

The deleted or unmounted files will now be hidden in recent category list.
But this happens not until the recent files were refreshed.

If a file is no longer available, but the recent list was not refreshed, there will be shown a warning in the files description, that the file is no longer available.

PR #5891, opened 2016-11-01
[Xlet settings] "entry" and "iconchooser" elements modifications.

Odyseus:
By default, "entry" and "iconchooser" editable fields are too small on xlet's setting windows. In addition, they don't expand their width when the window is resized. This makes those elements *uncomfortable to work with*.

This commit adds to the "entry" and "iconchooser" elements a new boolean option (optional) called "expand-width". When set to **true**, the editable fields for "entry" and "iconchooser" elements will occupy all available space in the row and at the same time will expand if the window is resized.

#### Comparisons

This is how "entry" and "iconchooser" elements looks like on current Cinnamon stable (3.0.7). Editable fields occupy all the available space and they expand on window resize.

entrywidthonstable

This is how those elements looks like on Cinnamon nightly with the new proposed option ("expand-width") set to **true** in the section called **Custom launcher 1** and with the same option omitted on the section called **Custom launcher 2**. The editable fields on section **Custom launcher 1** will expand when the window is resized.

entrywidthnormal

### Technicalities

I have chosen to add this new option instead of directly editing the python functions that creates the "entry" and "iconchooser" elements because I considerer it less intrusive and it doesn't affect existing elements using those functions (like the away message option on the Screensaver settings window).

### DIFF
--- a/docs/reference/cinnamon-tutorials/xlet-settings-ref.xml
+++ b/docs/reference/cinnamon-tutorials/xlet-settings-ref.xml
@@ -21,6 +21,7 @@
         <listitem><code>type</code>: should be <code>entry</code></listitem>
         <listitem><code>default</code>: default string value</listitem>
         <listitem><code>description</code>: String describing the setting</listitem>
+        <listitem><code>expand-width</code>:  (optional) true or false, or leave off entirely. Forces editable fields of <code>entry</code>'s elements to occupy the entire space available in a row</listitem>
       </itemizedlist>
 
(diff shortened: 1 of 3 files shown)


clefebvre:
Hi @Odyseus,

I can't reproduce the issue, or understand it...

I'm looking at master without your fix here, and the screensaver away message entry does expand already..

clefebvre:
ah sorry.. you meant Xlet configure UI. I got on the wrong track cause of the away message.

Issue #5647, opened 2016-09-05
New settings system color choosers without alpha


Odyseus:
I installed in one of my virtual systems the Cinnamon version hosted on GitHub specifically to test the new system settings and see if my applets were compatible with the new system. In short, it's an amazing redesign of the settings system and, for what I could test so far, it's perfectly retro-compatible. Very good job!!!

But, I found out that it isn't possible for the color choosers to select a transparent color. This may not be important in some places (the desktop background, for example), but for applets that have an option to set their background color is absolutely important.

Trying to find a fix, I compared how the color choosers worked on Cinnamon stable and _nightly_ (from GitHub). I found out where the color choosers were _created_ and did the following:
- I opened the `/usr/share/cinnamon/cinnamon-settings/bin/SettingsWidgets.py` file.
- After the line 936, inside the ColorChooser class, I added the following code:

``` python
self.content_widget = Gtk.ColorButton()
# The next line enables color choosers to have alpha.
self.content_widget.set_use_alpha(True)
```
- After restarting Cinnamon, color choosers were able to select colors with alpha.

Now, this _tweak_ enables alpha for color choosers globally, not just the color choosers on applets settings windows. In Cinnamon stable, that same line of code is used on the `/usr/share/cinnamon/cinnamon-settings/bin/XletSettingsWidgets.py` file and enables alpha on color choosers found only inside applets settings windows. I don't know if the Cinnamon developers omitted this on purpose, but the main issue remains. Applets need color choosers with alpha. Otherwise, it isn't possible to match the color of panels with gradients or, even worse, panels with transparency.
### My system information
- **Operating System:** Linux Mint 18 Cinnamon 64-bit
- **Cinnamon Version:** 3.0.7 **(latest version from GitHub)**
- **Linux Kernel:** 4.4.0-21-generic

PR #5892, opened 2016-11-02
[Xlet settings] Added option to "colorchooser" elements to enable alpha.

Odyseus:
By default, "colorchooser" elements can't choose a color with alpha (transparency). Contrary to what it occurs in the current stable Cinnamon version 3.0.7, where all "colorchooser" elements from xlet's settings windows are able to select a color with alpha.

This commit adds to the "colorchooser" elements a new boolean option (optional) called "enable-alpha". When set to **true**, "colorchooser" elements will be able to select a color with alpha. When set to **false** or ignored, "colorchooser" elements will not be able to select a color with alpha.

This commit closes linuxmint/Cinnamon#5647

### Technicalities

I first considered to enable alpha for color choosers globally by directly modifying the **ColorChooser** function (located in [this file](https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/cinnamon-settings/bin/SettingsWidgets.py#770)). But I didn't know if that would *break* something that uses a color chooser and it doesn't support a color with alpha. So, instead, I have chosen to add this new option, because I considerer it less intrusive and it doesn't affect existing elements using those functions (like the color choosers on the Backgrounds settings window). **If someone can confirm that enabling alpha for color choosers globally wouldn't *break* anything, I could *convert* this pull request or create a new one for that purpose.**

### DIFF
--- a/docs/reference/cinnamon-tutorials/xlet-settings-ref.xml
+++ b/docs/reference/cinnamon-tutorials/xlet-settings-ref.xml
@@ -44,6 +44,7 @@
         <listitem><code>type</code>: should be <code>colorchooser</code></listitem>
         <listitem><code>default</code>: default color string - can be <code>"red"</code> or <code>"rgba(x,x,x,x)"</code>, etc...</listitem>
         <listitem><code>description</code>: String describing the setting</listitem>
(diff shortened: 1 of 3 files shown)


clefebvre:
Hi @Odyseus,

It looks like this was omitted by mistake. I'll close this PR and simply re-add it instead.

If for any reason it needs to be disabled in some cases, then we'll make it configurable. Otherwise it seems overkill.

PR #5894, opened 2016-11-02
[window-list] context menu icons for workspaces

NikoKrause:
Added icons for "Visible on all workspaces" and "Only on this workspace"
bildschirmfoto vom 2016-11-02 15-10-40


clefebvre:
I'll close this one for now for the reasons below. Feel free to reimplement it in the future.

- Inconsistency with multi-monitor options (we would need an icon there as well).
- The chosen icons aren't specific enough. They don't look like workspaces. We would need to make our own probably?

Issue #5774, opened 2016-10-08
[BUG] Dialog gets softlocked when choosing an image for any menu applet


feren:
How to reproduce:
1. Try to configure 'Menu' in Cinnamon Unstable
2. Change the custom icon
3. Try to open the new icon in the dialog, the icon is now changed, but now the dialog is effectively softlocked and won't close when clicking on 'Cancel'...

JosephMcc:
I can reproduce this. I does close using the dialogs close button though.

feren:
For some reason, once, I was unable to close it without closing the parent window... :/

zquestz:
Yeah and escape works too, but I can also reproduce this.

feren:
(Fixed by Odyseus in Cinnamon 3.2.2)

PR #5922, opened 2016-11-09
[Xlet settings] Fix for icon choosers stuck dialogs

Odyseus:
[Xlet settings] Fix for icon choosers stuck dialogs

Currently, after choosing an icon file from a "filechooser" element on a settings window for any xlet, the file chooser dialog will not close itself after clicking the **Accept** button.

This commit fixes that behavior and other minor Gtk warnings.

Fixes linuxmint/Cinnamon#5774

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/SettingsWidgets.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/SettingsWidgets.py
@@ -962,11 +962,11 @@ class IconChooser(SettingsWidget):
             self.preview.set_from_icon_name(val, Gtk.IconSize.BUTTON)
 
     def on_button_pressed(self, widget):
-        dialog = Gtk.FileChooserDialog(_("Choose an Icon"),
-                                           None,
-                                           Gtk.FileChooserAction.OPEN,
-                                           (Gtk.STOCK_CANCEL, Gtk.ResponseType.CANCEL,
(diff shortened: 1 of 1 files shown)


JosephMcc:
Seems to work properly and solve the issue.

feren:
Can anyone consider merging this?

clefebvre:
@Odyseus why use _("_Cancel") and _("_Open")? Why not the stock values?

Odyseus:
Hello, everybody.

@clefebvre.
I just used the same **FileChooserDialog** as used in another part of that file ([around here](https://github.com/Odyseus/Cinnamon/blob/0fa68f9061441c18d20a8b03ae74da6e6927c324/files/usr/share/cinnamon/cinnamon-settings/bin/SettingsWidgets.py#L891)), not because I preferred it. Also, I liked the fact that passing the **transient_for** argument got rid of a Gtk warning, so I left it like that.

Do you want me to make the change to use **Gtk.STOCK_CANCEL** and **Gtk.STOCK_OPEN**? Should the other instances of  _("_Cancel") and _("_Open") be changed too?

JosephMcc:
Gtk.STOCK has also been deprecated for quite some time. Probably doesn't hurt to move to non-deprecated functionality.

clefebvre:
Oh yes, that's fine, I was just afraid of new msgids being introduced, I didn't realize there were already used all over the place.

Issue #5668, opened 2016-09-10
Missing translations mechanims on Json Setting Widgets (xlet)


lestcape:
There are not right now  a translation mechanism for the old xlet settings.

Issue #2245, opened 2013-07-09
settings-schema.json unicode and utf-8


ghost:
The real problem is the impossibility of put special characters in the file settings-schema.json, when the file settings-schema.json, is not utf-8, but is compatible with utf-8. 

This is an example of what happens:

Written in-------------------------Written---------------------Read ----------------------English

settings-schema.json--------detecciÃ³n---------------detección----------------detection
desklet.js-------------------------detecciÃ³n---------------detecciÃ³n---------------detection
settings-schema.json--------detección-----------------Error reading------------detection
desklet.js-------------------------detección-----------------detección-----------------detection

This is an example of what happens in json files:

Written in settings-schema.json-------------------It is written in MyGesklet-uuid.json(config folder)

detecciÃ³n-------------------------------------------------detección
detección---------------------------------------------------detecci\F3n

settings-schema.json and MyGesklet-uuid.json(config folder) need to be equals.

Cinnamon.get_file_contents_utf8_sync(filePath) It should return the converted file to utf-8 if it's possible. Maybe it's a problem of JSON, when it reads the file and calculate the md5 in /usr/share/cinnamon/js/ui/settings.js

When writing unicode characters (Example:\u00f1) in "/home/user/.local/share/cinnamon/desklets/MyDesklet-uuid/settings-schema.json" Cinnamon makes a copy of "settings-schema.json" file in "/home/user/.cinnamon/configs/MyGesklet-uuid/MyGesklet-uuid.json". This copy("MyGesklet-uuid.json") does not contain utf-8 format and therefore can not load cinnamon desklet.

Maybe related to:
https://bugzilla.gnome.org/show_bug.cgi?id=687095
https://bugzilla.gnome.org/show_bug.cgi?id=690716
https://bugzilla.gnome.org/show_bug.cgi?id=599860

Possible solution:
https://mail.gnome.org/archives/commits-list/2011-June/msg01574.html

if you run this command on terminal the problem goes away until needed again change the file MyGesklet-uuid.json
Command: iconv -f iso-8859-1 -t utf-8 "pathToConfig" -o "pathToConfig"

Where pathToConfig=home/user/.cinnamon/configs/MyGesklet-uuid/MyGesklet-uuid.json

Please fix this.
Thanks.

PR #5930, opened 2016-11-10
[Xlet settings] Added translation mechanism to xlets settings windows

Odyseus:
Xlets settings system currently doesn't have a translation mechanism to translate installed xlets. So, when an xlet with .po files is installed, its settings window wouldn't use the translated strings.

This commit adds the ability to use the corresponding translations for the installed xlet.

### Technicalities

- Since the **xlet-settings.py** file switched from using Python2 on Cinnamon 3.0.7 to use Python3 on Cinnamon 3.2.x and the **gettext** module doesn't have a **ugettext** method on Python3, ~~I used the **lgettext** method for getting the translations. As I understand it, they are equivalent(?)~~. **Edit:** I switched to the **gettext** method.
- I don't know if I used the **decode()** method the right way inside the **translate** function. Someone more skilled should take a closer look at that. **Edit:** I switched to use **decode("utf-8")**.
- Basically, this commit gets the job done. But someone more skilled with Python should take a **very close look** at the changes that I added.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py
@@ -52,10 +52,10 @@ def translate(uuid, string):
     #check for a translation for this xlet
     if uuid not in translations:
         try:
-            translations[uuid] = gettext.translation(uuid, home + "/.local/share/locale").ugettext
+            translations[uuid] = gettext.translation(uuid, home + "/.local/share/locale").gettext
         except IOError:
             try:
-                translations[uuid] = gettext.translation(uuid, "/usr/share/locale").ugettext
+                translations[uuid] = gettext.translation(uuid, "/usr/share/locale").gettext
             except IOError:
                 translations[uuid] = None
 
@@ -65,6 +65,12 @@ def translate(uuid, string):
 
     if translations[uuid]:
         result = translations[uuid](string)
+
+        try:
+            result = result.decode("utf-8")
+        except:
+            result = result
+
         if result != string:
             return result
     return _(string)
@@ -74,7 +80,7 @@ class MainWindow(object):
         self.type = xlet_type
         self.uuid = uuid
         self.selected_instance = None
-        
+
         self.load_xlet_data()
         self.build_window()
         self.load_instances()
@@ -180,7 +186,7 @@ class MainWindow(object):
             icon_path = os.path.join(self.xlet_dir, "icon.png")
             if os.path.exists(icon_path):
                 self.window.set_icon_from_file(icon_path)
-        self.window.set_title(self.xlet_meta["name"])
+        self.window.set_title(translate(self.uuid, self.xlet_meta["name"]))
 
         self.window.connect("destroy", self.quit)
         self.prev_button.connect("clicked", self.previous_instance)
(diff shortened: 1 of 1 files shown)


lestcape:
Hi @Odyseus thanks for that, i personal appreciate  this:
https://github.com/linuxmint/Cinnamon/issues/5668

Please wait for the opinions of others with more skill in python 3...

The gettext module in Python 3 does not have a ugettext method because everything's already unicodes. **u**gettext it's because **unicode**gettext.

I think the properly way it's try catch this, as python it's hacked in several distributions to then not need to port things. Probably:
```
  try:
        return t.ugettext(msg)
    except AttributeError:
        return t.gettext(msg)
```
Some people use gettext instead of ugettext, as:

**gettext.gettext(message)**
    Return the localized translation of message, based on the current global domain, language, and locale directory. This function is usually aliased as _() in the local namespace (see examples below).
**gettext.lgettext(message)**
    Equivalent to gettext(), but the translation is returned in the preferred system encoding, if no other encoding was explicitly set with bind_textdomain_codeset().
https://docs.python.org/3/library/gettext.html
So, this is a thing that depend of what you want to do, but i think it's preferable  gettext instead of lgettext.

I will try to be more explicit using the decode/encode functions:
result = result.decode('utf-8') instead of result = result.decode(), in python 3 decode/encode method used 'utf-8' as default, but this contribute to understand the code better and ensure what you try to do.

As finally, it's recommended in python 3 ensure to save the file as utf-8, as in Python 3.x all strings are Unicode by default.
f.write(filetext.encode('utf-8'))
Another alternative could be instead of do that convert all string data in the file to utf-8, after read it, that's it's what i recommended. Then work with utf-8 data instead, but whatever, and you can save the data finally without convert it to utf-8 (as you do that before).

lestcape:
For another hands cinnamon settings file it's the first that read the settings and it will ensure that data will be wrote to the working settings file in utf-8. I patch this to be in that way a lot time ago:
https://github.com/linuxmint/Cinnamon/pull/2855/commits

Then in the cinnamon, the settings it's reading as utf-8 https://github.com/linuxmint/Cinnamon/blob/master/js/ui/settings.js#L593
And will be write as it's https://github.com/linuxmint/Cinnamon/blob/master/js/ui/settings.js#L705 streamtostream, so you can be sure that what you read it's utf-8

Odyseus:
@everyone
I recovered momentarily my internet connection. I managed to make the commit with the changes that I mentioned in my previous message. Let me know if that is acceptable and if I should do a *squash*.

lestcape:
@Odyseus: a call to save the file:
https://github.com/Odyseus/Cinnamon/blob/04064186b024fcab1a173f750f792ab6e9081394/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py#L351

I thing the file it's ensure to be encode in utf-8:
https://github.com/linuxmint/Cinnamon/blob/77e246f45c24ea6defd2cfd48f5db536eabcbcf5/files/usr/share/cinnamon/cinnamon-settings/bin/JsonSettingsWidgets.py#L183

Maybe for some magical function it will encode...

Yes, I'm sure that **js/ui/settings.js** file with the **_saveToFile** function encode the file in utf-8. I was who do that in the origin.

lestcape:
Before was not necessary to encode the file as utf-8 in this function because I translate the file to utf-8 whatever i read it, now this not occurs, while the json it's reading, so i thing this need to be ensure in any place.

lestcape:
See this interesting discussion about ensure to dump a json as utf-8 in **Python 2**
http://stackoverflow.com/questions/18337407/saving-utf-8-texts-in-json-dumps-as-utf8-not-as-u-escape-sequence

lestcape:
@Odyseus:
It occurs to me now to leave doubts, just create a settings file without translation, put there non ascii characters, restart cinnamon and open xletsettings and modify any relate value and see if something goo wrong... If something go wrong, there are a missing ensure to save utf-8 characters in some place.

Odyseus:
Hello, @lestcape.

Lets focus in one problem at the time. Otherwise, we will never get this thing done.

This pull request is for adding the missing translation mechanism. And the translation mechanism have to read files that already are UTF-8 encoded and it doesn't have to deal with saving any data. And more importantly, people more capable than me should take care of files encodings and such. With my very limited knowledge, I would be wasting everybody's time. Just dealing with the differences between Python 2 and Python 3 almost made me give up on this pull request.

### So, lets focus on the translation mechanism.

Frankly, what I'm more worried about is the the place where the **translate** function is mainly used. Because I added a loop inside a loop that contains another loop. And all this occurs inside another loop inside a try/finally block.

Another note. Inside the **translate** function I added the **result.decode("utf-8")** just to avoid a Unicode warning triggered by the **if result != string:** comparison.

By the way, I'm doing my tests with a .po file that contains mixed translations (Spanish, Russian, Chinese, Korean and Hebrew words). And everything is displayed correctly so far.

captura de pantalla de 2016-11-12 02-50-14

lestcape:
@Odyseus i'm focus on not get this problem back again: https://github.com/linuxmint/Cinnamon/issues/2245

Odyseus:
@everyone
I *squashed* the commits of this pull request and updated the first comment.

PR #5942, opened 2016-11-12
[applet.js] Re added ability to display a separator on applets context menus.

Odyseus:
In Cinnamon version 3.0.7, when new menu items are added to the context menu of an applet/desklet, a separator automatically appears, dividing the default menu items (**About**, **Configure...** and **Remove...**) from the just added menu items. In Cinnamon version 3.2.x, the same behavior was observed until [this commit](https://git.io/v6rI7) removed it.

This commit brings back that behavior to the context menu for applets

### DIFF
--- a/js/ui/applet.js
+++ b/js/ui/applet.js
@@ -492,6 +492,10 @@ Applet.prototype = {
             this.context_menu_separator = new PopupMenu.PopupSeparatorMenuItem();
         }
 
+        if (this._applet_context_menu._getMenuItems().length > 0) {
+            this._applet_context_menu.addMenuItem(this.context_menu_separator);
+        }
+
         if (items.indexOf(this.context_menu_item_about) == -1) {
             this._applet_context_menu.addMenuItem(this.context_menu_item_about);
         }


brownsr:
I can't see any change.  Is there a bug that describes the problem you are trying to solve, or some other case where you see an issue ? How do I reproduce a problem case that this solves ?

Odyseus:
Hello, @brownsr.
Sorry for not being specific. I was having internet connection problems and didn't have the opportunity to update my initial comment.

The problem is simple. The code that I added back is responsible for showing the separator inside the context menu for applets when one adds to the context menu some custom menu items.

In a nut shell, the context looks like this...
context-not-ok

When it should look like this...
context-ok

NikoKrause:
Sorry, I don't get it either. The contextMenu of which applet is this?

JosephMcc:
I would like to see the normal context items put into a preferences submenu when an applet adds additional menu items to it. Something like was done here for the panel launchers applet: https://github.com/linuxmint/Cinnamon/commit/2271b0c73a5aedde667cd3b236f5411eaba2a924

collinss:
@JosephMcc I disagree. Often there's only one or two additional options there, and that seems a little bit clunky to me. For example, in the sound applet there is only mute input and output, plus the normal options. It doesn't look too long, and it's not hard to find what you're looking for. Putting those extra things in a submenu would make it slightly harder to access, and would make it less consistent. Plus, that would be a fairly significant change when the new version of Cinnamon has already been tagged.

Odyseus:
@NikoKrause.
The applet is irrelevant. I used and image of a context menu of an applet with custom items just to show the difference with the separator added and without it. I updated the first comment and the commit message with detailed information. Let me know if you understand it now.

@JosephMcc
I agree with **collinss** view on this one; to which I add. Context menus have to have easy and fast access at all times. In a context menu with sub-menus built with any toolkit, one can access those sub-menus quickly by hovering over them. The sub-menus built with Clutter(?) are, to cut it short, the epitome of clumsiness. Not just because it requires to click on them to open them, but also because they are constantly resizing the width of a menu. And if the menu that a sub-menu has to display contains a scrollbar, you will see the clumsiness I talked about in its maximum expression.

**About the Launchers applet:** In this applet a sub-menu for the default actions (About, Configure, etc.) in its context menu would make sense. But a sub-menu for actions that will be performed with the launchers, it's just annoying. When one right click on a launcher, 99% of the time it's because one wants to perform an action related to the launcher. The sub-menu is there just to add an extra click. In my opinion, that first sub-menu should be removed, or at least be open by default.

brownsr:
@Odyseus I won't get into a discussion as to whether sub-menus or separators are most appropriate.  I just wanted to test this out.  I thought I might be able to track down the thing that was actually changing from seeing your images (thanks for posting those), but I can't see how to reproduce these.  Perhaps it's an applet I don't have, or a route through the system I am not familiar with, but I simply can't reproduce it.  Could you possibly post back a very explicit method to reproduce the issue you see ?

Odyseus:
@brownsr: Sorry, I can't explain it better than I already did. The detailed description that I added to the first message and the code added by this commit speak by themselves.

@JosephMcc: Forget what I said about the launchers applet. Clearly, I don't use it. Because it makes perfect sense both sub-menus right how they are. Sorry about the confusion.

brownsr:
@Odyseus Sorry, our posts crossed.  I've put my testing results in the comment above.

Odyseus:
@brownsr: Very sorry for not pointing you to an specific applet to test this out. I almost do not use the default ones, and the default applets that I do use doesn't add extra items to their context menu. Plus, I didn't thought a good idea to linking you to a third party applet.

About the *separator vs sub-menu "debate"*. I didn't meant to start a debate. I just gave my point of view.

JosephMcc:
@brownsr Remember that some themes will use a transparent separator so you won't see it either way in that case.

JosephMcc:
> I agree with collinss view on this one; to which I add. Context menus have to have easy and fast access at all times.

That argument doesn't make sense in this case. The items below the separator are little used items. In most cases you barely touch them once the applet is set the way you want. The items specific to the applet are the ones that you will nearly always want.

In any case, I won't be the one deciding :)

PR #5943, opened 2016-11-12
[workspace-switcher] Fix buttons' size

germanfr:
Workspace applet buttons didn't get proper height (or width in vertical mode), like any other button, if the auto-scale font setting was not enabled.

Here is a screenshot of the problem:
screenshot from 2016-11-13 20-49-51 screenshot from 2016-11-15 21-07-48


brownsr:
Tests OK, no issues I can see. :)

Issue #5934, opened 2016-11-11
Main menu background no longer themable [3.2.1]?


petrucci4prez:
Using Cinnamon 3.2.1 on Arch. It seems it is no longer possible to separately theme the main menu applet's background using .menu-background; instead it takes the theme of .menu like all the other applet popup menus. This seems like a bug as the .menu-background style class is still in menu@cinnamon.org/applet.js, but for some reason is being ignored. This could break many 3rd party themes that attempt to theme the main menu differently than the applet popup menus.

Downgrading to cinnamon 3.0.x resolves the problem. Looking Glass inspector on 3.2.1 shows "StBin.menu bottom" instead of "StBin.popup-menu-boxpointer popup-menu menu-background" on 3.0.x. Perhaps .menu-background isn't being properly applied now that .popup-menu-boxpointer is gone in 3.2.x?

leigh123linux:
3rd party themes will need to adapt or die.

NikoKrause:
@leigh123linux 
That's not helping him and he's not saying that he doesn't want to adapt his theme.
He's asking if it is possible to theme the menu-applet differently from the other applets.
It's needed for example in the CinnXP theme, which by the way didn't used boxpointers before.
So that's a legitimate question.

JosephMcc:
I have a feeling I know what's happening here. I'll look into it over the weekend and see if I can make sure there is a way in place for you to do what you want. Can you show some sort of screenshot of what your trying to do?

lestcape:
I agree with @NikoKrause and also with @leigh123linux both things are true!!!

louisgv:
Just some screenshot on my issue.
What I want it to look like:
![image](https://cloud.githubusercontent.com/assets/6723574/20290328/6560c71e-aaad-11e6-9bb0-db6b7d6f0f03.png)

What it looks like after update:
![image](https://cloud.githubusercontent.com/assets/6723574/20290400/b07b6696-aaad-11e6-9a29-bca1bc78ef77.png)

JosephMcc:
There is already a PR waiting to fix this.

feren:
I hope it gets merged... I really miss that feature, and even that could seriously affect Cinnamon's reputation

feren:
@JosephMcc Erm, the picture of the vertical panels seems misleading, given you got rid of the box-pointers, and this image was made before the removal of those:
https://www.linuxmint.com/rel_serena_cinnamon_whatsnew.php

PR #5947, opened 2016-11-13
menu-applet: Make the 'menu-background' style class work again

JosephMcc:
The new popupMenu implementation allows theming the menus based on orientation.
This overwrites the menu-background style class added by the menu applet. To
get around this connect to the actors 'style-changed' signal so that anytime
popupMenu.js changes the style class we ensure the applets custom style class
gets re-added.

Fixes https://github.com/linuxmint/Cinnamon/issues/5934

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1150,7 +1150,7 @@ MyApplet.prototype = {
         this.settings.bind("activate-on-hover", "activateOnHover", this._updateActivateOnHover);
         this._updateActivateOnHover();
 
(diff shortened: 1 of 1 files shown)


collinss:
Ah man. You beat me to it. I actually wrote a fix for it this morning, but my internet has been down all day and I haven't been able to push. I did it a little different though. Instead of a signal which is intercepted by the menu, I added a new function to the popup menu which allows an additional (custom) style class to be defined.

JosephMcc:
@collinss if you have a better solution we can go with that. I have no issue with closing this one. It's just the way I thought of doing it. Doing it from popupMenu might be better.

collinss:
Ok, I pushed it.

https://github.com/linuxmint/Cinnamon/pull/5948

JosephMcc:
I like that better.

PR #5948, opened 2016-11-13
Restore ability to theme the menu applet menu separately from other m…

### LINKED ISSUE #5934: Main menu background no longer themable [3.2.1]?

collinss:
…enus

Fixes #5934

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1150,7 +1150,7 @@ MyApplet.prototype = {
         this.settings.bind("activate-on-hover", "activateOnHover", this._updateActivateOnHover);
         this._updateActivateOnHover();
 
-        this.menu.actor.add_style_class_name('menu-background');
+        this.menu.setCustomStyleClass('menu-background');
         this.menu.connect('open-state-changed', Lang.bind(this, this._onOpenStateChanged));
 
         this.settings.bind("menu-icon-custom", "menuIconCustom", this._updateIconAndLabel);
--- a/js/ui/popupMenu.js
+++ b/js/ui/popupMenu.js
@@ -2175,7 +2175,22 @@ PopupMenu.prototype = {
         this._orientation = orientation;
     },
 
-    /*
+    /**
+     * setCustomStyleClass:
+     * @className (string): the custom class name to add
+     *
+     * Adds a custom class name to the menu which allows it to be styled separately from other menus.
+     */
+    setCustomStyleClass: function(className) {
+        this.customStyleClass = className;
+        if (this.actor.get_style_class_name()) {
+            this.actor.set_style_class_name(this.actor.get_style_class_name() + "" + className)
+        } else {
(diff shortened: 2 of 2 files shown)


files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -1150,7 +1150,7 @@ MyApplet.prototype = {
         this.settings.bind("activate-on-hover", "activateOnHover", this._updateActivateOnHover);
         this._updateActivateOnHover();
 
-        this.menu.actor.add_style_class_name('menu-background');
+        this.menu.setCustomStyleClass('menu-background');
```
lots0logs: @collinss It'd be nice if all applets added a predictable custom style class to their menus. It won't happen overnight but I think this PR is a good place to start. What do you think?
```javascript
  this.menu.setCustomStyleClass('menu-background menu-for-applet--menu');
```

Then eventually...
```javascript
 /* Window List Applet */
  this.rightClickMenu.setCustomStyleClass('menu-for-applet--windowlist');
```
```javascript
 /* Sound Applet */
  this.menu.setCustomStyleClass('menu-for-applet--sound');
```

lots0logs: @collinss 
> In fact, we want all the other menus to be uniform in appearance. 

Sorry that doesn't make much sense to me. I mean why even have themes in the first place if you don't allow them to use whatever styles their designers want to use? Consistency is important, all of the official Cinnamon/Linux Mint themes should definitely be consistent. That can easily be achieved while at the same time allowing unofficial theme authors to be as creative as they like.  

The bottom line is that implementing what I described costs the project absolutely nothing thanks to the work you already did on this PR. That extra style class won't change anything for users of themes that don't utilize that class.

zagortenay333: Don't applets already have the ability to do this?

The weather applet does (because I requested it) and some of your applets @collinss do too (also because I opened a request.)

In many applets one can utilize this for some really cool effects.

Odyseus:
Hello, everybody.

@collinss: there is another instance of `this.menu.actor.add_style_class_name('menu-background');` inside the **on_orientation_changed** method. Shouldn't the `setCustomStyleClass` method be used in there too? I'm just pointing that out because I'm not sure if not using `setCustomStyleClass` would override all other classes the menu could have.

PR #5951, opened 2016-11-13
[applet.js] Added a mechanism to highlight applets with open menus.

Odyseus:
Since the removal of box pointers (menu arrows) from Cinnamon's UI, there isn't a visual guide that inform us to which applet the currently opened menu belongs to.

This commit adds a mechanism that will set or unset the "checked" pseudo class to applets with menus. When the main menu of an applet (or its context menu) is open, the "checked" pseudo class will be applied to the applet. When either of the previously mentioned menus are closed, the "checked" pseudo class will be removed.

Only *"standard"* applets are affected. *"Special"* applets like **Window list**, **Panel launchers** and **Workspace switcher** aren't affected and might need to be *"dealt with"* individually. (Read **Technicalities** for more details)

Cinnamon themes will need to be modified to style the .applet-box:checked selector.

This is an animation with the changes made by this commit in action.

recording

In this case, I have chosen to modify a Mint-X based theme to style the **.applet-box:checked** selector like this:

```css
.applet-box:checked,
.applet-box:hover {
    border-image: url("panel-assets/hover-1.png") 6;
}
```

## Technicalities

- **AppletContextMenu prototype changes:** I chose to pass **launcher.actor** as an argument because I don't know if **this.sourceActor** is always defined(?).
- **AppletPopupMenu prototype changes:**
    - I chose to pass **launcher** as an argument to take advantage of the **launcher instanceof Applet** condition. I did this so I can set/unset the "checked" pseudo class just to *"standard"* applets and ignore *"special"* applets like **Window list** and **Panel launchers**. I imagined that these *"special"* applets can have a custom method defined in them that will highlight the element to which the currently opened menu belongs to. Plus, I didn't think that the "checked" pseudo class would be the correct class to use on those elements. That is something that someone more capable than me would have to decide. Sadly, this also leaves "un-highlighted" applets like the **Workspace switcher**. **I will need feedback on how to solve this problem.**
    - I added the **if (!sourceActor._applet_context_menu.isOpen)** condition to the **_onOpenStateChanged** method because, without this check, if one opens the applet context menu while the main menu of an applet is also open, the main menu will remove the "checked" class **after** the context menu is open and it has already set the class. I didn't add such a check to the **_onOpenStateChanged** method inside the **AppletContextMenu** prototype for two reasons:
        1. It seems that the context menu always finish closing **before** the applet's main menu finish its opening sequence.
        2. I don't know if there is an standard method to check if an applet has a menu so I can check if that menu is open. **I would like to know this, so I can add such a check to make the process of setting/un-setting the "checked" class more reliable.**

### DIFF
--- a/js/ui/applet.js
+++ b/js/ui/applet.js
@@ -27,7 +27,7 @@ const AllowedLayout = {  // the panel layout that an applet is suitable for
     VERTICAL: 'vertical',
     HORIZONTAL: 'horizontal',
     BOTH: 'both'
-}
+};
 
 /**
  * #MenuItem
@@ -42,9 +42,9 @@ function MenuItem(label, icon, callback) {
 /**
  * #AppletContextMenu
  * @short_description: Applet right-click menu
- * 
+ *
  * A context menu (right-click menu) to be used by an applet
- * 
+ *
  * Inherits: PopupMenu.PopupMenu
  */
 function AppletContextMenu(launcher, orientation) {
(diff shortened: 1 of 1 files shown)


JosephMcc:
You might be better off using the :checked pseudo class. That has been the standard in gtk themes for some time now. It might also help avoid conflicts with existing applets that use :active for other purposes, such as the window list.

NikoKrause:
@Odyseus 
Do you think something like that could fix this issue https://github.com/linuxmint/Cinnamon/issues/5953 ?

Odyseus:
@NikoKrause
With the current state of this commit no, it will not fix that issue that you linked. I purposely avoided all *"special"* applets and focused on *"standard"* applets. Plus, the pseudo class used in there is "focus", no the "active" class. I just did some tests, and the behavior described in that issue (and the issue linked in that issue) happens in Cinnamon 2.8.6, 3.0.7 and 3.2. The "culprit" seems to be the **_onButtonPress** function inside the **AppMenuButton** prototype (Window list applet). Something like the following seems to fix the issue without negative repercussions.
```javascript
_onButtonPress: function(actor, event) {
    this._tooltip.hide();
    if (!this.alert && event.get_button() === 3) {
        this.rightClickMenu.mouseEvent = event;
        this.rightClickMenu.toggle();
    }
    if (this._hasFocus())
        this.actor.add_style_pseudo_class('focus');
}
```

@JosephMcc
I was *"fixated"* with the "active" pseudo class because I saw that class being used everywhere. But it would make much more sense to use the "checked" class, not just because *has been the standard in gtk themes for some time now*, but because it will also avoid conflicts with applets that toggle the "active" class (like the menu applet). Plus, I never saw any applet using the "checked" class (I have installed all the applets from Spices that doesn't require root privileges, so I have installed more than 100 applets).

Odyseus:
@everyone
For people that already read the initial post and commit message, this are the changes that I made.

- I updated the title to be more descriptive and less cryptic.
- Changed all instances of "active" to "checked". Now this commit sets/unsets the "checked" pseudo class.
- Added the third paragraph in the first message to briefly describe which type of applets are affected.
- I cleaned the code of trailing white spaces and added some missing semicolons.

PR #5957, opened 2016-11-14
Change popup-sub-menu style class to popup-submenu for consistency

### LINKED ISSUE #5956: Popup Menu :: Submenus :: Style Class Names Are Inconsistent
While working on documenting the structure of the menus (as it relates to theme development) I noticed that both `submenu` and `sub-menu` are used in style class names.  PR incoming....

lots0logs:
Fixes #5956
Requires: linuxmint/cinnamon-themes#45
Requires: linuxmint/mint-y-theme#49

### DIFF
--- a/js/ui/popupMenu.js
+++ b/js/ui/popupMenu.js
@@ -2610,7 +2610,7 @@ PopupSubMenu.prototype = {
 
         this._arrow = sourceArrow;
 
-        this.actor = new St.ScrollView({ style_class: 'popup-sub-menu',
+        this.actor = new St.ScrollView({ style_class: 'popup-submenu',
                                          hscrollbar_policy: Gtk.PolicyType.NEVER,
                                          vscrollbar_policy: Gtk.PolicyType.NEVER });
 


NikoKrause:
I'm not very familiar with the code, but wouldn't this commit break the themes?
What I mean, shouldn't this PR also contain this change for the themes (also creating PR for the other themes as mint-y?)

If I'm wrong, please ignore my comment.

lots0logs:
@NikoKrause Indeed, you are correct. 

homer-simpson-doh

JosephMcc:
No disrespect, but I don't think we should do this. Yes the name might be a bit better but we aren't normally in the habit of breaking things without a good reason. I'm not sure this would really qualify.

lots0logs:
@JosephMcc I disagree. Consistency is a damn good reason IMO. Also nothing will be broken since 3.2 hasn't been officially released yet.

clefebvre:
Many artists already updated their themes for 3.2.

lots0logs:
@clefebvre This is the kind of thing that you end up being stuck with in the name of not breaking things. You have't even officially released 3.2 yet so why not avoid being stuck in this case? Seems silly to me. IJS!

JosephMcc:
@lots0logs I don't totally disagree with you here. There is a part of me that would like to completely revamp a lot of the style classes used in Cinnamon. Partly to allow more consistency but also to allow more generic theming of certain items for reuse by third party xlet devs. I just don't think for this one small thing it's worth it and it's not the approach we normally take.

PR #5962, opened 2016-11-14
[sound-applet] remove show percentage label for slider option

NikoKrause:
Remove the option to show the percentage value besides slider, which was introduced in https://github.com/linuxmint/Cinnamon/pull/5854

If implementing something like that, it should be made consistent and globaly. Means the other applets, which use slider, e.g. brightness slider, should have this option too. It's not very clever to add this option individually for each applet. If it's done, it should be one switch, which turns on/off this option for every applet, and not having many switches for each applet on it's own.
Also this option has no significant additional benefit. A siginificant additional may (or may not?!) be, if the sound-applet gets the feature to turn the volume to more than 100%. So considering this maybe in the future and removing it for now, because of the first said reasons.


clefebvre:
@NikoKrause can you rebase it?

NikoKrause:
@clefebvre rebased

PR #5966, opened 2016-11-15
cinnamon-json-makepot: Added support to scan for JS files inside sub-dirs.

Odyseus:
Currently the `cinnamon-json-makepot` command, when run with the `--js` argument, it scans only the JavaScript files that are in the root folder of an xlet. In an xlet with "multiversion" enabled, all JavaScript files inside sub-directories will be ignored and the resultant .pot file will be incomplete.

This commit adds the ability to scan all JavaScript files inside an xlet folder by saving a list with all JavaScript files found into a temporary file. Then that file is used to pass the *--files-from* argument to `xgettext`.

## Technicalities

- **settings-schema.json** files found inside sub-folders are already *taken care of*. The *problem* is only with JS files inside sub-folders.
- ~~I'm *worried about* the *universality* of the `find` command and the **/tmp** folder. I don't know if that command and that folder are *guaranteed* to be fount in all Linux distributions~~. Still using the `find` command, but switched to use the **tempfile** Python module.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-json-makepot/cinnamon-json-makepot.py
+++ b/files/usr/share/cinnamon/cinnamon-json-makepot/cinnamon-json-makepot.py
@@ -126,7 +126,12 @@ class Main:
                 quit()
             print " "
             print "Running xgettext on JavaScript files..."
-            os.system("xgettext --language=C --keyword=_ --output=%s *.js" % (self.potname))
+
+            tmp = tempfile.NamedTemporaryFile(prefix="cinnamon-json-makepot-")
+            try:
+                os.system('find . -iname "*.js" > %s' % tmp.name)
+            finally:
+                os.system("xgettext --language=C --keyword=_ --output=%s --files-from=%s" % (self.potname, tmp.name))
 
         self.current_parent_dir = ""
 


lestcape:
Hi @Odyseus i don' t use multi-version, but to me this is a little confused, create a pot file where are joined several applet versions...

That's probably because i think, will added sentences to translate from the old versions that are not available in the new version. This thing can happen?

Odyseus:
Hello, @lestcape.

> That's probably because i think, will added sentences to translate from the old versions that are not available in the new version. This thing can happen?

Yes, that could happen. And I don't see a problem with that. The new strings still need to be translated and need to be added anyway. And the old strings also need to be translated to be used by the old versions of an applet. After all, having several **settings-schema.json** from several version of an applet also could have new strings that an old applet version may not need, and they are scanned and added to the .pot file by the `cinnamon-json-makepot` command right now.

lestcape:
Ok, probably what it's confused to me then it's the multi-version and not your pull... Anyway, i think if we want to generate one pot per versions, we can move others version out, generate the pot only for the specific version of the applet and then move again the others where they belong. In any case, it's more general your proposition and it's how the thing need to happens, just only the multi-version confuse me.

clefebvre:
You should use the python tempfile module for the tmp file path.

clefebvre:
Check the mintupdate source code for examples of tempfile.NamedTemporaryFile().

Odyseus:
@everyone
Updated commit to use the **tempfile** Python module. Updated all messages to reflect said change.

Hello, @clefebvre.
Thanks for pointing me in the right direction. That helped me a lot and it was a really easy change.

NikoKrause:
I have a `po` folder in the `2.8` folder.
 I used `cinnamon-json-settings -i` to install the localization, but got any error "metadata missing". Or something like that. Does this commit fix this issue as well?

Odyseus:
Hello, @NikoKrause.

The **po** folder should be at the root of the xlet directory and contain the translations for all the xlets versions (with multiversion enabled). The command to install is `cinnamon-json-makepot -i ./po/*`.

This pull request has nothing to do with your problem because it only affects the .pot creation, not the .po installations.

NikoKrause:
Thanks. I'll try to use your udated file. I haven't found it, before I created the translation template.

PR #5974, opened 2016-11-16
[3.4] window-list: Rearrange popup menu items

### LINKED ISSUE #5969: Window List menu - usability regression with top panel
I'm in the habit of closing windows by right clicking on the panel window list to open the menu and then clicking `close`. Under previous versions of Cinnamon this required the minimum of mouse mileage. With the panel at the bottom, close was the bottom (first) item on the menu, and with the panel at the top it was the top (still first) item. I found this particularly useful for closing unfocused windows quickly, but now the extra mouse mileage makes this much more fiddly.

With Cinnamon 3.2.1+betsy and the panel at the bottom `preferences` is the first item in the list and `close` is second.
With the panel at the top `preferences` is the last item in the list and `close` is second last.

This is both inconsistent with previous versions, and interferes with my workflow. I know it seems minor, but I'm finding it *really* annoying.. Please could the order of the menu items be either reverted or made customisable?

Old behaviour:
cinnamon_panel_window_list_menu_old

New behaviour:
cinnamon_panel_window_list_menu_new

germanfr:
Window actions should be the nearest to where the cursor is.
Pushed preferences item to the top and reverted order in the top panel.
Adding the elements at the top reverts the order while keeping the same
code structure.
Fixes #5969

**EDIT**: Re-commited changes. Only moved preferences to the top.

### DIFF
--- a/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
@@ -760,6 +760,26 @@ AppMenuButtonRightClickMenu.prototype = {
         let mw = this.metaWindow;
         let item;
         let length;
+        
+        // Applet preferences
+        let subMenu = new PopupMenu.PopupSubMenuMenuItem(_("Preferences"));
+        this.addMenuItem(subMenu);
+
+        item = new PopupMenu.PopupIconMenuItem(_("About..."), "dialog-question", St.IconType.SYMBOLIC);
+        item.connect('activate', Lang.bind(this._launcher._applet, this._launcher._applet.openAbout));
+        subMenu.menu.addMenuItem(item);
+
+        item = new PopupMenu.PopupIconMenuItem(_("Configure..."), "system-run", St.IconType.SYMBOLIC);
+        item.connect('activate', Lang.bind(this._launcher._applet, this._launcher._applet.configureApplet));
+        subMenu.menu.addMenuItem(item);
+
+        item = new PopupMenu.PopupIconMenuItem(_("Remove 'Window list'"), "edit-delete", St.IconType.SYMBOLIC);
+        item.connect('activate', Lang.bind(this, function() {
+            AppletManager._removeAppletFromPanel(this._launcher._applet._uuid, this._launcher._applet.instance_id);
+        }));
+        subMenu.menu.addMenuItem(item);
+        
+        this.addMenuItem(new PopupMenu.PopupSeparatorMenuItem());
 
         // Move to monitor
         if ((length = Main.layoutManager.monitors.length) == 2) {
@@ -884,25 +904,6 @@ AppMenuButtonRightClickMenu.prototype = {
             mw.delete(global.get_current_time());
         });
         this.addMenuItem(item);
-
-        this.addMenuItem(new PopupMenu.PopupSeparatorMenuItem());
-
-        let subMenu = new PopupMenu.PopupSubMenuMenuItem(_("Preferences"));
-        this.addMenuItem(subMenu);
-
-        item = new PopupMenu.PopupIconMenuItem(_("About..."), "dialog-question", St.IconType.SYMBOLIC);
-        item.connect('activate', Lang.bind(this._launcher._applet, this._launcher._applet.openAbout));
-        subMenu.menu.addMenuItem(item);
-
-        item = new PopupMenu.PopupIconMenuItem(_("Configure..."), "system-run", St.IconType.SYMBOLIC);
-        item.connect('activate', Lang.bind(this._launcher._applet, this._launcher._applet.configureApplet));
-        subMenu.menu.addMenuItem(item);
-
-        item = new PopupMenu.PopupIconMenuItem(_("Remove 'Window list'"), "edit-delete", St.IconType.SYMBOLIC);
-        item.connect('activate', Lang.bind(this, function() {
-            AppletManager._removeAppletFromPanel(this._launcher._applet._uuid, this._launcher._applet.instance_id);
-        }));
-        subMenu.menu.addMenuItem(item);
     },
 
     _onToggled: function(actor, isOpening){


JosephMcc:
I don't feel like you should be rearranging items inside a menu when the position changes. The menu should always have the same order. It's just confusing otherwise.

germanfr:
One does not keep changing from top to bottom panels so often. But the one who uses the top panel wants to have the "close" near and the one who uses the bottom panel wants it either. Both can't be satisfied without reverting the order. They won't get confused because they will use one or another. Nobody has two panels with the window list in both. Even if someone had both, its not even confusing. It's the expectable behaviour.

lestcape:
I agree with @JosephMcc.

JosephMcc:
> Nobody has two panels with the window list in both

Some people use more than one monitor. The window list is designed to work well with a window list per monitor.

> Even if someone had both, its not even confusing. It's the expectable behaviour for this concrete case.

No other menu that I can think of changes the order of it's items like this.

germanfr:
> No other menu that I can think of changes the order of it's items like this.

This is a special case where the order is important. It has been like that in Cinnamon before. It doesn't change the order randomly, it's a logical and a practical change.

I personally find quite annoying to have things between the "click" and close option. It's the most common action.

Alternatively, it could be a toggle in settings like "swap order". The user chooses once, and it would "consistent" through all panels while keeping top panel users happy.

lots0logs:
> This is a special case where the order is important. It has been like that in Cinnamon before. It doesn't change the order randomly, it's a logical and a practical change.

I agree. These menus are a special case. When they open, you read the items starting with the one closest to your mouse cursor and you go up/down depending on whether its a top or bottom panel. The behavior implemented in this PR was actually the behavior prior to 3.2. Let's try not to confuse this PR with changing the order of the buttons in the list. I know there have been a lot of requests for that but this PR is not that. The order of the buttons does not change. Only their position relative to the panel is changed in order to restore what was the previous (and, IMO, the most desirable behavior)

NikoKrause:
Waaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaay too much codelines! <- The word "Way" here has 69 "a". That's the number of codelines this PR has. ;)

Actually the code to revert the order of the contextMenu items on the Top panel is a **oneliner**. The line to add is:
```js
 _populateMenu: function() {
+        this.box.pack_start = this.orientation == St.Side.TOP;
```
Digging in the history shows the "fix" has been made here: https://github.com/linuxmint/Cinnamon/commit/763a9cfb2518b6b56719254d744cbeb7238b079b

I think there is a saying: "When something has been wrong for too long, it becomes right." (Or something like that.)
I also think this saying is wrong in many cases, but in this context(Menu <- pun intended) it has it's right, in my opinion.

It's one of the most (if not THE most) used applets. And "fixing this error" after all this time (not sure though how long it has been this way) may feel like the GWX "fix" to some users.

I'm personally don't really care about it. I'm using the windows-list at the bottom. Just wanted to say, that the code is waaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaay too complicated. :)

germanfr:
@NikoKrause thanks, that's a better solution. I didn't know that property. Also notice that one item that was below "close" was shifted to the top for both.

collinss:
There may have been a time where rearranging the menus based on orientation was OK to do, but we now have multi-monitor and vertical panel support. There is just no way we can support all possible configurations and still have it make sense unless we make them all the same, regardless of orientation.

lestcape:
There are now in my opinion a highly desktop with a lot of functionalities. So, the desktop can be configured in many ways and can follows different users expectations. This is in my opinion an amassing things. More than amassing fantastic!!!

The controversy it's them a consequence of that. I think that right now, more than before, cinnamon will need to decide with conscience any step about add another configuration option or create more particularity cases, or this will be the second version of KDE.

Third-party extensions are a powerful way to satisfy users, please considered it first before try to create particular cases.

In this direction, I will say again that for me it's more and more important over time, provide a general way to access of all popup menu manager api. This particular case could be resolved if we can accesses to all popup menu managers, then list all menus inside the manager and change the pack start of the menu that we want to change in an extension to do that.

Right now there are not a good way to create an extension for a thing like that.

lestcape:
If some one, want to create an extension to change all menu orientation or for return the boxpointer back, there are a bad way to do that currently. This will be make an engineer of all applet, desklets and panel classes.

This is an example for get all menu managers of all applets (inside are all applets menus):
```
function getMenuManagers() {
    var instances = [];
    for(let applet_id in AppletManager.appletObj) {
        let applet = AppletManager.appletObj[applet_id];
        if(applet) {
            for(let attribute in object) {
                if(object[attribute] instanceof PopupMenu.PopupMenuManager)
                    instances.push(object[attribute]);
            }
        }
    }
    return instances;
}
```

germanfr:
I'm not going to fight more for this because it's not worth it, although I feel it right. BUT the preferences item should be pushed to the top. Would you accept that part?

collinss:
@germanfr agreed. The order definitely needs work still. The major actions (i.e close, minimize, maximize) should be on the top or bottom, not in the middle surrounded by submenus.

lots0logs:
@mtwebster @clefebvre What say you regarding [this](https://github.com/linuxmint/Cinnamon/pull/5974#issuecomment-261070496)? :wink:

germanfr:
I changed the commit. Only change is moved preferences away to top, which I think is sensible.

JosephMcc:
If we accept that these items need to be rearranged based on orientation, then it needs to be done properly. This PR only takes into account users of a window list at the bottom. By your own explanations you've now broke it for anyone using a window list with a different orientation.

@NikoKrause the change you pointed to should absolutely NOT be undone. It fixed an issue where submenus were broken when a window list was placed in a top panel.

lestcape:
I disagree with this, but if will be done, please reconsider what it's an special case. In my opinion it's not the top panel, it's the bottom panel. In any other orientation different to bottom, if we follow the same criteria, the most close position will be at top of the menu always.

germanfr:
Major actions shouldn't be in the middle because it is annoying for every orientation.
They can be at the bottom (for bottom users), they can be at the top (for top and vertical users) or they could dinamically change.

@JosephMcc this doesn't break it for top panel users. It was already broken. It just improved it a bit for bottom users, which seem to be the target.

clefebvre:
Hi,

I don't feel strongly about it.

If I had to choose, I'd prefer things to be symetrical... actions being close to the panel. In other words, yes, I think the menu content should be mirrored/reversed when the applet is on the top panel.

That said, I don't like the solution proposed here. I'd rather see this done at cinnamon's level, for all applets, not for this one in particular. If we need exceptions, then let the applet "unreverse" their content I guess.

I don't really see this as a bug for 3.2, so I'm marking it 3.4 for it not to get in the way of point releases right now.

JosephMcc:
> If I had to choose, I'd prefer things to be symetrical... actions being close to the panel. In other words, yes, I think the menu content should be mirrored/reversed when the applet is on the top panel.

That said, I don't like the solution proposed here. I'd rather see this done at cinnamon's level, for all applets, not for this one in particular. If we need exceptions, then let the applet "unreverse" their content I guess.

I personally hope the decision isn't made to do this. As someone who uses multiple monitors with different panel layouts and multiple instances of applets with different orientations, this sounds like a horrible idea. It means two instances of the same applet could have the menu items in different orders.

germanfr:
> In this case you would do what, have the search box in the main cinnamon menu at the bottom in some cases? Or would you special case that one? For some reason deciding it should behave differently?

This is why I say this is a special case, not every applet is the same. The order is not important in any applet I know, but it is in this one. I think a global setting for this would be a mess.

I completely understand your point. This should be an optional feature (if it became a feature).

That said. I remember everyone, that feature is not part of this PR anymore. What it currently does is this static change:
screenshot from 2016-11-22 19-00-32

clefebvre:
Hi,

Sorry I'm a bit late to the party and things had to be taken care in emergency. I thought this would be ok for 3.2 but feedback on the BETA release proved otherwise. This was changed with https://github.com/linuxmint/Cinnamon/commit/c62b3d62882454c60b8c43d0b34f0e98dce4915b during that time.

PR #5976, opened 2016-11-16
panel-launchers: close the menu before destroying when removing launcher

### LINKED ISSUE #5968: Cinnamon widgets unable to gain focus
After these steps cinnamon is unable to give focus to the menu search input and Cinnamon dialog buttons become unresponsive to clicks.
 1. In the menu: add program to panel.
 2. In the panel launcher: **remove launcher shortcut**.
 3. Open the menu and try to search. Open any applet's about dialog and try to close it.

This is on Cinnamon 3.2.1

itzexor:
There may be a better/more correct fix for this. I'm not sure if this was introduced with the popupMenu.js changes or if it has existed for a while unnoticed.

If we destroy the menu while it is open then a callback set by pushModal removes
the grab when it recieves the actor's destroy signal. Shortly after, we attempt to
remove it again in popupMenu.js via the menu state change signal and fail. Closing
the menu before destroying the actor avoids this issue.

Fixes #5968

### DIFF
--- a/files/usr/share/cinnamon/applets/panel-launchers@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/panel-launchers@cinnamon.org/applet.js
@@ -95,6 +95,7 @@ PanelAppLauncherMenu.prototype = {
     },
 
     _onRemoveActivate: function(item, event) {
+        this.close();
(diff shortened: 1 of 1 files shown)


NikoKrause:
>  I'm not sure if this was introduced with the popupMenu.js changes or if it has existed for a while unnoticed.

I can reproduce this on LM 18 Cinnamon 2.8.8

PR #6013, opened 2016-11-24
[WIP] 6011 add gresource support to theme stylesheets

### LINKED ISSUE #6011: Themes :: Stylesheets :: Support GResource URIs
PR Incoming...

lots0logs:
### W.I.P...
Fixes #6011


germanfr:
Is this OK? If both are NULL it's going to return true. Is that the intended behavior?
```
if (file1 == file2) {
  return TRUE;
}

if ((file1 == NULL) || (file2 == NULL)) {
   return FALSE;
}
```

lots0logs:
Yes in that context I think its fine. We're checking if two files are equal. If null is passed for both values then the values are equal so returning false would be inconsistent with the purpose of the function.

Ref: https://github.com/GNOME/gnome-shell/blob/master/src/st/st-theme.c#L100

germanfr:
Great then :+1:

JosephMcc:
I'm not sure you can use G_DECLARE_FINAL_TYPE() here. Docs say it was added in GLib 2.44 but Debian Stable, and LMDE, only contain 2.42. You would need to confirm it would be supported.

lots0logs:
Yeah, this actually snowballed into waaaaaay more work than I anticipated. I do intend to finish it, but I'll have to work on it as I have time in my schedule.

clefebvre:
Closing it for now.

Issue #5028, opened 2016-01-21
Drag and Drop (Start) Menu entries to Desktop


Mintberry1:
I wanted to suggest that you can drag and drop menu entries of programs to the Desktop (To make an additional copy of the launcher on the Desktop) and the bottom Panel bar as well to make a copy.

I know you can rightclick on the entry and choose the option but drag and drop would be nice, easy and intuitively to use i think

lestcape:
And also is possible and easy do that.

JosephMcc:
You can already drag and drop applications to the panel launchers applet. Doing it to the desktop might be a different story since you are talking about two different UI toolkits.

lestcape:
@JosephMcc he say "menu entries", but you also can add panel buttons launchers and they are clutter objects...

He also only want to create a .desktop file (this is not a different UI toolkits), is just a file this a format specification that you can create when you want and put where you want over cjs.

Inside the desktop, there are a clutter object to. So, the desktop clutter object can capture the clutter buttons drop and ask for the delegate class of the buttons.

Then you can add a render_desktop_file function inside the delegate class with a the path parameter (in that case the desktop path) for all buttons.

And now, when you capture the drop inside the desktop object, ask for the delegate class of the button, call the render_desktop_file function of the button with the desktop path parameter and then accepted the drop.

**Done**, this is just what is needed. i don't see the  two different UI toolkits.

lestcape:
What could be difficult, is the possible drag and drop way, from the desktop to the menu, but is also possible. Who Nemo is who control the desktop, you can write a dbus function inside cinnamon Dbus with a path parameter of a desktop file. When nemo detect a drag  of .desktop file he can call the Dbus function and with the path of the file as a parameter, capture the response of the Dbus function and if is true stop the drag and drop and hide the object that is dragging.

As cinnamon receive the path of the desktop file inside the dbus function, he can construct a panel launcher button on fly and add it to the desktop object again and force the drag and drop class to begin a normal drag and drop of an a clutter object. So, as it's a clutter object you can drop it inside any component where you want that accept this type of button.

lestcape:
This last techniques, also could be generic, if the dbus function receive also as a parameter the class of the object that cinnamon will construct. Again, the result clutter object can be drop in any clutter object that accept this type of object.

This have a directly application in cinnamon settings, to drag and drop applets and desklets.

anandrkris:
[...]

Issue #3865, opened 2015-01-31
drag & drop for Menu and Panel


cptX:
It would be a great enhancement to support drag & drop from desktop icons to the Menu, from Panel to the Menu and from one folder of the Menu to another without the need for Menu Editor. In general to support full exchangeability of icons between these 3 groups (desktop, Menu, Panel).

Right now, the only thing that works is drag & drop an icon from a Menu to the Panel. The opposite is not possible, the drag & drop from and to Desktop is not possible and also the drag & drop from one folder to another in the Menu is not possible.

I discovered that by right clicking on an Icon/Link in a menu a new menu appears underneath with the following items:
Add to Panel
Add to Desktop
Add to Favorites
Uninstall

By clicking these actions the above requested functionality is achieved but not with drag & drop. I think it makes clear sense to support the same capabilites with drag & drop.

lestcape:
https://github.com/linuxmint/Cinnamon/issues/5028#issuecomment-262874381

lestcape:
I see grab_notifier in **src/cinnamon-global.c**: https://github.com/linuxmint/Cinnamon/search?q=grab_notifier&type=Code&utf8=%E2%9C%93
but the windows is never mapped or shown, i don't know if this Gtk windows cloud be used to also receive the Gtk drag and drop information and allow send it to cjs. But if will be possible, cinnamon could react to a gtk drag and drop and accept drops from gtk application like nemo.

JosephMcc:
Many thanks for contributing to Cinnamon. Your suggestion was reviewed. Your job is done and we'll take it from here :)

For more information on our workflow and feature requests, read https://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/faq.html

PR #6017, opened 2016-11-25
[3.4] Support to drag and drop operation from applications to the desktop

lestcape:
This pull add the ability to drag and drop any cinnamon object to the desktop, where the actor contain a delegate class with an app attribute. So, will support to drag and drop in a general way, elements that contains a GApplication information and create automatically the corresponding .desktop file of the application. Fix: https://github.com/linuxmint/Cinnamon/issues/5028 and partially: https://github.com/linuxmint/Cinnamon/issues/3865

### DIFF
--- a/js/ui/layout.js
+++ b/js/ui/layout.js
@@ -14,6 +14,8 @@ const Tweener = imports.ui.tweener;
 const EdgeFlip = imports.ui.edgeFlip;
 const HotCorner = imports.ui.hotCorner;
 const DeskletManager = imports.ui.deskletManager;
+const DND = imports.ui.dnd;
+const FileUtils = imports.misc.fileUtils;
 
 const STARTUP_ANIMATION_TIME = 0.5;
 const KEYBOARD_ANIMATION_TIME = 0.15;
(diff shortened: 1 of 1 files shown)


brownsr:
Appears to need the resulting files to be made executable
screenshot from 2016-12-11 17-23-30
screenshot from 2016-12-11 17-23-52

Odyseus:
Hello, @brownsr.

The .desktop files are made executable by [this line](https://github.com/linuxmint/Cinnamon/pull/6017/files#diff-73d8ed048dbdd9af429eba6537e2660bR74). It's inside a **try {}catch{}** block, so you may find the error in the **glass.log** file perhaps.

brownsr:
Good idea.  The right click option within the menu to add to desktop works fine by comparison, anyway, this is what is in the log:

info t=2016-12-11T19:45:44.691Z FileUtils.changeModeGFile is not a function
trace t=2016-12-11T19:45:44.691Z 
<----------------
NemoDesktopArea.prototype.acceptDrop@/usr/share/cinnamon/js/ui/layout.js:74
_Draggable<._dragActorDropped@/usr/share/cinnamon/js/ui/dnd.js:496
wrapper@resource:///org/gnome/gjs/modules/lang.js:169
_Draggable<._onEvent@/usr/share/cinnamon/js/ui/dnd.js:187
wrapper@resource:///org/gnome/gjs/modules/lang.js:169
_getEventHandlerActor/<@/usr/share/cinnamon/js/ui/dnd.js:56
---------------->

Odyseus:
As I suspected, you may not have a fully updated Cinnamon system. The **changeModeGFile** function was added to **FileUtils** [at the same time that was added to the menu applet](https://github.com/linuxmint/Cinnamon/pull/5972/files#diff-9730f14a54367b06bdc5d0af4a9a5706R145). The updated menu applet also uses **FileUtils.changeModeGFile** to set the permission of the .desktop files copied to the Desktop. Otherwise, you should have received the exact same error when adding from the menu if you were using an up to date Cinnamon.

But this is inconsequential right now. Since this new feature was pushed to be added to Cinnamon 3.4.

brownsr:
I had just spotted this myself.  Will retest.

lestcape:
Well i think this is incomplete anyway, there are not right now a way to support drag and drop between all applications, thats include nemo, for example. This ability need to be implemented from scratch in cinnamon_global.c, and i think it's possible implement this without a problem, but i have dubs abouts what could happen later when Wayland replace the X11 compositor.

Right now, cinnamon use the overlay windows to detect a drag and drop and because of this, we can only received xdnd event inside the clutter actors, but the drag and drop occurs before of that, only we can not detect it. This is an known problem of mutter and muffin. I have a partial solution, if we check the type of the windows, whatever a windows it's create, because when a drag and drop begin the source windows push a new windows with the type DND in X. So, we can check this and detect a drag and drop more faster, and we can have access to the draggable windows that represent the drag and drop.

There are another problem and it's related with how cinnamon_plugings works. When a xdnd event are capture for the atom side the event are not passed to clutter and that inhibit the possibility of use the standard implementation of cinnamon DND to handled xdnd, because we can not received motion event inside clutter in this case.

Finally the are another problem, this is how we can pass the ATOM information of the xdnd to the cinnamon side when we capture the xdnd information. This is a list of atom properties, some of then could have a list of uris or a binary images or whatever the source windows think that it's useful... So only in the client side and with the information about how the specific source works,  we can parse the data in a good way.

But again it's possible do all that and really support a general drag and drop operation, something that it's missing actually.

lestcape:
The necessary code to do all that it's here: https://github.com/swayfreeda/blender-2.77a/tree/5969d704f44952ea8cbecba2ba4150fb4a48e6de/extern/xdnd

Just need to be adapted to the cinnamon work-flow.

Issue #5603, opened 2016-08-26
desktop switcher applet does not respect wraparound


illumilore:
With number of virtual desktops set to 4, the workspace switcher applet does not wraparound when using the mouse wheel to change desktops and stops at 1 or 4, even though wraparound is set to true and when using keyboard shortcuts to move left and right, that wrapsaround.

Cos-mo:
"does not respect" implies, that this is a bug.
I think this would be a feature request for the applet, which has its own settings (but none option for wraparound).

As this applet can be added to the panel multiple times, it needs its own settings, separate from the workspace settings.

illumilore:
It is inconsistent for each applet to have its own settings and not follow the options that the user has selected. There is no reason for wraparound preference to only apply to keyboard shortcut switches and not mousewheel switches.

mtwebster:
Yes - it can be added multiple times, but I can't possibly imagine wanting to have multiple instances that behave differently in this respect - I think it's one of those things you either want, or not, no matter what.

germanfr:
Hi. 
Is it possible to restrict the rotation to different scroll times? When you have a touchpad it is almost impossible to scroll a unit to move just one workspace. Since both options are shared I get the following:
![ezgif com-debc914266](https://cloud.githubusercontent.com/assets/10391266/20641908/d6d2f71a-b402-11e6-88dd-5a49cf748332.gif)

I don't want to restrict the workspace rotation, because is useful for me on shortcuts. But leaving it enabled results in this. 

Can anything be done? Thanks.

JosephMcc:
I just tested this on my own machine. Switching by edge scrolling isn't too difficult at all. l wonder how much of a difference different hardware makes? Not sure how much could actually done about it.

germanfr:
Two finger scrolling here. The thing is that with just a tiny move it scrolls like 3 times or so. 

The old behavior would not be the solution, because it would only work with two workspaces. 
Could we limit the speed change? Define a small period of time since last scroll event where it ignores other, so that it does not scroll too fast.

JosephMcc:
I suppose that would depend on whether it affects the behavior when using the mouse scroll wheel. Currently that works perfectly and you would need to do it without breaking that. Best bet is probably to just experiment with the idea.

germanfr:
I'm going to try it, and if it works good with both I'll submit a PR :)

germanfr:
I've done it. What do you think?
[...]

PR #6021, opened 2016-11-26
workspace-switcher: avoid fast consecutive moves

germanfr:
Consecutives switches occuring on touchpad devices due to
uncontrollable scroll units. This avoids that by giving you a small
time to react and stop scrolling.
This does not apply when you go and come back from a workspace.

Relating #5603


JosephMcc:
I just tested this and it seems to work pretty well. Doesn't add enough of a delay to effect mouse wheel scrolling. I will say it doesn't seem to affect the behavior using the touchpad much for me. I tested using a higher delay but if you do anything much higher it starts affecting scrolling with the mouse scroll whee.

PR #6024, opened 2016-11-27
Change shebang to use system python

leigh123linux:


opticyclic:
Why would you change the version to a hard coded path instead of letting the interpreter pick the version from your PATH?
http://stackoverflow.com/questions/5709616/whats-the-difference-between-these-two-python-shebangs

If you hard code it then it will break for people who don't have python in that dir.

lots0logs:
Because some systems use Python 3 as the default and some use Python 2. Thus we must be specific about which Python we want. Furthermore, every distribution includes symlinks that point to the installed version of Python 2 and 3 (`/usr/bin/python2` and `/usr/bin/python3`).

leigh123linux:
@opticyclic  This change has already been discussed and agreed by the team.

> If you hard code it then it will break for people who don't have python in that dir.

Show me a distro that doesn't use /usr/bin/ for system python as standard!

clefebvre:
Hi,

This has been discussed over and over again. There are legit pros and cons on both sides of the argument.

The rationale here is that custom python installation are usually installed to do custom work, not to affect system software.

Issue #6049, opened 2016-12-04
Vertical panels' applet order on Cinnamon 3.2 is broken when using RTL system language


eladhen:
On Cinnamon 3.2, when setting the system language to Hebrew (a Right to Left language), and switching to a vertical panel, the order of the applets is pretty broken. The Mint menu and the app launchers are set on lower half of the panel, with the launchers set under the menu.

screenshots:
![2016-12-03 21-15-45](https://cloud.githubusercontent.com/assets/12398651/20868270/413eb2f2-ba60-11e6-96c6-a7fd9cc1e037.png)

![2016-12-03 21-16-53](https://cloud.githubusercontent.com/assets/12398651/20868268/413ca96c-ba60-11e6-9180-bd759711c8d7.png)

The windows list applet appears *under* those two, pushing them up as windows appear:

screenshot:
![2016-12-03 21-25-15](https://cloud.githubusercontent.com/assets/12398651/20868269/413dcb9e-ba60-11e6-9a42-95475c871c32.png)

This was tested on Linux Mint Cinnamon 18.1 Beta and on LMDE2 Cinnamon.

JosephMcc:
Thanks for the report. As far as I know none of the devs uses Cinnamon with a RTL language and we get very little feedback on these issues.

eladhen:
I'm not a developer, but I'll gladly help if you tell me what to do. I have a testing laptop so I will gladly test whatever you want.

JosephMcc:
@eladhen Are you running 64-bit Mint 18.1? I'll post some updated debs for you test if you're willing to try them out.

eladhen:
I am running the 64bit version. How can I get these debs?

eladhen:
Yes, I'll gladly test them.

JosephMcc:
@eladhen You can grab them here: https://dl.dropboxusercontent.com/u/71834451/share/cinnamon-deb.tar.gz

Extract the folder, then in Nemo right click the cinnamon-deb folder and choose "Open in Terminal".
You can then install them by running `sudo dpkg -i *.deb`

You'll need to restart Cinnamon for the changes to take effect.

eladhen:
It seems to work. Anything you want me to test with this?

![2016-12-06 08-45-06](https://cloud.githubusercontent.com/assets/12398651/20917201/bfc3505e-bb98-11e6-9d29-07f7151e8804.png)
[...]

PR #6058, opened 2016-12-05
remove rtl code in vertical panel

brownsr:
will require testing by 3rd party


JosephMcc:
This seems to do the trick. See the discussion here: https://github.com/linuxmint/Cinnamon/issues/6049

Issue #6016, opened 2016-11-25
Launcher Properties "OK" button not working


derlucas:
I wanted to edit a Launcher Icon in one of my Panels. Right click on the Icon (say Firefox) and choose "Options" "Edit". Then Change the Name and Click on the "OK" button does nothing. My changes are not stored.

![1480059415](https://cloud.githubusercontent.com/assets/1943040/20617560/5b7c2596-b2ea-11e6-97bd-a57874584bfb.png)

Versions in use:
local/cinnamon 3.2.1-4.1
local/cinnamon-control-center 3.2.0-1
local/cinnamon-desktop 3.2.0-1
local/cinnamon-menus 3.2.0-1
local/cjs 3.2.0-2
local/cinnamon-settings-daemon 3.2.0-1
local/cinnamon-session 3.2.0-1

leigh123linux:
The issue reproduces here.

Odyseus:
Hello, everybody.

The error described in this issue is due to several typos in the [cinnamon-desktop-editor.py](https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/cinnamon-desktop-editor/cinnamon-desktop-editor.py#L478) file.

It looks like this...
```python
settings = JSonSettingsWidgets.JSonSettingsHandler(self.json_path)
```

when it should look like this...
```python
settings = JsonSettingsWidgets.JSONSettingsHandler(self.json_path)
```

After fixing those typos, pressing the OK button will close the dialog, but it will still not save the changes. Someone more familiar with the code should take a look at it.

JosephMcc:
https://github.com/linuxmint/Cinnamon/commit/fdf572b9123d5a46a2e6b551a5844bbdf7139853

PR #6067, opened 2016-12-07
cinnamon-desktop-editor: Fixed panel launcher add/edit dialogs breakage.

Odyseus:
Currently, when pressing the **Ok** button of a dialog that adds a new panel launcher or edits an existent one, the changes will not be saved and the dialog will not close. This commit fixes that.

Fixes: linuxmint/Cinnamon#6016


Odyseus:
Closed because [this commit](https://github.com/linuxmint/Cinnamon/commit/fdf572b9123d5a46a2e6b551a5844bbdf7139853) actually does fix the problem.

Issue #6059, opened 2016-12-05
Please add 'disable screen rotation' setting


dcrystalj:
i have lenovo yoga 2 pro with screen rotation accelometer feature. It became automatic based on accelometer in version 3.2 and now i cannot disable or i cannot find a way to disable it. I have tried with disabling in keyboard shortucts but does not help

leigh123linux:
Try running this as standard user

```
gsettings set org.cinnamon.settings-daemon.peripherals.touchscreen orientation-lock  true
```

dcrystalj:
woow i think it's working, thank you!!!, how did you find this settings exists? I couldn't google it :-1:

leigh123linux:
I'm familiar with the code.
Perhaps it would be worth adding a setting to settings to make it easier to disable.

dcrystalj:
nice! I just want to mention that on my yoga there is special key to disable/enable rotation which is by default same as ctrl+o. Do you have any idea how to bind it to correct thing?

leigh123linux:
I think additional coding/backporting will be required for keybinding support.

https://github.com/GNOME/gnome-settings-daemon/commit/76b548e3dd4c52455e9c2684d2e64e41b9c5cd84

Cinnamon-settings-daemon currently has none

dcrystalj:
Thank you

brianjmurrell:
Thanks for all of this.  Screen rotation on a laptop is hideous.  Especially where it rotates 90, 180, even 270 degrees on a hair trigger but absolutely refuses to come back to 0 without opening up the Displays setting widget and trying to navigate it when things are 90 degrees off.

venik212:
Sadly, the key on the side of the Yoga 2 Pro does not work, at least not under lubuntu/cinnamon 18...

dcrystalj:
well its working for me. button and everything... some times it gets broken but after reboot its fine.

specs:
```
Cinnamon 3.2.8
Ubuntu 16.10
```

venik212:
We can't all be as lucky.  However, using dconf Editor to disable screen rotation worked for me, so I am happy.

leigh123linux:
@venik212 Did you try changing the keybinding to match you switch key code?

venik212:
We must be using different cinnamons, since I see no such setting in cinnamon-settings general tab.

leigh123linux:
I use 3.2.8

[...]

PR #6074, opened 2016-12-08
add orientation lock keybinding

leigh123linux:


leigh123linux:
Should address linuxmint/Cinnamon#6059

Issue #5525, opened 2016-08-05
[Archlinux, Cinnamon 3.0.7] Startup tray issues.


kolorafa:
**When booting up Cinnamon a lot of tray icons doesn't appear.
Looks like the icon applet start to late and other apps fail to add icons.**

For example:
Skype randomly show up (if it doesn't it sill running but without icon, i need to restart skype to fix it)
Synergy always die with message "System tray unavailable. Quitting." so i need to run Synergy one more time to show and run.

Is there a way to fix this other than making huge delay on startup apps?

I have spinning disk on that notebook so it slows down loading and probably make the difference that apps starts before tray applet, but that's my educated quess.

two-dogs:
@kolorafa , i would look at the output of 'looking glass' log

kolorafa:
Sorry for the delay.

I have this problem every boot. On both of my linux arch machines.

http://kolorafa.firehost.pl/test/snapshot-2016-08-19-19-21-23.png

![](http://kolorafa.firehost.pl/test/snapshot-2016-08-19-19-21-23.png)


attaching glass.log (don't see any errors)

like you see on screenshot, the skype is running but not showing
the syncthing-gtk is running but the icon is blank
i will install synergy on this machine too later

[glass.log.txt](https://github.com/linuxmint/Cinnamon/files/427538/glass.log.txt)

kolorafa:
Just to mention, the cinnamon was installed as a second Desktop Enviroment on those machines.

Also kadu, running but no icon like skype.

mainmachine:
The apps that don't show up - are they apps that require authentication or superuser creds to run?

I had a similar problem with Owncloud client - it would run but no icon in the system tray. I had to install libgnome-keyring for it to show up, IIRC. This is on Ubuntu, not sure what the package name would be on arch.

Do you have "system tray" listed in Settings > Applets?

kolorafa:
The problem is only about the timing

If i add delay in apps startup settings, so they start later, then the icons appear without any problem.
So it looks like cinnamon traybar starts to load, other apps starts to load, and the other apps try to "access" the tray  and got "not found" because they tray isn't loaded at that time.

The apps doesn't require any additional stuff, because skype and kadu don't use keyring, and Synergy clearly say in 'Window Message' that "No tray found, exiting" because it requires the tray, it will not work in background without tray icon.

But like i said - after adding  5(seconds?) delay in the control panel to those apps, they appear normally.

[...]

PR #6080, opened 2016-12-09
Start systray applet first

leigh123linux:
Starting systray applet before the menu applet should hopefully be enough to address

https://github.com/linuxmint/Cinnamon/issues/5525

### DIFF
--- a/data/org.cinnamon.gschema.xml.in
+++ b/data/org.cinnamon.gschema.xml.in
@@ -23,7 +23,7 @@
     </key>
     
     <key name="enabled-applets" type="as">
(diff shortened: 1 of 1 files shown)


kolorafa:
does the enabled applet change every time you modify(add/delete) the applets?

kolorafa:
and if someone accidentally move/remove the applet and readd it, will it still work correctly?

(and do i need a clean profile to test it?)

lestcape:
@leigh123linux, sorry, but in my opinion this will not solve the problem, just mitigate it, not more than that.
- How this is possible?
- Do you know what really happens?
- Can you explain what you detected?
- There are a dependency between the applets? 
- You know where the dependency is?
- Can be removed the dependency?
- If not, this will really solve the problem or we need to create something to handle dependencies between extensions?

leigh123linux:
@lestcape There is no true way to solve the autostart app issue in cinnamon code, the autostart code is in cinnamon-session.
What this pull request does is start the systray applet before the menu applet and as you know menu takes ages to start.
On my system it made nearly 500ms difference (I use ssd for / and /home), on slower systems it could be much more.

lestcape:
I understand now thanks. I can not reproduced the problem anyway.

leigh123linux:
@lestcape It also makes sense to have a sane default that minimizes it.

PR #6082, opened 2016-12-10
Partial revert of 864f632a and handle premature signals from the file…

### LINKED ISSUE #5985: Reset to defaults in settings of applets doesn't take effect
E.g. you are messing around with the options in the settings of the menu-applet.
You can click on the "burger" symbol in the top right corner and then "Reset to defaults".
You have to restart Cinnamon to get to the default state. In Cinnamon 2.8.8 you didn't had to restart. So I guess it's a regression.

I think the bug is hidden somewhere here https://github.com/linuxmint/Cinnamon/commit/3478db2822d2f4c839c7d0ef7b517f0aa1d50bae
(@collinss *cough cough* ;)

collinss:
… monitor gracefully

Fixes #5985

### DIFF
--- a/js/ui/settings.js
+++ b/js/ui/settings.js
@@ -258,7 +258,7 @@ XletSettingsBase.prototype = {
         this.bindObject = bindObject;
         this.uuid = uuid;
         if (this._get_is_multi_instance_xlet(this.uuid)) this.instanceId = instanceId;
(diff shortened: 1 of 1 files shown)


lestcape:
@collinss if you have time, take a look here: https://github.com/linuxmint/Cinnamon/issues/6086

PR #6118, opened 2016-12-20
Fix for non translated applet names on Remove context menu items.

Odyseus:
Currently, the applet name of the **Remove** context menu items of third party applets  aren't translated. This is because the translated string is currently fetched from Cinnamon's domain and not the third party applet's domain. Default Cinnamon applets doesn't have this problem.

This commit uses a mechanism similar to the used by the **SpicesAboutDialog** prototype to try to fetch the correct translated string.

Also, some trailing spaces were removed and some missing semicolons were added.

## Technicalities
- I moved the call to **Gettext.bindtextdomain** from the **modalDialog.js** file to the **appletManager.js** file so the applet's domain it's binded (don't know if I express this correctly) before the first call to the **finalizeContextMenu** function. This way, the translated strings are available when it's required (when the **this.context_menu_item_remove** is created and when the **About** dialog is opened). Like always, it's up to the real experts to decide if this pull request is the correct way to fix this issue.
- I wasn't sure if this bug would be considered serious or not serious at all. That's why I didn't added the **[3.4]** tag to this pull request title.

### DIFF
--- a/js/ui/applet.js
+++ b/js/ui/applet.js
@@ -15,6 +15,7 @@ const Mainloop = imports.mainloop;
 const Flashspot = imports.ui.flashspot;
 const ModalDialog = imports.ui.modalDialog;
 const Signals = imports.signals;
+const Gettext = imports.gettext;
 
 const COLOR_ICON_HEIGHT_FACTOR = .875;  // Panel height factor for normal color icons
 const PANEL_FONT_DEFAULT_HEIGHT = 11.5; // px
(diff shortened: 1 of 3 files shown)


NikoKrause:
A little correction:
The "Remove %s" item is translated for applets.
The issue is:
The "%s" in "Remove %s" is translated for Cinnamons default items, but it is not translated, if you install third party applets (when they have a translation of their applets name in their po files).

Odyseus:
Hello, @NikoKrause.

Thanks for pointing that out. I updated the commit message and first post with more precise details based on your observations.

clefebvre:
Please rebase this PR.

Odyseus:
Hello, @clefebvre.

Rebase done.

clefebvre:
Cheers :)

PR #6139, opened 2016-12-28
Menu applet: Pull functions out of loops in _refreshPlaces, _refreshApps...

jaszhix:
and _refreshRecent.

These anonymous functions are recreated on every iteration, which comes
at a performance cost.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -2202,7 +2202,7 @@ MyApplet.prototype = {
             let devices = this._listDevices();
             let places = bookmarks.concat(devices);
 
-            var handleEnterEvent = (button) => {
+            let handleEnterEvent = (button) => {
                 this._addEnterEvent(button, () => {
                     this._clearPrevSelection(button.actor);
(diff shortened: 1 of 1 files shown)


jaszhix:
Ah, sorry about that. How does this look?

brownsr:
Any before/after measurements ?

jaszhix:
No, I'm not sure how I would do that. It may or may not be noticeable to people, but these small in efficiencies can add up to a net effect. It doesn't hurt anything to use best practices in JS.

jaszhix:
Yes, I read the style guide. Why are random blocks without brackets, though? If we are solely using ```let```, then using let in these bracket-less blocks will cause a TypeError because it is block scoped.

Cobinja:
Let me rephrase that: We usually use "let" *in newly written code*.
In this case, "let" doesn't hurt a bit.
About your example: that line was last changed in 2012, and it used "var" even before then.

jaszhix:
I don't mind using let, and my point wasn't that it *hurts* anything, I actually think its a good idea. I am saying, though, if there are multiple contributors not using brackets, its going to lead to issues towards consistency, and cause scoping issues. JS isn't Python. Maybe the Mint team can produce an eslintrc file.

PR #6162, opened 2017-01-03
keyboard: destroy the key's boxpointer

brownsr:
from upstream https://github.com/GNOME/gnome-shell/commit/9d933356e171a1dc49b61b5864826531c0f24d97

Am not convinced I have exercised this code path in testing, so would be grateful if someone who understands 'extended keys' could check, or give me some idea on how to do it.


JosephMcc:
@brownsr you can activate the extended keys on certain ones by clicking and holding. This works on the "l" and "o" keys for example.

PR #6178, opened 2017-01-06
[3.4] dnd: Don't use reparent for adding actors to the uiGroup

brownsr:
upstream commit https://github.com/GNOME/gnome-shell/commit/9d2791d9f8f6d0bb62998ca94f4372b9d0438b5d#diff-902d56799170f0dc27500667c665b683

### DIFF
--- a/js/ui/dnd.js
+++ b/js/ui/dnd.js
@@ -260,7 +260,7 @@ const _Draggable = new Lang.Class({
 
         if (this.actor._delegate && this.actor._delegate.getDragActor) {
             this._dragActor = this.actor._delegate.getDragActor();
-            this._dragActor.reparent(Main.uiGroup);
+            Main.uiGroup.add_child(this._dragActor);
             this._dragActor.raise_top();
             Cinnamon.util_set_hidden_from_pick(this._dragActor, true);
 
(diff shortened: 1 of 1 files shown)


lestcape:
@brownsr don't do that... In cinnamon, you can start a drag and drop manually, also...

If you want to use add_child first check if the actor have a parent, remove the actor from the parent if exist and then add it to another container. Or use global.reparent instead of actor.reparent https://github.com/linuxmint/Cinnamon/blob/master/js/ui/main.js#L253

You can not be sure that the actor has or not a parent.

lestcape:
https://github.com/linuxmint/Cinnamon/pull/6178/commits/461a3e74d22bb01cfe8d21c5f031d38aaa8a1fd4#diff-902d56799170f0dc27500667c665b683R263

brownsr:
thanks for the comments @lestcape  I'm currently working through upstream commits looking for relatively straightforward changes to pull down.  I picked this because the original author highlighted that he believed it was bad practice and could cause bugs.  But if it does not work for us then I will just throw this change away.  Is it appropriate to pursue a change here in your opinion ?  Out of interest how does global.reparent work, is it a plug in replacement ?  You've got drag and drop changes queued up so I imagine you know this area very well.

lestcape:
I think it's appropriate as it's deprecate, but it's not appropriate for the Cinnamon dnd, how it's implemented right now. This line it's the line that can create problems: https://github.com/linuxmint/Cinnamon/pull/6178/commits/461a3e74d22bb01cfe8d21c5f031d38aaa8a1fd4#diff-902d56799170f0dc27500667c665b683R263 because my first comment...

I know you are doing that... I use the nightly or at less i was used the nightly repo. When cinnamon crash in some update, i see the cause and what you are doing here and inside cjs. thanks...

brownsr:
Closing then ...

PR #6184, opened 2017-01-07
[WIP] workspace switcher: remove hard coded scaling

### LINKED ISSUE #6148: Workspace-switcher applet doesn't extend full width in vertical panel
Is there any particular reason that you hardcoded the width of the applet to be .9 of the width of the panel?
https://github.com/linuxmint/Cinnamon/blob/3936ac1e355e9a5f1c2edd1aae4e6da9d2855f45/files/usr/share/cinnamon/applets/workspace-switcher%40cinnamon.org/applet.js#L194

It doesn't look good if the other applets extend to the edges:
screenshot-area-2017-01-01-130849 ____ screenshot-area-2017-01-01-130912

brownsr:
fixes #6148 but will force theme changes

### DIFF
--- a/data/theme/cinnamon.css
+++ b/data/theme/cinnamon.css
@@ -1352,8 +1352,7 @@ StScrollBar StButton#vhandle:hover {
 }
 
 .workspace-button {
-    width: 20px;
-    height: 10px;
+    width: 1em;
     color: #cccccc;
     border: 1px;
     border-color: #a6a6a6;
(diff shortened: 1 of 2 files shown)


zagortenay333:
Nice :smile: thx!

JosephMcc:
I think you might actually be better off removing that if/else block that sets a size completely. That allows the theme complete control over the size instead of hard coding to the size of the panel. Then using `width` and `height` you can create the look you want.

brownsr:
Taking this away for now until I have a good chunk of time to review and think through the options.

PR #6225, opened 2017-01-26
JS API: Add Promise and setTimeout polyfills to environment.js

jaszhix:
This adds a polyfill for Promise using the MIT licensed library from
https://github.com/taylorhakes/promise-polyfill. setTimeout is also
polyfilled because the Promise polyfill needs it. Adding it to the
global context will also help with compatibility for other third party
libraries. It is a wrapper around Mainloop.timeout_add.

On line 299, an extra return was added to avoid the "anonymous function
does not always return a value" CJS warning.

On line 9, an extra semicolon was removed.

This was tested with my own applets and confirmed working.

### DIFF
--- a/js/ui/environment.js
+++ b/js/ui/environment.js
@@ -333,9 +333,7 @@ function init() {
             };
 
         Promise._unhandledRejectionFn = function _unhandledRejectionFn(err) {
-            if (typeof console !== 'undefined' && console) {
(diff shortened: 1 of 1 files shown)


clefebvre:
Hi @jaszhix, I think this was discussed internally before you joined Slack. It wasn't met with enthusiasm, mostly because other devs didn't see the point.

I'll close the PR for now. Feel free to discuss the idea on Slack.

PR #6226, opened 2017-01-28
[WIP] Improve the look of xlet settings

Cobinja:
This puts all xlet settings widgets into a horizontal size group, making it look better.
Until now, all widgets had their own width, now they all share the same width.
Atm, it may look strange in the menu applet's settings, because there's a hotkey definition and a text entry, which are also put into the size group, this can be improved further, that's why I marked this WIP.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/SettingsWidgets.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/SettingsWidgets.py
@@ -359,7 +359,7 @@ class SettingsBox(Gtk.Frame):
         frame_style = self.get_style_context()
         frame_style.add_class("view")
         self.size_group = Gtk.SizeGroup()
-        self.size_group.set_mode(Gtk.SizeGroupMode.VERTICAL)
+        self.size_group.set_mode(Gtk.SizeGroupMode.HORIZONTAL)
 
         self.box = Gtk.Box(orientation=Gtk.Orientation.VERTICAL)
         self.add(self.box)
--- a/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py
@@ -273,6 +273,7 @@ class MainWindow(object):
                     elif settings_type in XLET_SETTINGS_WIDGETS:
                         widget = globals()[XLET_SETTINGS_WIDGETS[settings_type]](key, info["settings"], item)
                         section.add_row(widget)
+                        widget.add_to_size_group(section.size_group)
 
     def build_from_order(self, settings_map, info, box, first_key):
         page = SettingsPage()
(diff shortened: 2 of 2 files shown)


NikoKrause:
Screenshot? ;)

Cobinja:
Before:
screenshot-window-2017-01-28-225320
After:
screenshot-window-2017-01-28-225229
That's what it looks at this moment. If an xlet has other settings, e.g. a hotkey, the other elements look too wide, that's why it's WIP.

NikoKrause:
The height is different, isn't it.
What happens, when the word in the combobox is really long? Instead of Celsius something like FahrenheitTemperatureAndOtherWeirdStuff

Cobinja:
The height is different, yes.
With a really long entry in a combobox, it would make all others have the same width

JosephMcc:
I have to admit I'm not a big fan of that change. They no longer match the style of rest of cinnamon settings. It also totally breaks the styling of the switches. Those just don't look good. I'm also not sure exactly what behavior you would get with themes that use an image for GtkSwitch.

collinss:
I'm also a bit uncomfortable with this for usability reasons. Some of the widgets have to be larger in order to be functional (eg text entries, keybindings, file choosers, etc). As @JosephMcc pointed out, forcing some of the smaller widgets to be as large makes them look wrong. Also, I think this actually makes it slightly harder to follow.

Cobinja:
My personal preference would be to add the ability for settings-schema.json to define which items go into a size group and which don't. I'm just taking this as a starting point.
@collinss What do you mean by "harder to follow"?

JosephMcc:
@Cobinja I agree. That would be much better than forcing all widgets in to it. You also need to be sure nothing else is broken in cinnamon-settings by changing the size_group_mode in SettingsBox.

collinss:
@Cobinja It's a little hard to explain, but because everything is the same size my brain just sort of skips over them. My brain is weird though, so it could just be me.

Cobinja:
@JoseohMcc I did make sure. I used a previously unused group

Cobinja:
How about this?
screenshot-window-2017-01-29-112806
There seems to be a small allocation bug with the icon widget, will look into it.
For this screenshot, I put the switches and the spinbutton in one size group and the edit field and icon widget in another, configurable via settings-schema.json.

JosephMcc:
Personally I don't want to see switches placed in size groups ever. And like I said before this won't behave properly in themes that use an image to theme the switches. I wouldn't be against a dev being able to set size groups somehow using the settings API but that has a downside as well. It means some devs will probably do this well and others will break the look of items like switches in comparison to other cinnamon-settings pages.

Cobinja:
It's just an example. It would be up to the settings-schema, which items go into a group.

Cobinja:
Or I could manually exclude them

clefebvre:
Closing this for now.

PR #6271, opened 2017-02-12
St: support css margin property

brownsr:
From upstream https://github.com/GNOME/gnome-shell/commit/2935fd4ffe9176d28f68192bb34bdba1a0e57a2b
and https://github.com/GNOME/gnome-shell/commit/216e996f6680a3ffb797caafe1ed29990360c8f1

### DIFF
--- a/data/theme/cinnamon.css
+++ b/data/theme/cinnamon.css
@@ -1538,6 +1538,10 @@ StScrollBar StButton#vhandle:hover {
     font-weight: bold;
     color: #ccc;
 }
+.panel-top .applet-label,
+.panel-bottom .applet-label {
+    margin-left: 6px;
+}
 .applet-box:checked > .applet-label,
 .applet-box:hover > .applet-label {
     color: #fff;
--- a/js/ui/applet.js
+++ b/js/ui/applet.js
@@ -793,17 +793,19 @@ TextIconApplet.prototype = {
     /**
      * update_label_margin:
      *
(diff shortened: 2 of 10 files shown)


brownsr:
Gives more flexibility to theme designers, who can incorporate margin in places where the CSS margin property is inactive at the moment.  Most notable place is in the menu, where this PR permits margin to be used so the menu can be indented into the screen rather than butting right up against the panel

clefebvre:
Hi @brownsr, in the menu applet, this results in the panel icon and panel text "Menu" being stuck to each others.

Can you include the fix for this in this PR for the default theme?

We'll need two additional PRs (one on https://github.com/linuxmint/cinnamon-themes and one on https://github.com/linuxmint/cinnamon-spices-themes) to port that fix to the other themes.

clefebvre:
Actually... this happens for all IconText applets. I can see it with the Weather applet as well.

It might be a good idea to simply fix this by adding margin in the default CSS theme and setting that as important.

brownsr:
there's a routine update_label_margin in the main applet.js that this PR appears to be competing with.  I'll see if I can find a way for the two to play nice with each other.

brownsr:
Will need this adding to themes:

.panel-top .applet-label,
.panel-bottom .applet-label {
margin-left: 6px;
}

I don't think the important property will help, as themes will already have applet-label CSS.  I can ripple this through all themes after merging.

PR #6294, opened 2017-02-17
JS API misc/util.js: Add spawnCommandLineAsync

jaszhix:
This is a CLI utility method that invokes a callback after the process has exited cleanly, or an error callback if it does not. Neither callback will be called if the functions are not passed to the method. Having callbacks allows more precise handling of the asynchronous behavior involved with calling processes from JS.

This uses GLib.child_watch_add to return the callbacks after the process exits. https://people.gnome.org/~gcampagna/docs/GLib-2.0/GLib.child_watch_add.html

### DIFF
--- a/js/misc/util.js
+++ b/js/misc/util.js
@@ -107,15 +107,22 @@ function spawnCommandLine(command_line) {
 /**
  * trySpawn:
  * @argv: an argv array
+ * @doNotReap: whether to set the DO_NOT_REAP_CHILD flag
  *
  * Runs @argv in the background. If launching @argv fails,
  * this will throw an error.
  */
-function trySpawn(argv)
+function trySpawn(argv, doNotReap)
 {
-    let [success, pid]  = GLib.spawn_async(null, argv, null,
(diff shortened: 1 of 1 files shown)


itzexor:
Hi, I did some testing with this change and it needed some additional modifications to work. As per the documentation, we must pass the G_SPAWN_DO_NOT_REAP_CHILD flag since we are spawning it with g_spawn_async. This change worked for me, but is not heavily tested.

https://github.com/itzexor/Cinnamon/commit/712d72814c2a7ea2c1fdac0c9ae86521d332fdda

jaszhix:
Ah, thanks for testing this. I didn't have an issue when I tested, but I copied your changes to trySpawn over to my branch, and its working fine for me. Also, no problem, running external processes are asynchronous by their nature, and I think its needed.

itzexor:
For me, the callbacks were firing oddly and I was getting this:

> Window manager warning: Log level 16: GChildWatchSource: Exit status of a child process was requested but ECHILD was received by waitpid(). Most likely the process is ignoring SIGCHLD, or some other thread is invoking waitpid() with a nonpositive first argument; either behavior can break applications that use g_child_watch_add()/g_spawn_sync() either directly or indirectly.

PR #6299, opened 2017-02-20
Updated pkexec support

dg1727:
See https://github.com/linuxmint/mintsources/issues/75

In this pull request, file `files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py` is modified for the sake of the `mdm` GitHub project, not the Cinnamon GitHub project.

### DIFF
--- a/debian/cinnamon-common.install
+++ b/debian/cinnamon-common.install
@@ -1,7 +1,8 @@
 /etc
 /usr/share/cinnamon
+/usr/share/cinnamon-session
 /usr/share/desktop-directories
 /usr/share/glib-2.0
-/usr/share/cinnamon-session
-/usr/share/xsessions
 /usr/share/icons
+/usr/share/polkit-1/actions
+/usr/share/xsessions
--- a/files/usr/bin/cinnamon-settings-users
+++ b/files/usr/bin/cinnamon-settings-users
@@ -1,18 +1,25 @@
 #!/usr/bin/python3
 
-""" Launch the cinnamon users and groups editor utility
+""" Launch the Cinnamon users and groups editor utility
 """
 
 import os
-import subprocess
+import gettext
 
-def get_output_uni(*args):  # uni = Unicode 
-  return subprocess.check_output(*args, universal_newlines=True).rstrip()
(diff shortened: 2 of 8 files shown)


City-busz:
Please don't use the /usr/lib/linuxmint/common/mint-which-launcher.py script, because it's available only on Linux Mint. pkexec should be used instead of gksu.

dg1727:
Thanks for pointing out about `mint-which-launcher.py`.  I will look into it.  
>pkexec should be used instead of gksu.

That's exactly the point.  :-)  Due to your comment, I double-checked, and indeed my modified code doesn't use `gksu`.

City-busz:
>> pkexec should be used instead of gksu.
>
> That's exactly the point. :-) Due to your comment, I double-checked, and indeed my modified code doesn't use gksu.

The [mint-which-launcher.py](https://github.com/linuxmint/mint-common/blob/master/usr/lib/linuxmint/common/mint-which-launcher.py) uses gksu. So you didn't replace gksu, just made it impossible to run cinnamon-settings-users on other distributions.

leigh123linux:
I'm -1 for this PR to be merged if it uses mint-which-launcher.py

City-busz:
Why do you need gksu and kdesudo as fallback? pkexec is not enough? pkexec is available on all modern distributions.

dg1727:
>The mint-which-launcher.py uses gksu. So you didn't replace gksu

At the same time that I modified `usr/bin/cinnamon-settings-users` to call `mint-which-launcher.py`, I modified `mint-which-launcher.py` (https://github.com/linuxmint/mint-common/pull/22) to use `pkexec` if it's available, rather than `gksu`.  

>Why do you need gksu and kdesudo as fallback? pkexec is not enough? pkexec is available on all modern distributions.

I use `gksu` and `kdesudo` as fallback because that's what the existing `mint-which-launcher.py` did.  

Further comments from the development team are welcome.

leigh123linux:
@City-busz 

There is noting wrong with offering a choice.

> Why do you need gksu and kdesudo as fallback? pkexec is not enough? pkexec is available on all modern distributions.

dg1727:
Replaced by #6336.

PR #6303, opened 2017-02-21
xlet settings: add list widget

collinss:
* Allows adding/removing/editing/reordering entries in settings
* Allows a variable number of columns, each with it's own properties and data type
* Currently supports string, integer, float, bool, sound, icon, keybinding, and file variable types (more can be added easily - let me know if you have any ideas or requests)

list widget 2
list widget 3

### DIFF
--- a/data/icons/Makefile.am
+++ b/data/icons/Makefile.am
@@ -13,6 +13,7 @@ public_icons = \
 	hicolor_actions_scalable_pan-up-symbolic.svg \
     hicolor_actions_scalable_caps-lock-symbolic.svg \
     hicolor_actions_scalable_caps-lock-off-symbolic.svg \
+    hicolor_actions_scalable_list-edit-symbolic.svg \
     hicolor_actions_scalable_num-lock-symbolic.svg \
     hicolor_actions_scalable_num-lock-off-symbolic.svg \
 	hicolor_apps_scalable_cinnamon-panel-launcher.svg \
--- a/data/icons/hicolor_actions_scalable_list-edit-symbolic.svg
+++ b/data/icons/hicolor_actions_scalable_list-edit-symbolic.svg
@@ -1,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   id="svg7384"
+   height="16"
+   width="16"
+   version="1.1"
+   inkscape:version="0.91 r13725"
+   sodipodi:docname="list-edit-symbolic.svg">
+  <defs
+     id="defs9" />
+  <sodipodi:namedview
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1"
+     objecttolerance="10"
+     gridtolerance="10"
+     guidetolerance="10"
+     inkscape:pageopacity="0"
+     inkscape:pageshadow="2"
+     inkscape:window-width="1888"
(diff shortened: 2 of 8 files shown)


Odyseus:
Hello, @collinss.

I think that this is amazing! It isn't just useful, but absolutely needed!

~~I already can think of an applet that could make use of this ([Feeds Reader](https://github.com/linuxmint/cinnamon-spices-applets/tree/master/feeds%40jonbrettdev.wordpress.com)). It uses a 400 lines long Python script to create the widget to manage its feeds.~~ <- (You already mentioned [here](https://github.com/linuxmint/cinnamon-spices-applets/pull/154#issue-209815857)) In one of my applets, I could convert an entire setting page that holds more than 40 preferences of different types into just one list widget!

#### Suggestions/Feedback/Questions

1. ~~The **file** type can only select directories, not files.~~ But looking more closely, that can be changed in the settings-schema.json file (`"select-dir": true/false`). BTW, in the current settings-schema.json file that parameter is set as `"select-dir": 100` (an integer instead of a boolean).
2. Would it be possible to make all the entries in the list in-line editable? Not a must have, but it would facilitate the edition of large amount of data.
3. I would really like to also see available **iconchooser** and **keybinding** elements if possible.
4. I thing that all text boxes should be flexible. Maybe add the existent "expand-width" parameter that all entry types already have?
5. I added two long paths and the entire window resized horizontally beyond the limits of the screen and without the possibility to be resized nor maximized. I think that to fix this, the columns could be resizable and its content *ellipsized*.

collinss:
@Odyseus Thanks for the feedback, and for breaking my pr for me (no really, it helps).

1. You can leave the select-dir off and it will default to file select rather than directory. Also, ```"select-dir": 100``` was fixed as well (nice catch).
2. I think this might turn out to be not only be difficult to implement, but a bad idea as well. Certain items (eg. keybindings) have particular formatting, and an incorrect format could spell disaster.
3. Added. I also added soundchooser as well because I could and it was easy. I was going to add datechooser, but there are some challenges since the date is stored as a json style object rather than a basic type. I might still if I can figure out a good way to make it play nice with the Gtk tree widgets without changing the way datechoosers work.
4. It should be up to the applet dev to determine if it's needed or not. The "expand-width" property is already available. The individual authors should be the ones to decide if a particular field needs to be larger or not.
5. Fixed.

collinss:
I just updated the add/edit dialog to look a little more like the rest of settings:
list widget 4

collinss:
I think this is ready.

Odyseus:
Hello, @collinss

I tested your last changes and all seems to be working perfect.

> It should be up to the applet dev to determine if it's needed or not. The "expand-width" property is already available. The individual authors should be the ones to decide if a particular field needs to be larger or not.

I think that you misunderstood my request on point 4. I was talking about the entries on the following dialog:

testlistwidget2

See those highlighted entries? Those are the ones that should be able to expand. Right now, setting the *expand-width* property on *string* or *icon* column types does not make those entries expandable.

There is another thing that I'm curious about (not directly related to this pull request). Why the change and use of URIs (file:///path/to/file/or/folder) on *filechooser* elements, and now on the list widget's *file* type columns, when in the past the *filechooser* elements returned a path (/path/to/file/or/folder)? That is kind of inconsistent, especially considering that the *iconchooser* elements are still returning a path on the new settings system, just like the old settings system does. This change caused me a real headache when I first started using the new settings system, because the function I used was expecting a path, not an URI.

collinss:
Ah, I see what you mean. I didn't realize those didn't support the expand property (someone else added that after-the-fact - I had nothing to do with it). I'll see about adding them, but I'll probably do so in a separate pr as that's kind of outside the scope of this one. As for the uri/path issue, when I did the revamp of xlet settings, I basically used the gsettings version of the widgets if they already existed, and only ported from the applet settings if they didn't. Most likely there was already a filechooser, and I didn't realize the difference.

collinss:
@Odyseus Apparently I misunderstood you again. I just updated it again, and it should be working now. Let me know if it doesn't.

Odyseus:
Hello, @collinss.

Now we are cooking! This is how it looks now with the *expand-width* option set:

testlistwidget3

Yes, I know. It doesn't look pretty. But at least they are comfortable to work with. That's why [I added the *expand-width* option](https://github.com/linuxmint/Cinnamon/pull/5891) to *entry* and *iconfilechooser* elements (I'm the culprit of that. LOL).

By the way, thanks again for adding this new list widget. This adds so much more possibilities and adds a most needed *power* to Cinnamon's native settings system.

Driglu4it:
Hello, i have a question. How use a list widget in applet (how to get title of column and use it in applet)?
Sorry, for my bad English...

PR #6334, opened 2017-03-03
util.js: add documentation for spawn_async

collinss:


jaszhix:
No command I pass to this function, string or array, seems to work. I tried 'touch ~/test.html' and ['touch', '~/test.html']. The callback isn't firing for me, either.

Odyseus:
Hello, everybody.

@jaszhix: I use this function in almost all my xlets and it works as intended. But it seems that the callback function is triggered only if the command is successful. In your example, the command isn't successful because **~** isn't expanded. The following example works:

```javascript
Util.spawn_async(["touch", GLib.get_home_dir() + "/test.html"], Lang.bind(this, function() {
    Main.notify("Hello, there!!!", "How are you?");
}));
```

collinss:
@jaszhix as @Odyseus mentioned, for some reason the GLib utilities don't interpret ~/ as a shortcut for the user's home directory. If you want to make sure that it is interpreted correctly you could do a quick find and replace before passing the command to the spawn function. That's what I do in the places I use it. I suppose we could add this to util.js as well.

jaszhix:
Ah okay, I got this working. I think I tried just passing a string to it before which is why I never thought it worked, so I guess we don't need my other PR.

js/misc/util.js
```diff
@@ -78,6 +78,14 @@ function spawn(argv) {
 
 let subprocess_id = 0;
 let subprocess_callbacks = {};
+/**
+ * spawn_async:
+ * @args: the command to run asynchronously
```
jaszhix: Could you specify the type of parameter? Such as ```@args {array}```.

collinss: Oh, yeah.
Better?

jaszhix: Looks good. :+1:

PR #6336, opened 2017-03-04
Update pkexec support

dg1727:
This PR should fix the issue mentioned in https://github.com/linuxmint/mintsources/issues/75#issuecomment-282432902 
This PR replaces #6299

The branch name is different now because I wanted to rebase (rather than merge) against the other commits that have happened to the linuxmint/Cinnamon repo, and apparently the way to do that is to make a new branch.

### DIFF
--- a/debian/cinnamon-common.install
+++ b/debian/cinnamon-common.install
@@ -1,7 +1,8 @@
 /etc
 /usr/share/cinnamon
+/usr/share/cinnamon-session
 /usr/share/desktop-directories
 /usr/share/glib-2.0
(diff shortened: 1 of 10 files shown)


collinss:
You can rebase on the same branch. You just need to force-push to the remote (eg. ```git push -f origin my-branch```) because the histories will be divergent.

clefebvre:
Please rebase this PR.

Also make sure the tool works with pkexec. I've only done limited testing here, but it crashes for me... fails to get the display.

Fantu:
is there someone that can rebase and test it?

PR #6346, opened 2017-03-08
[WIP] Systray applet: make it easier to move in panel edit mode by replacin…

collinss:
…g the icons with a non-reactive box


collinss:
There is a problem here where clicks are still being passed to the status icons even when they are hidden. This is possibly due to how plugs and sockets work, but I haven't been able to track it down yet.

clefebvre:
Closing this for now. Let's create a new PR if we revisit.

PR #6348, opened 2017-03-10
[WIP]systray: add css class to permit theming

brownsr:


brownsr:
PR is fine as is, but review has made me think about whether theming should be hooked to the systray itself, or to the elements within it.  Or both.. Marked as WIP while I do some experimenting

collinss:
You should probably also add the important property to make sure it still looks ok. (I know we control the spices now, but how many people are going to forget to update their themes when they upgrade Cinnamon to the newer version, and it won't hurt anything.)

PR #6349, opened 2017-03-10
Switch to vp9

leigh123linux:
Based on https://git.gnome.org/browse/gnome-shell/commit/?id=d183f13456991d12ea57ad14ba38a1f7407d0037

From d183f13456991d12ea57ad14ba38a1f7407d0037 Mon Sep 17 00:00:00 2001
From: Adel Gadllah <adel.gadllah@gmail.com>
Date: Sun, 11 Jan 2015 12:16:40 +0100
Subject: recorder: Switch to vp9

Currently we have been using the vp8 codec because it was the best unencumbered codec at that time. With vp9 we now have a successor that leads to smaller
files at at the same video quality and has been supported by current browsers
for a while.

With the raise of hidpi and 4K displays we need a better codec that handles
those resolutions better, so switch to vp9.

https://bugzilla.gnome.org/show_bug.cgi?id=742744

### DIFF
--- a/src/cinnamon-recorder.c
+++ b/src/cinnamon-recorder.c
@@ -150,7 +150,7 @@ G_DEFINE_TYPE(CinnamonRecorder, cinnamon_recorder, G_TYPE_OBJECT);
  * (Theora does have some support for frames at non-uniform times, but
  * things seem to break down if there are large gaps.)
  */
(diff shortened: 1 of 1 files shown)


leigh123linux:
Closing for now to investigate high cpu usage

PR #6359, opened 2017-03-15
Make sure applets that don't support vertical panels don't get droped…

collinss:
… on a vertical panel. This was happening with at least some applets, causing them to 'disappear', and could only be recovered by removing them from applet settings and readding them again.


brownsr:
Tests out fine for me.  And tidies up the code :)

PR #6373, opened 2017-03-22
spawn_async: clean up the callbacks after we're done

itzexor:
This ensures that cinnamon-process-wrapper never fails due to an uncaught
exception when launching a process. With this, we can always call back into
cinnamonDbus.js and cleanup the subprocess_wrapper callback. The callback
is only invoked if the command runs successfully and exits with a status
code of 0. NFCI

Note: spawn_async can only be used if your command exits with a status of 0. If you're trying to differentiate between program exit states then you'll need to look elsewhere. An example is "gksudo", if the user cancels then you get status 255, incorrect password is status 1, and success is status 0. With this command you will not be able to tell if the user canceled, or typed the password wrong.

### DIFF
--- a/files/usr/bin/cinnamon-subprocess-wrapper
+++ b/files/usr/bin/cinnamon-subprocess-wrapper
@@ -11,8 +11,14 @@ import dbus
 
 if __name__ == "__main__":
     process_id = int(sys.argv[1])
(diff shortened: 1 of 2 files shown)


clefebvre:
true and false should be True and False.

But more importantly ALL PRs should be TESTED.

This is not cool, at all.

Issue #4503, opened 2015-08-12
Panel shadow disappear


darekdeo:
Hi, I have this issue since update to 17.2 Mint. Panel shadow disappear after both panel hide or intelii hide.

Here is how it looks:
before panel hide:
![Image of panel with shadow](http://i.imgur.com/gbUNFp8.png)
after panel hide:
![Image of panel without shadow](http://i.imgur.com/RcOB8Il.png)

DarekDeo:
Issue still present on 17.3 and Cinnamon 2.8.6. Other themes which use additional panel styling like Faience+ (curved corners like in gnome3) share the same issue of styling disappear after panel did hide.

Vahan86:
Can't reproduce it in Cinnamon 3.0.

JosephMcc:
I can. The problem with the panel shadow disappearing sometimes has been around for along time. I don't think anyone has ever pinned down the cause.

PR #6411, opened 2017-04-05
Panels: clean up _showPanel and _hidePanel and fix shadow with autohide

itzexor:
See commit messages for a detailed description.


JosephMcc:
This should fix: https://github.com/linuxmint/Cinnamon/issues/4503

Issue #2533, opened 2013-10-24
[cinnamon-settings → accessibility] high contrast toggle does not remember theme settings


VolMi:
In the accessibility module of cinnamon-settings, if I activate and deactivate the "high contrast" toggle, my previously defined custom theme settings are lost.

VolMi:
Still reproducible in 2.6.

hallamjeff:
This is reproducible in 2.8.6

JosephMcc:
Can still reproduce on 3.0 as well.

I think what's happening is when you switch to High Contrast and back it is setting your theme to the default value stored in org.cinnamon.interface. In the case of Linux Mint those are the Mint-X themes.

mtwebster:
What I'm going to do here is add a bit more functionality to our accessibility plugin in c-s-d to take care this (along with a key or keys to store the previously selected themes) - indeed right now it recalls the default when you turn off high-contrast (the same as our old module did)

PR #6415, opened 2017-04-06
cs_accessibility: Restore previous theme when disabling High Contrast

JosephMcc:
We want to go back to the theme that was being used before the High Contrast
was enabled. Not reset to the default theme.

Fixes https://github.com/linuxmint/Cinnamon/issues/2533


JosephMcc:
Requires https://github.com/linuxmint/cinnamon-desktop/pull/85

PR #6419, opened 2017-04-07
Panel highlighting and _updatePanelVisibility cleanup

itzexor:
The first change keeps track of panel highlight state so that we can avoid noop style changes and ensure panel visibility while highlighted.

The second change slightly refactors _updatePanelVisibility to stick the body of the code in a conditional based on whether we are forcing visibility or not. This avoids wasting time calculating autohide during panel edit mode or highlighting.

### DIFF
--- a/js/ui/panel.js
+++ b/js/ui/panel.js
@@ -1877,6 +1877,7 @@ Panel.prototype = {
 
         this._hidden = false;
         this._disabled = false;
+        this._highlighted = false;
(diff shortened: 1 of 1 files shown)


itzexor:
I'm having second thoughts about how I want to implement this. When a panel is force-shown during settings editing, then it becomes impossible for a user to easily test the auto hide/show delays.

PR #6438, opened 2017-04-14
Panel.js: _moveResizePanel refactor and cleanups

### LINKED ISSUE #6385: With autohiding top panel, right clicking flashes up a menu very briefly
When there is a top panel set to autohide, right clicking at the very top of the screen flashes up a menu very briefly. . and sometimes - i.e., seemingly on some themes but not others - flashes the panel, i.e. brings it down and then hides it again. I thought for a moment that my Conky - which sits at the vertical top of the screen - caused the problem, but I killed the Conky and the problem persisted.

Mint 18.1 x64.
Cinnamon 3.2.7.

itzexor:
This fixes most instances of hidden panel flashing, and also has a minor fix and some cleanups. See commit messages for detailed descriptions. Please test and comment if you find any bugs.

This branch exposes an issue in existing code which is fixed in #6460

Fixed issues that I don't know if are reported:
- Panel shadows show up on adjacent monitors
- Hidden panels flash on any settings or theme change
- Panel sometimes disappears when moving a panel with autohide enabled
- Hidden vertical panels partially show when activating a horizontal panel menu
- Recently added shadow clipping code in _showPanel doesn't handle all shadow types correctly

Existing issues:
- <s>Partial vertical panels don't have a shadow on top or bottom. These should be unclipped when the panel above or below hides. This is not worth the added complexity IMO. </s> I now believe that this is the intended behavior.
- Horizontal panels shadow overlaps vertical panels. I don't know how to solve this as the clip region would need to be non-rectangular.
- When activating a menu on a hidden panel, the menu can briefly appear behind the panel if the panel shows while the menu is active. The menu pops into place once hovered, so it is not a big issue.
- When activating a menu on a hidden panel, the menu may briefly appear on an adjacent monitor. I don't know the fix for this, but I don't think it is in the panel code.

Fixes #6385

### DIFF
--- a/js/ui/panel.js
+++ b/js/ui/panel.js
@@ -951,6 +951,7 @@ PanelManager.prototype = {
                 }
 
             } else { // Nothing happens. Re-allocate panel
+                this.panels[i]._monitorsChanged = true;
                 this.panels[i]._moveResizePanel();
             }
         }
@@ -1881,6 +1882,8 @@ Panel.prototype = {
         this._autohideSettings = null;
         this._themeFontSize = null;
         this._destroyed = false;
+        this._positionChanged = false;
+        this._monitorsChanged = false;
         this._signalManager = new SignalManager.SignalManager(this);
         this.margin_top = 0;
         this.margin_bottom = 0;
@@ -2028,6 +2031,7 @@ Panel.prototype = {
     updatePosition: function(monitorIndex, panelPosition) {
         this.monitorIndex = monitorIndex
         this.panelPosition = panelPosition;
+        this._positionChanged = true;
 
         this.monitor = global.screen.get_monitor_geometry(monitorIndex);
         //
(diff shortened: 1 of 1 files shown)


itzexor:
I need to squash this before merge.

js/ui/panel.js
```diff
@@ -3149,29 +3149,28 @@ Panel.prototype = {
         }
 
         // set up original and destination positions and add tween
-        // destination position paramater
+        // destination paramater
+        let origPos, destPos;
```
itzexor: this is defined twice

itzexor:
I have rebased this to fix some problems I noticed, as well as some stuff I broke while splitting it.

itzexor:
For some reason I still get flashes on one single panel on theme changes. It doesn't happen with another panel on a different monitor in the same position. I'll have to keep looking for the cause of that.

itzexor:
I'd actually prefer to keep this split now as it is much easier to understand the changes this way, but as they are built on top of eachother care must be taken when partial-reverting.

clefebvre:
@itzexor can you rebase this one please?

itzexor:
I have rebased it but I hit an issue with the new changes and I need to make sure that's resolved before testing if the rebase is fully working correctly.

itzexor:
This should be good to go, I tested it on master and everything seems to be working as intended with the new margin changes.

itzexor:
I'm removing the layout.js commit as I want to do it a different way, and I'm going to put it in a different PR.

Issue #6173, opened 2017-01-05
avatar is blurred on lockscreen


igordovgaluk:
user's avatar is blurred on lockscreen.
cinnamon 3.2.7 (Mint 18.1)

when avatar was set the selected image had 1000x1000 pixels.
maybe internally in cinnamon the reduced image is stored that is too small for the lockscreen.

JosephMcc:
Could you please move this here: https://github.com/linuxmint/cinnamon-screensaver/issues
Since the lock screen is completely new it would be nice to keep these gathered together.

igordovgaluk:
looks like same as https://github.com/linuxmint/cinnamon-screensaver/issues/187

PR #6452, opened 2017-04-16
cs_user.py: Save appropriately sized .face for cinnamon-screensaver

EnglishInfix:
I have upped the image.thumbnail resolution from 96x96 to 255x255. This will save an appropriately .face image for the new larger face thumbnail on the lockscreen for cinnamon-screensaver.

This push resolves bug linuxmint/cinnamon-screensaver#187 and closed duplicate #6173.


JosephMcc:
> Why 255?, shouldn't it be 256?

I was thinking the same thing.

EnglishInfix:
cinnamon-screensaver/src/widgets/framedImage.py has a comment that the maximum size for the image frame is 255px.

igordovgaluk:
i believe the bigger images should be supported. 1024+
otherwise new bugs will be opened for HiDPI diplays

PR #6453, opened 2017-04-16
cinnamon-settings: Fix some sliders being drawn backwards

JosephMcc:
Add a flipped argument to the cinnamon settings range widget. Gtk.Scale is a
little limited in what it allows. There is no way built in to map a range from
a big->small value without drawing the fill value in reverse. Add a bit of a
hack to allow working around this limitation.


JosephMcc:
Added fixes to allow resetting the acceleration and sensitivity controls back to the defaults. Requires https://github.com/linuxmint/cinnamon-settings-daemon/pull/158

PR #6454, opened 2017-04-17
power-applet: Add support for warning and error style classes

JosephMcc:
These allow us to recolor the applets icon depending on the current state of
the battery. Use a 'warning' style class when the percentage drops below 20 and
an 'error' class when it drops below 5.


clefebvre:
@JosephMcc can you port this to Mint and spices themes?

JosephMcc:
Yep.

PR #6474, opened 2017-04-23
st: margin: avoid unnecessary warning messages for margin:auto

brownsr:
the warning messages do not bring anything except distraction, so suppress them, and log a code FIXME to consider whether margin:auto would be a useful facility to add to the st library in the future

### DIFF
--- a/src/st/st-theme-node.c
+++ b/src/st/st-theme-node.c
@@ -1537,7 +1537,17 @@ do_margin_property (StThemeNode   *node,
         return;
       else if (decl->value->next == NULL) /* 1 value */
         {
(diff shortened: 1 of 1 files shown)


JosephMcc:
If `margin: auto` doesn't work and we're going to hide warnings saying as much, it's use should probably be removed from the default theme. Having it there just leads to confusion.

brownsr:
I agree. will have to do some testing just to make absolutely sure there is nothing reacting to it.  I did some grepping earlier and I could see no evidence that anything was, but I'd want to be absolutely sure.

brownsr:
On reflection I could not see a reason to suppress the warnings, removal of the margin:auto in CSS would be preferable.  Have pushed a PR for that in the default theme and the menu applet.

PR #6482, opened 2017-04-25
Update cs_power.py

mikee805:
was not getting  low battery setting on ubuntu 16.04 running cinnamon 3.2.8-1 since it has gir1.2-upowerglib-1.0 version 0.99.4-2ubuntu0.3 so the minor version should be <= to 99


JosephMcc:
Nice catch.

PR #6484, opened 2017-04-25
[RFC] window-list: remove hard coded CSS

brownsr:
(not switcher list)
There's nothing here that can't be handled in static CSS, so it makes sense to get rid of
the hard-coding so it actually behaves as themed

Two things : 1) Mint-Y and Mint-X have 1 pixel of top padding that would require some theme changes, as currently that would lead to an annoying single pixel unreactive edge strip in the top panel.  Not difficult, but it's a bit of forced work that accepting this PR would make necessary.
2) panel-top and top (etc. for left, right, bottom) are both used in themes.  I've allowed for both, but perhaps we ought to cut down to one.  panel-top ?

The preview box could benefit from similar removal of hard-coding, but that's not addressed here.

### DIFF
--- a/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
@@ -1030,35 +1030,28 @@ MyApplet.prototype = {
             this.actor.set_important(true);
         }
 
-        // Any padding/margin is removed on one side so that the AppMenuButton
-        // boxes butt up against the edge of the screen
-
         if (orientation == St.Side.TOP) {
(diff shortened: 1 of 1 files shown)


JosephMcc:
> 2) panel-top and top (etc. for left, right, bottom) are both used in themes. I've allowed for both, but perhaps we ought to cut down to one. panel-top ?

Actually we should get rid of both. If you want specific theming you can use `.panel-top .window-list-item-box`

brownsr:
Of course !  simply ensuring class window-list-item-box is set is all that is needed. Doh.  I'll take another look, it may be all of this existing code is monkey puzzle that can go. Closing for now.

PR #6486, opened 2017-04-27
main.js: improve logging of GLib.Errors and objects 

itzexor:
Prevents an error when calling the dbus method
org.Cinnamon.LookingGlass.GetErrorStack

g-io-error-quark: GDBus.Error:org.gnome.gjs.JSError.ValueError: Service
implementation returned an incorrect value type (36)

### DIFF
--- a/js/ui/main.js
+++ b/js/ui/main.js
@@ -927,7 +927,22 @@ function notifyError(msg, details) {
  * extension system as well as debugging.
  */
 function _log(category, msg) {
-    let text = msg;
+    if (msg == undefined) {
+        _log('error', _('logging failed: message was null or undefined'));
+        return;
+    }
+
+    let cat, text;
+    if (typeof category !== 'string')
+        cat = 'info';
+    else
+        cat = category;
+
+    if (typeof msg !== 'string')
+        text = msg.toString();
+    else
+        text = msg;
+
     if (arguments.length > 2) {
         text += ': ';
         for (let i = 2; i < arguments.length; i++) {
@@ -937,7 +952,7 @@ function _log(category, msg) {
         }
     }
     let out = {timestamp: new Date().getTime().toString(),
-                         category: category,
+                         category: cat,
                          message: text };
     _errorLogStack.push(out);
     if (lookingGlass)
@@ -954,7 +969,23 @@ function _log(category, msg) {
  * Returns (boolean): whether @obj is an error object
  */
 function isError(obj) {
-    return typeof(obj) == 'object' && 'message' in obj && 'stack' in obj;
+    if (obj == undefined) return false;
+
+    let isErr = false;
+    if (typeof(obj) == 'object' && 'message' in obj && 'stack' in obj) {
+        isErr = true;
+    } else if (obj instanceof GLib.Error) {
+        // Make existing logging functionality work as expected when passed
+        // a GLib.Error which doesn't normally have a stack trace attached.
+        let stack = new Error().stack;
+        // This is reached the first time isError is called by a _log function,
+        // so strip off this function call and the _log function that called us.
+        let strPos = stack.indexOf('\n', stack.indexOf('\n') + 1)  + 1;
+        stack = stack.substr(strPos);
+        obj.stack = stack;
+        isErr = true;
+    }
+    return isErr;
 }
 
 /**


clefebvre:
@itzexor could these variables be Null? should we test that before calling .toString() on them?

itzexor:
I'm not sure. I had the same concern, but I wasn't sure the best way to add a null check considering the `arugments` parsing that's going on.

The timestamp will never be null (I sure hope, at least). Category is easy null check. The part I'm not really sure about is if we are passed null for msg initially, but also have some extra arguments the message won't ever actually be null since we're at the very least appending strings. That's just my guess, but I'm not completely sure.

I think maybe `if (category == undefined || msg == undefined) return;` would do it. That would match undefined and null. I just don't know if anything depends on passing null msg + args (I would hope not, but I'm not totally sure on that.)

itzexor:
Let me know if you think this should be solved in a different way.

js/ui/main.js
```diff
@@ -927,6 +927,8 @@ function notifyError(msg, details) {
  * extension system as well as debugging.
  */
 function _log(category, msg) {
+    if (category == undefined || msg == undefined) return;
```
jaszhix: You could just use !category || !msg here because if they are null this won't catch it because you're only checking for undefined. As long as it is at least a variable, which parameters are, you don't need to check undefined.

In JS this checks

- null
- undefined
- NaN
- empty string ("")
- 0
- false

jaszhix:
Actually this might work better:

```js
function _log(category, msg) {
    category = typeof category == 'string' ? category.toString() : category;
    msg = typeof msg == 'string' ? msg.toString() : msg;
    
    let text = msg;
    if (arguments.length > 2) {
        text += ': ';
        for (let i = 2; i < arguments.length; i++) {
            text += JSON.stringify(arguments[i]);
            if (i < arguments.length - 1)
                text += ' ';
        }
    }
    let out = {timestamp: new Date().getTime().toString(),
                         category: category,
                         message: text };
    _errorLogStack.push(out);
    if (lookingGlass)
        lookingGlass.emitLogUpdate();
    if (can_log) lg_log_file.write(renderLogLine(out), null);
}
```
Because we don't want to skip a log just because strings aren't passed, assuming whatever function reading the ```_errorLogStack``` array isn't going to choke on null values.

Or if we're writing this for mozjs38:

```js
function _log(category, msg) {
    let text = msg;
    if (arguments.length > 2) {
        text += ': ';
        for (let i = 2; i < arguments.length; i++) {
            text += JSON.stringify(arguments[i]);
            if (i < arguments.length - 1)
                text += ' ';
        }
    }
    let out = {timestamp: new Date().getTime().toString(),
                         category: `${category}`,
                         message: `${text}` };
    _errorLogStack.push(out);
    if (lookingGlass)
        lookingGlass.emitLogUpdate();
    if (can_log) lg_log_file.write(renderLogLine(out), null);
}
```

Use string template syntax.

itzexor:
It has to support js24 and js38 for the time being. The `Main._errorLogStack` object/array is directly converted into a gvariant with the type string `aa{ss}` for use in a dbus method return. That happens inside lookingGlass.js and is called externally by Melange(python looking glass).

Each object in `Main._errorLogStack` must be a dictionary with keys and values both of string type. With js38 something in the stack is not a string and is now throwing a TypeError inside cjs when the gvariant is created. I don't know exactly what log line it is.

In js, at least here in cinnamon, `null == undefined` is true. It won't evaluate true if you use the type checking `===` operator, though.

jaszhix:
You're right, I'm used to strict equality (===). I was thinking about it and my first example after the review is wrong. If an object is passed it will show as '[object object]', so ```toString``` would work after being serialized. I'm curious why its doing that though in strict mode, if the inputs passed to JSON.stringify are being serialized or if its the category value.

itzexor:
The only args that should be passed to JSON.serialize are the arguments after msg like in `_log("info", "Got these objects", someObject, someString, someArray)` `"info"` and `"Got these objects`" would be used directly and `someObject, someString, someArray` woudl be serialized and turn the var `text` into something like `"Got these objects: [Object object] [string] [Object object]"`

jaszhix:
Yes. I think we are on the same page (I think?) with passing objects through JSON.stringify first. I am running mozjs38 again with CJS so I can try to reproduce this.

itzexor:
Ok I got around to tracking down the root cause of this issue. If you have a glibtop-using applet enabled when upgrading to mozjs38 and assuming the glibtop issue is still unresolved, a `Glib.Error` is passed as the `msg` parameter. This is why it's failing to be converted to a GVariant.

I added this code to diagnose what was going on:

inside main.js file/main/module scope or whatever you call this scope:
`let bad_text, bad_msg, bad_category, bad_arguments;`
after arguments parse code in `_log()`:
```javascript
    let categoryIsString = typeof category === 'string';
    let textIsString = typeof text === 'string';
    if (!categoryIsString || !textIsString) {
        bad_text = text;
        bad_arguments = arguments;
        let warningMsg = "Attempting to log non-string args: "
                         + (categoryIsString ? "" : "category ")
                         + (textIsString ? "" : "text ");
        _logWarning(warningMsg);
        _logTrace();
        return;
    }
```

This gave me the following output for the offending log call:
```
warning t=2017-04-28T16:15:51Z Attempting to log non-string args: text 
trace t=2017-04-28T16:15:51Z 
<----------------
_log@/usr/share/cinnamon/js/ui/main.js:952:9
_logError@/usr/share/cinnamon/js/ui/main.js:1059:9
Extension.prototype.logError@/usr/share/cinnamon/js/ui/extension.js:189:13
addAppletToPanels@/usr/share/cinnamon/js/ui/appletManager.js:336:9
finishExtensionLoad@/usr/share/cinnamon/js/ui/appletManager.js:60:18
loadExtension@/usr/share/cinnamon/js/ui/extension.js:451:17
onEnabledAppletsChanged@/usr/share/cinnamon/js/ui/appletManager.js:236:13
---------------->
error t=2017-04-28T16:15:51Z [Applet "cinnamem@exor"]: Failed to load applet: cinnamem@exor/138
error t=2017-04-28T16:15:51Z Could not load applet cinnamem@exor
info t=2017-04-28T16:15:51Z Adding systray: steam (24x24px)
info t=2017-04-28T16:15:51Z Adding systray: unknown (24x24px)
error t=2017-04-28T16:15:51Z applet._extension is undefined
trace t=2017-04-28T16:15:51Z 
<----------------
removeAppletFromPanels@/usr/share/cinnamon/js/ui/appletManager.js:260:9
onEnabledAppletsChanged@/usr/share/cinnamon/js/ui/appletManager.js:209:17
---------------->
error t=2017-04-28T16:15:51Z Failed to refresh list of applets
info t=2017-04-28T16:15:51Z Adding systray: steam (24x24px)
info t=2017-04-28T16:15:51Z Adding systray: unknown (24x24px)
info t=2017-04-28T16:15:53Z Resized steam with normalized size (16x16px)
info t=2017-04-28T16:15:53Z Resized unknown with normalized size (16x16px)
```
and the vars I stored in Main can be inspected and reveal:
```
Main.bad_arguments.length = 2
Main.bad_arguments[0] (category) = "error"
Main.bad_arguments[1] (msg) = [boxed instance proxy GIName:Glib.Error jsobj@address native@address]
Main.bad_text = [boxed instance proxy GIName:Glib.Error jsobj@address native@address]
```

So there are at least two problems here: we're not handling non-string arguments properly in `Main._log()` and we're passing a non-string argument in the extension code, which seems to also be breaking, possibly for other reasons, at the same time.

After looking into this more, I think we should fix the extension code to only pass a string, and check `typeof category and msg === 'string'` inside `Main._log`

Sorry for the super long message. Let me know what you think.

itzexor:
Also my guess is that this probably would've broken in js24 the same way, if we ever had an extension throw a GLib.Error, maybe when the library wasn't installed? I haven't seen this though.

itzexor:
Ok, dug even deeper. The reason we are receiving a `GLib.Error` as the message, is because those errors do not have a `stack` property. Our error logging code expects every error to have that property, if it doesn't it's treated as a message.

I don't see a clean way out of this one... The most effective, but probably dirtiest, method I can think of would be to somehow ensure all errors that make it past cjs into user code (from cjs's perspective) are JS native errors. This is dirty as hell, and I'm not sure if it's even possible. It would probably require all kinds of different error type handling code, and would make us diverge more from upstream.

Another fix would be to change logging in main.js to properly handle errors without a stack. This would touch almost every logging function inside Main.js from what I have seen. One way this could be done is to use a helper function wherever we are grabbing the stack that would return an error's stack if it exists, or create a new stack trace (and somehow intelligently clip the useless last few function calls).

I have hacked together a partial implementation of the second method there. It does work, and the messages are displayed as originally intended in appletManager.js/extension.js, but I don't like this solution so I'll wait for any better suggestions or just leave everything as is for now.

itzexor:
And of course the simplest solution is to just convert any non-string, non null/undefined, non-error `message` into its string representation, or serialize it. String representation makes the logged error completely useless from a support perspective. A serialized object representation (when not strictly intended) could end up being extremely long(and will it even work on non-js natives?).

jaszhix:
Maybe we could just check the stack property in ```_LogTraceFormatted``` and ```_logTrace```? If the missing stack prop is the cause of the error, I think it could be straight forward.

```js
function _logTrace(msg) {
    if (!msg.stack) {
      msg.stack = ''
    }
    if(isError(msg)) {
        _LogTraceFormatted(msg.stack);
    } else {
        try {
            throw new Error();
        } catch (e) {
            // e.stack must have at least two lines, with the first being
            // _logTrace() (which we strip off), and the second being
            // our caller.
            let trace = e.stack.substr(e.stack.indexOf('\n') + 1);
            _LogTraceFormatted(trace);
        }
    }
}
```

itzexor:
That sounds like a good solution, but I think we should also get a trace for non js-native errors.

mtwebster:
It's either going to be a string or GLib.Error - can't we just check the type (`instanceof`)  of msg and react accordingly?

itzexor:
Yes that is the method I used in the partial implementation I spoke of. Checking `instanceof GLib.Error` does work as expected. I wasn't aware that was the only other error we would need to handle.

itzexor:
This seems hacky, changing it in a function meant to check it, but it works.

```javascript
/**
 * isError:
 * @obj (Object): the object to be tested
 * 
 * Tests whether @obj is an error object. If the object
 * is a GLib.Error then we add a stack trace to it.
 *
 * Returns (boolean): whether @obj is an error object
 */
function isError(obj) {
    let isErr = false;
    if (typeof(obj) == 'object' && 'message' in obj && 'stack' in obj) {
        isErr = true;
    } else if (obj instanceof GLib.Error) {
        let stack = Error().stack;
        stack = stack.substr(stack.indexOf('\n') + 1);
        obj.stack = stack;
        isErr = true;
    }
    return isErr;
}
```

mtwebster:
I'd be ok with that, it still answers the question "is it an error?" without complicating all the points this is checked - we just need to make sure the error has a stack.  GLib.Error is the only error type I can think of dealing with - any library that uses gobject (including wrapper libraries for libs that don't) throws GErrors, or some implementation of them.

itzexor:
I have rebased this.

PR #6490, opened 2017-04-27
[WIP] load xlet after version check anyway

NikoKrause:
E.g. you use Cinnamon 3.2
If you want to add an xlet, which has the cinnamon-version flag in it's metadata.json file, but Cinnamon 3.2 is not included in this.
You get the following message:
> Extension %s is not compatible with current version of cinnamon. Using it may break your system. Load anyway?
> Yes, No

If you press yes, the xlet won't load anyway.

Allow xlet to load, if clicked Yes. Don't throw an error, log warning into LookingGlas instead.


NikoKrause:
@collinss @clefebvre 
I'm not very familiar with python so please doublecheck this.
I tested it on LM 17.3, just added the [WIP] so you are aware to better test it befor merging.

@dalcde 
You added the line
```
else 
    uuid = "!" + uuid
```
Do you remember why? I don't know why it's there. Removing it, seems to solve my issue mentioned above.

PR #6491, opened 2017-04-28
js/ui/tooltips.js: Check if this.mousePosition is defined

jaszhix:
Encountered this when calling a PanelItemTooltip instance from a child class in ITM.

```sh
(cinnamon:10020): Cjs-WARNING **: JS ERROR: TypeError: this.mousePosition is undefined
PanelItemTooltip.prototype.show@/usr/share/cinnamon/js/ui/tooltips.js:318
open@/home/jason/.local/share/cinnamon/applets/IcingTaskManager@json/specialMenus.js:653
hoverOpen@/home/jason/.local/share/cinnamon/applets/IcingTaskManager@json/specialMenus.js:635
_onMenuEnter/<@/home/jason/.local/share/cinnamon/applets/IcingTaskManager@json/specialMenus.js:610
setTimeout/<@/home/jason/.local/share/cinnamon/applets/IcingTaskManager@json/__init__.js:12
```

### DIFF
--- a/js/ui/tooltips.js
+++ b/js/ui/tooltips.js
@@ -206,6 +206,8 @@ Tooltip.prototype = {
         this.desktop_settings = new Gio.Settings({
             schema_id: DESKTOP_SCHEMA
         });
+
+        this.mousePosition = null;
     },
 
     hide: function() {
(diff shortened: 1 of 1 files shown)


brownsr:
Not sure about the additional checks on mouse position in showing.  The tooltip is reused in the accessibility applet for sticky keys in a vertical panel, so this may kill that, and what about key navigation without using the mouse ? Not tested this at all though.

jaszhix:
I tested the sticky keys feature in the accessibility applet, its still working. I'm not sure what feature you're referring to with the later. This just defines this.mousePosition when the constructor is initialized like the rest of the props. It doesn't get defined until an event handler is called, and that is causing an issues when code is deferred or asynchronous.

PR #6507, opened 2017-05-01
appletManager.js: don't error removing broken applet

itzexor:
Avoids an error when the applet is added and then immediately removed
upon failure.

error t=2017-04-30T21:28:42Z applet._extension is undefined
trace t=2017-04-30T21:28:42Z 
<----------------
removeAppletFromPanels@/usr/share/cinnamon/js/ui/appletManager.js:261:9
onEnabledAppletsChanged@/usr/share/cinnamon/js/ui/appletManager.js:209:17
---------------->
error t=2017-04-30T21:28:42Z Failed to refresh list of applets

### DIFF
--- a/js/ui/appletManager.js
+++ b/js/ui/appletManager.js
@@ -257,7 +257,8 @@ function removeAppletFromPanels(appletDefinition, deleteConfig) {
             applet._panelLocation = null;
         }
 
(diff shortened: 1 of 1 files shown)


itzexor:
didn't realize jaszhix already got this one in his series

PR #6511, opened 2017-05-05
checkBox.js: remove extra unused argument which was causing a warning…

collinss:
… with new cjs/mozjs

### DIFF
--- a/js/ui/checkBox.js
+++ b/js/ui/checkBox.js
@@ -32,7 +32,7 @@ CheckBoxContainer.prototype = {
         this.label = new St.Label();
         this.label.clutter_text.set_line_wrap(false);
         this.label.clutter_text.set_ellipsize(Pango.EllipsizeMode.NONE);
-        this.actor.add_actor(this.label, { y_fill: true, y_align: St.Align.END });
+        this.actor.add_actor(this.label);
 
         this._spacing = 0;
     },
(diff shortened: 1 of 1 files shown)


js/ui/checkBox.js
```diff
@@ -52,7 +52,7 @@ CheckBoxContainer.prototype = {
         let min = minBoxWidth + minLabelWidth + this._spacing;
         let nat = natBoxWidth + natLabelWidth + this._spacing;
         [min, nat] = node.adjust_preferred_width(min, nat);
```
In-line: This is unnecessary change

js/ui/checkBox.js
```diff
@@ -52,7 +52,7 @@ CheckBoxContainer.prototype = {
...
-        
+
```
In-line: This is unnecessary change.

leigh123linux: I don't see any issue removing white-space.

JosephMcc: I don't see the issue here either.

mtwebster: There's no issue with removing nearby line that is just whitespace.. we try to leave code cleaner than it was - so picking up the trash around the area you're working is ok.  It's the needless reformatting of swaths of code for no other reason than appearance that is frowned upon.

PR #6519, opened 2017-05-10
Use libnm >= 1.2, instead of deprecated libnm-glib

besser82:
This goes hand in hand with https://github.com/linuxmint/cinnamon-control-center/pull/173.


mtwebster:
There's way more that needs fixed in the applet than just these imports. I'll come back to it when I get a chance.   The build-time changes here can be completely removed (in configure and makefile) - we only need the gir package in debian/control for runtime dependencies (and remove any other old network-related deps there also)

Fantu:
@clefebvre you merged https://github.com/linuxmint/cinnamon-control-center/pull/173 but you missed to merge also this PR about the applet

mtwebster:
This completely breaks the applet in its current form, as mentioned above.

clefebvre:
Well the CCC PR seemed ready and it could be merged on its own.

jbicha:
@clefebvre This issue still exists. Should this PR be reopened or should someone open an Issue?

Fantu:
main job about is done in cinnamon-control-center https://github.com/linuxmint/cinnamon-control-center/commit/1966c6281cb7a8792a88edd40aa0b09f1c2fed0f
here is about the applet looking files in master is still present
mtwebster tell that broke the applet, these minimal changes in configure, makefile and import itself should not cause issue but api seems completly different: https://wiki.gnome.org/Projects/NetworkManager/Developers
In this case require major changes to the network applet code not present in this PR and not fast.

PR #6524, opened 2017-05-13
sound-applet: remove extraneous arguments on add_actor calls

itzexor:
With js38 these throw a warning because add_actor() takes only one argument,
the actor to add to the container add_actor() is called on. I don't believe
these ever had any effect, so I've removed them.

Quiets 5 of these on MPRIS player open with js38:
Too many arguments to method Clutter.Container.add_actor: expected 1, got 2


brownsr:
I would think these properties would want to be added with separate calls if this is done.

PR #6585, opened 2017-06-06
Progress bar text initializes and resets

binhonglee:
Existing behavior for installation progress bar text :
- Before installation : _no changes_ (_empty_ when initialized, stays at `100%` when continued from previous installation instead of reset to _empty_ or to `0%`)

Behavior after this change : 
- Before installation : initializes (or reset) to `0%` along with all other relevant components (`progressbar.set_fraction` and `progresslabel`)


JosephMcc:
It would be nice to see a description of what you are fixing as part of the actual commit message. It makes things more clear when you are looking at the history down the road.

Example: https://github.com/linuxmint/Cinnamon/commit/be31728a694a6c4c1522b32b464ee0c1265b0ec8

PR #6593, opened 2017-06-09
[WIP] xlets about dialog: move icon to top and center labels

NikoKrause:
screenshot-area-2017-06-09-151354

### DIFF
--- a/js/ui/modalDialog.js
+++ b/js/ui/modalDialog.js
@@ -451,22 +451,23 @@ SpicesAboutDialog.prototype = {
         this.uuid = metadata.uuid;
 
         let contentBox = new St.BoxLayout({vertical: true, style_class: "about-content" });
+        contentBox.set_style("text-align: center");
         this.contentLayout.add_actor(contentBox);
 
-        let topBox = new St.BoxLayout();
+        let topBox = new St.BoxLayout({vertical: true});
         contentBox.add_actor(topBox);
 
         //icon
         let icon;
         if (metadata.icon) {
-            icon = new St.Icon({icon_name: metadata.icon, icon_size: 48, icon_type: St.IconType.FULLCOLOR, style_class: "about-icon"});
+            icon = new St.Icon({icon_name: metadata.icon, icon_size: 48, icon_type: St.IconType.FULLCOLOR});
         } else {
             let file = Gio.file_new_for_path(metadata.path + "/icon.png");
             if (file.query_exists(null)) {
                 let gicon = new Gio.FileIcon({file: file});
-                icon = new St.Icon({gicon: gicon, icon_size: 48, icon_type: St.IconType.FULLCOLOR, style_class: "about-icon"});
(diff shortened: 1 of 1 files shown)


brownsr:
Would be grateful if you could test against some spices themes to give an idea how many will have to have changes made as a result. Hopefully none, but there may be some breakage.

Odyseus:
Hello, everyone.

@NikoKrause: I'm totally and absolutely against this change.

### Reasons

1. Reading centered text is absolutely annoying and it doesn't fit with anything in Cinnamon's UI.
2. Expanding the elements vertically is against "widescreen styling guidelines". This change makes a lot of vertical space usage, and it wastes a lot of horizontal space.
3. Piling up all those elements steals vertical space from the scroll box at the bottom.

NikoKrause:
Maybe something like this?
screenshot-area-2017-06-10-105533
Right now it looks like this:
screenshot-area-2017-06-10-105948

The scrollbox is a little bit smaller, but I think having the icon only together with Title,UUID,Timestamp but not with description looks better.

NikoKrause:
@brownsr 
I tested TOP 10 spices themes. Looks good with most. Only issue I see, is that the scrollbox with some themes is too small, but those themes look ugly with the default about design as well, e.g.:
screenshot-area-2017-06-10-112936

Odyseus:
Hello, @NikoKrause.

This one...

This one

...is definitely an improvement (from my point of view, of course). It separates the short description from the icon, title, etc. and allows for the short description to have a little more horizontal space. Plus, the icon, title, etc now looks infinitely better (again, from my point of view) without the short description expanding the containers height. I would risk to say that this is exactly the change the about dialog for xlets needed.

If you don't want to update your pull request with this appearance, could you share the **modalDialog.js** file that produces the about dialog shown on the picture above? I would like to test it with several themes.

About themes that may need to change. The ones that needs immediate changes are the default **Mint-X** family of themes and some of its derivatives. In these themes, the about dialog is very small by default, even without applying the changes proposed here. Look image bellow.

about

In my own fork of the **Mint-X** theme, [I already expanded the about dialog size](https://user-images.githubusercontent.com/3822556/27004547-1b955b04-4de1-11e7-88c2-aeba40fe4f47.png).

NikoKrause:
@Odyseus 
You habe to add_actor(desc) into contentBox instead of topTextBox. That's all.

You might want to look into the changes of Mint-Y and Mint-X themes, which JosephMcc did yesterday.
He did some refinements to the about dialog through themeing. Looks good to me.

JosephMcc:
Here is a quick screenshot of what they look like in Mint-Y now. The look in Linux Mint and Mint-X was updated to a similar layout.
selection_011

The only things I would like to see improved:
* Move the icon up and inline with the box that holds the title and uuid.
* Remove the "Comments" label. No reason for it
* Hardcode some of this layout. It shouldn't be so easy to break in the theme and the theme really only needs control over styles and colors.

NikoKrause:
Closing this. See new PR https://github.com/linuxmint/Cinnamon/pull/6608

PR #6596, opened 2017-06-09
ChooserButtonWidgets.py: Fix Python 3 error

jaszhix:
Lack of parenthesis is preventing xlet-settings from starting.

```sh
  File "xlet-settings.py", line 9, in <module>
    from JsonSettingsWidgets import *
  File "/usr/share/cinnamon/cinnamon-settings/bin/JsonSettingsWidgets.py", line 2, in <module>
    from SettingsWidgets import *
  File "/usr/share/cinnamon/cinnamon-settings/bin/SettingsWidgets.py", line 15, in <module>
    from ChooserButtonWidgets import *
  File "/usr/share/cinnamon/cinnamon-settings/bin/ChooserButtonWidgets.py", line 167
    print message
                ^
SyntaxError: Missing parentheses in call to 'print'
```

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/ChooserButtonWidgets.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/ChooserButtonWidgets.py
@@ -164,7 +164,7 @@ class PictureChooserButton(BaseChooserButton):
         if pixbuf:
             self.button_image.set_from_pixbuf(pixbuf)
         else:
-            print message
+            print(message)
             self.button_image.set_from_file("/usr/share/cinnamon/faces/user-generic.png")
 
     def set_button_label(self, label):


Odyseus:
Hello, @jaszhix.

I wonder if the error that you get isn't triggered by not having Python 2 installed in your system. Because the **ChooserButtonWidgets.py** file *shebang* isn't specified, but the use of `from __future__ import division` (among other things) suggests that it expects Python 2.

I also wonder if it isn't time to move all Python code to Python 3 and let Python 2 RIP (I can already feel the headaches of the g@ys in charge of that. LOL).

mtwebster:
@Odyseus xlet-settings.py is a python3 shebang, cinnamon-settings is python2, but both are sharing some other files :).  Hopefully we can get everything uniform on 3 during the next dev cycle.

Odyseus:
Hello, @mtwebster.

Thanks for the clarification (my *newbieness* reveals itself). I didn't considered the *"inheritance"* situation.

> Hopefully we can get everything uniform on 3 during the next dev cycle.

Nice to know that. For a newbie like myself, sometimes Python 2 and 3 feels like two completely different languages.

PR #6606, opened 2017-06-11
panel: avoid cjs giving a warning that dummyPanels[i][j] is undefined

brownsr:
Checked out on single monitor, and when adding and removing a second monitor


lestcape:
Short:
```
while (this.dummyPanels.push([true, true, true, true]) < monitorCount);
```

brownsr:
Thanks @lestcape that's definitely neater.

PR #6608, opened 2017-06-12
xlets-about-dialog: improvements for about dialog

NikoKrause:
* The version number shows up only, when the timestamp is not shown/does not exist. So move the version number to where the timestamp is, i.e. under the uuid. -> Remove `about-version` style_class.
* Move the icon up and inline with the box that holds the title, uuid and timestamp/version.
* Remove the "Comments" label.
* Add tab before website link like it's done for contributors.

left column is new, and right column is old, i.e. how it looks right now:

screenshot-area-2017-06-12-124000

### DIFF
--- a/js/ui/modalDialog.js
+++ b/js/ui/modalDialog.js
@@ -475,19 +475,8 @@ SpicesAboutDialog.prototype = {
         topBox.add_actor(topTextBox);
 
         /*title*/
-        let titleBox = new St.BoxLayout();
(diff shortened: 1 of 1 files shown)


NikoKrause:
@clefebvre This should be merged into 3.4 imo. It's not only an improvement, it's a bugfix: compare last applet in the screenshot above. The description is currently not readable.

PR #6621, opened 2017-06-16
[3.6] Make cinnamon-settings unaware of installation location

Cobinja:
This makes cinnamon-settings independent from the installation location,
making it easier for developers, because you don't have to install
Cinnamon for testing during development. You can just run it from where
it is.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_backgrounds.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_backgrounds.py
@@ -20,7 +20,8 @@ import gi
 gi.require_version("Gtk", "3.0")
 from gi.repository import Gio, Gtk, GObject, Gdk, Pango, GLib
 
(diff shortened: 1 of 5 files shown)


Cobinja:
Removed WIP-tag

files/usr/share/cinnamon/cinnamon-settings/modules/cs_backgrounds.py
```diff
@@ -20,7 +20,8 @@
 gi.require_version("Gtk", "3.0")
 from gi.repository import Gio, Gtk, GObject, Gdk, Pango, GLib
 
-sys.path.append('/usr/share/cinnamon/cinnamon-settings/bin')
+curPath = os.path.dirname(os.path.abspath(__file__))
```
mtwebster: why don't we use config.currentPath here?

Cobinja:
Sry, missed that during development.
- rebased from current master
- fixed a leftover reported above by @mtwebster 
- squashed into one commit

PR #6647, opened 2017-06-25
[3.6] st-theme-node-drawing: correct logic for calculation of cover and con…

brownsr:
…tain

fixed upstream a long time ago https://github.com/GNOME/gnome-shell/commit/3dbf06420df0ad669f12939a788ed44061dcc1e1


brownsr:
How to test/reproduce problem: was causing grey bars in tall pop-up menus, theme 'Photocopy', particularly apparent when expanding a sub-menu.  Example: right click panel menu, expand 'Modify panel' sub-menu.

PR #6651, opened 2017-06-26
panel-launchers: fix undeclared variable warning

itzexor:
The original declaration was recently removed, add it back here.


itzexor:
I always do these quick fixes before even checking if someone else already has, oops...

PR #6671, opened 2017-07-03
Added header and inline vanilla screenshot

milesflo:
Small change. Adds a large header with the project name, and after the subhead includes a screenshot of the vanilla desktop.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,4 +1,9 @@
+====
+Linux Mint Cinnamon
+====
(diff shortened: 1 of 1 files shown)


NikoKrause:
Cinnamon is not bounded to Linux Mint. There are a lot of Distros using Cinnamon: Arch, Fedora, etc.

And the screenshot is old. The menu shows a boxpointer. Those have been removed.

PR #6690, opened 2017-07-09
windows-quick-list: wrap menu in scrollable sub-menu container

### LINKED ISSUE #2491: Windows Quick List: Add scrollbar if list doesn't fit on screen
I can't access all windows from applet windows quick list if the list doesn't fit on screen. Please add a scrollbar. A close button for each window in the list would be great too.
quick-windows-list

jaszhix:
This wraps the menu items in a PopupSubMenuMenuItem menu. PopupSubMenuMenuItem was modified so its label and Clutter.Point can be skipped in its _init if no label text is passed to it. The classes for PopupSubMenuMenuItem's actor and menu actor are reset in the applet's _init method, so the applet retains its proper theming.

Closes #2491


mtwebster:
This works until the window list becomes tall enough to require a scroll bar, or until you remove enough windows to the point where one is no longer required.  At either of those points, however, the applet breaks.  There's nothing in the logs either - the only thing I did see, but it was only once, was:

```
Cjs-Message: JS LOG: The property _arrowRotation doesn't seem to be a normal object property of [0x562b1f9c22f0 StScrollView:last-child] or a registered special property
```

Not sure that's related or not.

jaszhix:
I added checks in ```on_applet_clicked``` and ```activateWindow``` for the sub menu container's open state. The problem could be the on_applet_clicked function get invoked multiple times, so the sub menu is opening and closing before the parent menu has a chance to open, so it renders a hidden actor.

In PopupSubMenuMenuItem ```this._triangle``` is set to null when its being used as a generic scroll-able container, as PopupSubMenu has all the checks needed for it to work. Another check in PopupSubMenu's _init method was added that ensures its _arrow property doesn't get assigned in this scenario. I didn't see the warning, but I'm assuming the binding API doesn't like null properties.

PR #6694, opened 2017-07-12
Make touchpad settings compatible with libinput

AlbertJP:
Requires linuxmint/cinnamon-settings-daemon#181.

The scrolling work is compatible with either driver. Clickpad support works differently: synaptics does not disable button emulation in any case, at least on the laptop I have at hand.

If you need a reminder why libinput makes sense: see https://fedoraproject.org/wiki/Changes/LibinputForXorg
TL;DR other distros are moving towards it, and it can handle a number of modern input devices better. The synaptics driver was for instance not built for laptops where buttons need to be emulated and its default (and only) way of handling those devices causes issues for some (especially Lenovos).

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_mouse.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_mouse.py
@@ -94,7 +94,7 @@ class Module:
             switch = GSettingsSwitch(_("Disable touchpad while typing"), "org.cinnamon.settings-daemon.peripherals.touchpad", "disable-while-typing")
             settings.add_row(switch)
 
-            clickpad_list = [[0, _("Left click only")], [1, _("Emulate mouse buttons")], [2, _("Use multiple fingers for right and middle click")]]
+            clickpad_list = [[0, _("Left click only")], [3, _("Automatic")], [1, _("Emulate mouse buttons")], [2, _("Use multiple fingers for right and middle click")]]
 
             combo = GSettingsComboBox(_("Clickpad click action:"), "org.cinnamon.settings-daemon.peripherals.touchpad", "clickpad-click", clickpad_list, valtype="int")
             settings.add_row(combo)
@@ -105,7 +105,7 @@ class Module:
             switch = GSettingsSwitch(_("Reverse scrolling direction"), "org.cinnamon.settings-daemon.peripherals.touchpad", "natural-scroll")
             settings.add_row(switch)
 
-            clickpad_list = [[0, _("No scrolling")], [1, _("Two-finger scrolling")], [2, _("Edge scrolling")]]
+            clickpad_list = [[0, _("No scrolling")], [3, _("Automatic")], [1, _("Two-finger scrolling")], [2, _("Edge scrolling")]]
             combo = GSettingsComboBox(_("Scrolling method:"), "org.cinnamon.settings-daemon.peripherals.touchpad", "scrolling-method", clickpad_list, valtype="int")
             settings.add_row(combo)
             switch = GSettingsSwitch(_("Horizontal scrolling"), "org.cinnamon.settings-daemon.peripherals.touchpad", "horizontal-scrolling")
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_mouse.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_mouse.py
@@ -94,15 +94,8 @@ class Module:
             switch = GSettingsSwitch(_("Disable touchpad while typing"), "org.cinnamon.settings-daemon.peripherals.touchpad", "disable-while-typing")
             settings.add_row(switch)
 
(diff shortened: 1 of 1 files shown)


clefebvre:
I don't mind improving this and I don't even mind considering a dependency on libinput (i.e. that would result in basically always using that driver), but no matter happens, things have to work well out of the box.

Here are my findings on a Macbook Pro:

with LibInput:

- two-finger scrolling works by default but can't scroll with one finger only
- right-click is done by clicking invisible right part of touchpad, two-finger emulates middle-button, so does 3finger click
--> works if clickpad action is set to use multiple fingers.

with Synaptic:

- two-finger scrolling works by default but can't scroll with one finger only
- can't right-click, can't middle-click
--> works if clickpad action is set to use multiple fingers.

As you can see the default situation isn't acceptable with synaptic.

Losing the ability to edge-scroll is OK if and only if selecting 2-finger-scrolling by default works on touchpads which don't support it (i.e. if it falls back to edge-scrolling for these).

I'll have to test that with another laptop to make sure.

leigh123linux:
@clefebvre  Maybe defer this till Mint rebases to the next LTS then remove synaptics support completely?

clefebvre:
Well, we don't really need to wait. We just need to figure out how to make things work for the most people out of the box.

AlbertJP:
@clefebvre, thanks for the review. Your Mac does right clicks differently than my ThinkPad X1 Carbon, so it's good to check the behaviour on multiple laptops. On my laptop Synaptics emulates the bottom mouse buttons in any case.

You're right that it is worth testing the code on a laptop that lacks 2-finger scroll support, but those are probably hard to find. My previous laptop, a 2006 ThinkPad supports it (the touchpad is, however, pretty bad at detecting 3-finger gestures). I'll see if I can find another old laptop to check the touchpad behaviour.

AlbertJP:
Using Mint 17, I can confirm that on the HP laptop, synaptics does _not_ fall back to edge scrolling if I set the requested configuration options manually (that was already reported the previous time we introduced a combobox for scrolling: https://github.com/linuxmint/cinnamon-settings-daemon/commit/1eba56f1e89920c7cf5125ad9a21c704a48baf91).
It turns out the laptop has a number of hardware issues including a faulty

AlbertJP:
I managed to boot the dv4245EA using Plop Boot Manager and the USB port was - by sheer luck - working stably for long enough to nearly install Mint 18 on the machine, even though dmesg was full of hardware errors. I only had to do a grub install afterwards as the installer locked up when installing language packs. Will try installing libinput later today.

AlbertJP:
Test results with the HP Pavilion dv4245EA (2005), touchpad: Alps Glidepoint

Right-click method: makes no difference, since the device has physical buttons.

Libinput:
- edge scrolling by default

Synaptics driver:
- no scrolling by default, can be set to edge scrolling

Your laptop has issues with the default right-click mode, this old one with the default scrolling mode when Synaptics is used. I am thinking of adding a mode: Automatic or Default, which will use the default values of libinput (which are quite sensible and depend on the device's capabilities) and will enable all methods in synaptics as is currently default in Mint 18.

AlbertJP:
I have just implemented an automatic mode in c-s-d which takes libinput's default & enables all scroll & click methods on synaptics. Please test with your Mac, @clefebvre. It will probably not be until next week that I have access to the old HP again.

clefebvre:
So...

This new automatic mode works perfectly on the macbook, with both drivers.

clefebvre:
On an old HP laptop, automatic works perfectly with synaptics... but I get no scrolling at all with libinput.

I think that touchpad only supports edge scrolling, so when scrolling-method is set to automatic and libinput is installed, we get no scrolling at all.

I know libinput is supposed to be better... but that kinda sucks.

clefebvre:
With libinput in automatic mode, could we maybe detect the capabilities of the touchpad? If it doesn't support two-finger scrolling, we should set the scrolling method to edge-scrolling.

clefebvre:
ok, don't mind the previous comment...

There is a "libinput Scroll Methods Available" property, but I just found out the touchpad supported two-finger scrolling.

It's just a really bad touchpad and it barely works... if your two fingers are too close, it just sees them as one finger and moves the cursor :)

OK, so anyway...

Long story short, the code is fine and works well with synaptic in auto mode.

We then need to decide whether to ship with libinput or not in the 18.3 Cinnamon edition, and it looks like we might be able to :)

Thanks for your patience and your work @AlbertJP!

AlbertJP:
> There is a "libinput Scroll Methods Available" property, but I just found out the touchpad supported two-finger scrolling.
>
> It's just a really bad touchpad and it barely works... if your two fingers are too close, it just sees them as one finger and moves the cursor :)

I'd like to be able to measure how comfortable a touchpad works :) the Alps Glidepoint which I encountered in an HP laptop wasn't exactly a nice touchpad either, but libinput correctly detects it capabilities. You could file a bug report with libinput to see if the default can be changed or two-finger detection improved, they appear to be quite helpful with issues like these.

Issue #6706, opened 2017-07-17
Feature Request: Better Method for Updating Spices


mfreeman72:
```
 * Cinnamon version 3.4.4
 * Distribution - Mint 18.2
 * Graphics hardware *and* driver used: AMD A8 (Radeon R5) and Open Source Mesa driver
 * 64 bit
 ```

**Issue**

Feature Request: Better Method for Updating Spices

It is currently a relatively involved process to keep Applets, Desklets, Extensions, and Themes updated. I would like to see any one of the following methods implemented to maintain an up-to-date feature set:

1. Test and then package up all of the Spices into .deb files in the Mint repositories, according to compatibility with Cinnamon version included in each supported Mint version. Any new updates get tested and updated in the repos so that they are updated through the standard packaging system. Users can have the option of installing and maintaining them system-wide through this method, or installing and maintaining them on a per-user-account basis using the current method. This packing can be done with volunteers, if necessary. Or perhaps run a community repository through the LM community website.

2. Include a section of the Mint Update tool that monitors the Spices website and offers updates to those that are installed on the currently logged-in user account, as well as an optional auto-update feature.

3. Create a Spices Update tool that runs on a per-user-account basis, rather than system-wide, that monitors the Spices website and offers updates to those that are installed on the user account, as well as an optional auto-update feature.

4. Have Cinnamon itself optionally check the Spices website for updates and automatically update them on a user-selected time interval.

JosephMcc:
We have other open issues requesting this same thing. I'm going to close some of those and linked them to this one since it's a bit nicer.

This is something that gets discussed internally though I'm not sure if any decisions have been made. The solution should be cross distro though since Cinnamon is not limited to Mint or deb based distros in general.

mfreeman72:
@JosephMcc Thank you, and I apologize. I did a search (although the search feature is not very intuitive on github), and did not see these. I'll be more careful next time.

icarter09:
Many thanks for submitting your feature request for Cinnamon. Your feature request has been documented and will be reviewed by the Linux Mint team.

For more information on our workflow and feature requests, read https://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/faq.html.

PR #6708, opened 2017-07-18
ui/modalDialog: Show the version when time stamps are present

### LINKED ISSUE #6705: About menu of applet does not show version
```
 * Cinnamon version 3.4.4
 * Distribution - Mint 18.2
 * Graphics hardware *and* driver used - Intel with Intel driver
 * 64 bit
 ```

**Issue**
The About menu of the applet Icing Task Manager does not show the version.

Jason Hicks author of the applet Icing Task Manager told me to open an issue here rather than on Cinnamon Spices. 

I said: "The About menu of Icing Task Manager should contain the current version number. At the moment it says: 2017-07-07 08:25:06 UTC which doesn't help much as when I updated from an older version the date did not change."

Jason said: "They changed this in Cinnamon. Please open an issue there."
https://github.com/linuxmint/cinnamon-spices-applets/issues/1035

**Steps to reproduce**
Go to About menu of Icing Task Manager

**Expected behaviour**
Shows a version

**Other information**

jaszhix:
Partial revert of 7f019cf55c921138c334e8c9881448cc05bdd4b9
Closes #6705

### DIFF
--- a/js/ui/modalDialog.js
+++ b/js/ui/modalDialog.js
@@ -492,12 +492,12 @@ SpicesAboutDialog.prototype = {
 
             let lastEdited = new St.Label({text: dateUTC, style_class: "about-uuid"});
             topTextBox.add_actor(lastEdited);
-        } else {
-            //version
-            if (metadata.version) {
-                let version = new St.Label({text: "v%s".format(metadata.version), style_class: "about-uuid"});
-                topTextBox.add_actor(version);
-            }
+        }
+
+        //version
+        if (metadata.version) {
+            let version = new St.Label({text: "v%s".format(metadata.version), style_class: "about-uuid"});
+            topTextBox.add_actor(version);
         }
 
         //description


NikoKrause:
I would remove metadata.version completely.

Odyseus:
Hello, everybody.

@NikoKrause: The *"version"* field on the about dialogs should never have been hidden/removed. The *"last-edited"* field is completely useless for feedback/debugging purposes.

For example, if an applet cannot be loaded into a panel (or a desklet cannot be loaded into the desktop), its about dialog can't be accessed. With extensions is worse, because they lack about dialogs all together. By looking at the **metadata.json** file, the *"last-edited"* field is a *non human-readable* number which we (the xlets developers) are forced to convert into a *human-readable* format for it to make sense. And even with that *human-readable* "version" at hand, it's still completely useless, because we (at least I) keep the **metadata.json** files *clean* (with the *"last-edited"* field and other fields removed, otherwise, we are bothered to remove them from our pull requests). And without something to compare to the "version" provided by the *feedback giver*, trying to determine the version used by him/her will simply be a soothsaying act!

Like I said, the *"version"* field should never have been hidden/removed. If it exists, **display it always**.

NikoKrause:
The last-edited timestamp gets added automatically if a users installs an applet through Cinnamon. So it's always there. You can convert the timestamp to readable UTC form, e.g. with online converters. Then you have the date of the commit the user is running and can look it up in the commit history. So it's pretty helpful, even when the About dialog does not appear.(which should happen very rarely anyway)

jaszhix:
The ultimate question is, do we want to deprecate the version property? I don't think we should, semantic versioning is one of the most basic things that can be available to xlets, and software in general. Its good to have an exact time stamp, but Odyseus makes a good point that it adds extra effort to cross reference it to find the actual commit. If an xlet is updated twice in one day, especially so. Its nice that the time stamp adds additional information, and is used internally by Cinnamon, but for figuring out if someone is on the latest version, we can't know by just looking at it.

collinss:
@jaszhix applet versioning was never really an official thing as far as Cinnamon is concerned. When I first wrote the about dialog, I included it there because a handful of applet devs (including myself) were including a version in the metadata. It was a way for the few of us who used it to more easily determine whether the user was up to date, at a time when there wasn't really a good way to do so otherwise. Most of the applets still don't include that information (I picked about 5 or 6 at random and none of them did). Now we have a system that not only is more reliable, but also works for all applets, without the need for the author to do anything. And since we are talking about official, the official version scheme for applets is now to use the timestamp, since each commit constitutes a new version. Finally, versioning of the 'traditional' form (eg. 1.2.3) really only makes sense for software that has periodic releases. For rolling releases (such as betas, nightlies, etc.) which the applets more closely resemble, it is more traditional to use a timestamp, or a modified version thereof.
I think a much better way of handling the issue would be to add a button to the about dialog which would copy all the relevant information (ie. uuid, timestamp, author, etc) to the clipboard.

jaszhix:
I like the idea about copying metadata to the clipboard. Really, I'm just trying to preserve the ability for people to share what version they're on so applets can be easily debugged. If I start receiving dates in issues, that will create extra work for me. I still don't understand what's "official" or not. It seems like if something gets merged into Cinnamon, regardless of how it ended up there, it then becomes official.

NikoKrause:
> The "last-edited" field is completely useless for feedback/debugging purposes.

> trying to determine the version used by him/her will simply be a soothsaying act!

Hello and welcome to FAQ about timestamps and how to find out if the user is running the latest version of my applet.

1. _Q: The timestamp is useless. How does it tell me, which version the user is running?_
    A: For example the user is running Icing Task Manager (ITM) and the user is telling you that the About dialog of his ITM applet says `2017-07-18 14:15:06 UTC`. Open the [cinnamon-spices website of ITM](https://cinnamon-spices.linuxmint.com/applets/view/269) applet, e.g. by pressing the `More info` button in the about dialog. There you click on the [last-commit](https://github.com/linuxmint/cinnamon-spices-applets/commits/master/IcingTaskManager@json) link. Now you see the commit history of ITM only.
screenshot-area-2017-07-22-105644
You'll see, the user is running the second latest version. There is already a newer version from July 19, to which the user can update.
2. _Q: But there are more than 1 commit on the same date. How can I be sure, which version the user is running?_
    A: Hover over the date under the commit message and there'll appear a tooltip with the full date:
screenshot-area-2017-07-22-110925
screenshot-area-2017-07-22-110938
3. _Q: But the timestamp on the GitHub website in you screenshot doesn't say UTC it says GTM. How am I supposed to know which time it is in UTC?_
     A: The interesting part in the date `12 Mai 2017, 22:38 GMT+2` is the `+2` at the end. Subtract two hours from the date and you get the UTC, so it's `20:38 UTC`.
4. _Q: But there are 7 commits in less than 5 minutes and I noticed that the timestamps shown in the about dialog and those shown on the website are +-10 minutes difference. How am I supposed to know, which of those commits the user is running, for sure?_
      A: Nice catch. The cinnamon spices website is updating every 5 minutes. So the minutes of the timestamps in the about dialog are divisble by 5, i.e. "21:05, 21:10, 21:15, 21:20, etc."
          So if there are e.g. 3 commits at `20:11`, `20:12` and `20:13`, then there is no chance the user is running the first two commits, since the website updates at `20:10` and `20:15` and therefore the first two commits are never available through Cinnamon settings.
5. _Q: But in question 4 I can't be 100% sure, if the user is running the very latest version, can I?
      A: If this case really happens, that there are 13 commits in 13 mintues and you want to be 100% sure the user is running the very latest version, then download the latest version of the applet on your own and compare the  your about dialog timestamp with the users timestamp._
6. _Q: But the user says, that he cannot add the applet to the panel and therefore he cannot open the about dialog. How am I supposed to read the timestamp?_
      A: Ask the user to send you the `last-edited` line in the metadata.json file, e.g. by using the terminal:
      ```
        grep last-edited ~/.local/share/cinnamon/applets/printers@linux-man/metadata.json
      ```
      And the output is:
      ```
      "last-edited": 1499365506,
      ```
     You can convert this by an online timestamp converter or by using the terminal command:
      ```
      date -u -d @1499365506
      ```
     Which gets you:
      ```
       6. Jul 18:25:06 UTC 2017
      ```
7. _Q: Extensions lack about dialogs all together. How am I supposed to know, which version the user is running?_
     A: That's no argument for using metadata.version. If you can't access the about dialog, you can't see the timestamp, but you can't see the metadata.version either. It's a tie! To get the timestamp anyway see question No. 6.
8. _Q: The about dialog of my ITM applet says `Last published: 2017-05-10 14:15:06 UTC` But I can clearly see on the website, that there have been a few updates of ITM in June and July. Why does my about dialog says, last published version of ITM is in May, when I can clearly see that the latest version of ITM is on July?_
      A: I think it's just a bad wording. It should supposed to mean, that you are running the version published on 10 May.

jaszhix:
I rebased my branch and removed the date identifier.

My point is ordinary users shouldn't have to go to Github just to find out if they're on the latest version. Most's won't, I think. Look at the amount of text you have to write just to explain this to people. Its too complicated for a version. What we need is a boolean value, and maybe it could tie into xlet-settings - where Cinnamon knows whether or not each applet is up to date. They click the About dialog and it says "Up to date". If not, it says "An update is available". I don't think there's any use for the time stamp if that were done.

NikoKrause:
Oh. The FAQ is not meant for the users. It's meant for the devs.
Ordinary users don't need to go to GitHub to find out, if they are running the latest version. Ordinary users do refresh the list and then see available updates through Cinnamon settings.
And some ordinary users don't have a clue, that they need to refresh the list. Those users go to GitHub anyway to report an issue. Then you need to tell them to update, based on their timestamp.

jaszhix:
Even worse, I don't have time for that much cross referencing because we don't have better update logic in Cinnamon. Its not fair to people who are already doing all the work. I'm going to leave my PR as is, and bow out of the discussion, as I don't think such a small reversal is worth this much conversation unless it turns into a conversation on how to make this more streamlined for everyone.

NikoKrause:
I just wanted to show, that it's not soothsaying to say which version the user is running using timestamps.

I'm not disagreeing with you, that checking for up-to-date version couldn't or shouldn't be improved.
I'm disagreeing with you to use metadata.version for that. Because not every developer uses/updates the metadata.version. Hence it's unreliable and inconsistent. Timestamps are autogenerated, consistent for all spices and updated for each commit. (I know you and Odyseus are updating it every time, but you are the exception and not the norm. A solution for this, should be something, that is consistent and reliable for ALL spices)

Improving this, shouldn't rely on showing the metadata.version.
There are a few things, which could be improved, as you said.
1. Showing the latest timestamp on the spices website as suggested here: https://github.com/clefebvre/cinnamon-spices/issues/14
2. Adding a "Check for update" in the About dialog.
3. Creating an update notifier for spices, as you mentioned in the past. 

A cool feature would be, if the user gets notified about an update and where he gets an offer to look at the changelog, which shows him the changes between his current version and the latest version.

collinss:
@jaszhix there already is an easy way for the user to tell if they're up to date: open applet/desklet/extension/theme settings, go to the download page, hit the refresh button. The problem is that currently, the user has to update the cache to tell if there is an update. If we implement an auto-refresh of the cache, it will be easier.

jaszhix:
@NikoKrause The version being there does not affect any other part of the UI. Its just making the UI inconsistent for people when the version disappears after they change the install method.

@collinss There has been interest in a way to do this so its more convenient. Such as #6706.

NikoKrause:
I looked at LineageOS. They have nightly builds and show the version number as well as a build date.
Also versions are more common than timestamps. So I created a PR https://github.com/linuxmint/Cinnamon/pull/6808 here, to clarify what the timestamp means and called it `Build date`. And added the version back in to show always if exists.
There are some valid points, the one that convinced me is, that by a version number it's more obvious if it's a bigger change or a minor change of the applet. You don't see this by timestamp.

Also to make the timestamp even more useful, I asked the team of the spices website, if they can show the latest timestamp in the website as well. So it makes more easy to compare the timestamp by hitting the `More info` button in the about Dialog.

jaszhix:
Looks good, I'll close this since #6808 addresses it more thoroughly.

Issue #4374, opened 2015-06-27
Translation: Some strings need to be fetched with ngettext


gunchleoc:
Please fetch the following strings with ngettext, so that they can be translated properly:

```
 (%s seconds ago)
```

Located in: files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js:314

This one:

```
 (%s minute ago)
```

Located in:  files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js:317

Can be combined with:

```
 (%s minutes ago)
```

Located in:  files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js:320

This one:

```
%d updates available!
```

Located in: files/usr/lib/cinnamon-settings/bin/ExtensionCore.py:1126

Can be combined with:

```
%d update available!
```

Located in:  files/usr/lib/cinnamon-settings/bin/ExtensionCore.py:1128

ManIVIctorious:
Is this still an issue in cinnamon 3.0?

gunchleoc:
Yes, number ranges are still hard-coded in https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/applets/notifications%40cinnamon.org/applet.js#L318

NikoKrause:
PR created to fix this issue

PR #6710, opened 2017-07-19
Translation: fetch some strings with ngettext

NikoKrause:
fixes https://github.com/linuxmint/Cinnamon/issues/4374

### DIFF
--- a/files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js
@@ -11,6 +11,7 @@ const MessageTray = imports.ui.messageTray;
 const Urgency = imports.ui.messageTray.Urgency;
 const NotificationDestroyedReason = imports.ui.messageTray.NotificationDestroyedReason;
 const Settings = imports.ui.settings;
+const Gettext = imports.gettext.domain("cinnamon-applets");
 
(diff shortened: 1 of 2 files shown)


files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js
```diff
@@ -314,13 +315,12 @@ function timeify(orig_time) {
             str += _(" (Just now)");
             break;
         case (diff > 15 && diff <= 59):
-            str += _(" (%s seconds ago)").format(diff.toString());
+            diff_seconds = diff.toString();
+            str += Gettext.ngettext(" (%s second ago)", " (%s seconds ago)", diff_seconds).format(diff_seconds);
```
gunchleoc: I think we can save a line here, because ngettext takes a number:

    str += Gettext.ngettext(" (%s second ago)", " (%s seconds ago)", diff).format(diff.toString());

And maybe we can avoid calling toString() manually be using the  % notation below, in combination with %d?

NikoKrause: Changed

gunchleoc: Thanks, looks good to me now :)

Issue #4833, opened 2015-11-16
The notification that the applets list has been refreshed is sometimes too fast to see 


CottonEaster:
(. . although at other times it times out and gives an error).

Go to system settings. Click 'applets'. Select 'available applets (online). Click 'refresh'. Then you may see a brief flash. That is the notification - that one does not get time to read - that the list has been refreshed.

What should happen: one should have time to read the notification.

Also: I don't really see the point of the animation between the installed applets and the ones available online, especially since (1) sometimes that animation stalls a little on my system, half way through, (2) the animation does not seem under the control of any animation settings.

Vahan86:
@ACinnamonUserInUK, is this still an issue for you in Cinnamon 3.0?

ACinnamonUserInUK:
@Vahan86: I am afraid so, yes.

I am on Cinnamon 3.0.7.

NikoKrause:
The applets settings have been changed in Cinnamon 3.6

Issue #3728, opened 2014-12-04
Different mouse cursor when navigating online extensions, desklets, and applets


raidermax:
I am using Cinnamon 2.4.5 and whenever I go into System Settings > Applets or Extensions or Desklets and go into the tab that allows me to view and install online applets, extensions, or desklets, the mouse cursor changes from my default cursor to a very weird looking one.

I would post a screenshot but whenever I do take a screenshot, the pointer is of the cursor that I like and not the strange one that only appears in the places that I listed above. The pointer is rather skinny and is tilted id say 45 degrees to the right.

raidermax:
This is still occurring on the latest Cinnamon release.

dalcde:
What cursor theme are you using? Also you might try taking a photo of your computer monitor.

Vahan86:
@raidermax, is this still an issue for you in Cinnamon 3.0.x?

Odyseus:
Hello, everybody.

@Vahan86: This happens to me too. But I don't think that it is a problem with Cinnamon. I used see this *strange* cursors in certain parts of XFCE too. It seems to only happen with cursor themes created for KDE.

This image is of the [Capitaine](https://www.gnome-look.org/p/1148692/) cursor theme on the **Available applets (online)** section:

![cursor-capitaine-on-applets-window](https://cloud.githubusercontent.com/assets/3822556/18817475/66adc95c-8337-11e6-883f-558cf645bc0a.png)

And this is an image of the Breeze-AIO cursor theme (don't remember where I got this theme) on the same window:

![cursor-breeze-on-applets-window](https://cloud.githubusercontent.com/assets/3822556/18817477/76209d1a-8337-11e6-92b7-94e4a291f259.png)

When the correct cursor should look like this:

![cursor-capitaine-correct](https://cloud.githubusercontent.com/assets/3822556/18817478/8022b8ac-8337-11e6-9ba3-1137b9510972.png)

With the Adwaita cursor theme the **Available applets (online)** window displays the correct cursor. My guess is that there must be a missing file inside the **cursors** folder of the cursor themes that we are using. This doesn't bother me that much, so I never investigated the issue further.

JosephMcc:
@Odyseus Thanks for the additional info.

Vahan86:
@Odyseus, thanks for info.

NikoKrause:
I investigated this issue a little bit.
[...]

PR #6714, opened 2017-07-21
cinnamon-settings <spices>: rewrite to simplify code and implement a …

collinss:
…more modern ui

### DIFF
--- a/data/org.cinnamon.gschema.xml.in
+++ b/data/org.cinnamon.gschema.xml.in
@@ -21,7 +21,15 @@
         override version check
       </_description>
     </key>
-    
+
+    <key name="extension-cache-updated" type="i">
+      <default>0</default>
+      <summary>Timestamp of last cache refresh</summary>
+      <description>
+        The timestamp of the last time the index cache was updated from the spices website
+      </description>
+    </key>
+
     <key name="enabled-applets" type="as">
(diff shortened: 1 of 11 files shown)


NikoKrause:
Maybe closes https://github.com/linuxmint/Cinnamon/issues/4833

NikoKrause:
Fixes https://github.com/linuxmint/Cinnamon/issues/3728

Odyseus:
Hello, @collinss.

I have been testing this pull request for various days now. I had a list of issues that I wanted to report, but almost all of them have been fixed in subsequent commits.

- When trying to uninstall an xlet from the **Download <xlets>** tab, the following error is thrown when clicking **Yes** to proceed with the xlet removal.

```shell
  File "/usr/share/cinnamon/cinnamon-settings/bin/ExtensionCore.py", line 890, in uninstall
    self.disable_extension(extension_row.uuid, extension_row.name, 0)
AttributeError: 'DownloadSpicesPage' object has no attribute 'disable_extension'
```

I changed that line to the following now it seems to be working as expected.

```python
self.spices.disable_extension(extension_row.uuid)
```

Odyseus:
@collinss.

I found a minor issue with the recently implemented **About** button. The about dialog opened from the xlets manager windows never uses the icon file defined for the xlet (icon.png) because the `metadata` object passed to the `SpicesAboutDialog` prototype doesn't contain a `path` key. So, it always uses the generic `cs-xlet` icon.

I don't think that it is possible to *"fix"* this. I just brought this up to your attention in case your *more experienced eyes* can see a possible solution. LOL

collinss:
@Odyseus nice catch, thanks! That uninstall button was working at one point, I'm not sure what happened. Anyway, I've pushed a fix for it now. As for the missing icon in the about dialog, I did notice it, but I thought it was a general problem with the spices about dialog. I'll see if I can figure something out.

collinss:
@Odyseus the icon issue in the about dialog is now fixed as well.

Odyseus:
@collinss: Tested and working. 👍

Issue #6550, opened 2017-05-20
Right-click Menus stuck in place until logoff


piers109uk:
Apparent bug: Entering panel edit mode and right clicking on a window causes the right-click menu to become fixed in place until log off. This persists after exiting panel edit mode.

I can replicate this issue and have attached screenshots. The only fix I have found is logging off and on again.

These menus are fixed at the front of the screen regardless of the window I pull up (seen in second screenshot). Oddly, one of the fixed menus disappeared and I have no idea why. 

I have also attached the output of inxi -fxr in a text file as was suggested in Linux-help chat.
EDIT: Closing the associated window gets rid of the pop-up menu, which makes this a very non-urgent bug, but a bug nonetheless.

![screenshot from 2017-05-20 11-22-19](https://cloud.githubusercontent.com/assets/28823694/26275178/c8c44078-3d52-11e7-9f1d-e81ca7fd156b.png)

![screenshot from 2017-05-20 11-55-04](https://cloud.githubusercontent.com/assets/28823694/26275193/35e18472-3d53-11e7-98f1-e68f2103805e.png)

[inxi-Fxr.txt](https://github.com/linuxmint/Cinnamon/files/1016088/inxi-Fxr.txt)

JosephMcc:
I can't reproduce this in Cinnamon 3.4. What is that thing running in the upper right hand corner?

collinss:
I can't reproduce this either. I don't even get that menu in panel edit mode - right click just brings up the standard applet context menu as it's supposed to in panel edit mode.

@JosephMcc are you referring to the devtools desklet, or the system monitor?

germanfr:
I can reproduce it since long time ago. But I can close them right-clicking again on each window button.

JosephMcc:
@collinss I'm talking about the devtools desklet. Didn't realize that's what it was since it isn't themed like the other cinnamon chrome.

piers109uk:
@germanfr Thanks. I'm surprised I hadn't tried right clicking after exiting edit mode before.

I'm on 3.2.7.

Should I close the issue now that there's an easy fix, or left open until the bug itself is fixed?

JosephMcc:
You can leave it open. If you plan to upgrade to Mint18.2 with Cinnamon3.4 maybe retest at that point and report back.

brownsr:
readily reproducible in Cinnamon 3.4,  just right click to launch the context menu of an applet when edit mode is on.

[...]

PR #6720, opened 2017-07-22
[RFC] popupMenu.js: cancel tweens on open/close if we are already animating

itzexor:
This fixes the stuck menus from #6550, but not the cause of multiple menus.

This prevents a race condition that can occur if a menu is opened and closed
in rapid succession. I don't believe this is reproducible by a user event, but
was exposed by a case where we had 2 menu activations, one immediately after
the other. The first menu would be closed by a grab loss signal during or
before the open animation started. When this happens the menu becomes stuck
open and unreactive as it has no grab.

The open tween is added on idle sometime after open() returns, so I'm not
sure of the exact order of events that cause the close animation to fail
and the menu to become stuck open with no grab.

### DIFF
--- a/js/ui/popupMenu.js
+++ b/js/ui/popupMenu.js
@@ -2028,6 +2028,10 @@ PopupMenuBase.prototype = {
         });
     },
 
+    get animating() {
+        return this.animating_for_open || this.animating_for_close;
+    },
+
     get firstMenuItem() {
         let items = this._getMenuItems();
         if (items.length)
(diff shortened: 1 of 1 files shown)


itzexor:
The window list and panel launchers double menu trigger was fixed in master.

Stuck menus can still be tested on certain sytems by right clicking and left clicking the sound applet at the same time, or by pressing menu shortcut key and right clicking the panel at the same time.

PR #6736, opened 2017-07-28
layout.js: special case handling of panel actors

itzexor:
This replaces the clip region based input region calculations with a panel-specific special case and also special cases monitor detection for panel actors.


itzexor:
I fully reworked the input region logic here, now it just intersects the panel rect with the monitor rect if actor is instanceof Panel.

PR #6743, opened 2017-07-28
recent-applet: always show icon in panel-edit-mode

NikoKrause:


files/usr/share/cinnamon/applets/recent@cinnamon.org/applet.js
```diff
@@ -69,14 +70,26 @@ MyApplet.prototype = {
...
+    _on_panel_edit_mode_changed: function () {
+        if (global.settings.get_boolean(PANEL_EDIT_MODE_KEY)) {
+            if (!this.actor.visible) {
+                this.actor.show();
+            }
+        } else {
+            this._refreshRecents();
+        }
+    },
+
```
mtwebster: Regarding _on_panel_edit_mode_changed() can we do something more like:

```
if (global.settings.get_boolean(PANEL_EDIT_MODE_KEY)) {
    this.actor.show();
} else {
    this.actor.visible = this._recentButtons.length > 0;
}
```

This gives us something tangible to check (length of _recentButtons) which is established at the end of refreshRecents if no recent files exist.  This keeps us from needing to run _refreshRecents as well.

files/usr/share/cinnamon/applets/recent@cinnamon.org/applet.js
```diff
@@ -87,7 +100,12 @@ MyApplet.prototype = {
     },
 
     _display: function() {
-        this._refreshRecents();
+        if (global.settings.get_boolean(PANEL_EDIT_MODE_KEY)) {
+            this._refreshRecents();
+            this._on_panel_edit_mode_changed();
+        } else {
+            this._on_panel_edit_mode_changed();
+        }
```
mtwebster: This can just be:
```
this._refreshRecents();
this._on_panel_edit_mode_changed();
```
The panel-edit-mode key is already being checked in this._on_panel_edit_mode_changed(), we don't need to do it here also.

NikoKrause:
PR changed

PR #6747, opened 2017-07-30
layout.js optimizations and docs

itzexor:
Wide range of optimizations and documentation fixes for layout.js. See individual commits for more details.

Optimizations that may improve performance:
- from 2 to 1 connections to screen signal windows-restacked
- remove unused var and if conditional from windows-restacked handler
- on screen windows-restacked - from 2 to 1 calls to _queueUpdateRegions if popup window is visible
- on screen monitors-changed - from 2 to 1 calls to hotCornerManager.updatePosition
- on layoutmanager monitors-changed -  from 2 to 1 calls to both _queueUpdateRegions and _updateVisibility
- when popupmenu is visible, one less comparison in Chrome actors loop during _updateRegions

Non-performance optimizations:
- call _queueUpdateRegions directly in _updateVisibility since we call it directly after, avoids an extra line of code everywhere that it is called
- replace clip region based input region correction with panel specific fix that instead intersects panel actor and monitor rect to determine on-monitor input region

New functionality:
- new "public" method LayoutManager.updateChrome() for externally forcing Chrome to update input region/visibility/etc. reason: panel.js called directly into chrome/layoutmanager underscore methods and i broke it during this series.
- new hideable-panel-specific special case in Chrome.getMonitorInfoForActor() to return the most up to date monitor. prevents returning adjacent monitor when panel is hidden.
- Chrome.findMonitorIndexForActor() now returns the primary monitor index if the index to be returned was invalid. this brings the behavior in line with the non-index variant. these are exposed through LayoutManager, so I wanted them to be consistent.

### DIFF
--- a/js/ui/layout.js
+++ b/js/ui/layout.js
@@ -64,8 +64,6 @@ LayoutManager.prototype = {
         this.hideIdleId = 0;
         this._chrome = new Chrome(this);
 
-        this._isPopupWindowVisible = false;
-
         this.enabledEdgeFlip = global.settings.get_boolean("enable-edge-flip");
         this.edgeFlipDelay = global.settings.get_int("edge-flip-delay");
 
@@ -425,6 +423,7 @@ Chrome.prototype = {
 
         this._monitors = [];
         this._inOverview = false;
+        this._isPopupWindowVisible = false;
         this._updateRegionIdle = 0;
         this._freezeUpdateCount = 0;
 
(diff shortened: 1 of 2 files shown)


itzexor:
Hold off on merging this, thanks.

I think I'm going to cherry pick my layout test branch into this one, make it a big split commit PR for review, then squash it back down to a reasonable amount of commits if accepted for merge.

itzexor:
Still WIP, few docs fixes needed.

itzexor:
Don't really like the way I have duplicated the params docs, but that shouldn't change often.

A preview of the new LayoutManager doc page is available here:
https://itzexor.github.io/cinnamon-js/cinnamon-js-ui-layout-LayoutManager.html
I manually added this, the rest of the docs there are not up to date.

PR #6753, opened 2017-08-01
panel: correct margin calculation in hidpi

brownsr:
+ minor correction to regression from 718ce025dc360d62c6e510b5b57483553bf80c1c

### DIFF
--- a/js/ui/panel.js
+++ b/js/ui/panel.js
@@ -2612,10 +2612,10 @@ Panel.prototype = {
             this._monitorsChanged = false;
         }
 
-        // calculate new panel sizes
+        // calculate new panel sizes.  NB margin is already scaled for hidpi
         let newVertPanelHeight = this.monitor.height - this.toppanelHeight - this.bottompanelHeight
-                                 - global.ui_scale*(newMarginTop + newMarginBottom);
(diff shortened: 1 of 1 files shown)


brownsr:
github has changed a plus to a highlight, changing the meaning of the comment.  There's a fix to margin calculation in hidpi here, and also a fix to a minor regression from a previous commit that I tripped over.  I now see there's a separate PR in for that regression fix as well.

js/ui/panel.js
```diff
@@ -2675,8 +2675,8 @@ Panel.prototype = {
             if (Main.panelManager && horizontal_panel) {
                 let panels = Main.panelManager.getPanelsInMonitor(this.monitorIndex);
                 for (let p = 0, len = panels.length; p < len; p++) {
-                    if (panels[i].panelPosition == PanelLoc.left || panels[i].panelPosition == PanelLoc.right)
-                        panels[i]._moveResizePanel();
+                    if (panels[p].panelPosition == PanelLoc.left || panels[p].panelPosition == PanelLoc.right)
+                        panels[p]._moveResizePanel();
```
jaszhix: Just a head up, I have this fixed in #6751.

brownsr: yep, that's what I noticed :)  I can remove it from here if needed, but as the changes are identical I imagine there won't be a problem if both are merged.

PR #6774, opened 2017-08-05
main: Improve logging

jaszhix:
This allows _logInfo to accept unlimited arguments. In the base _log function, it iterates arguments like before, but now uses a new function called formatLogArgument to format types, inspect nested objects, and format them with indentation. The original logging format with colons for two arguments has been preserved as well.

### DIFF
--- a/js/ui/main.js
+++ b/js/ui/main.js
@@ -916,6 +916,62 @@ function notifyError(msg, details) {
 }
 
 /**
+ * formatLogArgument:
+ * @arg (any): A single argument.
+ * @recursion (int): Keeps track of the number of recursions.
+ * @depth (int): Controls how deeply to inspect object structures.
+ *
+ * Used by _log to handle each argument type and its formatting.
+ */
+function formatLogArgument (arg, recursion=0, depth) {
+    // Make sure falsey values are clearly indicated.
+    if (arg === null) {
+        arg = 'null';
+    } else if (arg === undefined) {
+        arg = 'undefined';
+    // Ensure strings are distinguishable.
+    } else if (typeof arg === 'string'
+        && recursion > 0) {
+        arg = `'${arg}'`;
+    }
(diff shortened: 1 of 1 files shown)


js/ui/main.js
```diff
@@ -916,6 +916,62 @@ function notifyError(msg, details) {
...
 
 /**
+ * formatLogArgument:
+ * @arg (any): A single argument.
+ * @recursion (int): Keeps track of the number of recursions.
+ * @depth (int): Controls how deeply to inspect object structures.
+ *
+ * Used by _log to handle each argument type and its formatting.
+ */
+function formatLogArgument (arg, recursion=0, depth) {
```
Hanmac: this does break cinnamon:
> JS ERROR: SyntaxError: parameter(s) with default followed by parameter without default

jaszhix: Which version of cjs are you using? From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters):

> Prior to Gecko 26 (Firefox 26 / Thunderbird 26 / SeaMonkey 2.23 / Firefox OS 1.2), the following code resulted in a SyntaxError. This has been fixed in [bug 777060](https://bugzilla.mozilla.org/show_bug.cgi?id=777060) and works as expected in later versions. Parameters are still set left-to-right, overwriting default parameters even if there are later parameters without defaults.

Hanmac: @jaszhix:
cinnamon nightly build from there https://launchpad.net/~gwendal-lebihan-dev/+archive/ubuntu/cinnamon-nightly
cjs: 3.2.0-20170426020213-zesty

jaszhix: Yeah its cjs dependency is out of date. Cinnamon 3.6 will be using cjs depending on mozjs38. You can try [building Cinnamon](https://github.com/linuxmint/mint-dev-tools) instead.

PR #6784, opened 2017-08-09
trash applet: Fix arguments warnings

jaszhix:
Cjs-Message: JS WARNING:
[/usr/share/cinnamon/applets/trash@cinnamon.org/applet.js 37]: Too many
arguments to method Gio.File.monitor_directory: expected 2, got 3
Cjs-Message: JS WARNING:
[/usr/share/cinnamon/applets/trash@cinnamon.org/applet.js 80]: Too many
arguments to method Gio.FileEnumerator.next_file: expected 1, got 2

### DIFF
--- a/files/usr/share/cinnamon/applets/trash@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/trash@cinnamon.org/applet.js
@@ -34,7 +34,7 @@ MyApplet.prototype = {
 
             this._onTrashChange();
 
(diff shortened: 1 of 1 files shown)


mtwebster:
I'm pretty sure, like in python, GError slots in calls like this are supposed to be handled instead with try/catch, which will yield a GLib.Error type of error.  I'm not sure if it's important enough here to do, since it won't cause any real catastrophic issues.

PR #6786, opened 2017-08-10
Spices settings: Fix the trasition animation when switching to the do…

collinss:
…wnloads page


NikoKrause:
I see a translation problem with using `%ss` (%s being applet, desklet, extension I guess?).
It works in English, but it fails in other languages, e.g. German
Applet**s**, Desklet**s**, Erweiterung**en**

It's a similar issue as in your devTools desklet. I did a translationDummy back then.
Remind me to look into this later.

PR #6790, opened 2017-08-10
extension: Use imports.searchPath instead of imports.addSubImporter

jaszhix:
This replaces the usage of ```imports.addSubImporter``` with ```imports.searchPath```. In main, both the global and local Cinnamon data directories are added to the search path. This allows us to import those xlet directories directly, e.g. ```imports.applets```. This is then aliased to each xlet's UUID property on the importObject object in extension.js, instead of the old import object the CJS patch provided.

The good part is we can remove this extra functionality from CJS and move a little closer towards upstream GJS. ~~The bad part is, and why this is a work in progress, is it breaks the ```imports.applet``` import object~~. The majority of xlets on Spices import files through appletManager, which should behave like normal.

Edit: I will be updating all Spices that break because of this PR. Spices maintainers can do it too if they would like to preempt the change. 

Applets:
```imports.ui.appletManager.applets[UUID]```

Desklets:
```imports.ui.deskletManager.desklets[UUID]```

Extensions:
```imports.ui.extensionSystem.extensions[UUID]```

This PR also includes an [example change](https://github.com/jaszhix/Cinnamon/blob/57776fbc2124728af432c4c9da7f921c896cb8af/files/usr/share/cinnamon/applets/calendar%40cinnamon.org/applet.js#L11) that is shorter to type, but is _not_ intended for applets that are not multi-version, as it will break xlets on older versions of Cinnamon.

### DIFF
--- a/docs/reference/cinnamon-tutorials/importer.xml
+++ b/docs/reference/cinnamon-tutorials/importer.xml
@@ -146,10 +146,10 @@
       The best way is using <code>imports.ui.manager.xlet[UUID]</code>.
       <code>manager</code> is one of <code>appletManager</code>, <code>deskletManager</code>, or <code>extensionSystem</code>.
       <code>xlet</code> is the xlet type pluralized and is one of <code>applets</code>, <code>desklets</code>, or <code>extensions</code>.
-      <code>UUID</code> is a string of your xlet's UUID.
+      <code>UUID</code> is the string of your xlet's UUID.
 
       <para>
-        Importing applets:
+        Importing for applets:
         <informalexample>
           <programlisting>
             const Module = imports.ui.appletManager['foo@bar'].module; // get module.js in your applet directory
@@ -158,7 +158,7 @@
       </para>
 
       <para>
-        Importing desklets:
+        Importing for desklets:
         <informalexample>
           <programlisting>
             const Module = imports.ui.deskletManager['foo@bar'].module; // get module.js in your desklet directory
@@ -167,7 +167,7 @@
       </para>
 
       <para>
-        Importing extensions:
+        Importing for extensions:
         <informalexample>
           <programlisting>
             const Module = imports.ui.extensionSystem['foo@bar'].module; // get module.js in your extension directory
@@ -176,10 +176,10 @@
       </para>
     </para>
     <para>
-      In Cinnamon 3.6+, you can also import xlets by using imports.xletTypePluralized[UUID]. This method is not compatible with previous versions of Cinnamon.
+      In Cinnamon 3.6+, you can also import xlets by using <code>imports.xletTypePluralized[UUID]</code>. This method is not compatible with previous versions of Cinnamon.
 
       <para>
-        Importing applets:
+        Importing for applets:
         <informalexample>
           <programlisting>
             const Module = imports.applets['foo@bar'].module;
@@ -188,7 +188,7 @@
       </para>
 
       <para>
-        Importing desklets:
+        Importing for desklets:
         <informalexample>
           <programlisting>
             const Module = imports.desklets['foo@bar'].module;
@@ -197,7 +197,7 @@
       </para>
 
       <para>
-        Importing extensions:
+        Importing for extensions:
         <informalexample>
           <programlisting>
             const Module = imports.extensions['foo@bar'].module;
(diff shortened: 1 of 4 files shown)


Odyseus:
Hello, @jaszhix.

Can you confirm if the `imports.applet.module_name` method will be removed/broken when this PR is merged? I'm asking because I saw the strikethrough text on your PR message. In addition, I cannot test this PR because my test environment isn't very reliable at the moment (it's a long story).

If the `imports.applet.module_name` method will be effectively removed/broken, please, also update the [Cinnamon documentation](https://github.com/linuxmint/Cinnamon/blob/master/docs/reference/cinnamon-tutorials/importer.xml#L142). Because the only example about importing xlets modules in the documentation is precisely the method that this PR might remove/break. And it would be cool if the docs can include all the ways to import *custom modules*, not just one.

Just to see if I got it right. The *right way* to do *custom modules imports* is as you specified in your edit (e.g., for applets: `imports.ui.appletManager.applets[UUID].module_name`, etc.)? I said the *right way* because I tested this method in all versions of Cinnamon that I have (2.8.6, 3.0.7 and 3.4.x) and in all of them seems to be working as expected. I never used this method because it isn't documented in Cinnamon docs.

> This PR also includes an example change that is shorter to type, but is not intended for applets that are not multi-version, as it will break xlets on older versions of Cinnamon.

Your *double negative* just killed me there (LOL). You say that the *shorter to type* method *"is **not** intended for applets that are **not** multi-version"*, but the linked example is effectively using that shorter method in a non multi-version xlet. To *whom* do I believe in?

jaszhix:
Yeah, I was cringing as I wrote that sentence but it was the best way I could put it at the time. :) I updated the docs. I think because there was a shorter alternative it wasn't necessary to document the manager imports, but at least its a little shorter with this PR if you are working on a multi-version xlet in 3.6+. I usually assign ```imports.ui.appletManager.applets['uuid@foobar']``` to a variable, and use that object to import the files I need.

MrElectroNick:
@jaszhix Can you please sync this page also:
https://github.com/linuxmint/linuxmint.github.io/blob/master/reference/git/cinnamon-tutorials/importer.md

lestcape:
@jaszhix, you can create an alias for `imports.applet.module_name` in cjs and set there the **imports** of the sub-module (xlet), that you catch here: https://github.com/linuxmint/Cinnamon/commit/34d9e451dccfcdf4654782d112dac53acdf68615 and then use the global Proxy class and the Reflect static class if you are in ES2016. You can find more information here: https://github.com/lestcape/Classic-Gnome/issues/20

jaszhix:
@lestcape  Yeah, but now Cinnamon can run with GJS if needed. Not immediately necessary, but adds more flexibility for the future. There is also a notable [performance penalty](https://jsperf.com/proxy-cost/8) with proxying methods, especially slow on Spidermonkey. I have overhauled the module system in a [PR](https://github.com/linuxmint/Cinnamon/pull/6878) planned for 3.8.

lestcape:
Yes of course, a proxy is more slow that not have one that is clear, but you don't use the proxy a lot of times, just when is needed. Also the implementation you have in this pull, from sure will be more slow as is required not just a proxy, but more additional steps. I have not doubts about the usefulness of what you're doing though, I just do not like that kind of syntax of Node, but is a personal taste, i suppose...

jaszhix:
The main reason for following the node conventions is because most editors support the commonjs spec, and have syntax and plugin support for it. For instance, a `require` statement will match the regex for an editor/plugin to start suggesting file names to import into a file. It may also make it easier to get [typing info](https://github.com/sammydre/ts-for-gjs) from TS definitions. Its not really about my opinion so much as helping us join what's essentially a defacto standard outside GJS. In addition, it makes it unnecessary to use `var` for imports, which is a new restriction in GJS 1.50+.

I'm most interested in the move to async loading of xlets though, and hopefully more async handling of operations within Cinnamon. I think the additional steps towards that will be a net positive in the long run. I did use proxies in that PR, but realized later they weren't needed for getting the correct source mapping in stack traces, and removed them.

lestcape:
Not in all context you can avoid a proxy, because avoid the proxy mean that you need to create an alternative way (syntax and/or functionality) to do your steps and like me you probably not want that, but some time is more the improve and is ok do it.

Async loading of xlets will be really an amazing thing, especially for the session start. I will follow your improve in that area for sure. As i say, `I have not doubts about the usefulness of what you're doing`. Also yes, you can move from node a lot of things that currently exist and thats wonderful ofcourse, but also have a price to be pay, like all in this world.

Issue #6607, opened 2017-06-11
JS WARNING reported by the use of CJS 3.4.x on Cinnamon 3.4.x


Odyseus:
Hello, everybody.

The following is a warning logged into the **.xsession-errors** file when using CJS 3.4.1 on Cinnamon 3.4.1.

> Cjs-Message: JS WARNING: [/usr/share/cinnamon/js/ui/popupMenu.js 1125]: reference to undefined property this._ornament.child._delegate

> System:    Kernel: 4.4.0-21-generic x86_64 (64 bit gcc: 5.3.1)
>            Desktop: Cinnamon 3.4.1 (packages from ppa:gwendal-lebihan-dev/cinnamon-nightly built on 2017-06-11)
>            Distro: Linux Mint 18.1 Serena
>            Cjs: 3.4.1 (built from source on 2017-06-06)

### Edit

I didn't realize that almost no applet makes use of the **PopupIndicatorMenuItem** prototype. Only I am using it on three of my applets ([Argos for Cinnamon](https://cinnamon-spices.linuxmint.com/applets/view/289), [Popup Translator](https://cinnamon-spices.linuxmint.com/applets/view/279) and [Simple ToDo List](https://cinnamon-spices.linuxmint.com/applets/view/290)). I thought that this would be worth mention it in case someone wants to reproduce the warning before fixing it.

The above warning only shows up upon logging in with any of those three applets placed into a panel. It triggers only once regardless if there is only one of the applets placed or all of them. Restarting the shell (Cinnamon) doesn't trigger the warning.

two-dogs:
@Odyseus  https://gist.github.com/anonymous/003cac0ada514f44701231fc82d96c37

Odyseus:
Hello, @lestcape.

I know what's causing it, how to fix it and that it is just a warning. But lately, the Cinnamon developers have been *"fixing"* all CJS messages, regardless of them being actual errors or just warnings.

@ everyone: I added some more details to my initial message.

Odyseus:
Fixed by  #6801.

PR #6801, opened 2017-08-13
popupmenu: fix an undeclared variable

brownsr:


Odyseus:
Hello, @brownsr.

I thing that this PR might fix #6607.

I can't test it right now because I *trashed* (LOL) my test virtual machine with Cinnamon 3.4.x and CJS 3.4.x. Could you confirm?

brownsr:
I tried the simple to do list applet.  It's not showing any warning message with this PR.  Testing this change is a bit awkward, as it really needs an indicator with suitable adornments, and I do not know of suitable examples [there may be some, I just don't know them].  But at least the warning message is treated and there's no regression  apparent.

Odyseus:
Thanks for taking the time to checking it out, @brownsr.

I managed to setup a new, more or less stable, test environment (the package versions on the nightly PPA are kind of messed up). I can confirm that this PR eliminates the warning reported on the issue that I linked above.

I will also work on the errors/warnings that you found with my applets.

Thanks again. :+1:

PR #6811, opened 2017-08-17
[RFC][WIP] Peek panels for 1.5 seconds while highlighted when hidden

itzexor:
This helps to bring attention to the panel since the highlighted state is not easily noticeable without this. Before my _moveResizePanel refactor a while back, the panel would be immediately moved into showing state on style change (not intended). This restores that behavior for the highlight style change, except does it with a smooth animate in/out.

This likely has merge conflicts with other open PRs. I will rebase as needed.

### DIFF
--- a/js/ui/panel.js
+++ b/js/ui/panel.js
@@ -3015,66 +3015,74 @@ Panel.prototype = {
      * position of mouse/active window. It then calls the _queueShowHidePanel
      * function to show or hide the panel as necessary.
      *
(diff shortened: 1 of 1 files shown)


itzexor:
This is still not as useful as it should be. For the applet add case, you probably want the panel to always show. That is not easy to detect and would require quite a bit more complexity, so I feel that it's necessary at this time. The reason it doesn't always force show when highlighted is that it would prevent users from testing autohide settings in panel settings.

PR #6845, opened 2017-08-24
Bump version for muffin build dep

leigh123linux:
Needed because of https://github.com/linuxmint/muffin/commit/fd1349b5ba2a8aab7aa703bf2fd8fc843c4d0634


mtwebster:
The reason I didn't do this is because the debian *install* dependencies require equal or greater than current version - I'd have to tag all cinnamon-related stuff in addition to this to be able to install it, and I didn't feel like doing that.

leigh123linux:
I can revert the control file change for now but the configure.ac change should be added to prevent FTBFS reports.

PR #6854, opened 2017-08-28
boxpointer: Don't hide when we're already hidden

brownsr:
https://github.com/GNOME/gnome-shell/commit/d77fc015801e60bb491eac5c5e65cbc347ebb0f4

Do we still use boxpointer ?


itzexor:
We do still use it in the on screen keyboard, though I haven't figured out how to test that part yet in order to test removal. It's used for extended keys. My best guess is that when you activate that mode it brings a little popup with the extra keys inside it.

PR #6878, opened 2017-09-06
Refactor the extension system and add commonjs compatible module importation

jaszhix:
Depends on: https://github.com/linuxmint/cjs/pull/54

This cleans up some cyclical references and moves state from nested objects to an array of objects in extension.js. Global properties were made read-only, in the unlikely event they are re-assigned and crash Cinnamon. This also restructures the extension code for asynchronous loading using [Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). 

The main advantage of loading extensions asynchronously is non-blocking I/O. Currently, when an extension is loaded, the UI locks up until the extension is finished loading. Worse, this can cause Cinnamon to freeze up entirely if a mistake is made in the xlet's code. It also opens up the possibility of off-loading smaller tasks to dedicated worker threads, such as reading files.

A new module importer is added that brings fixes for xlets not getting fully removed from the JS context when removed from the panel, or reloaded, and brings closer compatibility to libraries written for node. 

Example usage of the module importer from inside an xlet:

```js
const Cinnamon = require('gi.Cinnamon');
const DND = require('ui.dnd');
const gettext = require('gettext');
const myModuleA = require('./myModuleA.js');
```
Node conventions such as ```require```, ```module.exports```, ```__dirname```, and ```__filename``` only exist in the extension context.

No breaking changes to the API are introduced with this PR outside of the internal extension utilities.

### DIFF
--- a/docs/reference/cinnamon-tutorials/importer.xml
+++ b/docs/reference/cinnamon-tutorials/importer.xml
@@ -176,31 +176,49 @@
       </para>
     </para>
     <para>
-      In Cinnamon 3.6+, you can also import xlets by using <code>imports.xletTypePluralized[UUID]</code>. This method is not compatible with previous versions of Cinnamon.
+      In Cinnamon 3.6+, there is now a simpler way to import files using <code>require</code>. This method is not compatible with previous versions of Cinnamon.
 
       <para>
-        Importing for applets:
+        Importing GI modules:
         <informalexample>
           <programlisting>
-            const Module = imports.applets['foo@bar'].module;
+            const Cinnamon = require('gi.Cinnamon');
           </programlisting>
         </informalexample>
       </para>
 
       <para>
-        Importing for desklets:
+        Importing Cinnamon modules:
         <informalexample>
           <programlisting>
-            const Module = imports.desklets['foo@bar'].module;
+            const DND = require('ui.dnd');
           </programlisting>
         </informalexample>
       </para>
 
       <para>
-        Importing for extensions:
+        Importing other JS files (".js" extension is optional):
         <informalexample>
           <programlisting>
-            const Module = imports.extensions['foo@bar'].module;
+            const MenuItem = require('./menuItem.js').MenuItem;
+          </programlisting>
+        </informalexample>
+      </para>
+
+      <para>
+        You can also manually export your modules by adding them to <code>module.exports</code> at the end of the file.
+        Inside your module file:
+        <informalexample>
+          <programlisting>
+            function MenuItem() {...}
+            ...
+            module.exports = MenuItem;
+          </programlisting>
+        </informalexample>
+        Then from the file you want import MenuItem from:
+        <informalexample>
+          <programlisting>
+            const MenuItem = require('./menuItem');
           </programlisting>
         </informalexample>
       </para>
--- a/js/misc/fileUtils.js
+++ b/js/misc/fileUtils.js
@@ -141,7 +141,6 @@ function unloadModule(index) {
 }
 
 function createExports({path, dir, file, size, JS, returnIndex, reject}) {
-    JS = `${JS};`;
     // Import data is stored in an array of objects and the module index is looked up by path.
     let importerData = {
         size: size,
@@ -171,6 +170,7 @@ function createExports({path, dir, file, size, JS, returnIndex, reject}) {
         moduleIndex = LoadedModules.length - 1;
     }
 
+    JS = `${JS};`;
     // Regex matches the top level variable names, and appends them to the module.exports object,
     // mimicking the native CJS importer.
     let modules = []
@@ -188,6 +188,7 @@ function createExports({path, dir, file, size, JS, returnIndex, reject}) {
     }
     // Return the exports object containing all of our top level namespaces.
     JS += `return module.exports;`;
+
     try {
         // Create the function returning module.exports and return it to Extension so it can be called by the
         // appropriate manager.
@@ -208,7 +209,6 @@ function createExports({path, dir, file, size, JS, returnIndex, reject}) {
             dir,
             file.get_basename()
         );
-
         return returnIndex ? moduleIndex : importerData.module;
     } catch (e) {
         let error = requireModuleError(path, e);
@@ -247,38 +247,30 @@ function requireModule(path, dir, async = false, returnIndex = false) {
             path = `${dir}/${path}`;
         }
     }
-    let success, JS, info, size;
+    let success, JS;
     let file = Gio.File.new_for_commandline_arg(path);
     if (!file.query_exists(null)) {
         throw requireModuleError(path, e);
     }
-    let flags = Gio.FileQueryInfoFlags.NOFOLLOW_SYMLINKS;
-    let priority = GLib.PRIORITY_DEFAULT;
 
     if (!async) {
-        info = file.query_info('*', flags, null);
-        size = info.get_size();
         [success, JS] = file.load_contents(null);
         if (!success) {
             throw requireModuleError(path, new Error('Unable to query file info.'));
         }
-        return createExports({path, dir, file, size, JS, returnIndex});
+        return createExports({path, dir, file, size: JS.length, JS, returnIndex});
     }
     return new Promise(function(resolve, reject) {
-        file.query_info_async('*', flags, priority, null, function(file, result) {
-            info = file.query_info_finish(result);
-            size = info.get_size();
-            file.load_contents_async(null, function(object, result) {
-                try {
-                    [success, JS] = file.load_contents_finish(result);
-                    if (!success) {
-                        throw new Error('Unable to load file contents.');
-                    }
-                    resolve(createExports({path, dir, file, size, JS, returnIndex, reject}));
-                } catch (e) {
-                    reject(requireModuleError(path, e));
+        file.load_contents_async(null, function(object, result) {
+            try {
+                [success, JS] = file.load_contents_finish(result);
+                if (!success) {
+                    throw new Error('Unable to load file contents.');
                 }
-            });
+                resolve(createExports({path, dir, file, size: JS.length, JS, returnIndex, reject}));
+            } catch (e) {
+                reject(requireModuleError(path, e));
+            }
         });
     });
 }
--- a/js/misc/fileUtils.js
+++ b/js/misc/fileUtils.js
@@ -102,7 +102,18 @@ function findModuleIndex(path) {
 }
 
 function getModuleByIndex(index) {
+    if (!LoadedModules[index]) {
+        return;
+    }
     return LoadedModules[index].module;
+}
+
+function unloadModule(index) {
+    if (!LoadedModules[index]) {
+        return;
+    }
+    LoadedModules[index] = undefined;
+    LoadedModules.splice(index, 1);
 }
 
 function createExports(path, dir, file, size, JS, returnIndex) {
@@ -126,7 +137,7 @@ function createExports(path, dir, file, size, JS, returnIndex) {
         // Module already exists, check if its been updated
         if (size === LoadedModules[moduleIndex].size) {
             // Return the cache
-            return LoadedModules[moduleIndex].exports;
+            return returnIndex ? moduleIndex : LoadedModules[moduleIndex].module;
         }
         // Module has been updated
         LoadedModules[moduleIndex] = importerData;
@@ -172,14 +183,10 @@ function createExports(path, dir, file, size, JS, returnIndex) {
             dir,
             file.get_basename()
         );
-        if (returnIndex) {
-            return moduleIndex;
-        }
-        return importerData.module;
+        return returnIndex ? moduleIndex : importerData.module;
     } catch(e) {
         // Remove the module from the index
-        LoadedModules[moduleIndex] = undefined;
-        LoadedModules.splice(moduleIndex, 1);
+        unloadModule(index);
         throw requireModuleError(path, e);
     }
 }
--- a/js/misc/fileUtils.js
+++ b/js/misc/fileUtils.js
@@ -3,6 +3,7 @@
 const Gio = imports.gi.Gio;
 const GLib = imports.gi.GLib;
 
+const LoadedModules = [];
 const FunctionConstructor = Symbol();
 const Symbols = {};
 Symbols[FunctionConstructor] = 0..constructor.constructor;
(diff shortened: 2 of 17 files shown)


Odyseus:
Hello, @jaszhix.

If you don't mind, I would like to ask you a couple of questions because I just hit a dead end from which I don't see an exit.

Can you tell me what's the alternative way of getting the xlets meta data post Cinnamon 3.6.x? `imports.ui.extensionSystem.extensionMeta[ExtensionUUID]`, `imports.ui.appletManager.appletMeta[AppletUUID]` on pre Cinnamon 3.6.x. It used to give an object containing all the data from the xlet metadata.json file plus "path" and other added data.

I tried `imports.ui.extension.getExtension(AnyTypeOfXletUUID).meta`. But I get the error *imports.ui.extension.getExtension(...) is null*. It seems to work if I call **getExtension** inside the **init** or **enable** functions in an extension (not tested on applets/desklets), which is absolutely useless for me in there.

I also tried `imports.ui.extension.getMetadata(AnyTypeOfXletUUID, "xlet-type")`. But it returns an object that has nothing to do with any xlet meta data. (Example: `{state: ['REJECTED',],queue: [],outcome: {},handled: ['UNHANDLED',],}`).

In short, I can't seem to find a way to get an xlet meta data from outside the xlets objects. Every single one of my xlets depends on a way to get that meta data SYNCHRONOUSLY. I would like to know if this is possible post Cinnamon 3.6.x. And if so, how.

jaszhix:
The metadata is passed to the applet constructor, you can get it in the `_init` method. I would have regression tested against your applets, but they disappeared.

Odyseus:
Hello, @jaszhix.

> The metadata is passed to the applet constructor, you can get it in the _init method.

I already knew that. And in fact, I already modified all my xlets to use that meta data (it was an absolute nightmare!!!). In some cases, I had to pass that data through countless functions and/or classes just to make use of one of its properties where was really needed. And I even was forced to make that data a property of a class (where that data isn't needed/used) just to be able to pass it to another class/function.

Still, that doesn't answer the question. Is there a way to get that meta data **SYNCHRONOUSLY** outside the xlets objects from a sub-module that isn't applet.js, extension.js, etc.?

I have a couple more questions to ask you if you don't mind.

1. Could you confirm that exists only one way to import JavaScript files now (post Cinnamon 3.6.x)? Example: `require("./file.js")`. Is there any other way?
2. It is still possible to use `imports.gi.GLib` on post Cinnamon 3.6.x. Are there any plans for removing this method and force us to use `require("gi.GLib")`? Or we can still use `imports.gi.GLib` without future surprises?

jaszhix:
I did test a few of your applets, I only modified the meta object in appletManager and removed the properties that weren't in use by any xlets. None of your applets were using the meta object for anything but the path, which is preserved. Asynchronous loading has nothing to do with this, if it does, you should provide an example. All xlet files are loaded asynchronously, the timing of this should be irrelevant to the applet's themselves though, as far as they're concerned their execution is synchronous, unless they themselves are loading something asynchronously.

> It is still possible to use imports.gi.GLib on post Cinnamon 3.6.x. Are there any plans for removing this method and force us to use require("gi.GLib")? Or we can still use imports.gi.GLib without future surprises?

The old importer still works, but reloading doesn't work unless you use `require`.

Odyseus:
Before continuing, I would like to report what I think is an error.

I get the following error when trying to use the function `imports.ui.extension.getExtension(ExtensionUUID)`.


```shell
[0Test@domain.test]: this.formatError is not a function
[0Test@domain.test]: Error importing extension.js from 0Test@domain.test
trace t=2018-02-22T07:39:45.894Z 
<----------------
Extension.prototype.checkProperties@/usr/share/cinnamon/js/ui/extension.js:317:39
Extension.prototype.validateMetaData@/usr/share/cinnamon/js/ui/extension.js:287:9
Extension.prototype._init/<@/usr/share/cinnamon/js/ui/extension.js:219:17
---------------->
```

The *problematic* line is [this one](https://github.com/linuxmint/Cinnamon/blob/master/js/ui/extension.js#L317). Just to test, I removed the `this` from the call to `formatError` (because the Extension prototype doesn't have a `formatError` method). But I also noted that the call in this line to the `formatError` method passes **3 arguments**, while the `formatError` function defined at [L112](https://github.com/linuxmint/Cinnamon/blob/master/js/ui/extension.js#L112) receives just **2 arguments**.

I'm surprised, with how *anal* that CJS lately is, it hasn't noticed all of this.

After fixing all those possible errors, **there is still an error when calling `imports.ui.extension.getExtension(ExtensionUUID)` at the beginning of a file**.

```shell
imports.ui.extension.getExtension(...) is null
trace
<----------------
anonymous@/home/user_name/.local/share/cinnamon/extensions/0Test@domain.test/extension.js:7:16
createExports@/usr/share/cinnamon/js/misc/fileUtils.js:210:31
requireModule/</<@/usr/share/cinnamon/js/misc/fileUtils.js:286:25
---------------->
```

This is why I mentioned "synchronously" in my previous messages. One of the lines where the trace points to (**fileUtils.js:286**) is inside a **Promise**. I thought that the calls to `createExports` contained the wrong number of arguments, but then I saw that it is one whole object. So, I got lost here and I couldn't debug it further.

The point is, if I call `imports.ui.extension.getExtension` at the beginning of a file for the purpose of getting an xlet meta data, it will not work. But calling `imports.ui.extension.getExtension` inside `init` or even `enable`, it works.

Here you have a very minimal extension with some `try{}catch{}` blocks depicting what I described above.

[0Test@domain.test.zip](https://github.com/linuxmint/Cinnamon/files/1747077/0Test.domain.test.zip)
***

This is possibly off-topic.

I observed another possible error. And it might not have anything to do with this PR (I didn't check all files that this PR modifies). The `dangerous` flag is not being set for any xlet of neither type. Not even if the xlet is actually *dangerous*.

Let me know if I should open a new issue or if this is by design.

jaszhix:
Okay, I see what you're talking about now. The error you're seeing is indeed because the promise hasn't resolved until the init method has been called. It would be possible to make the meta object apart of the xlet context, but there are still some regressions related to applets I am looking into.

This is the output I get when running your test:

```
Cjs-Message: JS LOG: Top of the file __init__.js:
Cjs-Message: JS LOG: imports.ui.extension.getExtension(...) is null
Cjs-Message: JS LOG: 
<----------------
anonymous@/home/jason/.local/share/cinnamon/extensions/0Test@domain.test/__init__.js:5:16
createExports@/usr/share/cinnamon/js/misc/fileUtils.js:210:31
requireModule@/usr/share/cinnamon/js/misc/fileUtils.js:277:16
require@/usr/share/cinnamon/js/misc/fileUtils.js:220:24
anonymous@/home/jason/.local/share/cinnamon/extensions/0Test@domain.test/extension.js:3:18
createExports@/usr/share/cinnamon/js/misc/fileUtils.js:210:31
requireModule/</<@/usr/share/cinnamon/js/misc/fileUtils.js:286:25
---------------->
Cjs-Message: JS LOG: Top of the file extension.js:
Cjs-Message: JS LOG: imports.ui.extension.getExtension(...) is null
Cjs-Message: JS LOG: 
<----------------
anonymous@/home/jason/.local/share/cinnamon/extensions/0Test@domain.test/extension.js:22:16
createExports@/usr/share/cinnamon/js/misc/fileUtils.js:210:31
requireModule/</<@/usr/share/cinnamon/js/misc/fileUtils.js:286:25
---------------->
Cjs-Message: JS LOG: InitFile constant:
Cjs-Message: JS LOG: {
}
Cjs-Message: JS LOG: Inside init:
Cjs-Message: JS LOG: {
    uuid: '0Test@domain.test',
    url: '0Test@domain.test',
    cinnamon-version: [
        '3.0',
        '3.2',
        '3.4',
        '3.6',
        '3.8',
    ],
    name: '0Test@domain.test',
    description: '0Test@domain.test',
    state: 0,
    path: '/home/jason/.local/share/cinnamon/extensions/0Test@domain.test',
    error: '',
}
Cjs-Message: JS LOG: true
```

Not able to reproduce the first error you mention. `getExtension` is not intended to be a public API in my mind, its an internal function Cinnamon uses for the xlet loading process. I don't intend to make it work that way myself, but feel free to submit a PR if you think you can make it work.

Odyseus:
You should really look at the line I linked above ([this one](https://github.com/linuxmint/Cinnamon/blob/master/js/ui/extension.js#L317)). **The error is there**. You didn't see it with the example extension that I linked because I added the **url** property to the extension metadata.json file. That was what revealed the error to me.

I repeat:

> Just to test, I removed the `this` from the call to `formatError` (because the Extension prototype doesn't have a `formatError` method). But I also noted that the call in this line to the `formatError` method passes **3 arguments**, while the `formatError` function defined at [L112](https://github.com/linuxmint/Cinnamon/blob/master/js/ui/extension.js#L112) receives just **2 arguments**.

**Summarized: There is a call to a function that doesn't exists (`this.formatError`) and with the wrong number of arguments passed (3 instead of 2) [right here](https://github.com/linuxmint/Cinnamon/blob/master/js/ui/extension.js#L317).**

jaszhix:
Ah, I missed that. Sorry. I'll fix it up when I'm able to, within the next few days.

Odyseus:
Hello, @jaszhix.

> ...but feel free to submit a PR if you think you can make it work...

Before starting to work on a PR, I would like to know what's the motive for removing the full meta data from `appletMeta` and `deskletMeta` from the applet/desklet managers (leaving only the "path") and completely removing the `extensionMeta` variable from `extensionSystem`? Because if there aren't any security or performance issues, then I'm inclined to add the full meta data back to all managers.

That none of the xlets found on the Spices repositories aren't using all that meta data isn't reason enough to eradicate it from Cinnamon's core. First, because the xlets found in Spices aren't the only xlets out there. And second, because most of the applets in Spices are using the imported meta data from inside the applet module (which is pointless due to the meta data being already available as an argument).

I think that the situation that I described in one of my first messages would be reason enough to keep the full meta data available. And I cite myself:

> And in fact, I already modified all my xlets to use that meta data (it was an absolute nightmare!!!). In some cases, I had to pass that data through countless functions and/or classes just to make use of one of its properties where was really needed. And I even was forced to make that data a property of a class (where that data isn't needed/used) just to be able to pass it to another class/function.

**Summarized: if there aren't any security or performance repercussions by having the full xlet meta data available, and if you don't have any objections, I can prepare a PR to add that meta data back to all xlets managers.**

jaszhix:
I don't recall objecting to adding it. I think you might be reading too much into its removal (from global scope), it was purely done for performance. Originally a lot of circular references were occurring with the previous meta object, it wasn't just applet metadata, the entire applet structure was referenced to it, and because it was just a secondary reference from the source (`this.meta` on extension instances), it was easiest to prune that seeing no applets actually use it.

It sounds like you don't want to [look](https://github.com/linuxmint/Cinnamon/pull/6878/commits/91df80ceda5e00b40ccde6e08bcce89d04f76a1f) at my [commits](https://github.com/linuxmint/Cinnamon/pull/6878/commits/31c6380386ecc750dea91d0c8489b42074db2100), where I'm trying to detangle some of the circular references. It was not done without a good reason originally. The Type object in extension.js was messy and made logging/debugging these object structures impossible. There is a way I had in mind of adding the metadata back through the extension scope directly. I was going to do this in a separate PR, but I would appreciate a little patience because I have to stay busy with my other work too.

Odyseus:
> It sounds like you don't want to look at my commits,

I did see at your commits. In fact, between the *"nodejs(isms)"*, the *"ES6(isms)"* and the fact that I'm not a professional developer with a deep understanding of JavaScript, it took me more than 2 weeks to understand *very superficially* what it is being done. And once I did, I had to move/add hundreds of lines of code just to achieve the same functionality that in every single version of Cinnamon prior to 3.6.x can be done with just one line of code.

Please, do not take my words as criticism. I'm just narrating my experience caused by this *simple change*. And please, strike from your though any intent of *pressure* towards you to do anything. The insistence or impatience that you sense in all my previous messages is just my intent (with my own non-professional developer words) to make you see that that meta data is *important* to be made available globally. If not for the current available xlets, for the ones that could come. It's just that that global availability simplifies so much obtaining that data.

> ...I think the meta object should be shallow object...

And here is where you will see that I'm not a professional developer.

I was about to do the following to add the full meta data back:

```javascript
type.legacyMeta[uuid] = this.meta;
```

But I assume that that is precisely what you want to avoid (?).

So, the correct way (I'm assuming again) would be (?):

```javascript
type.legacyMeta[uuid] = Object.assign({}, this.meta);
```

My research revealed that this is the correct and more direct way of creating a shallow copy of an object.

I already have the [PR ready to submit](https://github.com/Odyseus/Cinnamon/commit/5c12b36518734a117f47ea3447c58f29fb034f5a).

jaszhix:
Ah, okay, I didn't realize the extension meta object was completely missing, that was a mistake.

What I will be PRing will probably result in the replacement of those objects. I think moving away from code duplication in the xlet managers would be the best strategy. It can be fixed in ui/extension.js and misc/fileUtils.js by passing the original metadata through to the function constructor as `__meta` in the xlet's scope. Like how I already did it with the other node-like [variables](https://github.com/linuxmint/Cinnamon/blob/2dc04b8a26fa7a672c579ed62a70dca6c6e4b295/js/misc/fileUtils.js#L214), `__dirname` and `__filename`. That would be a cleaner solution IMHO instead of more object references.

jaszhix:
On the other hand the meta object _is_ passed to the init method for extensions. I think you're missing the point with async loading. If the promise hasn't resolved then you can't do anything inside the extension anyway. We're probably talking about the 1-10ms range between the time the file is loaded from disk and the `init` method being called. 

> And once I did, I had to move/add hundreds of lines of code just to achieve the same functionality that in every single version of Cinnamon prior to 3.6.x can be done with just one line of code.

Sounds like hyperbole. You still are vague about what is causing you to write hundreds of lines of code. I would have spent that time researching Cinnamon's code and figuring out how to work with it. I mean I'm sorry I broke your flow, but APIs always evolve. Its like writing angry letters to Mozilla because they updated their extension API.

Odyseus:
> Like how I already did it with the other node-like variables, __dirname and __filename.

I think that that would be amazing!

While checking those variables, I found what it might be a very serious bug (emphasis on might). In a **multiversion** enabled xlet (only did tests on a minimal test extension), both **extension.js** files seem to be executed/loaded, the **extension.js** file found on the root folder of the xlet, and the **extension.js** file found on the version sub-folder (3.6 in this case).

[Here is an example extension](https://github.com/linuxmint/Cinnamon/files/1754260/0Test.domain.test.zip).

jaszhix:
That's an incorrect file structure for multi-version extensions.

Odyseus:
> That's an incorrect file structure for multi-version extensions.

That's the exact same folder structure that I have been using since Cinnamon 2.8. In a **multiversion** enabled xlet, if an xlet sub-folder correspond to the version of Cinnamon used, that will be the xlet loaded. Else, the xlet files found in the root of the xlet. And several of the xlets that I ckecked in the Spices repository confirm that. ([binaryclock@entelechy](https://github.com/linuxmint/cinnamon-spices-applets/tree/e698e72a30d5670ec8e27b1dfe9dfc39d8c09319/binaryclock%40entelechy/files/binaryclock%40entelechy) - [search-box@mtwebster](https://github.com/linuxmint/cinnamon-spices-applets/tree/390cd8e78661546cbb47d3331626ca35f7282ae9/search-box%40mtwebster/files/search-box%40mtwebster) -[sticky@scollins](https://github.com/linuxmint/cinnamon-spices-applets/tree/fe2f8027f7084570fdd4b3aa9e5bbbf044e2fae0/sticky%40scollins/files/sticky%40scollins))

jaszhix:
Well its not correct, if you're going to do multi-version its supposed to be in their own folders. We only support the current version of CInnamon. I feel like I've spent enough of my limited personal time and energy trying to reason with you. I'm no longer interested in this back and forth since you don't listen.

PR #6886, opened 2017-09-09
xlet-settings: translate title in list widget

NikoKrause:
The title in the list widget, which has been introduced with Cinnamon 3.4, is not translated.
Gotta catch (and translate) 'em all.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py
@@ -262,6 +262,10 @@ class MainWindow(object):
                                     continue
                                 new_opt_data[translate(self.uuid, option)] = opt_data[option]
                             settings_map[setting][key] = new_opt_data
+                        elif key in "columns":
+                            columns_data = settings_map[setting][key]
+                            for column in columns_data:
+                                column["title"] = translate(self.uuid, column["title"])
             finally:
                 # if a layout is not explicitly defined, generate the settings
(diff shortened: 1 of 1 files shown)


files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py
```diff
@@ -262,6 +262,10 @@ def load_instances(self):
                                     continue
                                 new_opt_data[translate(self.uuid, option)] = opt_data[option]
                             settings_map[setting][key] = new_opt_data
+                        elif key in "columns":
```
Odyseus: Hello, @NikoKrause.

Replace that entire block of code that you added with the following...

```python
elif key in "columns":
    col_data = settings_map[setting][key]

    for col_idx in range(0, len(col_data)):
        try:
            settings_map[setting][key][col_idx]["title"] = translate(self.uuid, col_data[col_idx]["title"])
        except:
            pass
```

...that should make it work.

NikoKrause: Where's the difference?

Odyseus: > Where's the difference?

I have no idea. I stopped reading your code when I saw the use of the `_()` function (which shouldn't be used) and I "mashed up" some code for you to test.

NikoKrause:
Removed the `_()`

PR #6888, opened 2017-09-10
Guard against failure of cairo_pattern_get_surface()

brownsr:
https://github.com/GNOME/gnome-shell/commit/6d22670307928dfdce75a7536598486fa25d51d7#diff-8a01b40c2c1be22ef97916811be0ab65

Stops a potential fault condition causing a crash.


leigh123linux:
Defend was poor wording on the original commit IMO, using the term guard would have been better.

brownsr:
reworded :)

Issue #6185, opened 2017-01-08
[feature request][power manager] Add more sleep and screen-off time options


elegantandrogyne:
The current 5/10/15/30/45min/1hr times are not enough for those users who watch movies, listen to music etc. without moving the mouse for a long time, but still would like to use automatic sleep in case e.g. they fall asleep themselves.

Add more options: 1.5, 2, 2.5, 3 hours (or maybe even more); it may be more practical to use sliders rather than drop-down lists to set the times.

PR #6900, opened 2017-09-13
Update cs_power.py

Schleehauf:
Allow for longer times till the system goes to sleep - added 2 and 3 hours. 
This is useful when listening to albums ( > 60 min) without having to move the mouse or watching movies. 
Maybe this could/should be fixed else where for all different media/video players and maybe browsers (html5 or flash) but given the simplicity of this solution I guess this is useful already - See feature request  #6185

I tested the "fix" (actually just additional options) but the **translation(s) for "hours" are missing** afterwards. 
**Where or how can I add translations?**

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py
@@ -40,6 +40,8 @@ SLEEP_DELAY_OPTIONS = [
     (1800, _("30 minutes")),
     (2700, _("45 minutes")),
     (3600, _("1 hour")),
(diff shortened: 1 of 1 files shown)


itzexor:
translations are handled on launchpad: https://translations.launchpad.net/linuxmint/latest/+pots/cinnamon/

but the new string won't show up until after this lands and the pot files are regenerated

Schleehauf:
Thanks for the quick reply!
So I am done for the moment - I am not responsible for providing any translation to get this merged, right?

mtwebster:
No.  Thanks

Issue #6324, opened 2017-02-26
Power management applet - showing wrong power status


eastcat:
Using Cinnamon 3.2.7 (with Mint 18.1) on Asus UX410UAK laptop. This laptop will not charge the battery until its level drops below a given threshold.

When the device is plugged in but not charging the battery, the power applet shows "Using battery power" status and the same icon as if the laptop was running on battery. It should be something like "On AC power (not charging)" with a distinct tray icon.

Screen capture of the displayed status:

![battery_applet_wrong_status](https://cloud.githubusercontent.com/assets/26039869/23340265/1f945a46-fc2b-11e6-8478-589f7094bb14.png)

Commands showing the actual power status (AC adapter online):

```
mint@mint` ~ $ acpi -a -b -i
Battery 0: Discharging, 96%, discharging at zero rate - will never fully discharge.
Battery 0: design capacity 4240 mAh, last full capacity 4123 mAh = 97%
Adapter 0: on-line
```

```
mint@mint ~ $ upower -i /org/freedesktop/UPower/devices/line_power_AC0
  native-path:          AC0
  power supply:         yes
  updated:              Sun 26 Feb 2017 01:21:35 PM UTC (1708 seconds ago)
  has history:          no
  has statistics:       no
  line-power
    warning-level:       none
    online:              yes
    icon-name:          'ac-adapter-symbolic'

```

```
mint@mint ~ $ upower -i /org/freedesktop/UPower/devices/battery_BAT0
  native-path:          BAT0
  vendor:               ASUSTeK
  model:                ASUS Battery
  power supply:         yes
  updated:              Sun 26 Feb 2017 01:49:35 PM UTC (29 seconds ago)
  has history:          yes
  has statistics:       yes
  battery
    present:             yes
    rechargeable:        yes
    state:               discharging
    warning-level:       none
    energy:              45.531 Wh
    energy-empty:        0 Wh
    energy-full:         47.013 Wh
    energy-full-design:  48.336 Wh
    energy-rate:         0 W
    voltage:             11.4 V
    percentage:          96%
    capacity:            97.2629%
    technology:          lithium-ion
    icon-name:          'battery-full-symbolic'
```

furui:
Similar result from my Asus GL702VM running Cinnamon 3.2.7 on Linux Mint 18.1.

```
user@GL702VM ~ $ uname -a
[...]

PR #6913, opened 2017-09-18
[WIP] Fix UPower state not refreshing

iia:
The issue seems to be resulting from this UPower issue, https://bugs.freedesktop.org/show_bug.cgi?id=79200. Main reason being battery state not updated and reported by UPower when charger is connected/disconnected but the charger connected/disconnected events are fired and handled.

As a solution, for each charger connected/disconnected event I save a list of available devices and schedule a manual UPower DBUS Refresh() method call on these devices after 2 seconds from the event occurrence. As a result the battery status is refreshed each time charger is connected/disconnected and the state is properly updated on the applet.

Here is the actual reported issue, https://github.com/linuxmint/Cinnamon/issues/6324.

### DIFF
--- a/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
@@ -10,11 +10,14 @@ const Pango = imports.gi.Pango;
 const Main = imports.ui.main;
 const Settings = imports.ui.settings;
 const GnomeSession = imports.misc.gnomeSession;
+const Mainloop = imports.mainloop;
 
 const BrightnessBusName = "org.cinnamon.SettingsDaemon.Power.Screen";
 const KeyboardBusName = "org.cinnamon.SettingsDaemon.Power.Keyboard";
 
 const PANEL_EDIT_MODE_KEY = "panel-edit-mode";
+
+const DELAY_SEC_REFRESH_DEVICES = 2;
 
 const UPDeviceType = {
     UNKNOWN: 0,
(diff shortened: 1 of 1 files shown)


mtwebster:
I'm really not keen on having everything reconstructed every 8 seconds, this will eat up processor and memory.  Was there a reason the initial commit wasn't reliable?

Aren't we getting a `g-properties-changed` anyhow?  It's just that the on-battery property hasn't been updated by this time.  Our properties changed because we added a device (ac brick?) - couldn't we just delay the entire devicesChanged callback, and include your Refresh?

around line 298, instead of this:
```
this._proxy.connect("g-properties-changed", Lang.bind(this, this._devicesChanged));
```
more like this:

```
...
this._refreshId = 0;

this._proxy.connect("g_properties-changed", Lang.bind(this, function() {
    if (this._refreshId > 0) {
        Mainloop.source_remove(this._refreshId);
        this._refreshId = 0;
    }

    this._refreshId = Mainloop.timeout_add_seconds(DELAY_SEC_REFRESH_DEVICES, Lang.bind(this, function() {
        this._refreshAllDevice();
        this._devicesChanged();

        this._refreshId = 0;
        return false;
    }));
}));
...
```
This would listen for g-properties-changed, and set up the two-second delay to refresh devices then call devicesChanged.  If another properties-changed comes in during that two-seconds, the timer starts over (this prevents a bunch of these signals coming all in a cluster and causing problems with too many refreshes one after the other.

Does the dbus refresh call trigger g-properties-changed itself?  Or does it just make sure the properties are current?  If it's helpful you can see which properties changed in the g-properties-changed callback (though admittedly a pain in the butt because they're GVariants).

I'm not in any way saying my suggestion there is *the* way to go, I haven't tested anything.  I'm just definitely against an infinitely recurring timer - hopefully there's a way we can get back to just reacting better to the proxy signals.

iia:
Hi,

I agree that always refreshing on a fixed interval is wasteful. This is why initially I didn't implement the solution that way. But the problem seems to be that occasionally when power is connected or disconnected the properties changed signal isn't triggered. This could be a laptop/charge controller specific issue, I'm not sure. In this case refreshing and updating states in fixed interval seemed reliable.

I tried your suggested solution which is reasonable and seems to work. It is also less wasteful.

clefebvre:
Hi @mtwebster, @iia can we get this PR adapted so it can get into 3.6?

PR #6931, opened 2017-10-07
st-private: Update shadow APIs for Cogl deprecations

brownsr:
https://github.com/GNOME/gnome-shell/commit/1c1f63a7d7a5f0be5a796679d029a6b0c600b838#diff-578ccc5ce70c4a11d7f30e3f38002f0bR168
also
st: Adapt the mutter implementation for create_texture_material
https://github.com/GNOME/gnome-shell/commit/05f9f991d8d5796b77def6c096a36786b5307351#diff-578ccc5ce70c4a11d7f30e3f38002f0b
This was adapted to make use of the static cogl context in st-cogl-wrapper

No regressions that I can see.
Presumably for Cinnamon 3.8 at this stage in the cycle

### DIFF
--- a/src/st/st-cogl-wrapper.c
+++ b/src/st/st-cogl-wrapper.c
@@ -15,8 +15,8 @@ hardware_supports_npot_sizes (void)
     if (cogl_context != NULL)
         return supports_npot;
 
-    supports_npot = cogl_has_feature (st_get_cogl_context(),
-                                      COGL_FEATURE_ID_TEXTURE_NPOT);
+    cogl_context = st_get_cogl_context();
+    supports_npot = cogl_has_feature (cogl_context, COGL_FEATURE_ID_TEXTURE_NPOT);
 
     g_message ("cogl npot texture sizes %s", supports_npot ? "SUPPORTED" : "NOT supported");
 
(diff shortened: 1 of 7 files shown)


src/st/st-cogl-wrapper.c
```diff
@@ -12,20 +12,27 @@ static gboolean supports_npot = FALSE;
...
-    supports_npot = cogl_has_feature (cogl_context, COGL_FEATURE_ID_TEXTURE_NPOT);
+    supports_npot = cogl_has_feature (st_get_cogl_context(),
+                                      COGL_FEATURE_ID_TEXTURE_NPOT);
 
     g_message ("cogl npot texture sizes %s", supports_npot ? "SUPPORTED" : "NOT supported");
 
     return supports_npot;
 }
 
+CoglContext * st_get_cogl_context(void)
```
JosephMcc: Return type and function name go on separate lines. Look at every other function in this file.

brownsr:
I need to check the use of a static for context is thread safe.

melmelissa:
and the bug of the menu that behaves in a strange way, it will be corrected him ??
http://pix.toile-libre.org/upload/original/1514309481.png

or is it still a unique and exceptional bug, as for boot - reboot crashes ??
and it will be left out because not serious for you??

PR #6964, opened 2017-10-28
Move inhibit switches to power and notification applets

germanfr:
It makes more sense to have these functionalities in their respective applets instead of in a separate misc. icon. Also in consistency with the sound applet with its respective mute buttons in its context menu.

_Example: right sound applet, left and center notifications and power applet now respectively._
screenshot from 2017-10-28 17-47-43


collinss:
This doesn't make sense to me. I don't use the power applet on my desktop - and it doesn't even show an icon since I don't have a battery for anything. This means I would be unable to access these settings. I also don't usually use the notification applet (because I don't like my tray filling up with notifications that I've already seen), but this doesn't stop my computer from showing notifications. Also, there is a setting to hide the notification applet when the tray is empty, meaning there are at least 2 different situations in which a user would be unable to temporarily disable notifications without significant extra hassle.

IMO this change would make these features less accessible to a certain segment of the Cinnamon users without any real gain for anyone.

germanfr:
That's a good point @collinss I didn't think about. But what about adding the power applet to your panel to control the power in place of the inhibit applet? I don't know, this is just a suggestion. What I mean is that the inhibit applet has no purpose by itself, it's more like a collection of features of other applets that can be enabled/disabled if used or not.

The gain here would be more organized functionalities and one less icon in the already crowded toolbar.

collinss:
I could 'add' the power applet to my panel, but it wont show anything because I don't have a battery or a brightness controllable screen. There is no applet to click.
As for the inhibit applet, the purpose is to allow the user to (temporarily) quickly and easily disable several features (such as turning off the monitor and showing notifications) when you don't want them (ie. watching a movie, giving a presentation, etc.) In other words, these 2 switches are usually used together, so it makes more sense to have them together than apart.
All that being said, I don't really see any reason not to have those switches available in the respective applets (other than potential redundancy of settings), but IMO getting rid of the inhibit applet is a bad idea.

ManIVIctorious:
I think you have both a good point, would it be against any design principles to have the settings in both places? So you could add the inhibit applet and would have the functionality without it

germanfr:
I think keeping the inhibit applet is a good idea too. There wouldn't be redundancy of settings because the use cases are mutually exclusive (if you have the power applet you don't use the inhibit applet). It would be perfect if there was a way to reuse the code in both applets.

brownsr:
Sorry, I also don't get it.  If you are in the power applet you are only a click away from the power settings where you can make the power settings whatever you want. And the trouble about putting anything in the notifications applet is that it only shows when there are notifications.  The inhibit applet was invented as a sort of global 'do not disturb' switch. These things make sense in that context.

JosephMcc:
Also, if I'm not mistaken, the inhibit applet shows an indication of when an application like a movie player is doing things to inhibit things from happening. So I don't really feel like the other options are too out of place. I also agree with the others. Both the power applet and notifications applet can be hidden for many reasons and leave these inaccessible. I can't really see what the big problem would be with having them available in both places.

clefebvre:
Hi,

Nice idea, but I think @collinss and @JosephMcc summarized the issues quite well, it wouldn't work in some cases and it lacks information the inhibit applet shows already.

I use that inhibit applet a lot personally.. and I've no power applet here (no battery) and I never go into the notifications (which don't always show up btw).

We can revisit that idea if we decide to work on some mega-status applet... like the kind of things we see in GNOME/Budgie.

PR #6978, opened 2017-11-04
TreeListWidgets.py: edit elements with double click

germanfr:
Double clicking in an row launches the edit dialog as when clicking the edit button.


brownsr:
Could you point out which component this change would be visible in ?  It's not obvious to me just how to test this.

germanfr:
Oh yes. It's the new list widget of xlet-settings. For example found in the Stark Menu settings.

germanfr:
Wait, I found an issue when the list is empty.

germanfr:
I think it's ready. It now uses a signal that is meant for this.

PR #6988, opened 2017-11-06
Improving a piece of code in cinnamon-settings.py

abdouskamel:
Hello everybody. I just want to propose a small improvement of a piece of code.

The removed code sorts the array _unsortedSidePages_ using the name of the side pages as a key. To do so, he first extracts those names and stores them in a new array, sorts this array, and retrieves the elements from _unsortedSidePages_ (lines 263-267).

I propose to replace all of this by one line. This line sorts directly the elements of the original array by name, without extracting those names. It is smaller and more efficient.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
@@ -258,16 +258,7 @@ class MainWindow:
                 self.unsortedSidePages.append((samodule.sidePage, samodule.name, samodule.category))
 
         # sort the modules alphabetically according to the current locale
-        sidePageNamesToSort = map(lambda m: m[0].name, self.unsortedSidePages)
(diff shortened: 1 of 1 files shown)


clefebvre:
Hi,

It's good, but it also requires a better comprehension and to be used to python specifics. This isn't the case with all developers, some of which jump across and use many languages on a daily basis. It also makes it harder to troubleshoot the loop and/or modify it.

We try to discourage this kind of mappings and one-liners and usually prefer traditional constructs. This one isn't too bad.. we're far from the kind of things we see in JS, but still...

abdouskamel:
Hi, thank you for response. The issue i wanted to fix is that the current code does some kind of double work. It acts like this : 

Suppose you have an array of tuples (name, nickname), and you want to sort this array according to the field "name" of the elements. What the current code do is : extract the names in a new array, sort this new array, and replace the result using a kind of searching. 

I've just removed the part "extract the name and replace them" by replacing it with "sort the elements directly in the array". If the problem is in the one-liner, i would be happy to write it in a more expressive way.

PR #6992, opened 2017-11-07
cinnamon-settings: move everything to python 3

collinss:
There are a couple of issues that I know of right now:
* ~~cs_backgrounds.py isn't saving/loading the thumbnail cache correctly (the pickle library changed significantly from python 2 to 3 and isn't backward compatible)~~ (fixed) for some reason the cache files generated by the py2 version aren't working in the py3 version, but the py3 version files work fine in py2, so I changed the code to dump the cache file if it fails, and then regenerate it. This will be backward compatible with older versions of Cinnamon. The only side-effect is that even with upgrades and migrated files, the images will be slow to load the first time.
* ~~The cinnamon settings window sometimes opens really small
Also, I'm not sure if there are any other changes that need to be made to the dependencies in debian/control~~ (Fixed)


collinss:
@Odyseus done

clefebvre:
@collinss I get this in cinnamon-settings-applets:

Failed to load extension Cinnamenu@json: 'GNUTranslations' object has no attribute 'ugettext'
Failed to load extension weather@mockturtl: 'GNUTranslations' object has no attribute 'ugettext'

Can you rebase this PR also (to trigger CI on it).

clefebvre:
python3-pam not present in LMDE 3.

clefebvre:
I still get the same error:

```
Failed to load extension sound150@claudiux: 'GNUTranslations' object has no attribute 'ugettext'
Failed to load extension Cinnamenu@json: 'GNUTranslations' object has no attribute 'ugettext'
Failed to load extension weather@mockturtl: 'GNUTranslations' object has no attribute 'ugettext'
```

clefebvre:
These applets are installed but do not appear in the list with your PR.

PR #7025, opened 2017-11-16
[3.8] Code clean up on systray.js, workspace.js, and workspacesView.js

icarter09:
Removed unnecessary code. Removed unused parameters. Added spaces for better readability for conditionals. Added missing semicolon.

### DIFF
--- a/js/ui/accessibility.js
+++ b/js/ui/accessibility.js
@@ -58,7 +58,7 @@ A11yHandler.prototype = {
             this._signalManager.disconnect("bell");
     },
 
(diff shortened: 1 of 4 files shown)


clefebvre:
I'm not sure I agree with removing arguments in callbacks. You couldn't do that in C or Python and I'm not keen on language-specifics. Also, it's good to know what arguments you are called with, whether or not you're using them right now.

icarter09:
@clefebvre Thank you for explaining. I'll keep that in mind for the future to leave them be.

PR #7036, opened 2017-11-17
 Fix custom format not working and add tooltip 

RavetcoFX:
Fixes
 "NEWLY INTRODUCED BUG:
After updating to Cinnamon 3.6.3 a new bug is introduced that screws up the calendar’s time and date appearance if it is set to “use a custom date format.” I’ve had the same result after running the updates on two different computers and also in a virtual box hosted in the first of the two. I always set my time and date format to %a %e %b %Y @ %H:%M:%S. It displayed fine on both machines and in the vbox before I ran the update."


mtwebster:
What am I missing?  The format works fine for me.

PR #7044, opened 2017-11-18
dbusMenu.js: Formatted code to allow for better readability. 

icarter09:
Added spaces when necessary.

Fixed indentation to show the logic being implemented by conditional statement.

Added parenthesis to show order of logic in conditional statement.

### DIFF
--- a/js/ui/dbusMenu.js
+++ b/js/ui/dbusMenu.js
@@ -467,7 +467,7 @@ DBusClient.prototype = {
 
     sendEvent: function(id, event, params, timestamp) {
         if (this._proxyMenu) {
-            if (!params) 
+            if (!params)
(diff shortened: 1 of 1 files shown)


brownsr:
there's a trailing space on line 470, and the formatting change just after that does not look right. Not sure what the change on line 138 brings.

icarter09:
The change on 138 just allows for viewing of the concatenation that is taking place.

The trailing space on line 470 is my bad. Thanks for the heads up on that.

The formatting after that is to align the logic that is being implemented by the conditional.

icarter09:
Thanks for the heads up @brownsr. I removed the trailing space and removed the unnecessary spacing as well.

PR #7046, opened 2017-11-18
st-theme.c: st_theme_resolve_url could assert with a theme issue

brownsr:
Changed to how it was before - gives a warning and returns null.  Much friendlier.


clefebvre:
https://github.com/linuxmint/Cinnamon/issues/7021

Issue #7076, opened 2017-11-25
Information about GNOME Online Accounts panel is still English when use Hungarian language


KAMI911:
```
 * Cinnamon version
Cinnamon 3.6.6
 * Linux Mint 18.3 beta
 ```

**Issue**
The Information about GNOME Online Accounts panel is still in English however the Hungarian translation is complete.
![google_acc_hu](https://user-images.githubusercontent.com/5292264/33227968-c92e1db2-d1af-11e7-9227-0010eb747d08.png)

**Steps to reproduce**
0) Use Hungarian language
1) Go to system settings
2) Select Online accounts
3) Push "Information about GNOME Online Accounts" button

**Expected behaviour**
Since these strings are fully localized to Hungary it should be in Hungarian, not English.

**Other information**

NikoKrause:
Confirmed. Created a PR to fix this.

PR #7088, opened 2017-11-28
Online Accounts: Use translations

NikoKrause:
The help dialog of Cinnamon Online Accounts was translated, but it is not used currently.
Tell the gladefile, where to look for translations.

Fixes https://github.com/linuxmint/Cinnamon/issues/7076


NikoKrause:
This window doesn't use the translations:
screenshot-window-2017-11-28-144301

PR #7111, opened 2017-12-01
Notifications: Add support for a bottom position option

### LINKED ISSUE #4892: Moving Notification pop-up on Cinnamon desktop
Hello,
Greetings from Turkey.
As I know, users can change notifications pop-up position in MATE but we can't in Cinnamon.
It always shows at the top-right corner but I would like to see at right-bottom area, over on clock, sound applets for example. Many of users want to change the place too.
In previously versions of Cinnamon's, users can change this situation via to modify Cinnamon.css but in 17.2 we can't even modify.
If you give some permissions to modify Cinnamon.css it would be great for us.
If you add an option to user interface of Notification Settings can change it like in MATE, it would be greater than.

Thanks in advance.
Best Regards,
dirent

jaszhix:
Depends on: https://github.com/linuxmint/cinnamon-desktop/pull/100

Closes #4892

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_notifications.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_notifications.py
@@ -61,6 +61,9 @@ class Module:
         switch = GSettingsSwitch(_("Have notifications fade out when hovered over"), "org.cinnamon.desktop.notifications", "fade-on-mouseover")
         settings.add_reveal_row(switch, "org.cinnamon.desktop.notifications", "display-notifications")
 
+        switch = GSettingsSwitch(_("Show notifications on the bottom side of the screen"), "org.cinnamon.desktop.notifications", "bottom-notifications")
+        settings.add_reveal_row(switch, "org.cinnamon.desktop.notifications", "display-notifications")
+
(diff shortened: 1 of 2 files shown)


Marcosdw:
Hello,

i agree totaly

-------
Sorry, i use a oneline translator French/English

MegaV0lt:
When willit be released?

jaszhix:
Its for CInnamon 3.8, so probably sometime next spring/summer.

Marcosdw:
Is there an alternative before cinnamon 3.8?

jaszhix:
Not that I know of. If you are feeling confident in editing code and are okay with possibly having to re-install Cinnamon, can try this:

Look at the PR, in `js/ui/messageTray.js`, change [this line](https://github.com/linuxmint/Cinnamon/pull/7111/files#diff-065d2df89b17629dfc9c8771310bd705R1477) to `true` and remove this [block of code](https://github.com/linuxmint/Cinnamon/pull/7111/files#diff-065d2df89b17629dfc9c8771310bd705R1478) (3 lines). That should get it working on Cinnamon with CJS 3.4+. Copy the file to `/usr/share/cinnamon/js/ui/messageTray.js`. Restart Cinnamon. If there's a point release and you upgrade, the change will be lost.

PR #7130, opened 2017-12-08
Fix cinnamon-js doc build w/ gtk-doc 1.27

itzexor:
<s>Can't think of a better way to do this. The gen_doc_ran file should probably be cleaned on make clean, but I don't know how to do that. I initially made gen_doc_ran a .PHONY target, but that made it regenerate the docs during make install.</s>


itzexor:
This is caused by a new build target added to gtk-doc.make which is included in our makefiles.

https://github.com/GNOME/gtk-doc/commit/c08cc78562c59082fc83b55b58747177510b7a70

itzexor:
Ok I think I learned enough about automake/gtk-doc/etc to make it work again in a slightly cleaner way than my original patch. I have rebased it and would appreciate testing.

clefebvre:
@itzexor please rebase to trigger CI.

PR #7134, opened 2017-12-08
window-list: unique style class and new preview options

itzexor:
We now use the the "window-list-preview" style class for the preview
in order to allow fine grained control instead of sharing the window
switcher's style class.

Adds support for user to disable the preview label and to change the
thumbnail size from 75% to 150% of default.

<s>The thumbnail may have some aspect ratio issues right now, and the default theme may need a few tweaks, but this is the basic idea.</s>

Support will need to be added to all existing themes. Until then the default theme as shown <s>here</s> below will be displayed.

### DIFF
--- a/data/theme/cinnamon.css
+++ b/data/theme/cinnamon.css
@@ -1274,6 +1274,17 @@ StScrollBar StButton#vhandle:hover {
     background-gradient-end: rgba(255,144,144,0.5);
 }
 
+.window-list-preview {
+    background: rgba(80,80,80,0.8);
+    border: 2px solid #a5a5a5;
+    border-radius: 8px;
+    font-size: 9pt;
+    color: white;
+    text-shadow: black 0px 0px 2px;
+    padding: 8px;
+    spacing: 4px;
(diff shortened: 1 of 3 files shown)


itzexor:
I will need to edit the commit message to include that the icon has also been removed.

itzexor:
Fixed up the theme and the slight aspect ratio error.

current before/after:
before-after

brownsr:
Looks good to me.  I added support for this CSS to one of my themes on the assumption it gets pulled.  Gives another theme to check against, anyway.

itzexor:
I have squashed this and reworded the commit message. Should be ready to go if there are no objections.

Lucapz:
@itzexor 

Does this mean it's not going to happen?

itzexor:
It hasn't been approved or rejected yet. The failing check is a false positive, and the conflict is a simple deletion that I haven't gotten around to yet.

Lucapz:
Cool , thanks for explaining. :)

PR #7149, opened 2017-12-20
Create new cinnamon-xlet-makepot utility for automattically extractin…

collinss:
…g strings for translations. This file replaces the old cinnamon-json-makepot which was specifically designed to extract translatable data from metadata.json and settings-schema.json. However, that file was not replaced because there are commonly used scripts that rely on it.

The following changes were made to the cinnamon-xlet-makepot file in terms of functionality:
* A folder is now passed as the last arguement to the script. The old script used the cwd instead. This gives the script a little more flexability. All necessary information is extrapolated based on the folder path (uuid, po folder, etc). In addition, the <applet-repo>/<uuid>/ or <applet-repo>/<uuid>/files/<uuid>/ folder may be used.
* Javasript files are now scanned by default. They can still be skipped by adding the -j or --skip-js flags in the command-line.
* Python files will now be scanned for translatable strings. As with Javascript, this can be disabled with the -p or --skip-python flags.
* The output pot file path is no longer required, though an alternate location for the pot file can still be specified using the -o or --potfile flags. The default is po/<uuid>.pot in the main xlet folder.
* The location comments that appear in the generated potfile for the extracted json strings were improved for clarity and readability.

In addition to these changes, cinnamon-xlet-makepot now uses python3 rather than python2.

The cinnamon-json-makepot script was turned into a wrapper around to the new cinnamon-xlet-makepot script in order to ensure the continued functioning of scripts that call it (such as the one in the applets repo).


files/usr/bin/cinnamon-xlet-makepot
```diff
@@ -0,0 +1,274 @@
...
+"""
+
+try:
+    import polib
+except:
+    print("""
+
+    Module "polib" not available.
+
+    Please install the package "python-polib" and try again
```
Odyseus: Hello, @collinss.

I have been "playing around" with the new `cinnamon-xlet-makepot` script and I found a couple of "inconsistencies".

Since the script now uses Python 3, the package to install now (in Debian based systems) is called `python3-polib`. Although, I thing that it should also be added to that message something like *or whatever name the package for the **polib** Python 3 module is called in your distribution*.

files/usr/bin/cinnamon-xlet-makepot
```diff
@@ -0,0 +1,274 @@
...
+        quit()
+
+    if args.out_file is not None:
+        pot_path = os.path.abspath(args.out_file)
+    else:
+        pot_path = os.path.join(dir, 'po', uuid + '.pot')
+
+    pwd = os.getcwd()
+    os.chdir(dir)
+    if args.js or args.py:
```
Odyseus: `args.py` should be changed to `args.python`.

files/usr/bin/cinnamon-xlet-makepot
```diff
@@ -0,0 +1,274 @@
...
+        try:
+            import subprocess
```
Odyseus: `subprocess` is already imported at the beginning of the script.

files/usr/bin/cinnamon-xlet-makepot
```diff
@@ -0,0 +1,274 @@
...
+        js_files = []
+        for root, dirs, files in os.walk(dir):
+            rel_root = os.path.relpath(root)
+            js_files += [os.path.join(rel_root, file) for file in files if file.endswith('.js')]
+        js_files
+        if len(js_files) == 0:
+            print('none found')
+        else:
+            print('found %i file(s)\n' % len(js_files))
+        os.system("xgettext --language=JavaScript --keyword=%s --output=%s %s" % (
```
Odyseus: The `xgettext` command is called even if there are no files listed in the `js_files` variable. I think that those this call should be made inside the `else` block (only when there are files listed) to avoid the following message.

```
xgettext: no input file given
Try 'xgettext --help' for more information.
```

files/usr/bin/cinnamon-xlet-makepot
```diff
@@ -0,0 +1,274 @@
...
+
+        py_files = []
+        for root, dirs, files in os.walk(dir):
+            rel_root = os.path.relpath(root)
+            py_files += [os.path.join(rel_root, file) for file in files if file.endswith('.py')]
+        if len(py_files) == 0:
+            print('none found')
+        else:
+            print('found %i file(s)\n' % len(py_files))
+        os.system("xgettext -j --language=JavaScript --keyword=%s --output=%s %s" % (
```
Odyseus: Same as in the previous call to the `xgettext` command.

files/usr/bin/cinnamon-xlet-makepot
```diff
@@ -0,0 +1,274 @@
...
+
+def do_install(uuid, dir):
+    podir = os.path.join(dir, "po")
+    files_installed = 0
+    for root, subFolders, files in os.walk(podir):
+        for file in files:
+            locale_name, ext = os.path.splitext(file)
+            if ext == '.po':
+                lang_locale_dir = os.path.join(LOCALE_DIR, locale_name, 'LC_MESSAGES')
+                GLib.mkdir_with_parents(lang_locale_dir, 755)
```
Odyseus: With the move to Python 3, maybe it isn't needed to use `GLib.mkdir_with_parents` and just use `os.makedirs(this_locale_dir, exist_ok=True)` here? One less module to import (**GLib**).

Furthermore, I think that the mode `755` passed to `GLib.mkdir_with_parents` is wrong. Since it's supposed to create folders with **read/write/exec** permissions for the current user, but instead it creates folders with only **write/exec** permissions for the current user. Using the mode `0o755` will create folders with the right permissions.

files/usr/bin/cinnamon-xlet-makepot
Odyseus: Additionally, the `-j` (`--join-existing`) argument should be passed to the `xgettext` command **only** if the POT file already exists. If the scanning of the JS files didn't find any strings to translate (or if the `--skip-js` argument is used), a POT file will not be created and executing the `xgettext` command with the `--join-existing` argument with a non existent file for the `--output` argument will simply fail.

files/usr/bin/cinnamon-xlet-makepot
```diff
@@ -0,0 +1,274 @@
...
+        dir = os.path.join(dir, 'files', uuid)
+
+    if args.install:
+        do_install(uuid, dir)
+        quit()
+
+    if args.remove:
+        do_remove(uuid)
+        quit()
+
```
Odyseus: I think that it is needed to add around here the following line:

```python
os.makedirs(os.path.join(dir, "po"), exist_ok=True)
```

This is needed because if there were no strings found by `xgettext` on JS and/or Python files (or the `--skip-XX` arguments were used) and the **po** folder doesn't exist inside the xlet folder, then the **polib** module will fail to create the POT file when calling `pot_file.save()`.

files/usr/bin/cinnamon-xlet-makepot
collinss: Good point. I'll still include the package name for debian-based systems, but I'll change the wording and update the package name.

files/usr/bin/cinnamon-xlet-makepot
collinss: Ah, good catch. I guess I didn't run into a case that would fail here in my testing.

files/usr/bin/cinnamon-xlet-makepot
collinss: Ah, yeah, I think I copied that line from the other script

files/usr/bin/cinnamon-xlet-makepot
collinss: hmm... yeah, that would simplify things a bit. I didn't realize that option had been added to python 3. That could also be used in cinnamon-settings to simplify things in a couple of places. I think I'll add it to my python 3 branch.

Odyseus: Yes, I remember seeing the use of `mkdir` (the GNU coreutils command) in one of the other Python files. I found it weird, knowing that the `os.makedirs` method also exists in Python 2, although without the `exist_ok` parameter.

I didn't mention this in my previous message. But, if it is important for the directories to be generated with a **755** mode, then the code would be `os.makedirs(this_locale_dir, mode=0o755, exist_ok=True)`. The default mode used by `os.makedirs` is `0o777`.

collinss: Yeah, I saw that in the docs. I already fixed it locally. I'm going to do a bit more testing and then I'll push my changes.

collinss:
@Odyseus There, I pushed my changes. If you have any other suggestions for how to improve this, let me know. I don't really have much experience with translations. I only did this because I made some changes to one of my applets (that has strings from python) and needed to regenerate my potfile. When I realized that python wasn't supported, I decided to add it, and kind of got carried away... :)

files/usr/bin/cinnamon-xlet-makepot
```diff
@@ -0,0 +1,283 @@
...
+    pwd = os.getcwd()
+    os.chdir(dir)
+    if args.js or args.python:
+        try:
+            output = subprocess.check_output(["xgettext", "--version"])
+        except OSError:
+            print("xgettext not found, you may need to install the gettext package")
+            quit()
+
+    os.makedirs(os.path.basename(pot_path), mode=0o755, exist_ok=True)
```
Odyseus: In this line, `dirname` should be used instead of `basename`. Otherwise, a directory with the POT file name is created inside the current working directory.

files/usr/bin/cinnamon-xlet-makepot
```diff
@@ -0,0 +1,283 @@
...
+
+    pot_exists = False
+    if args.js:
+        print("Scanning JavaScript files...")
+
+        js_files = []
+        for root, dirs, files in os.walk(dir):
+            rel_root = os.path.relpath(root)
+            js_files += [os.path.join(rel_root, file) for file in files if file.endswith('.js')]
+        js_files
```
Odyseus: I forgot to mention this line before. The `js_files` variable is floating there without doing anything. LOL

files/usr/bin/cinnamon-xlet-makepot
```diff
@@ -0,0 +1,283 @@
...
+
+try:
+    import polib
+except:
+    print("""\
+
+Module "polib" not available.
+
+You will need to install this module using your disrtibution's package manager
+(in debian-based syatems "apt-get install python3-polib")
```
Odyseus: Typos:

- **disrtibution's** to **distribution's**
- **syatems to** **systems**.

files/usr/bin/cinnamon-xlet-makepot
```diff
@@ -0,0 +1,283 @@
...
+import json
+import subprocess
+import collections
+import argparse
+
+LOCALE_DIR = os.path.join(os.path.expanduser('~'), '.local/share/locale')
+
+USAGE_DESCRIPTION = """\
+Parses files in an applet, desklet or extension's directories, \
+extracts translatable strings, and then generates a ,pot file for them.
```
Odyseus: Typo: **,pot** to **.pot** (There is a comma in front of "pot").

files/usr/bin/cinnamon-xlet-makepot
collinss: Apparently I don't have spellcheck enabled in sublime on my laptop...
Usually I use my desktop, but I'm having hard drive issues.

files/usr/bin/cinnamon-xlet-makepot
collinss: facepalm... I even looked it up to make sure I got the right one.

files/usr/bin/cinnamon-xlet-makepot
collinss: lol

files/usr/bin/cinnamon-xlet-makepot
```diff
@@ -0,0 +1,282 @@
...
+
+try:
+    import polib
+except:
+    print("""\
+
+Module "polib" not available.
+
+You will need to install this module using your distribution's package manager
+(in debian-based syatems "apt-get install python3-polib")
```
Odyseus: Typo: **syatems** to **systems**.

files/usr/bin/cinnamon-xlet-makepot
```diff
@@ -0,0 +1,282 @@
...
+            print('none found')
+        else:
+            print('found %i file(s)\n' % len(py_files))
+            command_args = [
+                'xgettext',
+                '--language=Python',
+                '--keyword=%s' % args.keyword,
+                '--output=%s' % pot_path
+            ]
+            if pot_exists:
```
Odyseus: There is a problem with the implementation of the **pot_exists** variable. This variable will be set to **True** whether strings to translate where found or not while scanning JS files. Running the script for the first time on an xlet **without** strings to translate on its JS files and **with** strings to translate on its Python files will fail due to the use of the `--join-existing` argument in conjunction with a path to a non-existent POT file.

I think that the easiest and *safest* way of doing this is by directly checking the existence of the POT file to decide if the `--join-existing` argument should be added or not.

```python
if os.path.exists(pot_path):
    command_args.append('-j')
```

collinss: The problem with doing it that way is that if the script is run on an applet that already has a pot file, it will just add to the existing one instead of regenerating it like I would expect. I don't think it matters though. AFAICT none of the xlets we have currently meet that criteria, and the -j option would work if there were.

Odyseus: > ...it will just add to the existing one instead of regenerating it like I would expect...

I was about to ask you why would you want to regenerate the POT file, but then I remembered how dumb `xgettext` can be sometimes (it doesn't remove obsolete strings and it doesn't update the line reference from a *msgid* whose source code has been moved). And I suddenly remembered why I remove the POT files when I'm doing POT updates in bulk.

👍

germanfr:
Hey @collinss, I'm getting an error with your script (also with the old one):
```
xgettext: Non-ASCII comment at or before ./extension.js:4.
          Please specify the source encoding through --from-code
```
The reason is that there is a comment in my extension with my name, which has a tilde. The solution would be to add `--from-code=UTF-8` to the args list to xgettext, since most languages use non-ASCII characters. Could you include that? I'd open an issue but since the old one is probably going to be replaced by this I thought asking for that here is a better idea.

collinss:
@germanfr fixed

clefebvre:
@collinss please modify makepot accordingly.

collinss:
@clefebvre sorry! I forgot to check if my changes affected the makepot script. It's fixed now. I noticed that some of the entries from the json files get reordered in the pot file, but it seems to do that for me on master too, so I suppose it's just the randomness of os.walk which already existed in the old version. If you want me to add it into my pr, I could probably do a sort on the files so that it's more consistent going forward.

Issue #5617, opened 2016-09-01
Title bar settings not universally honoured


Fjodor42:
I have 3 buttons on the left-hand-side of the title bar.

The following is a, probably incomplete, list of applications that do not seem to honour that setting:
- gnome-calculator
- eog
- gedit

Screenshots attached of gnome-terminal which does the right thing, and gnome-calculator, which doesn't.
![screen002](https://cloud.githubusercontent.com/assets/1410995/18162073/ec80527c-7034-11e6-95f8-bf434f6c414c.png)
![screen001](https://cloud.githubusercontent.com/assets/1410995/18162074/ec828fc4-7034-11e6-89a3-397741b350f3.png)

Fjodor42:
Hmmm, incidentally, it would seem that ```import``` from ImageMagick cannot see the Cinnamon tool bar - perhaps that should also be reported as an Issue?

Pjotr123:
What Mint version and what Cinnamon version are you using? Note that in Mint 18 you're supposed to use the default X-apps, which do submit themselves to the general settings:
https://www.linuxmint.com/rel_sarah_cinnamon_whatsnew.php

JosephMcc:
@Pjotr123 We get bug reports from a lot of non Mint Cinnamon users here. Even in Mint this is valid with applications like Calculator.

JosephMcc:
I could have sworn there was already an open issue for this problem but I can't find it now.

feren:
There is a fix for it, but **it needs to be implemented into Cinnamon Settings for it to work**:
There's a dconf value called `org.cinnamon.desktop.interface gtk-decoration-layout` and that is what controls window button layouts in headerbars on Cinnamon

Coeur-Noir:
Ubuntu 16.04 here with Unity.
Cinnamon installed from embrosyn ppa.
CSD windows do not have buttons on left : 
![capture du 2017-01-15 17-47-18](https://cloud.githubusercontent.com/assets/1895947/21964369/bbf676b4-db4a-11e6-9fad-38621bbe511d.png)

clefebvre:
Fixes:

https://github.com/linuxmint/cinnamon-desktop/commit/2c67d16f5a62f95c20a7ea7f19a1315c774f56a4
https://github.com/linuxmint/muffin/commit/89d1f65a48b7f978506d5de475dd5f58197efd79
[...]

Issue #5136, opened 2016-03-16
GNOME application do not respect Windows button settings


jtlpa:
If you open Windows and set window-buttons to just one button on left side, many GNOME default packages do not respect the setting (all three buttons remain on the right side).

![screenshot from 2016-03-16 15 13 08](https://cloud.githubusercontent.com/assets/11615699/13827326/ddd1383e-eb89-11e5-923d-ec6dd417bfc3.png)

Mowalle:
I had this problem too a few days ago. You can work around this by changing the respective key in `gsettings`:
`gsettings set org.cinnamon.desktop.interface 'close,minimize,maximize,menu:'` 
(everything before the `:` will be on the left side). Or use `dconf-editor` to set the same key via GUI.

I think setting the window buttons in Cinnamon Settings should maybe change this key as well as a solution?

jtlpa:
Mowalle
Thanks for quick reply. I tried your recommendation, but gsettings give error that no **key-name** was supplied. Could you double check your fix for key and key-name.

Thank you
JTL

jtlpa:
Whoops, hit the wrong button, didn't mean to close
JTL

Mowalle:
Right I'm stupid, sorry...

The command should be `gsettings set org.cinnamon.desktop.interface gtk-decoration-layout 'close, minimize, maximize, menu:'`. You can of course sort them differently if you want.

leigh123linux:
Gnome apps are broken by design so please close this report and bug the gnome devs instead.

jtlpa:
Mowalle
Thanks, but didn't work. Here is terminal output.
gsettings set org.cinnamon.desktop.interface gtk-decoration-layout 'close,minimize,maximize,menu:'
No such key 'gtk-decoration-layout'

Mowalle:
@leigh123linux I don't think this issue should be closed, as it has to do with a Cinnamon-specific key not being set. 

@jtlpa That is weird, I just double checked it and it works fine for me. Which distro and Cinnamon version are you running?

jtlpa:
Hi Mowalle
I will leave the issue open and agree with you on that. As to the version:
Debian Jessie 8.3 64bit, Cinnamon 2.2.16

Hope that helps

collinss:
[...]

PR #7162, opened 2017-12-21
[cinnamon-settings] Set org.cinnamon.desktop.interface gtk-decoration-layout

tanderson-google:
This should fix issue 5136 and issue 5617

To test, run gtk3-widget-factory and change the titlebar button layout.  The app should immediately update it's layout.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_windows.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_windows.py
@@ -145,6 +145,11 @@ class TitleBarButtonsOrderSelector(SettingsBox):
         self.settings = Gio.Settings.new(self.schema)
         self.value = self.settings.get_string(self.key)
 
+        self.gtk_schema = "org.cinnamon.desktop.interface"
+        self.gtk_key = "gtk-decoration-layout"
+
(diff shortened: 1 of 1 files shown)


tanderson-google:
https://github.com/linuxmint/Cinnamon/issues/5136
https://github.com/linuxmint/Cinnamon/issues/5617

tanderson-google:
This is needed because csd-xsettings reads org.cinnamon.desktop.interface gtk-decoration-layout:
https://github.com/linuxmint/cinnamon-settings-daemon/blob/686fa5df0fd44e245ee34012b4a87a8cfd5332a7/plugins/xsettings/csd-xsettings-manager.c#L356

Now that I think about it, this patch will only help users who change the setting, but it won't retroactively fix the existing layouts.  Could we just change csd-xsettings to read from org.cinnamon.muffin button-layout instead?

clefebvre:
We should make muffin, csd and cinnamon all point to the same key, ideally one that is in cdesktop. Thanks for pointing this out. Fix on the way.

clefebvre:
Fixed by

linuxmint/cinnamon-desktop@2c67d16
linuxmint/muffin@89d1f65
linuxmint/cinnamon-settings-daemon@073fa6d
5c9848a

PR #7187, opened 2018-01-10
sound@cinnamon.org: Volume control up to 150% of nominal volume. 

claudiux:
Hi,

I propose an improvement of the sound@cinnamon.org applet.

An option allows you to control the volume up to 150% of its nominal value.

The icon and the slider are coloured according to the volume when it is greater than 100%. I added the stylesheet.css file for that.

New option in Settings:
sound-settings

From 0 to 100%: standard icons.
sound_100

From 101 to 115%: yellow icon.
sound_115

From 116 to 130%: orange icon.
sound_130

From 131 to 150%: red icon.
sound_150

Best regards,

Claudiux

### DIFF
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -156,7 +156,7 @@ VolumeSlider.prototype = {
     },
 
     _update: function(){
-        let value = (!this.stream || this.stream.is_muted)? 0 : this.stream.volume / this.applet._volumeMax;
+        let value = (!this.stream || this.stream.is_muted)? 0 : this.stream.volume / this.applet._volumeNominal;
         let percentage = Math.round(value * 100) + "%";
 
         this.tooltip.set_text(this.tooltipText + percentage);
@@ -164,7 +164,7 @@ VolumeSlider.prototype = {
         if (this.app_icon == null) {
             this.icon.icon_name = iconName;
         }
-        this.setValue(value);
+        this.setValue(value/this.applet.pcMaxVolume);
 
         // send data to applet
         this.emit("values-changed", iconName, percentage);
@@ -174,7 +174,7 @@ VolumeSlider.prototype = {
         if (isNaN(value))
             throw TypeError('The slider value must be a number');
 
-        this._value = Math.max(Math.min(value/this.applet.pcMaxVolume, 1), 0);
+        this._value = Math.max(Math.min(value, 1), 0);
         this._slider.queue_repaint();
     },
 
@@ -182,10 +182,10 @@ VolumeSlider.prototype = {
         let direction = event.get_scroll_direction();
 
         if (direction == Clutter.ScrollDirection.DOWN) {
-            this._value = Math.max(0, this._value - VOLUME_ADJUSTMENT_STEP/this.applet.pcMaxVolume)*this.applet.pcMaxVolume;
+            this._value = Math.max(0, this._value - VOLUME_ADJUSTMENT_STEP/this.applet.pcMaxVolume);
         }
         else if (direction == Clutter.ScrollDirection.UP) {
-            this._value = Math.min(1, this._value + VOLUME_ADJUSTMENT_STEP/this.applet.pcMaxVolume)*this.applet.pcMaxVolume;
+            this._value = Math.min(1, this._value + VOLUME_ADJUSTMENT_STEP/this.applet.pcMaxVolume);
         }
 
         this._slider.queue_repaint();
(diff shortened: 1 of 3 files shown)


claudiux:
Hi,

It seems to be OK.

Some new strings have to be translated in settings-schema.json:
"% of nominal volume"
"Maximum volume control"
"You can try to increase the volume of your sound card up to 150% of its nominal value. Use with caution. You can also set a value less than 100%, for example on a child's computer."

As an indication, I attach a .pot file and es.po, fr.po and it.po.

[es.po.txt](https://github.com/linuxmint/Cinnamon/files/1637551/es.po.txt)
[fr.po.txt](https://github.com/linuxmint/Cinnamon/files/1637552/fr.po.txt)
[it.po.txt](https://github.com/linuxmint/Cinnamon/files/1637553/it.po.txt)
[sound@cinnamon.org.pot.txt](https://github.com/linuxmint/Cinnamon/files/1637554/sound.cinnamon.org.pot.txt)

Best regards,

claudiux

claudiux:
Hi @NikoKrause,

(1) I have just solved the problem of size.
(2) A new option in settings allows to don't use colors.
(3) I prefer that there is a counter. I installed this applet on my daughter's computer (7 years old) and set the maximum volume at 70%. With a checkbox, this would no longer be possible.
(4) At 153%, the sound becomes very painful. I prefer to leave this margin. If this feature was really required, it would be easy to modify the settings-schema.json file.

The mark at 100% already exists in v1.2.0.

The v1.3.0 with all these new features is coming soon.

collinss:
Another issue with this pr is that you have a stylesheet included in the applet directory. The styling for system applets should instead be added to the default cinnamon theme (found in [data/theme/cinnamon.css](https://github.com/linuxmint/Cinnamon/blob/master/data/theme/cinnamon.css)) as well as the other mint cinnamon themes. Also, I would recommend using pseudo classes instead of having a separate style class for each state.

JosephMcc:
I'd have to agree with 2 issues pointed out by others. The margin at the left of the volume icon could be improved and the color changing part should be removed.

clefebvre:
Hi @claudiux, many thanks for working on this. This is something we committed to fixing for the next release. We weren't quite sure how to tackle it. Your ideas and implementation helped a lot.

We've been discussing this a lot lately. Your idea to set a max volume was excellent. We weren't quite sure how to implement the marker and magnetism in clutter, that was brilliant too.

We need a global setting rather than an applet setting because we want this to affect not only the applet, but the sound settings (in Cinnamon) and the media-keys OSD (in CSD).

I've got an implementation for this now and a couple of PRs. I'll link them here and close this one.

clefebvre:
Oh, I forgot to comment on the 153%... 

This is an arbitrary value in PulseAudio. It doesn't correspond to any particular device, it's just the percentage value which corresponds to a gain of +11dB.

https://freedesktop.org/software/pulseaudio/doxygen/volume_8h.html#a353b44010677b005abd390af732eb86e

https://github.com/pulseaudio/pavucontrol/commit/297af52ae5581f5373bb008f60c7aac858bb365a

Unless exactly +11dB is proved to be somewhat a key value in terms of acoustics, it really doesn't matter whether it's 153% or 150%... the former looking really weird from a user point of view.

So with that in mind, 150% makes more sense.

Theoretically you could set the volume to whatever you want. PA won't stop you from going over 153%... in practice you'll quickly get clipping and sound distortion, and you can also damage your hears and/or your speakers... so it's not a great idea to let people go over a certain value.

I've done like @claudiux in my implementation, I've used 150% as the upper limit.

clefebvre:
cdesktop: https://github.com/linuxmint/cinnamon-desktop/commit/d35987157f5d35c3b19db9d5f5adc19633f49589

clefebvre:
CSD: https://github.com/linuxmint/cinnamon-settings-daemon/pull/215

clefebvre:
Cinnamon (applet and settings): https://github.com/linuxmint/Cinnamon/pull/7238

claudiux:
Hi @clefebvre, @NikoKrause, @collinss and @JosephMcc,

I'm happy that parts of my code are right for you and that you're integrating them into Cinnamon.

For my part, I continue to develop the applet sound150@claudiux.

I considered the points of view and criticisms expressed here.

I deleted the stylesheet.css file; the colors are changed in the code; the style sheets of the themes are respected, including after a change of theme. The problem of margins is thus solved. In addition, a switch allows the user not to display the colors, if they do not suit him.

I made sure to replace the sliders with comboboxes; but I do not think that this solution is very elegant because it forces the user to make fixed choices.

The maximum percentage of the nominal volume is now determined by the code, but if it is still 153% and it is better to limit it to 150% (which I think from the beginning, as the name sound150 @ claudiux of my applet indicates it) that presents only little interest.

The time to redo the translations of the modified messages, and I make a commit and a PR of my applet on cinnamon-spices-applets.

Best regards.

PR #7196, opened 2018-01-13
[WIP] applet: set full colour icon size from theme if not in scale mode

brownsr:
At the moment turning scale mode off on panels should let sizes be fixed by the theme.  This isn't the case for full colour icons, which use a fixed default.  This change lets these icons use the theme default.  Further work will be needed on two applets which similarly use hard-coded defaults, panel-launchers and systray.  That could well be in a later PR once I've worked out how best to address that hard-coding.


clefebvre:
Please rebase.

Also include the applet changes in the PR if possible, and have PRs ready for any affected spices.

I'll mark this as WIP for now.

brownsr:
closing for now until I can work on this actively.

PR #7230, opened 2018-01-27
:D

harrycode:
You can remove the secret type, I just wanna see if you can implement the funny messages. Maybe randomize them or add more!


leigh123linux:
What does this commit do, I'm tempted to close as the description and commit messages suck.

leigh123linux:
temptation was to great

PR #7239, opened 2018-01-31
Optimize appFavorites

Connormiha:
This PR contains the following changes.
1) Optimize generating favorites map. Use `reduce` for array. Only 1 loop instead of 3 loops.
2) Use native functions like `Object.keys`, `Object.values` instead of `for loop`

### DIFF
--- a/js/ui/appFavorites.js
+++ b/js/ui/appFavorites.js
@@ -25,24 +25,18 @@ AppFavorites.prototype = {
     _reload: function() {
         let ids = global.settings.get_strv(this.FAVORITE_APPS_KEY);
         let appSys = Cinnamon.AppSystem.get_default();
-        let apps = ids.map(function (id) {
-                let app = appSys.lookup_app(id);
-                return app;
-            }).filter(function (app) {
-                return app != null;
-            });
-        this._favorites = {};
-        for (let i = 0; i < apps.length; i++) {
-            let app = apps[i];
-            this._favorites[app.get_id()] = app;
-        }
+        this._favorites = ids.reduce((favorites, id) => {
+            const app = appSys.lookup_app(id);
+            if (app) {
+                favorites[app.get_id()] = app;
+            }
+
+            return favorites;
+        }, {});
     },
 
     _getIds: function() {
-        let ret = [];
-        for (let id in this._favorites)
-            ret.push(id);
-        return ret;
+        return Object.keys(this._favorites);
     },
 
     getFavoriteMap: function() {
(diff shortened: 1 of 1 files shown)


js/ui/appFavorites.js
```diff
@@ -25,35 +25,26 @@ AppFavorites.prototype = {
...
     getFavoriteMap: function() {
         return this._favorites;
     },
 
     getFavorites: function() {
-        let ret = [];
-        for (let id in this._favorites)
-            ret.push(this._favorites[id]);
-        return ret;
+        return this._getIds().map((id) => this._favorites[id]);
```
germanfr: You would wan't to use `this._favorites` here because you don't need to copy the array to read the contents.

Connormiha: @germanfr  Sorry, I do not understand what you mean `copy the array`.
This method returned list of apps.

Original method creates new array
```js
let ret = [];
for (let id in this._favorites)
    ret.push(this._favorites[id]);
return ret;
```
From PR
```js
    return this. _getIds().map((id) => this._favorites[id]);
    //this._getIds() === Object.keys(this._favorites)
```
`_favorites` is map
```js
this._getIds() // list of ids, ['1', '2', '3']
.map() // to list of apps [app1, app2, app3]
```

germanfr: I mean that instead of
```js
this._getIds().map((id) => this._favorites[id])
```
you could do 
```js
this._favorites.map((id) => this._favorites[id])
```
because `this._getIds()` is basically copying the content of `this._favorites` if I understood it.

germanfr: Well no, I read it again and you're copying the id's array, which in any case is unnecessary. The original function was more efficient clear. The rest is fine.

Connormiha: `this._favorites` is Object and hasn't `map` method.
In this change we have 1 string instead of 4. And both methods have only 1 loop. Do you prefer classic for loop instead of functional methods?

germanfr: Yes I misunderstood that part (the code loses clarity with this). But anyways, i'ts not that I prefer imperative vs functional, it's that the new function generates a whole array of ids to loop over it and get the apps, while the old one just looped over the object and got the apps.

Connormiha: I just came up with an easier way ))
```js
    return Object.values(this._favorites)
```

Connormiha: `Object.values` form es2017 standart. Supports from firefox 47.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values
In which versions of the spider Monkey engine should the code work?

clefebvre: Hi @Connormiha, we're on mozjs52.

PR #7251, opened 2018-02-02
Speed up windows mapping

clefebvre:
The goal of this PR is to make it faster for Cinnamon to open windows.

Using the following test script:

```
#!/usr/bin/python3

import gi
gi.require_version('Gtk', '3.0')
from gi.repository import Gtk, GObject

import time, threading

# Used as a decorator to time functions
def print_timing(func):
    def wrapper(*arg):
        t1 = time.time()
        res = func(*arg)
        t2 = time.time()
        print('%s took %0.3f ms' % (func.__name__, (t2 - t1) * 1000.0))
        return res
    return wrapper

# Used as a decorator to run things in the background
def async(func):
    def wrapper(*args, **kwargs):
        thread = threading.Thread(target=func, args=args, kwargs=kwargs)
        thread.daemon = True
        thread.start()
        return thread
    return wrapper

# Used as a decorator to run things in the main loop, from another thread
def idle(func):
    def wrapper(*args):
        GObject.idle_add(func, *args)
    return wrapper

class MyApp ():

    def __init__(self):
        self.windows = []
        mainwindow = Gtk.Window()
        box = Gtk.Box()
        button = Gtk.Button()
        button.set_label("Open windows")
        button.connect("clicked", self.make_windows)
        box.pack_start(button, False, False, 0)
        button = Gtk.Button()
        button.set_label("Close windows")
        button.connect("clicked", self.close_windows)
        box.pack_start(button, False, False, 0)
        mainwindow.add(box)
        mainwindow.show_all()
        mainwindow.connect("destroy", Gtk.main_quit)

    @print_timing
    def make_windows(self, widget):
        for i in range(200):
            window = Gtk.Window()
            self.windows.append(window)
            window.show()

    @print_timing
    def close_windows(self, widget):
        for window in self.windows:
            window.destroy()
        self.windows = []


MyApp()
Gtk.main()
```

We look at the reported figures in the output and measure the time it takes (with a stopwatch) for the main window to defocus and the 200 windows to map themselves completely.

We observe a significant difference in performance in Cinnamon compared to other DEs.

Typical values: 

cinnamon: 4s and 22s
cinnamon without window-list: 2,7s and 10,5s
muffin: 1,3s and 10s
marco: 1s and 28s
metacity: 1s and 6s

The goal is to reduce both numbers as much as possible.

The first commit stops checking the window title and reduces the defocus time a bit.. from 22s to 20s.

The second commit simplifies the window list and reduces the defocus time from 20s to 12s.

### DIFF
--- a/files/usr/share/cinnamon/applets/panel-launchers@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/panel-launchers@cinnamon.org/applet.js
@@ -239,20 +239,18 @@ PanelAppLauncher.prototype = {
     },
 
     launch: function() {
-        let allocation = this._iconBox.get_allocation_box();
-        this._iconBox.width = allocation.x2 - allocation.x1;
-        this._iconBox.height = allocation.y2 - allocation.y1;
+        if (this.isCustom()) {
+            this.appinfo.launch([], null);
+        }
+        else {
+            this.app.open_new_window(-1);
+        }
         this._animateIcon(0);
-        if (this.isCustom()) this.appinfo.launch([], null);
-        else this.app.open_new_window(-1);
     },
 
     launchAction: function(name) {
-        let allocation = this._iconBox.get_allocation_box();
-        this._iconBox.width = allocation.x2 - allocation.x1;
-        this._iconBox.height = allocation.y2 - allocation.y1;
-        this._animateIcon(0);
         this.getAppInfo().launch_action(name, null);
+        this._animateIcon(0);
     },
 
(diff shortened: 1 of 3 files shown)


clefebvre:
With the addition of https://github.com/linuxmint/muffin/pull/308, we go down from 12s to about 7s.

clefebvre:
With the timeout_add to create the window list buttons, we're now down to 6s for defocus/map, and the creation delay goes from 4s to about 300ms.

PR #7252, opened 2018-02-03
Don't clear notifications when an app becomes focused

tanderson-google:
Apps like Chromium and Firefox create notifications with a reasonable
expectation that the user should see them before they disappear.  The
old behavior was to clear notifications when the app became focused.
If the user was away when notifications were generated, they would be
stored in the message tray.  If the user then focuses the app, all of the
notifications would disappear (even if the notifications had critical
urgency), never giving the user a chance to see them.

This behavior is not intuitive and can lead to loss of important notifications,
so this CL removes this behavior altogether.  Users can still use the "clear
all notifications" button in the tray.

### DIFF
--- a/js/ui/notificationDaemon.js
+++ b/js/ui/notificationDaemon.js
@@ -121,11 +121,6 @@ NotificationDaemon.prototype = {
         }
         setting(this, this.settings, "boolean", "removeOld", "remove-old");
         setting(this, this.settings, "int", "timeout", "timeout");
-
-        Cinnamon.WindowTracker.get_default().connect('notify::focus-app',
-            Lang.bind(this, this._onFocusAppChanged));
-        Main.overview.connect('hidden',
-            Lang.bind(this, this._onFocusAppChanged));
     },
 
    // Create an icon for a notification from icon string/path.
@@ -523,20 +518,6 @@ NotificationDaemon.prototype = {
             Config.PACKAGE_VERSION,
             '1.2'
         ];
-    },
-
-    _onFocusAppChanged: function() {
-        let tracker = Cinnamon.WindowTracker.get_default();
-        if (!tracker.focus_app)
-            return;
-
-        for (let i = 0; i < this._sources.length; i++) {
-            let source = this._sources[i];
-            if (source.app == tracker.focus_app) {
-                source.destroyNonResidentNotifications();
-                return;
-            }
-        }
     },
 
     _emitNotificationClosed: function(id, reason) {


clefebvre:
Please describe a use case, with browser and web-app, detailing how the current behavior is an issue and how the new one would help.

tomKPZ:
Here are some ways notifications could go missing:

1. Three web apps create notifications at around the same time.  The user clicks on the first one, which focuses Chrome.  The other 2 notifications never get shown and don't go to the message tray.

2. The user receives three notifications while they're away.  The log into their machine and focus Chrome.  They notice an icon disappear out of the corner of their eye.  It was the notification tray indicating there were 3 unread notifications.

3. Three web apps create notifications which all go to the message tray.  The user opens the message tray and notices they all have "OK" and "Cancel" buttons on them.  They want to click OK on all of them.  They click OK on the first one, which focuses Chrome, but also causes the other two to disappear.

By eliminating the unintuitive clear-notifications-on-focus behavior, all of these issues would go away.

clefebvre:
That's not what I mean. Please describe an actual use case this would solve, with the actual website and the content of the notifications.

tomKPZ:
In Google Calendar, notifications are typically generated 10 minutes before events.  Lets say there are 2 overlapping events: one is office hours with your manager that you usually don't attend.  The other is a real meeting that you are planning on attending.  Cinnamon displays notifications one-at-a-time, so lets say you see the office hours notification first.  You click on it to dismiss it, but this causes the meeting notification to disappear.  If users are relying on notifications as reminders, they may miss the meeting.

clefebvre:
That's a fair point, although I don't totally agree.

If you're in front of the computer you see both. If you're away, by the time you come back you see none, and your notification tray has both for you to see before you click anything.

Also, they both have the same purpose... to bring your attention to the calendar itself.

One last point, if one particular event isn't important and another one is, you'd expect the calendar to treat them differently. The important event would typically use a CRITICAL notification, so that it would remain on the screen for you to see, whereas the other one would simply end up in your tray.

beverloo:
Browsers display notifications on behalf of different websites. The fact that Chrome gets focused, either because the user manually launches the app or because they activate a notification from a particular site (e.g. Calendar), does not mean that they're aware of events that happened elsewhere. There is no 1:1 relationship between a notification shown by a browser and its conceptual sender.

Web Developers can show notifications not just from the visible tab, but also from backgrounded tabs and/or windows. In fact, there is no requirement for there to be a tab at all if they display a notification in response to a push message. (This is by design.)

If there is no foreground window or tab for a particular site (e.g. Twitter), how do we expect the user to be aware of activity if the notification gets dismissed when they just start the browser, or when they click on the notification of a completely different site?

clefebvre:
Isn't that the job of the browser itself?

Firefox improved for instance by showing which tab was emitting sounds lately. You also see Twitter indicate the presence of new messages in the tab title itself.

beverloo:
Yes, and we do, but that's orthogonal.

When a user returns to their computer with a particular goal in mind (for example to find a recipe) they'll launch the browser and it's reasonable to expect that they'll miss the notification icon in the tray—context from those notifications is now lost, even if the tab title might've been updated.

Furthermore, as I mentioned, there's no requirement for there to be a tab at all in either browser. Notifications _are_ the UI surface where such updates are portrayed. Giving the user the ability to interact with each individual notification at a time that's convenient for them is critical.

schachmat:
TBH, I'd rather see Chrome use a similar approach to X11 where windows can have an urgency hint set that goes away after selecting them. In chrome, tabs could have this urgency hint, highlighting them in a different color/flashing/whatever in the tab bar and when the user selects the tab the browser UI could show a "drawer" (similar to the notification shade pull down on android) that has the contents of the notification and optional interaction items. This would help to unify notification experience and between desktop chrome and android.

Additionally I can imagine chrome having just one system-native notification that draws the users attention to the chrome window if there are notifications available within chrome.

beverloo:
Thank you for the additional input. Unfortunately the fact that there doesn't have to be a tab for a website to display a notification makes such an association impractical.

Chrome used to ship its own notification center, also on Linux, but we removed this following vast amounts of feedback from users who felt that the duplication with the system's notification server was confusing at best. (This was supported by usage metrics.)

jaszhix:
Sometimes an app will be notifying about something that's occurring in a temporary manner, like a chat message. If the notifications disappears from the applet, the app is focused, and the user isn't alerted to it, they may never know it occurred. I think the whole point of native browser notifications is to make the browser more integrated - if the browser is expected to have its own notification system on top of the native notifications, it will be more complicated - more things to check and click through.

Maybe there needs to be a solution though to prevent the notifications from piling up too much, like a setting for notifications to be auto-pruned from the applet after a specified amount of time. I think that would be better than clearing on focus, because that assumes the user is going to always check the applet first, which isn't guaranteed.

clefebvre:
Hi everyone,

I'd like to thank you all for your patience and for your feedback. We've been talking about this and notifications a lot internally. It's not easy to think of what might work best for everybody. You're always afraid to wave off significant needs/use-cases.. and at the same time you don't want to fall into a niche and you certainly want it to work a certain way without requiring the user to fine tune settings he might or might not understand and certainly shouldn't need to worry about.

Note: this also got in: https://github.com/linuxmint/Cinnamon/pull/7336.

It introduces a close button on all notifications and removes the fade semi-transparency effect on hover. Clicking a notification closes it and focuses the caller app. Clicking the little close button on the other hand, only closes the notification, without focusing the caller app. This isn't directly related to this PR (it was done to fix the case where you don't want to lose focus on your current app or when you just want to dismiss a notification without being interrupted in your current activity).

tanderson-google:
Thanks for coming back to this, Clement.  Glad to see it merged!

Issue #7091, opened 2017-11-28
Feature Request: Color chooser for themes


Swiftpaw:
Are there any plans to have a more dynamic support for colors and themes?  Right now for example, the Mint-Y theme lacks the color variants that the Mint-X themes enjoy.  If there was an ability to select the "main color" and/or other colors for any theme instead of requiring separate themes, I think that would be a neat feature, and would let you get rid of all those duplicate themes that merely have color differences.

Also, having a "master theme" selector could also be nice, instead of having to go through five different options just to change to a new theme (or new color).

Just some thoughts for you all!  Mint rocks, keep up the great work, love supporting you all! :wolf:

JosephMcc:
> Are there any plans to have a more dynamic support for colors and themes? Right now for example, the Mint-Y theme lacks the color variants that the Mint-X themes enjoy. If there was an ability to select the "main color" and/or other colors for any theme instead of requiring separate themes, I think that would be a neat feature, and would let you get rid of all those duplicate themes that merely have color differences.

This isn't going to happen. There isn't an easy and straightforward way to do that with gtk3 themes. Mint-Y will get color variations at some point.

> Also, having a "master theme" selector could also be nice, instead of having to go through five different options just to change to a new theme (or new color).

Mate does it this way and I really dislike it. Not only that, judging by most of the screenshots I see people post, it seems mixing and matching is what a lot of users like to do.

Issue #3356, opened 2014-07-12
Allow themes to be customized


mintystark:
I am wondering if there is any word on allowing themes to be customized in the same since as Applets.  I am thinking that in the Theme Manager where you can see the themes and "Apply theme" that when the theme is selected there is a "Customize Theme".  Nothing fancy, just a simple settings popup like the Applets do. Same settings fields like text, checkbox, select, etc.  With the only difference of having a color picker.  Then an Apply button which would reload the theme.

The easiest thing would then be to have the values in the theme css file.
Something like $setting['setting_name'].

So like:
# main-background {

```
 background-color: $setting['mb_background_color'];
```

}

We would also need to be able to set defaults just like the Applets Settings have.

I am thinking of making an Applet that would do such a thing for specific themes, but don't want to waist my time if you in the works of having this.

Thanks

coryhilliard:
Is there a voting system in place for this feature?  I find this missing feature is THEE number one missing feature since the demise of Gnome 2.  I have built my own theme, but I would love to have just a simple pointy and clicky (as most users of operating systems would too).  Just because I can code CSS, doesn't mean it's fun all the time.

yro84:
well, where can I see a proper tutorial about how to customize the default cinnamon theme? I just want it to have the blue line above the name of the program window on the main panel, I just want this panel to be fully black (real black) and I just want to change the main selection colour to pick orange or something similar to a darker blue. Just simple things like that. Where can I get the how to? Atleast where can I have informations about these things I wanna change on the main theme?

cruftytheclown:
I would like to be able to control the transparency of individual panels.

Then I would use a normal non-transparent panel at the bottom of my screen, and a fully transparent panel at the top of the screen where I would put notifier applets in the top right corner.

germanfr:
@cruftytheclown this may be what you need: [https://cinnamon-spices.linuxmint.com/extensions/view/42](https://cinnamon-spices.linuxmint.com/extensions/view/42). The individual transparency is not a feature yet though.

rustedwolf:
If a color custom prime color selection would be implemented, that would be a very big plus for cinnamon. Then again it there would be no need for large preset list, like it was in Mint-X. Just select style > then select a color from preset or custom color. 
Right now Mint-Y theme (as I look in my settings) has only green color theme, so they won't match up with Mint-X icon other than green color presets...

JosephMcc:
[...]

Issue #3180, opened 2014-05-22
[Feature request] Change panel/menu color and opacity directly without changing theme


sergeypd:
It would be great to have such a possibility like in Mate or Xfce, because it is sometimes hard to find cinnamon theme, which would match your preference.

dalcde:
Special case of #3356. Closing

Issue #2022, opened 2013-05-13
tool to edit cinnamon themes and gtk3 themes


fcole90:
I think that it would be wonderful if as it was in the Gnome2 era it could be possible to edit the themes over the author's settings. Maybe starting with simple things like color chooser or similar.

PR #7255, opened 2018-02-04
[WIP] Support for environment colors in themes

germanfr:
Colors can be defined like if they were "variables" and picked by the user at runtime. This is done via a CSS function (similar to rgb) called "env". It can be used as `env("name")` where name is one of the supported "variables", inspired by the _getenv_ function in C. This can be used in borders, backgrounds
or anywhere a color is supported.

This only supports rgb colors, because typically the theme wants to play with the alpha channel rather than always using a fully opaque color. So the "env" function has an optional second parameter where you can define the alpha like in rgba. For example: `env("name", 0.5)`.

This behavior can be disabled by the user (and it's disabled by default), so that one can use the theme defaults if they want. To define a default color the theme just needs to make use of the fallback feature of CSS.

I have defined two "variables": _accent_ and _text_. _"Accent"_ is the color that stands out in the theme. Also exists _"text"_ so that one can pick a font color that is visible against the accent color if it's used as a background color.

This is my first contribution to the C part so it's going to be wrong. Specially names and descriptions suck, but I hope you get the idea. Everything is open to change, names, defaults, etc.

A more detailed usage example:
```css
.window-list-item-box:hover {
    background-color: #8fa876; /* Fallback, when user defined colors is disabled */
    background-color: env("accent", .8);

    color: #fafafa; /* Fallback, when user defined colors is disabled */
    color: env("text");
}
```

Then the user would come here and pick the colors, and the theme automatically updates.
screenshot from 2018-02-04 20-38-22

This opens a whole world of possibilities, like automatically setting the color based on the wallpaper via scripts.

Finally, I'd like to mention all the issues I found requesting something like this: 
#675 #2022 #3180 #3356 #7091

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_themes.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_themes.py
@@ -69,6 +69,18 @@ class Module:
             page = SettingsPage()
             self.sidePage.stack.add_titled(page, "options", _("Settings"))
 
+            settings = page.add_section(_("Desktop color options"))
+
+            invert = lambda b: not b
+            widget = GSettingsSwitch(_("Use default colors"), "org.cinnamon.theme.colors", "enable", map_get=invert, map_set=invert)
+            settings.add_row(widget)
+
+            widget = GSettingsColorChooser(_("Accent color"), "org.cinnamon.theme.colors", "accent")
+            settings.add_reveal_row(widget, "org.cinnamon.theme.colors", "enable")
+
+            widget = GSettingsColorChooser(_("Text color"), "org.cinnamon.theme.colors", "text")
+            settings.add_reveal_row(widget, "org.cinnamon.theme.colors", "enable")
+
             settings = page.add_section(_("Miscellaneous options"))
 
             widget = GSettingsSwitch(_("Show icons in menus"), "org.cinnamon.settings-daemon.plugins.xsettings", "menus-have-icons")
--- a/src/st/st-theme-node.c
+++ b/src/st/st-theme-node.c
@@ -616,7 +616,6 @@ get_color_from_context (CRTerm         *term,
                         ClutterColor   *color)
 {
   CRTerm *arg = term->ext_content.func_param;
-  gchar *name;
   const ClutterColor *envcolor;
 
   if (arg == NULL || arg->the_operator != NO_OP)
@@ -627,7 +626,7 @@ get_color_from_context (CRTerm         *term,
            arg->content.str->stryng &&
            arg->content.str->stryng->str)
     {
-      name = arg->content.str->stryng->str;
+      gchar *name = arg->content.str->stryng->str;
       envcolor = st_theme_context_get_color (context, name);
       if (!envcolor)
         return VALUE_NOT_FOUND;
@@ -637,17 +636,12 @@ get_color_from_context (CRTerm         *term,
       /* Alpha can be overriden with an optional second parameter */
       if (arg->next != NULL)
         {
-          double alpha;
-
           arg = arg->next;
-          if (arg->the_operator != COMMA)
-              return VALUE_NOT_FOUND;
-
-          if (arg->type != NUM_GENERIC)
-            return VALUE_NOT_FOUND;
-
-          alpha = CLAMP(arg->content.num->val, 0, 1);
-          color->alpha = color_component_from_double(alpha);
+          if (arg->the_operator == COMMA && arg->type == NUM_GENERIC)
+            {
+              double alpha = CLAMP(arg->content.num->val, 0, 1);
+              color->alpha = color_component_from_double(alpha);
+            }
         }
 
       return VALUE_FOUND;
--- a/src/st/st-theme-node.c
+++ b/src/st/st-theme-node.c
@@ -611,6 +611,52 @@ get_color_from_rgba_term (CRTerm       *term,
 }
 
 static GetFromTermResult
+get_color_from_context (CRTerm         *term,
+                        StThemeContext *context,
+                        ClutterColor   *color)
+{
+  CRTerm *arg = term->ext_content.func_param;
+  gchar *name;
+  const ClutterColor *envcolor;
+
+  if (arg == NULL || arg->the_operator != NO_OP)
+    return VALUE_NOT_FOUND;
+
+  if (arg->type == TERM_STRING &&
+           arg->content.str &&
+           arg->content.str->stryng &&
+           arg->content.str->stryng->str)
+    {
+      name = arg->content.str->stryng->str;
+      envcolor = st_theme_context_get_color (context, name);
+      if (!envcolor)
+        return VALUE_NOT_FOUND;
+
+      *color = *envcolor;
+
+      /* Alpha can be overriden with an optional second parameter */
+      if (arg->next != NULL)
+        {
+          double alpha;
+
+          arg = arg->next;
+          if (arg->the_operator != COMMA)
+              return VALUE_NOT_FOUND;
+
+          if (arg->type != NUM_GENERIC)
+            return VALUE_NOT_FOUND;
+
+          alpha = CLAMP(arg->content.num->val, 0, 1);
+          color->alpha = color_component_from_double(alpha);
+        }
+
+      return VALUE_FOUND;
+    }
+
+  return VALUE_NOT_FOUND;
+}
+
+static GetFromTermResult
 get_color_from_term (StThemeNode  *node,
                      CRTerm       *term,
                      ClutterColor *color)
@@ -633,10 +679,12 @@ get_color_from_term (StThemeNode  *node,
   else if (term->type == TERM_FUNCTION &&
            term->content.str &&
            term->content.str->stryng &&
-           term->content.str->stryng->str &&
-           strcmp (term->content.str->stryng->str, "rgba") == 0)
+           term->content.str->stryng->str)
     {
-      return get_color_from_rgba_term (term, color);
+      if (strcmp (term->content.str->stryng->str, "rgba") == 0)
+        return get_color_from_rgba_term (term, color);
+      else if (strcmp (term->content.str->stryng->str, "env") == 0)
+        return get_color_from_context (term, node->context, color);
     }
 
   status = cr_rgb_set_from_term (&rgb, term);
--- a/data/org.cinnamon.gschema.xml.in
+++ b/data/org.cinnamon.gschema.xml.in
@@ -679,6 +679,35 @@
         The timestamp of the last time the index cache was updated from the spices website
       </description>
     </key>
+
+    <child name="colors" schema="org.cinnamon.theme.colors"/>
+  </schema>
+
+  <schema id="org.cinnamon.theme.colors" path="/org/cinnamon/theme/colors/"
+        gettext-domain="@GETTEXT_PACKAGE@">
+    <key name="enable" type="b">
+      <default>false</default>
+      <_summary>Enable support for custom colors</_summary>
+      <_description>
+        Whether themes should use user defined colors or their defaults
+      </_description>
+    </key>
+
+    <key name="accent" type="s">
+      <default>'#4c4c4c'</default>
+      <_summary>Accent theme color</_summary>
+      <_description>
+        Dynamic color used by themes as an accent in any CSS format
+      </_description>
+    </key>
+
+    <key name="text" type="s">
+      <default>'#ffffff'</default>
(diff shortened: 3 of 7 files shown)


files/usr/share/cinnamon/cinnamon-settings/modules/cs_themes.py
```diff
@@ -69,6 +69,18 @@ def on_module_selected(self):
             page = SettingsPage()
             self.sidePage.stack.add_titled(page, "options", _("Settings"))
 
+            settings = page.add_section(_("Desktop color options"))
+
+            invert = lambda b: not b
+            widget = GSettingsSwitch(_("Use default colors"), "org.cinnamon.theme.colors", "enable", map_get=invert, map_set=invert)
```
germanfr: Also note this. I wasn't sure if this was the best way to do it. I could have used `not_` from the operator module, but I didn't want to add an import.

src/st/st-theme-node.c
```diff
@@ -610,6 +610,52 @@ get_color_from_rgba_term (CRTerm       *term,
   return VALUE_FOUND;
 }
 
+static GetFromTermResult
+get_color_from_context (CRTerm         *term,
+                        StThemeContext *context,
+                        ClutterColor   *color)
+{
+  CRTerm *arg = term->ext_content.func_param;
+  gchar *name;
```
JosephMcc: I didn't do a real review or test this. I don't even understand everything you're doing here but this gchar * should probably be freed when you're done with it. Since you only use it in the if() block below I would just declare it there and then free it right after you set the envcolor.

JosephMcc:
Not saying we shouldn't do this but I do have a couple of questions and reservations about it. You're opening a huge can of worms I'm not sure we want to :) Does this work with properties like we use for gradients? The -gradient-start/-gradient-end type properties? What if a theme is using background-image and image assets instead? This also won't effect other image based assets such switch, checkbox, etc.

On top of that this effects the Cinnamon theme only (and only partially). That really makes it quite limited since it changes very little. It also won't work with any theme that doesn't specifically support it or any 3rd party code added to an xlet that doesn't support it. While we control many of the themes, most of the large and popular themes are not part of the spices ecosystem because they are part of large theme packages.

I can see this leading to a lot of confusion and bug reports from users because it isn't working and doing what they probably expect it to do.

src/st/st-theme-node.c
germanfr: That is just a reference to a string in the term struct, which would get freed (I suppose) whether I reference it or not.

germanfr:
> Does this work with properties like we use for gradients? The -gradient-start/-gradient-end type properties?

Yes it works there also. I've just checked.


> What if a theme is using background-image and image assets instead? This also won't effect other image based assets such switch, checkbox, etc.

They should try to use CSS whenever possible. Also something that could help a lot would be the ability to use background colors behind border-images, I've struggled with this when moving my theme from images to CSS. Switches are the little stone in my shoe. Maybe in the future they should have a slightly more idiomatic structure than just a box with a background.


> I can see this leading to a lot of confusion and bug reports from users because it isn't working and doing what they probably expect it to do.

Now you say, I would add a disclaimer with something like "not all themes support it", similar to what's in the dark theme switch there.

germanfr:
Regarding the switches and the checkboxes @JosephMcc mentioned, it's pretty simple to theme them benefiting from this. One can simply set the background color and the border radius of the switch/checkbox with CSS, and overlay the borders, shades or whatever with a background-image. An example:
```css
.toggle-switch {
    width: 26px;
    height: 16px;
    border-radius: 10px;
    border: 1px rgba(0,0,0,.5);
    background-color: #999999;
    background-image: url("assets/switch-off.svg");
}
.toggle-switch:checked {
    background-color: #9ab87c;
    background-color: env("accent");
    background-image: url("assets/switch-on.svg");
}
```
Where the images overlay are [these](https://github.com/linuxmint/Cinnamon/files/1699947/switch-assets.tar.gz):
screenshot from 2018-02-06 17-47-53

Most themes I can think of can do that, and if not they can stay as they are now. This is something that is adding value, never subtracting.

Regarding that, I have in mind something to detect if the current theme supports the use of this feature and so enable/disable the option, but I'd need help with the Python part of disabling the controls.

clefebvre:
My two pence on it. It's a cool idea, but if it doesn't support the GTK theme, it's not useful enough to be considered.

clefebvre:
We're getting ready to switch to Mint-Y and so we're getting ready to create color variations for it. Alternatively, we could provide a tool which generates that variation based on a custom "accent" color code, and have a hook in Cinnamon, for Cinnamon to detect the fact that the selected theme supports custom colors... it's just an idea and it complements what you started here.

I'm not sure this would go anywhere, I'd suggest leaving the implementation aside for now and focusing on the design first so we can decide whether it's interesting or not to pursue this idea.

clefebvre:
From a distribution point of view, what we currently do is provide a neutral Cinnamon theme and accentuated icon and gtk themes... they're the ones which should be customizable, assuming it's a good idea.

germanfr:
Gtk is one "program" and Cinnamon is another "program". You can't change the behaviour of Gtk from here unless it exposes an API for that (which I don't know if exists). I don't understand why you insist that much in that it must support Gtk or nothing, I mean, it'd be great but it's up to them. Besides they implement it or not I'd be a great feature to have in Cinnamon and it adds value. Unity (RIP) supported it only for the panel and it didn't cause any confusion although it was automatic. Maybe its a good idea to leave the schema for advanced users and scripts as an open possibility for the future and remove the switches in settings.

germanfr:
Also from my point of view as a theme developer, it's a pain in the ass to have to maintain a number of variants with just a different color, and always have people asking for a colour variant. There's some people that create their own forks and then leave it unmaintained, which in general degrades the user experience polluting the pool of possibilities with themes that just don't work. 

That said, I don't know any drawback on having this feature.

smurphos:
Some thoughts. 

If Cinnamon shipped with a pre-processessor such as sassc and provided a convention for cinnamon recognised variable names in  `_colors.scss` then a settings modules could scan the theme for compatible `_colors.scss` and present the end user with a list of themes that can be edited and the variables that can be changed for that specific theme. Any changes would be written to the themes `_colors.scss'  result in sassc rebuilding the cinnamon theme and/or the GTK theme. If the theme lacks support then the cinnamon module would simply not list it. 

It's up to the theme designer what variables they name so as to be recognised by cinnamon as editable. 

Theme A might just have one or two colors that can be edited and then calculates other colors in the SCSS, whilst theme B gives access to a dozen base colors and variables for gradient, transparency, border radius etc.

JosephMcc:
> That said, I don't know any drawback on having this feature.

I don't want to just sound like a downer but I do. One being the constant bug reports we will get because this theme doesn't work with that thing or that thing doesn't work with this theme. It's easy to say themes just shouldn't use image based assets but that isn't the right answer. I've dealt with this trying to make mate-panels color changing stuff work with Mint-X and it' very limiting and a huge PITA. 

> Any changes would be written to the themes _colors.scss result in sassc rebuilding the cinnamon theme and/or the GTK theme.

Again you are totally ignoring the existence of image based assets.

The only way I could really seeing this being viable is to create a custom theme and a tool for modifying it. It will always be an imperfect solution just trying to make it work with any general theme.

smurphos:
> Again you are totally ignoring the existence of image based assets.

Not entirely but the workarounds would be messy.

> The only way I could really seeing this being viable is to create a custom theme and a tool for modifying it. It will always be an imperfect solution just trying to make it work with any general theme.

If for example the new Mint Y was designed in this way and the tool bundled with Mint then themers could write new themes compatible with the tool or port existing themes to do so. 

I'm not oblivious to the difficulty. I use scripts forked from Oomox to build my current themes and am toying with the idea of trying to port other themes to use the same tools,

germanfr:
> The only way I could really seeing this being viable is to create a custom theme and a tool for modifying it. It will always be an imperfect solution just trying to make it work with any general theme.

For example, someone could create a third party extension that would write to the settings schemas or directly use the JS API. The thing is that it would be hidden to normal users unless they explicitly installed it, and that extension would carry all those possible bug reports and may have more warnings and indications than just a switch in settings. From a themer's point of view, it'd be kind of theming an xlet, you can support it or not, but with color properties instead of classes.

But that wouldn't be possible without support from a Cinnamon backend like this (but without the settings part in python).

smurphos:
> For example, someone could create a third party extension that would write to the settings schemas or directly use the JS API

I'm intending to try and make something along these lines - whether I get there or not and how long it will take me is another question... ;-)

lestcape:
@germanfr just an idea.

Instead of trying to set a constant color, probably is better set just a key. You can then set the value of this key that you use in the css later, inside the theme context by calling a function and stored it in a hash. In that way the solution you use can be extended to other properties and not just colors in a more easy way.

Also using that keys, you can define latter what theme properties this key will be disable. For example:
If a key **bgcolor** is set inside an style class, you can disable the properties **background-image** and **border-image** of this style class, in that way you ensure your property will take effect also if there are images.

Latter, you can use also a hash on gsettings to set an array of pairs of (key, value) if you want to export that functionality to be controller externally.

**Please see:** Ideas about of render a css from a scss are also not good at all without a way to modify specific properties only, because a change of the whole theme will required a lot of processing and then the user will see the render, when what you want is a transparent "transition" of specific properties only. So, render a css from a scss is not a good idea instead a directly usage of scss will be better, because then the theme context can provide the functions to change the values of the scss variables dynamically. This ofcourse will be in opposition of the idea of use the same mechanism from Gtk and Cinnamon, because also Gtk will need to support a way to handled scss directly instead of css.

germanfr:
@lestcape I think that is a bit complicated and I'm not sure why the user would want to customize anything other than colors. If you already support custom colors you just don't use an image for the default case.

The basic scss color functions would be fairly easy to implement at runtime, although I wonder what would happen with those of us that already write our themes in scss. We would need something to tell Sass to not compile certain parts, and I think that doesn't exist.

lestcape:
@germanfr is exactly because will be a bit complicate, that will be a bit more general. Like i say, this was an idea. If you don't want to extend it for simplicity, is also ok.

For my personal opinion @JosephMcc have a good points against things like this: 
"If you already support custom colors you just don't use an image for the default case". 

I will explain you why with some examples:
1- Do you want to force also extensions to not use an image if they are not really belong to an specific theme?
2- How an extension can be integrate with a property without semantic? The extension will not look like out of the place?

germanfr:
> 1- Do you want to force also extensions to not use an image if they are not really belong to an specific theme?
2- How an extension can be integrate with a property without semantic? The extension will not look like out of the place?

An extension that doesn't use standard classes, whether they use images or CSS, already looks out of place. The images they use won't match the theme anyways. A popular example is gTile.

lestcape:
@germanfr Your comment partially resolved the point  1, but not 2, because the standard classes not exist... Just exist the specific classes for the specific things (the shell is not organized like Gtk)... So, while one extension can select the style-class-1, another extension can select the style-class-2. When you change the property, that is only set in style-class-1, one extension is updated while the another not adjusting colors to full-fill a good aspect will be a task for a Titan.

germanfr:
@lestcape but again, that has nothing to do with this. It's a Cinnamon problem. As you said this is not Gtk.

lestcape:
@germanfr You are right on this. It's how the shell is and this is not a problem that you will introduce here, but it is part of reality and therefore can make a good solution become not so good.

Let me be clear, what you already done, is i thing that i was wander to do for a lot time ago, but I never found a good solution (this is why my comments)... You have something working, thats is amassing of curse.

feren:
If you add environment colours, then it would maybe be nice to also have a colour value that used the GTK Highlight Colour...

germanfr:
> If you add environment colours, then it would maybe be nice to also have a colour value that used the GTK Highlight Colour...

That could be done through a Cinnamon extension over this. This is just the base for all that.

feren:
@germanfr Yeah, but the problem with relying on a Cinnamon Extension is that without the Cinnamon Extension, the theme might break down, and there's no easy way to fit 'Install _______ from Extensions > Get More Extensions' in the Themes > Get More Themes listing description bit without compromising the description of the theme, or label length in that bit...

germanfr:
@feren This shouldn't break any theme if they use this feature properly as I explained in the description and themes that don't use this will still work. That hypothetical extension would just simply set the colors in the schemas I've defined here, just as a user would do manually. Themes should not rely on that extension to be installed or any color to be used just the same as Cinnamon doesn't rely on any particular theme to be installed to work. If a color looks good or not is up to the user. Imagine that on Mint-Y for example.

PR #7292, opened 2018-02-17
Improvements to autogen.sh

ageis:
Addresses missing aclocal macros and downloads config.guess/config.sub if internet is available (that's GPLv3 code, along with isc-posix.m4 so I don't know if we should just commit them to version control instead.. I have them in separate commits though.


leigh123linux:
-1 to this commit being merged

clefebvre:
Thanks @ageis, I'll close this for now. We're moving to meson.

PR #7298, opened 2018-02-21
appletManager.js: fix getRunningInstancesForUuid()

itzexor:
Since we removed the discrete instances object appletObj and merged
instances into the definition object, we now have null instances.
This occurs during startup, for definitions that are on removed panels,
and likely for applets that fail to load.

To return the previous behavior we just need to filter instances that
aren't actually running at the moment, aka null.

### DIFF
--- a/js/ui/appletManager.js
+++ b/js/ui/appletManager.js
@@ -781,9 +781,14 @@ function pasteAppletConfiguration(panelId) {
 }
 
 function getRunningInstancesForUuid(uuid) {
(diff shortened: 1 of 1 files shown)


itzexor:
mint19 build failure is not related to this PR as far as I can tell:
`E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/d/dbus-glib/libdbus-glib-1-dev_0.110-2_amd64.deb  Connection failed [IP: 91.189.88.149 80]`

PR #7301, opened 2018-02-23
appletManager, deskletManager: Fix incorrect handling of adding/removing multiple xlets of the same UUID

jaszhix:
- Removes the call to unloadExtension in `onEnabledAppletsChanged`. It was calling a function that can only remove by UUID. In cases where multiple applets of the same UUID are on the panel, and only one instance needs to be removed, this prevents all of them from being removed.
- Re-added addAppletToPanels in `onEnabledAppletsChanged`. `initEnabledApplets` will try to initialize enabled applets, but the `Extension` constructor checks its cache and resolves true if an xlet of the same UUID is already loaded. This is ideal normally because instances are tracked in appletManager, but it skips the actual panel loading event.
- Fixes incorrect invocation of `formatError`, caught by @Odyseus 
- `_onEnabledDeskletsChanged` no longer gets called when desklets are dragged

### DIFF
--- a/js/ui/appletManager.js
+++ b/js/ui/appletManager.js
@@ -248,6 +248,7 @@ function appletDefinitionsEqual(a, b) {
 function onEnabledAppletsChanged() {
     let oldDefinitions = definitions.slice();
     definitions = getDefinitions();
+    let addedApplets = [];
     let removedApplets = [];
     let unChangedApplets = [];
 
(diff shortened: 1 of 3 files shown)


itzexor:
Just a note: I have been running with this branch applied for like 2 weeks now and it seems fine. This does indeed fix the 'clear all applets' issue I had reported privately. :+1:

PR #7303, opened 2018-02-24
xlet settings: add a generic tooltip to the keybinding widget

itzexor:
...that explains how to interact with it.

bind-tooltip


itzexor:
This PR is on hold for now, but I'm going to leave it open so I don't forget about it... See #7305

itzexor:
This should be OK now. I've replaced the first comment here and the name to accurately reflect the current changes.

PR #7304, opened 2018-02-24
Bring back xlet meta data

Odyseus:
This PR brings back the xlets meta data object that was removed with the #6878 PR.

An in-depth discussion can be read in the previously linked PR about this change.

### DIFF
--- a/js/ui/extension.js
+++ b/js/ui/extension.js
@@ -181,7 +181,7 @@ Extension.prototype = {
 
         const finishLoad = () => {
             // Many xlets still use appletMeta/deskletMeta to get the path
-            type.legacyMeta[uuid] = {path: this.meta.path};
+            type.legacyMeta[uuid] = Object.assign({}, this.meta);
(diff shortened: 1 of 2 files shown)


js/ui/extension.js
```diff
@@ -181,7 +181,7 @@ Extension.prototype = {
 
         const finishLoad = () => {
             // Many xlets still use appletMeta/deskletMeta to get the path
-            type.legacyMeta[uuid] = {path: this.meta.path};
+            type.legacyMeta[uuid] = Object.assign({}, this.meta);
```
jaszhix: Can you please point to a piece of code that warrants this use case? We are storing the full metadata in every extension and xlet instance if we do this. Its inefficient. There is no discernible difference between the time _init is called and whatever you want to do outside the xlet function constructor. After our conversation in #6878 I'm still convinced this is more about principal, not what's best for Cinnamon technically.

PR #7306, opened 2018-02-25
Fix multi-version xlets loading twice, address metadata availability

jaszhix:
- Removes left over code that imported multi-version xlets with the cjs importer
- Re-added `extensionMeta`
- Made the meta object available to the xlet scope with the variable `__meta`

### DIFF
--- a/js/misc/fileUtils.js
+++ b/js/misc/fileUtils.js
@@ -140,7 +140,7 @@ function unloadModule(index) {
     }
 }
 
-function createExports({path, dir, type, file, size, JS, returnIndex, reject}) {
+function createExports({path, dir, meta, type, file, size, JS, returnIndex, reject}) {
     // Import data is stored in an array of objects and the module index is looked up by path.
     let importerData = {
         size,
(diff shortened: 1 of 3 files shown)


Odyseus:
- [extensionSystem.js](https://github.com/linuxmint/Cinnamon/pull/7306/commits/98042c7b64d00391b35d623c619d807ac09931a3#diff-de241cee02b95735161e571507296766R39): `extensionMeta` is defined inside the `prepareExtensionUnload` function when it should be defined inside the `init` function.
- The `__meta` variable isn't available when used inside an imported JavaScript file/module. It's only available inside the main xlet's files (applet.js, etc.).

[Test extension](https://github.com/linuxmint/Cinnamon/files/1755768/0Test.domain.test.zip).

Odyseus:
- [extension.js#L251](https://github.com/linuxmint/Cinnamon/blob/master/js/ui/extension.js#L251).
    It reads:

    ```javascript
    throw new Error(`${type.name} ${uuid}: Could not create applet object.`);
    ```

    It would be better/less confusing if a more generic term like **xlet** is used:

    ```javascript
    throw new Error(`${type.name} ${uuid}: Could not create xlet object.`);
    ```

    Or a specific one:

    ```javascript
    throw new Error(`${type.name} ${uuid}: Could not create ${this.lowerType} object.`);
    ```

PR #7325, opened 2018-03-04
Fix bad allocation of panels with borders

### LINKED ISSUE #6440: window-list: inactive at terminal edge if in a panel with border on the opposing side
Issue found in master, not checked to see if present in production 3.2.
If you have a panel with border on the inward facing edge, then the terminal edge side becomes unreactive in the window list.  There appears to be a strip of unreactive pixels at the edge of the screen.

germanfr:
Children were allocated at 0 over the top-left borders and paddings, leading to dead zones in the opposite side.

Fixes #6440

Visual explanation:
screenshot from 2018-03-04 19-45-12

```css
.panel-bottom {
    background: red;
    border-top: 8px solid blue;
}
```
**I need to test it in RTL mode**. I did my best but I could not find a way to enable it. If you tell me how to do it I'll check if it works with that (it should).

### DIFF
--- a/js/ui/panel.js
+++ b/js/ui/panel.js
@@ -3041,11 +3041,9 @@ Panel.prototype = {
         let allocHeight  = box.y2 - box.y1;
         let allocWidth   = box.x2 - box.x1;
 
-        let childBox = new Clutter.ActorBox();
-        childBox.x1 = box.x1;
-        childBox.x2 = childBox.x1 + allocWidth;
-        childBox.y1 = box.y1;
-        childBox.y2 = childBox.y1 + allocHeight;
+        /* Panel left/center/bottom will fit inside this box, which is equivalent
+           to the CSS content-box (imaginary box inside borders and paddings) */
+        let childBox = box.copy();
 
         /* The boxes are layout managers, so they rubber-band around their contents and have a few
            characteristics that they enforce on their contents.  Of particular note is that the alignment
@@ -3059,6 +3057,8 @@ Panel.prototype = {
 
         if (this.panelPosition == PanelLoc.left || this.panelPosition == PanelLoc.right) {
 
+            /* Distribute sizes for the allocated height with points relative to
+               the children allocation box, inside borders and paddings. */
             let [leftBoundary, rightBoundary] = this._calcBoxSizes(allocHeight, allocWidth, true);
             leftBoundary += box.y1;
             rightBoundary += box.y1;
@@ -3115,6 +3115,8 @@ Panel.prototype = {
             }
         } else {           // horizontal panel
 
+            /* Distribute sizes for the allocated width with points relative to
+               the children allocation box, inside borders and paddings. */
             let [leftBoundary, rightBoundary] = this._calcBoxSizes(allocWidth, allocHeight, false);
             leftBoundary += box.x1;
             rightBoundary += box.x1;
--- a/js/ui/panel.js
+++ b/js/ui/panel.js
@@ -3041,6 +3041,12 @@ Panel.prototype = {
         let allocHeight  = box.y2 - box.y1;
         let allocWidth   = box.x2 - box.x1;
 
+        let childBox = new Clutter.ActorBox();
+        childBox.x1 = box.x1;
+        childBox.x2 = childBox.x1 + allocWidth;
+        childBox.y1 = box.y1;
+        childBox.y2 = childBox.y1 + allocHeight;
+
         /* The boxes are layout managers, so they rubber-band around their contents and have a few
            characteristics that they enforce on their contents.  Of particular note is that the alignment
            - LEFT, CENTER, RIGHT - is not independent of the fill as it probably ought to be, and that there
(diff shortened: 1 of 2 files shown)


brownsr:
Excellent.  That dead strip with a border had bugged me before, but I never got to the bottom of it.  This seems to all work fine for me.  Suggest you check all OK is on multiple monitors, I can't check that readily.  Could you add a brief comment in the code, or the logic is going to be incomprehensible in the future.  Testing RTL is difficult, you need to set locale to a RTL language such as Arabic, Hebrew etc.  I suspect there may well be RTL issues lurking because of the challenge of doing this, I don't think any of the frequent developers know RTL languages so if by any chance you do then that would be great.

germanfr:
Yes, I just checked and it works with multiple monitors.

I downloaded the Hebrew locale, I enabled it globally in settings and I also tried this command, but now way it to enable RTL.
```shell
LANGUAGE=he_IL.utf8 cinnamon --replace
```

JosephMcc:
@germanfr I posted this for brownsr somewhere else earlier but I'll copy it here. Not sure what distro you run but this is how I test it.

```
ok the way I did this in Mint was use mintlocale to install an arabic language
make sure all the language packs are installed
create a new user account for testing
I set the new account to use the arabic language
then reboot. A simple log in and out isn't enough for the new language to fully take effect
boot into your new user account and it will be using the new language

simply doing it with cinnamon --replace in a current sesssion doesn't seem to make it fully take effect
once everything is installed you can do a `LANGUAGE=ar_AE.UTF-8 cinnamon --replace` but it doesn't seem to update everything
it flips your panel around and updates some things so may work in his case but a better test is probably the dummy account
```

JosephMcc:
@germanfr I posted this for brownsr somewhere else earlier but I'll copy it here. Not sure what distro you run but this is how I test it.

```
ok the way I did this in Mint was use mintlocale to install an arabic language
make sure all the language packs are installed
create a new user account for testing
I set the new account to use the arabic language
then reboot. A simple log in and out isn't enough for the new language to fully take effect
boot into your new user account and it will be using the new language

simply doing it with cinnamon --replace in a current sesssion doesn't seem to make it fully take effect
once everything is installed you can do a `LANGUAGE=ar_AE.UTF-8 cinnamon --replace` but it doesn't seem to update everything
it flips your panel around and updates some things so may work in his case but a better test is probably the dummy account
```

germanfr:
@JosephMcc I tried that, everything is in Arabic but RTL is not enabled in any program. I installed `ar_SA` and also started cinnamon with `LANGUAGE=ar_SA.UTF-8 cinnamon --replace`, but nothing.

germanfr:
I added some commits to be squashed. They fix wrong positions for the panel corners too and I added some comments (not many, I can't explain it better than the code itself).

brownsr:
Looking good.  I thought I had spotted an error, but then I found it was there in the master version too.  You may be able to see the cause as you are in amending this section. Using theme faience which has visible corners, I set up two side panels, then added a new one at top (or bottom), all fine so far, then move the horizontal panel to the other side of the screen.  That's definitely not fine, with corners left around, not replaced, and some blanked out or overwritten bits of panel towards the corners of the screen.

germanfr:
@brownsr Yes, I can reproduce that in master too. It's a completely different issue not related to this and I have no idea how to fix it, sorry.

JosephMcc:
I know this had been discussed internally but the panel corner stuff could really die. It's hardly used anywhere and overcomplicates the panel code for virtually no benefit.

smurphos:
Glad this came up - I've been been attempting to work around this by various methods for a while trying to get bordered objects in panels to look OK in bordered panels and it's never quite right.

As the subject has come up I get other strange behaviour attempting to get a single outside border on all 4 panels. Can others reproduce this? Should I raise a separate issue?

This works.
```
.panel-bottom {
  border-top: 1px solid white;
}
```
This doesn't
```
.panel-top {
  border-bottom: 1px solid white;
}
```
but this does
```
 .panel-top {
    border: 1px solid white;
    border-top: 0px;
    border-left: 0px;
    border-right: 0px;
}
```
Similar stunts are needed for left and right panels.

smurphos:
With this fix and a thick border are tooltips rendering across the panel border? It would be good to fix that at the same time. In general having a selector available to modify applet tooltip position from css would also be good to have.

germanfr:
> With this fix and a thick border are tooltips rendering across the panel border? Edit - Yes they are....

I added a commit to fix popup menus being drawn behind borders.

> As the subject has come up I get other strange behaviour attempting to get a single outside border on all 4 panels. Can others reproduce this? Should I raise a separate issue?

Regarding those border issues, it's not related to this, it's from the C part. If I'm right, that was because it doesn't support different colors for each side and it uses the color from the top border. So when you type `border-bottom: 1px solid red`, the top border color is transparent. You can check that doing this:

```css
/* Doesn't work */
.panel-bottom {
    border-bottom: 1px solid red; /* Border color is transparent */
}

/* Works */
.panel-bottom {
    border-bottom: 1px solid red; 
    border-top-color: yellow; /* Border color is now defined */
}
```

brownsr:
Is this still Work In Progress ?

germanfr:
I don't think I need to add anything, but I haven't been able to test it in RTL mode as I said. I think it's correct but I can't confirm.

germanfr:
I finally managed to test it in RTL and it worked, although I found some more old issues that I solved in my last commit. I think it is ready. Tell me if you want me to squash the commits.

JosephMcc:
It can always be squashed during merge so you probably don't need to worry about it too much.

germanfr:
I modified the last commit because I fixed the panels alignment in the wrong place. It is so easy to get wrong with this. I must say that it feels so stupid to flip the panels depending on the language, because they are... panels. It complicates a

JosephMcc:
> I must say that it feels so stupid to flip the panels depending on the language, because they are both... panels.

I've always kind of felt that way also but that's how it works in other apps as well. In xed for example the left pane switches to the right for RTL. So I suppose it's consistent the way it is.

germanfr:
> In xed for example the left pane switches to the right for RTL. So I suppose it's consistent the way it is.

In xed moving the pane to the right makes... having the pane in the right, but flipping the panels makes... nothing. For a user with RTL it's like it isn't flipped. But anyways, this is a bit off topic. I just wanted to know your opinions.

PR #7340, opened 2018-03-10
Systray: remove the resize delay in scale mode

brownsr:
This causes a strange two stage display sequence for the user. Precautionary delay for the application to load is already baked in before this.

It's not clear to me why there would be a timing difference between scale mode (which has this extra delay) and having the icon size set by default/theme (no additional delay).  There's already a 1 sec initial delay for most applications which I presume is precautionary to give them time to load, so the current code gives a 1 second delay in most cases before an unscaled icon shows- and then there is a re-scaling of the icon to the expected size half a second later.  That's a bit weird, and gives an unhelpful impression of Cinnamon being slow and laggy.  Unless someone knows of a really good reason why the half second additional delay is in there, I propose removing it.

### DIFF
--- a/files/usr/share/cinnamon/applets/systray@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/systray@cinnamon.org/applet.js
@@ -349,10 +349,7 @@ MyApplet.prototype = {
         icon._rolePosition = position;
 
         if (this._scaleMode) {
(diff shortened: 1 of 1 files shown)


clefebvre:
This has been empirical from the start. Systray's worked fine for a while now... we can try and remove it and if something goes wrong we'll know what to revert :)

PR #7351, opened 2018-03-14
inhibit applet: use more common icon system-suspend

NikoKrause:


JosephMcc:
Just did a little digging and both of these are actually poor choices since neither exists in Adawaita/Gnome. Probably need to do a bit of digging to find a good one or ship the one we want to use ourselves.

PR #7354, opened 2018-03-16
Add desklet visibility toggling to the Show Desktop applet context menu, update with class syntax

jaszhix:


files/usr/share/cinnamon/applets/show-desktop@cinnamon.org/applet.js
```diff
@@ -33,13 +31,21 @@ MyApplet.prototype = {
 
         this.set_applet_icon_name("user-desktop");
         this.set_applet_tooltip(_("Show desktop"));
-    },
 
-    on_applet_removed_from_panel: function() {
+        this.showDeskletsOption = new PopupMenu.PopupIconMenuItem(
+            _('Show Desklets'),
```
NikoKrause: _("Show Desklets")

files/usr/share/cinnamon/applets/show-desktop@cinnamon.org/applet.js
```diff
@@ -108,9 +114,18 @@ MyApplet.prototype = {
         }
         this._did_peek = false;
     }
+
+    toggleShowDesklets() {
+        if (Main.deskletContainer.isModal) {
+            Main.deskletContainer.lower();
+            this.showDeskletsOption.label.set_text(_('Show Desklets'));
```
NikoKrause: _("Show Desklets")

files/usr/share/cinnamon/applets/show-desktop@cinnamon.org/applet.js
```diff
@@ -108,9 +114,18 @@ MyApplet.prototype = {
...
+        } else {
+            Main.deskletContainer.raise();
+            this.showDeskletsOption.label.set_text(_('Hide Desklets'));
```
NikoKrause: _("Hide Desklets")

NikoKrause:
Not sure if it's still the case, but in the past the makepot script didn't recognized single quotes. So all the default applets use double quotes for translatable strings.

jaszhix:
Ah, I think it should work because we passed JavaScript as the language to [xgettext](https://github.com/linuxmint/Cinnamon/blob/5a9de050d453140340fd0424811a751f367015ef/makepot#L8).

PR #7366, opened 2018-03-17
ExtensionCore: show the number of available updates in the 'Update al…

### LINKED ISSUE #7128: Show number of available updates in Settings>Applets>Download
```
 * Cinnamon version (cinnamon --version) 3.6.6
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...) Mint 18.3 Cinnamon 64-bit
 ```

**Issue**

The previous version of `Settings>Applets` showed the number of available updates, whereas the new version (Mint 18.3) does not show them anymore.

**Steps to reproduce**

1. Start `Settings>Applets`
2. Switch to `Download`
3. `Refresh`
4. If there are any new applets, `Update all` becomes active, but does not show the number of available updates.

**Expected behaviour**

I would like to see the number of available updates somewhere. Another option would be to have a filter which shows only the updateable applets. (Thinking of it, that might be even the better option.)

**Other information**

The same behaviour might be found in `Settings>Desklets` and `Settings>Extension`, but I could not test it, because there were no updated desklets/extension.

### PR DESCRIPTION
**Pull request by collinss**
…l' button tooltip

Fixes #7128

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/ExtensionCore.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/ExtensionCore.py
@@ -814,7 +814,6 @@ class DownloadSpicesPage(SettingsPage):
 
         self.update_all_button = Gtk.Button.new_from_icon_name("software-update-available-symbolic", Gtk.IconSize.MENU)
         self.update_all_button.set_size_request(50, -1)
-        self.update_all_button.set_tooltip_text(_("Update all"))
         self.update_all_button.connect('clicked', self.update_all)
         box.add(self.update_all_button)
         self.update_all_button.set_sensitive(False)
@@ -926,7 +925,13 @@ class DownloadSpicesPage(SettingsPage):
             self.extension_rows.append(row)
             self.list_box.add(row)
 
-        self.update_all_button.set_sensitive(self.spices.are_updates_available())
+        updates_available = self.spices.get_n_updates()
+        self.update_all_button.set_sensitive(updates_available)
+        if updates_available > 0:
+            msg_text = _("Update all") + ' (' + ngettext("%d update available", "%d updates available", updates_available) % updates_available  + ')'
+        else:
+            msg_text = _("No updates available")
+        self.update_all_button.set_tooltip_text(msg_text)
         self.refresh_button.set_sensitive(True)
 
     def get_more_info(self, *args):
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: files/usr/share/cinnamon/cinnamon-settings/bin/ExtensionCore.py [hunk 1]
```diff
@@ -926,7 +925,13 @@ def build_list(self, *args):
             self.extension_rows.append(row)
             self.list_box.add(row)
 
-        self.update_all_button.set_sensitive(self.spices.are_updates_available())
+        updates_available = self.spices.get_n_updates()
+        self.update_all_button.set_sensitive(updates_available)
+        if updates_available > 0:
+            msg_text = _("Update all") + ' (' + gettext.ngettext("%d is available", "%d are available", updates_available) % updates_available  + ')'
```
NikoKrause:
```
msg_text = gettext.ngettext("%d update available", "%d updates available", updates_available) % updates_available
```

NikoKrause:
Add those lines at the preamble:
```
# i18n
APP = 'cinnamon'
LOCALE_DIR = "/usr/share/locale"
gettext.bindtextdomain(APP, LOCALE_DIR)
gettext.textdomain(APP)
_ = gettext.gettext
```

Compare here: https://github.com/linuxmint/mintupdate/pull/329/files

NikoKrause:
I'm not entirely sure if you need `ngettext` for `%d available`, because it's missing a noun, where this would be relevant (at least for the languages I know)

collinss:
@NikoKrause There's no guarantee that all languages would be able to translate that without needing a plural and still make sense, so I did it that way just in case.

NikoKrause:
Yeah, I'm no language expert. My feeling just tells me, that you need a noun for it. But there are so many languages and I know just a handful, so whatever :)

clefebvre:
"%d available" isn't correct... you are missing a noun in there indeed. You need to use ngettext alright, but you also need a noun to dictate whether the adjective should be masculine or feminine.

collinss:
@clefebvre fixed. There is now a noun in the translation. I didn't think about needing a noun in the translation for adjective gender. I speak Spanish, though, so I should have.

germanfr:
Hey @collinss there is a typo with the singular form. It should be _update_ not _updates_.

collinss:
Face-palm. Ok, fixed now.

Issue #7375, opened 2018-03-19
sound@cinnamon.org applet - Make the 100% mark really magnetic


claudiux:
```
 * Cinnamon version : 3.6.7, built with mint-dev-tools
 * Distribution : Mint 18.3
 * 64 bit
 ```

**Issue**
(With the volume max set to 150% in sound settings.)
When the volume slider is on the 100% mark, the real volume is nearly of 96%.

**Steps to reproduce**
Click on the icon of the sound applet, drag the slider to 88% then scroll up one time.

**Expected behaviour**
Set the real volume to 100%, as the slider is.

**Other information**
Proposed solution : 
In the _update function, replace the line:
let value, visible_value, delta = VOLUME_ADJUSTMENT_STEP;
by the line:
let value, visible_value, delta = VOLUME_ADJUSTMENT_STEP * this.applet._volumeMax / this.applet._volumeNorm;

claudiux:
(With the volume max set to 150% in sound settings.)

claudiux:
This issue is solved by PR #7376.

claudiux:
I close this issue because PR #7376 is now merged.

PR #7376, opened 2018-03-19
sound@cinnamon.org applet - Solves issue #7375

claudiux:
Hi,

This PR is made to solve the issue #7375.

Best regards.

claudiux

### DIFF
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -166,7 +166,7 @@ VolumeSlider.prototype = {
         // visible_value: percentage of volume_norm (shown to the user)
         // these only differ for the output, and only when the user changes the maximum volume
         let volume = (!this.stream || this.stream.is_muted) ? 0 : this.stream.volume;
-        let value, visible_value, delta = VOLUME_ADJUSTMENT_STEP;
+        let value, visible_value, delta = VOLUME_ADJUSTMENT_STEP * this.applet._volumeMax / this.applet._volumeNorm;
 
         if (this.isOutputSink) {
             value = volume / this.applet._volumeMax;
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -150,6 +150,9 @@ VolumeSlider.prototype = {
             muted = true;
         } else {
             muted = false;
+            //100% is magnetic:
+            if (volume != this.applet._volumeNorm && volume > this.applet._volumeNorm*(1-VOLUME_ADJUSTMENT_STEP/2) && volume < this.applet._volumeNorm*(1+VOLUME_ADJUSTMENT_STEP/2))
+                volume = this.applet._volumeNorm;
         }
         this.stream.volume = volume;
         this.stream.push_volume();
@@ -188,6 +191,7 @@ VolumeSlider.prototype = {
         return false;
     },
 
+
     _update: function(){
         // value: percentage of volume_max (set as value in the widget)
         // visible_value: percentage of volume_norm (shown to the user)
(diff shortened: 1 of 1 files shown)


itzexor:
This doesn't fix it for me. The actual percentage still is set to wherever you let the mouse button off at.

claudiux:
Did you:
 - restart Cinnamon?
 - use the mouse scroll wheel on the slider?

If you simply drag the slider, the 100% mark isn't magnetic. This isn't a bug, but a feature ;)

JosephMcc:
> If you simply drag the slider, the 100% mark isn't magnetic. This isn't a bug, but a feature ;)

It being magnetic was intentional. Regardless of anything else you shouldn't remove that.

claudiux:
I know it was intentional, I am the one who proposed the code to make this mark magnetic.
I don't want to remove that, I want to improve it.

I start with a sound level at 89%. Then I scroll on the slider with the mouse scroll wheel.

## Before the change that I propose:
Init: without_89

First step: without_97

Second step: without_104

Where is the magnetic effect ? Nowhere.

## After changing:
Init: with_89

First step: with_100

Second step: with_107

100% is _now_ magnetic.

itzexor:
What I mean by it doesn't work for me is that when you're dragging the
slider with the mouse, even though it's magnetic it's still doesn't always
end up at 100%. It could be anywhere from 96 to 104% depending on where you
let off the mouse button after you are done dragging after it's snaps to
the mark. Not using the scroll wheel. So I still think it needs more
improvements.

claudiux:
I can't reproduce that. Please watch these videos: 

https://github.com/claudiux/docs/blob/master/sound/sound_applet-2018-03-20_22.16.30.mkv 

and 

https://github.com/claudiux/docs/blob/master/sound/sound_applet-2018-03-20_22.32.14.mkv

claudiux:
There is no problem, OK ?

itzexor:
Hey sorry i was really busy yesterday didn't have time to get back to you, but I did investigate the behavior more, and I can't reproduce the issue I was talking about anymore so I'm not sure what happened there... But you can just disregard that.

However, I did notice that this makes the scroll increment more than 5% around 100% when the volume is not a multiple of 5. When i was testing it was 8%. I think that might be too big of a jump. Can we make it do < 5%  instead somehow? Like instead of 87 -> 92 -> 100 i think 87 -> 92 -> 97 -> 100 would be better. What do you think?

itzexor:
Ok I can reproduce what I was talking about before, but it's not from your changes. It is the same with or without your changes, but if you drag the slider with the mouse and snap it to 100%, even if the sound applet says 100% you can open sound settings and see that the volume is not actually 100%

I think there are some other issues here that need to be fixed up...

claudiux:
OK. I feel like I have seen where it does not work. I will continue tomorrow.

claudiux:
I seem to have solved the problem. In fact, I was trying to correct what the "thermometer" indicated without worrying about the "temperature". Now, the real sound volume (shown in sound settings) is what is indicated in the applet and vice versa.

Can you make some tests, please ?

Thank you.

claudiux:
The result in video:

https://github.com/claudiux/docs/blob/master/sound/sound_applet-2018-03-28_00.33.21.mkv?raw=true

claudiux:
There is no problem now, OK ?

files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
```diff
@@ -150,6 +150,9 @@ VolumeSlider.prototype = {
             muted = true;
         } else {
             muted = false;
+            //100% is magnetic:
+            if (volume != this.applet._volumeNorm && volume > this.applet._volumeNorm*(1-VOLUME_ADJUSTMENT_STEP/2) && volume < this.applet._volumeNorm*(1+VOLUME_ADJUSTMENT_STEP/2))
```
collinss: Do we need the check ```volume != this.appet._volumeNorm``` here? It will not change the value in any way, and it adds complexity to the condition, making it harder to read. I suppose it might avoid some calculations if the value is 100% already, but the calculations are pretty minor, so I can't imagine it would have a noticeable impact on performance.

claudiux: Yes, we need it. Without this test, which is done quickly because it compares two integers, there are two comparisons with floats that are made at each loop. If you try removing this test, you will notice a loss of performance and your fan will run out.

claudiux: Thanks for your tests, @collinss !

PR #7382, opened 2018-03-20
cinanmon-settings: don't rely on the presence of cinnamon-control-center

eli-schwartz:
The existence of the cinnamon-control-center binary does not even tell us what control center libraries are available. So instead, do the same thing we did for cinnamon-settings modules, and check if the library exists.

AFAICT this doesn't even affect anything except whether cinnamon-settings prints "Unknown module %s" and possibly first tries to proxy out to gnome before running cinnamon-settings anyway...

It seems incongruous to simply accept absolutely anything just because the control center was installed. Also I'm not entirely sure what the control center binary actually does, as it seems to just be an inferior version of cinnamon-settings and its desktop file says not to display it... Is there any reason to actually install it? The previous Arch maintainer stripped it out and I'm not sure whether I should add it back in.

### DIFF
--- a/files/usr/bin/cinnamon-settings
+++ b/files/usr/bin/cinnamon-settings
@@ -7,6 +7,7 @@ Usage:  cinnamon-settings [optional module name]
 
 import os
 import sys
+import sysconfig
 
 if len(sys.argv) > 1:
     module = sys.argv[1]
@@ -14,7 +15,7 @@ if len(sys.argv) > 1:
         os.execvp("/usr/share/cinnamon/cinnamon-settings/xlet-settings.py", (" ", module[0:-1]) + tuple(sys.argv[2:]))
     if os.path.exists("/usr/share/cinnamon/cinnamon-settings/modules/cs_%s.py" % module):
         os.execvp("/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py", (" ",) + tuple(sys.argv[1:]))
-    elif os.path.exists("/usr/bin/cinnamon-control-center"):
+    elif os.path.exists(os.path.join(sysconfig.get_config_var("LIBDIR"), "cinnamon-control-center-1/panels/lib%s.so" % module)):
         os.execvp("/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py", (" ",) + tuple(sys.argv[1:]))
     elif os.path.exists("/usr/bin/gnome-control-center"):
         print ("Unknown module %s, calling gnome-control-center" % module)


clefebvre:
The binary isn't used by users, it's a bit like muffin... it's not needed by users but it's handy for us when troubleshooting. It allows us to quickly run CCC panels without needing cinnamon.

Your change won't work in Mint, the panel path is arch dependent /usr/lib/x86_64-linux-gnu/cinnamon-control-center-1/panels/libcolor.so.

leigh123linux:
This PR would also break fedora

```
$ rpm -ql cinnamon-control-center |grep lib64/
/usr/lib64/cinnamon-control-center-1
/usr/lib64/cinnamon-control-center-1/panels/libcolor.so
/usr/lib64/cinnamon-control-center-1/panels/libdate_time.so
/usr/lib64/cinnamon-control-center-1/panels/libdisplay.so
/usr/lib64/cinnamon-control-center-1/panels/libnetwork.so
/usr/lib64/cinnamon-control-center-1/panels/libonline-accounts.so
/usr/lib64/cinnamon-control-center-1/panels/libregion.so
/usr/lib64/cinnamon-control-center-1/panels/libwacom-properties.so
/usr/lib64/libcinnamon-control-center.so.1
/usr/lib64/libcinnamon-control-center.so.1.0.0
```

eli-schwartz:
Hmm, I forgot that libdir is different everywhere. :(

What about using `os.path.dirname(sysconfig.get_path('stdlib'))`

eli-schwartz:
So how does this new approach work -- also I've filed a pull request to not install the cinnamon-control-center binary by default.

clefebvre:
Hi @eli-schwartz,

That might work but there's no real point to it. It's not as simple as the current solution and the binary is there always for everybody.

eli-schwartz:
This change is completely independent of whether the binary exists. The existence of the binary doesn't tell us whether the *module* exists.

I'm not too bothered by your rejection of the other pull request as it is somewhat opinionated, but this one is simply a matter of technical correctness IMHO.

Please reconsider. This doesn't even really add much complexity at all.

collinss:
The only way there is a c-c-c module missing, is if someone removes it manually or if there is a version mismatch between c-c-c and Cinnamon. In the first case, whoever it is they probably know what they're doing. The second case should never happen, as there are many other issues that could arise from that too. Therefore, this is still not necessary.

eli-schwartz:
... or the module never existed in the first case, which is why there is a code path immediately after that tries to call gnome-control-center instead, and if that still fails it logs an "unknown module %s" warning before still calling cinnamon-settings.py

Currently this is all dead code (unless cinnamon-control-center is not installed at all, but gnome-control-center is, which is *much* worse than mismatched versions), you might as well not check anything whatsoever and go straight to cinnamon-settings.py without checking what exists. Would you prefer that?

collinss:
If cinnamon is installed, but c-c-c isn't then it should break. c-c-c is a dependency of Cinnamon, and so Cinnamon should never be installed without it. If anything, we should remove the g-c-c code as I can't see a use case at this point where we would want it.

eli-schwartz:
Or if I use `cinnamon-settings sdfsdfsdfs` then it could in theory give me a console log stating that the module is unknown but cinnamon-settings is trying to load it anyway just in case. Currently this whole file is pretty confusing in terms of "what is it doing, and why don't we just check to see if xlet-settings should be used, otherwise unconditionally call cinnamon-settings.py".

Perhaps logging for unknown modules could be incorporated directly into cinnamon-settings.py, actually.

collinss:
Yeah, that would probably be an improvement. Though I will say that the file is rarely called directly anyway (which is probably why the file has remained largely unchanged in the last 6 years or so since it was written).

PR #7386, opened 2018-03-23
Window list applet: fix bad allocation of icons and labels

germanfr:
* Place icon and label according to their parent borders and paddings (aka don't draw these over the borders).
 * Use correct spacing of elements.
 * Update default theme and remove unused class.

Basically, avoid things like this (before/after):
screenshot from 2018-03-23 00-58-09


smurphos:
I think the `#appMenuIcon` selectors can go as-well in the default theme.

JosephMcc:
> I think the #appMenuIcon selectors can go as-well in the default theme.

The default theme probably needs a good clean up but it needs to be in it's own PR.

PR #7388, opened 2018-03-23
tooltips.js - base #PanelTooltip position on panel width / height

### LINKED ISSUE #7267: Workspace applet tooltip inconsistent positioning.
```
Cinnamon version - 3.6.7
Distribution - Linux Mint 18.3 Sylvia
Card: Intel 2nd Generation Core Processor Family Integrated Graphics Controller
Display Server: X.Org 1.18.4 drivers: (unloaded: fbdev,vesa)
64 bit
 ```

Workspace applet tooltips are aligned to the edge of the applet workspace icon and not the edge of the panel so they overlap the panel slightly.  Other applets tooltips by default are aligned to the panel edge. I think for visual consistency the workspace applet tooltip should align in the same way as other panel item tooltips.

### PR DESCRIPTION
**Pull request by smurphos**
Hi,

Fixes #7267 & #7266 
Complements #7325

A similar problem to that addressed in #7325 affects tooltip positions in panels with a border.
e.g. 
screenshot from 2018-03-23 10-55-55

This commit changes the calculation of the tooltip vertical position to be based on the panel height for horizontal panels, and panel width for vertical panels. A one pixel buffer is added to ensure that there is some separation between a panel with borders and a tooltip with borders.

After - 
screenshot from 2018-03-23 10-56-56

Please be gentle if I've done anything really stupid in the JS - it's been a number of years since I've done any JS work and I am an amateur by any definition of the word.

### DIFF
--- a/js/ui/tooltips.js
+++ b/js/ui/tooltips.js
@@ -327,16 +327,16 @@ PanelItemTooltip.prototype = {
             if (!panel.getIsVisible()) continue;
             switch (panel.panelPosition) {
                 case PanelLoc.top:
-                    panelTop = panel.actor.height + 1;
+                    panelTop = panel.actor.height;
                     break;
                 case PanelLoc.bottom:
-                    panelBottom = panel.actor.height + 1;
+                    panelBottom = panel.actor.height;
                     break;
                 case PanelLoc.left:
-                    panelLeft = panel.actor.width + 1;
+                    panelLeft = panel.actor.width;
                     break;
                 case PanelLoc.right:
-                    panelRight = panel.actor.width + 1;
+                    panelRight = panel.actor.width;
                     break;
             }
         }
--- a/js/ui/tooltips.js
+++ b/js/ui/tooltips.js
@@ -315,59 +315,39 @@ PanelItemTooltip.prototype = {
         let tooltipWidth = this._tooltip.get_allocation_box().x2 - this._tooltip.get_allocation_box().x1;
 
         let monitor = Main.layoutManager.findMonitorForActor(this._panelItem.actor);
-        
-        let panels = Main.panelManager.getPanelsInMonitor(Main.layoutManager.monitors.indexOf(monitor));
-        
-        let panelTop = 0;
-        let panelBottom = 0;
-        let panelLeft = 0;
-        let panelRight = 0;       
-        
-        for (let panel of panels) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by germanfr** [comment]
I guess that is not a bug but a feature, and that's the reason I didn't change that too. Also this is adding some overhead to show a tooltip and is inconsistent with how they behave in other scenarios, like for example when they are not in the panel.
screenshot from 2018-03-23 13-01-29
I think they are supposed to be next to their source, not outside their parent :)

**Comment by germanfr** [comment]
A cheaper way to do it would be to just use the panel height from `this._panelItem` as you already know the position. For example:
```js
switch (this.orientation) {
            case St.Side.BOTTOM:
                tooltipTop = monitor.height - tooltipHeight - this._panelItem.actor.height;
```

smurphos:
Thanks for the feedback. `this.panelItem.actor` is the applet, not the panel unfortunately. I've have just double-checked and using it in that way appears to replicate the original code in terms of what you see on the screen.

I'll see if I can work out how to do it without looping through all the visible panels - it should be possible given as you say we know which orientation we get to `switch (this.orientation)`

Re bug or feature this has in any case 'bugged' me for ages and I do enjoy the learning experience of trying to fix the little things. ;-)

smurphos:
Removed the 1px buffer added to panel height/width as this was inconsistent with the position of GTK controlled tooltips for systray icons.

germanfr:
> this.panelItem is the applet

Then you could do:
```js
switch (this.orientation) {
     case St.Side.BOTTOM:
         tooltipTop = monitor.height - tooltipHeight - this._panelItem.panel.actor.height;
```
Although it looks a bit dirty.

smurphos:
Thanks,

Just tried some tests with that - it works but not in all cases. It throws an x-sessions error with panel-launcher tooltips - I guess because in that case the parent is the panel launcher applet and not the panel it sits on.

```
(cinnamon:3339): Cjs-WARNING **: JS ERROR: TypeError: this._panelItem.panel is undefined
PanelItemTooltip.prototype.show@/usr/share/cinnamon/js/ui/tooltips.js:339:17
TooltipBase.prototype._onShowTimerComplete@/usr/share/cinnamon/js/ui/tooltips.js:112:13
```

smurphos:
Amended to use getPanel in switch statement to get the correct panel rather than getPanelsInMonitor and getting panel dimensions for irrelevant panels.

PR #7389, opened 2018-03-23
Applets, desklets: switch to class syntax, remove try-catch blocks

jaszhix:
- Updates stock xlets with class syntax. `_init` needed to be replaced with `constructor` for cases where multiple levels of classes are subclassed for this to work, and the usage was adapted everywhere else for consistency. `constructor` is the equivalent of `function MyApplet(...) {...}` and `super` is the equivalent of `Applet.Applet.prototype.call(this, ...)`. Super calls the parent object.

- Remove large try-catch blocks found in _init methods and a couple other spots. This was done because there are already two or more levels of try-catch blocks wrapping xlet execution. Try-catch isn't free performance-wise, and can make debugging more difficult. Xlet exceptions are always caught by either an xlet manager, extension.js, or fileUtils.js.

- In the files where a lot of lines were already changed and the indentation looked rough, I ran [prettier-eslint](https://github.com/prettier/prettier-eslint) on the file. This auto-formats the code using the .eslintrc.json file in the Cinnamon repo.

- Updated fileUtils.js so classes can be imported from the xlet importer easily, for the Calendar applet. This cannot currently be done in non-xlet Cinnamon files without lexical scope warnings occurring from GJS' module implementation.

- Made the TooltipBase class a subclass-able object since it is being used that way in the window list applet - doesn't work with class syntax otherwise.

- Fixed SystemButton from the menu applet subclassing PopupSubMenuMenuItem and intializing with PopupBaseMenuItem, for class support. PopupSubMenuMenuItem inherits PopupBaseMenuItem.

- Excluded the network applet because of its complexity and outstanding PRs still open. Don't mind rebasing if other conflicting PRs get merged.

- Gives xlet constructors unique names.

- Fixes "reference to undefined property Symbol.toPrimitive" warning from the sound applet.

### DIFF
--- a/files/usr/share/cinnamon/applets/a11y@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/a11y@cinnamon.org/applet.js
@@ -24,11 +24,11 @@ const HIGH_CONTRAST_THEME = 'HighContrast';
 
 const Keymap = Gdk.Keymap.get_default();
 
-function MyApplet(metadata, orientation, panel_height, applet_id) {
+function CinnamonA11YApplet(metadata, orientation, panel_height, applet_id) {
     this._init(metadata, orientation, panel_height, applet_id);
 }
 
-MyApplet.prototype = {
+CinnamonA11YApplet.prototype = {
     __proto__: Applet.TextIconApplet.prototype,
 
     _init: function(metadata, orientation, panel_height, instance_id) {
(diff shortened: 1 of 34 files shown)


brownsr:
All working fine for me :-)

brownsr:
Ah ... cinnamon process was up to 600mb a couple of hours later, way more than I normally see. Might want to check for memory leaks.

jaszhix:
I currently have the window list and menu added to a second panel on another monitor for testing. Memory looks steady around 450MB, haven't restarted Cinnamon in a few hours. Will keep monitoring, thanks for testing.

jaszhix:
That got me curious, so I started using the menu more, and I think that's where the leak is occurring. Went to 575 MB pretty quickly while moving between categories and searching. That applet was the hardest to get working because there's a lot of subclassing going on. Will try to look into more this weekend.

PR #7393, opened 2018-03-24
Default menu - fix unused selector .menu-category-button-button

smurphos:
Hi,

`.menu-category-button-button` is unused and may simply be an uncaught typo from many years ago.

`..menu-category-button-selected` is valid and makes sense in this context and as written does distinguish visibly between the selected category and a hovered and selected category. Whether the visual effect is desirable I'll leave for others to decide.

### DIFF
--- a/data/theme/cinnamon.css
+++ b/data/theme/cinnamon.css
@@ -1151,8 +1151,6 @@ StScrollBar StButton#vhandle:hover {
 
 }
 .menu-category-button-selected:hover {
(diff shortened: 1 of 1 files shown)


clefebvre:
Hi @smurphos, it's definitely a fix. The result looks awful though :)

Can you remove the radius/backgorund-color?

smurphos:
Good call... ;-) Done.

PR #7395, opened 2018-03-24
Menu applet: autoselect exact matches

Tak:
If there's an exact match for the user's search string (application names or places), select (and scroll to) that entry in the list instead of the first alphabetical entry.


Tak:
The things that CodeFactor found seem to be issues in the preexisting code

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -3216,22 +3227,26 @@ MyApplet.prototype = {
         }
     },
 
+    _matchNames: function(names, pattern){
+        var res = [];
```
jaszhix: Although older code is using `var`, in new code we use lexical scoping with `let`.

Tak:
Is there more action I need to take on this?

clefebvre:
Hi @Tak, could you rebase this PR please?

clefebvre:
You're missing something here..

`if (apps.indexOf(appId)) != -1) {`

NikoKrause:
@Tak

Tak:
Oops, must have missed that notification, fixed.

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -2971,10 +2972,15 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
             }
         } else if (apps) {
             for (let i = 0; i < this._applicationsButtons.length; i++) {
-                if (apps.indexOf(this._applicationsButtons[i].app.get_id()) != -1) {
-                    this._applicationsButtons[i].actor.show();
+                let button = this._applicationsButtons[i];
+                let appId = button.app.get_id();
+                if (apps.indexOf(appId) != -1) {
+                    button.actor.show();
+                    if (appId == exactMatch) {
```
marcus-lundgren: Don't use `==`. Use `===` instead

marcus-lundgren: @clefebvre Can you make checks like these, e.g. not using `==`, a part of the CI/code style enforcement in javascript?

Tak: Updating. FWIW, the existing code doesn't use ===/!==

marcus-lundgren: Yeah, I know. And those should definitively be fixed at some point (in a different PR). It is a buggy operator that should never be used, and is scary to change. Making it at least better in the area we touch goes a long way :-)

Tak:
Thanks!

PR #7396, opened 2018-03-24
hicolor icon: Add removable-drives icon

NikoKrause:
I want it to use for the removable-drives applet.
It's the icon `usermount` from the Mint-X theme


germanfr:
Hey, why don't you use `drive-removable-media`? It's a standard icon.
Adwaita also has `media-removable` but it isn't standard.

NikoKrause:
Hmmm... I didn't liked those icons. But I guess it's the wiser choice to use those icons o.O?

PR #7399, opened 2018-03-25
Inhibit applet icon: fix uncolored parts

germanfr:
The ZZ's inside where not paths and they weren't affected by the color property. I converted them to paths.

Screenshot of Mint-X: 
screenshot from 2018-03-25 21-53-37


JosephMcc:
Ah cool. I forget about that sometimes. Nice catch.

PR #7402, opened 2018-03-26
Sound applet: fix missing muted icon

germanfr:
`microphone-sensitivity-none` does not exist in Adwaita, it's `microphone-sensitivity-muted`.


collinss:
This should also be changed in mint-x-icons. https://github.com/linuxmint/mint-x-icons/blob/master/usr/share/icons/Mint-X/status/scalable/microphone-sensitivity-none-symbolic.svg

germanfr:
@collinss done.

PR #7405, opened 2018-03-27
environment/overrides: Harden St.Label.prototype.set_text

jaszhix:
- Warns about and prevents non-strings from being passed to `set_text`
- Prevents `set_text` from being called if the text hasn't changed
- Cleans up `overrideDumpStack`

### DIFF
--- a/js/ui/environment.js
+++ b/js/ui/environment.js
@@ -103,6 +103,21 @@ function init() {
         return St.describe_actor(this);
     };
 
+    let originalSetText = St.Label.prototype.set_text;
+    St.Label.prototype.set_text = function(text) {
+        // Warn about non-strings being passed to set_text.
+        if (typeof text !== 'string') {
+            global.dump_gjs_stack(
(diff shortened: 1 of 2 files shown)


js/ui/environment.js
```diff
@@ -103,6 +103,21 @@ function init() {
...
+    let originalSetText = St.Label.prototype.set_text;
+    St.Label.prototype.set_text = function(text) {
+        // Warn about non-strings being passed to set_text.
+        if (typeof text !== 'string') {
+            global.dump_gjs_stack(
+                `Invalid value of type ${typeof text} passed to St.Label.prototype.set_text.`
+            );
+            return false;
+        }
+        // Only call set_text if the text has changed.
```
mtwebster: This is already done in st_label_set_text() - https://github.com/linuxmint/Cinnamon/blob/master/src/st/st-label.c#L364-L365

jaszhix: Okay, I removed it since its not needed. I could have left the warning in, but don't think the extra function call is worth it.

PR #7408, opened 2018-03-28
Repaint node when background position changes

germanfr:
For example when you change the handle position of a switch.

### DIFF
--- a/src/st/st-theme-node.c
+++ b/src/st/st-theme-node.c
@@ -3862,6 +3862,17 @@ st_theme_node_paint_equal (StThemeNode *node,
   if (g_strcmp0 (node->background_image, other->background_image) != 0)
     return FALSE;
 
+  if (node->background_image)
+    {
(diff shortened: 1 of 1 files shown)


clefebvre:
Hi @germanfr,

Can you describe what this fixes? How to reproduce it?

germanfr:
I think this for example will expose the issue. I'm not at home right now and I can't test it. But it's obvious that if two nodes have different background possitions they should not `st_theme_node_paint_equal`.

```css
.toggle-switch {
	width: 32px;
	height: 16px;
	border-radius: 8px;
	background-color: gray;
	background-image: url("switch-handle.svg");
	background-position: 0px 0px;
}
.toggle-switch:checked {
	background-position: 16px 0px;
}
```
The switch won't update on `:checked` unless you change another visible property. It's just a dummy example, I had a real issue with this but it was more complicated.

PR #7414, opened 2018-03-29
menu applet - if no selected item when adding search provider results make first search provider result selected item

### LINKED ISSUE #7412: Cinnamon Menu: Search keyboard navigation not working with search provider results
```
 * Cinnamon version 3.6.7
 ```

When using the search function of the menu, you can use the arrow keys to navigate through the results. If all the results returned originate from search providers, the arrow keys do nothing. However, in the latter case, if you hover over one of the results with the mouse and the press an arrow key, the navigation works.

smurphos:
Just reorders code so that search providers results are added before selecting first result as selected item.

Fixes #7412 

@NikoKrause - Stark Menu has the same issue and the same fix I think would apply. Cinnamenu doesn't have this issue.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -3345,7 +3345,7 @@ MyApplet.prototype = {
             }
             }catch(e){global.log(e);}
         }));
- 
+
         return false;
     },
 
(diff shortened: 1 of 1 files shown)


collinss:
This only works if it returns results are returned synchronously. Many of the search providers do not return synchronously (such as apt, wikipedia, etc.), and so it doesn't work for those.

smurphos:
This works OK for me - tested with Wikipedia, Apt, Tracker and my Google Search Provider in combination and individually. I'm not sure if there is a better/cheaper way to get the index value given the check should only return true just after adding the first search provider result?

PR #7417, opened 2018-03-29
debian/control: use python3-pampy instead of python3-pam

Fantu:
python-pam is old and python3-pam is not present in stretch
I not tested build for now but other changes should not be needed FWIK
because fedora already use it without additional patches


clefebvre:
  File "/usr/share/cinnamon/cinnamon-settings/modules/cs_user.py", line 371, in _on_current_password_changed
    self.auth_pam() if pam else self.auth_PyPAM()
  File "/usr/share/cinnamon/cinnamon-settings/modules/cs_user.py", line 352, in auth_pam
    if not pam.authenticate(GLib.get_user_name(), self.current_password.get_text(), 'passwd'):
AttributeError: module 'pam' has no attribute 'authenticate'

PR #7419, opened 2018-03-30
xlet-settings: improvements to key dependency handling

collinss:
* Settings with dependencies now use a revealer rather than changing the sensitive property.
* Sections can now have dependencies in addition to just widgets.
* Dependencies can now be inverted by preceding the key name with !.
* Simple expressions using boolean operators can be used to compare the value of the key to other values than just true and false.
* The settings example applet was updated to showcase these additional features.
* The settings api reference tutorial was updated to include these changes.


clefebvre:
I'm getting random issues with this...

After adding the example settings applet, it fails to load and I get:

error t=2018-04-23T13:43:53Z Could not set up binding - settings object was not initialized successfully for settings-example@cinnamon.org
error t=2018-04-23T13:43:53Z Could not set up binding - settings object was not initialized successfully for settings-example@cinnamon.org
error t=2018-04-23T13:43:53Z 
[settings-example@cinnamon.org]: Expected type filename for Argument 'path' but got type 'undefined'
[settings-example@cinnamon.org]: Failed to evaluate 'main' function on applet: settings-example@cinnamon.org/24

When removing it from the panel I get this:

  File "/usr/share/cinnamon/cinnamon-settings/xlet-settings.py", line 113, in _on_proxy_ready
    proxy.highlightXlet('(ssb)', self.uuid, self.selected_instance["id"], True)
TypeError: 'NoneType' object is not subscriptable
  File "/usr/share/cinnamon/cinnamon-settings/xlet-settings.py", line 437, in quit
    proxy.highlightXlet('(ssb)', self.uuid, self.selected_instance["id"], False)
TypeError: 'NoneType' object is not subscriptable

Note: I'm out of file-handles, so files monitors are not working here on this machine atm. Not sure if it's related or not. I don't get these issues on master though.

PR #7420, opened 2018-03-30
Add a script to install 3rd party spices from a local folder. This is…

collinss:
… meant to be a helper script to aid in applet development.

### DIFF
--- a/files/usr/bin/cinnamon-install-spice
+++ b/files/usr/bin/cinnamon-install-spice
@@ -2,6 +2,7 @@
 
 import sys
 import os
+import argparse
 
 import gi
 gi.require_version('Gtk', '3.0')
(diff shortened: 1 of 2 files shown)


files/usr/bin/cinnamon-install-spice
```diff
@@ -0,0 +1,44 @@
...
+
+if spice_type not in ['applet', 'desklet', 'extension', 'theme', 'search_provider']:
+    print('invalid argument %s' % spice_type)
+
+if not os.path.exists(source_dir):
+    print('invalid directory')
+    quit()
+
+if spice_type is 'theme':
+    theme_file = os.path.join(source_dir, 'Cinnamon/cinnamon.css')
```
germanfr: Shouldn't this be `cinnamon/cinnamon.css` all in lower case?

collinss: Oh yeah, probably. I wrote this pretty late last night, and haven't really done a lot of testing yet (part of the reason it's labeled WIP :) ). I'll fix it later when I get the chance.

files/usr/bin/cinnamon-install-spice
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/python3
+
+import sys
+import os
+import argparse
+
+import gi
+gi.require_version('Gtk', '3.0')
```
maxyz: You don't seem to be using the gi import in this file.

collinss: Those lines prevent a warning when Spices.py is imported.

maxyz: I see, then, shouldn't these lines be added to Spices.py?

collinss: Probably. They weren't there before because Spices.py was always only imported from cinnamon-settings, which already had those lines. Another option could be to see if we can refactor it so that we don't need to use Gtk as an import in Spices.py.

PR #7426, opened 2018-04-02
Added "shutdown immediately" entries for lid close and power critical.

Vanov:
This change introduces the option to make your machine turn off completely when the lid is closed, as well as turning of when the power reaches a critical level.
I always find myself adding these lines manually in all my machines, so I figured I might not be the only one.


clefebvre:
Thanks @Vanov, quick question.. why "immediately"?

Vanov:
@clefebvre I was simply mimicking the other lines. If you look, every other line that calls for a shutdown says immediately for some reason.

PR #7444, opened 2018-04-08
[WIP] background effects: rework and turn on.

brownsr:
This reworks, enables, and extends the previous background effect facility that was turned off last release as it was broken.

There are two possible background effects - bumpmap and blur.  Both are turned on by CSS, and can be layered on each other if desired.
example CSS 

    background-bumpmap: url("/usr/share/cinnamon/bumpmaps/hex.png");
    background-blur: 1px;

Note that the size of the blur is treated as an iteration count.  So background-blur: 1px is the basic effect, while background-blur: 2px  applies the effect again on top of itself, and so on.  The basic blur is relatively mild, twice is stronger, anything above 3 is overkill IMO.
Speeds seem fine with no noticeable speed penalty unless a large blur count is put in for test purposes.

The blur effect is probably the more useful.  The bumpmap effect looks dramatic on a background, but rubbish over something like a text file being edited.

The effect works by snapshotting the screen underneath an actor, applying the effect to it, and then drawing it out as the lowest layer in the process of drawing an actor.  There are no restrictions as to what can be layered on it, though clearly you won't want to put something without transparency over it or the effect would be hidden.

In the screenshots below the bumpmap is shown with nothing layered on it.  The  blur is shown with a semi-transparent background-gradient on top of it.

Still to do:

1) Mask out the corners where there is a border radius.  You can see the effect bleeding outside the border in the images below.
2) Consider other potential effects
3) Optimise

screenshot from 2018-04-08 13-59-20

screenshot from 2018-04-08 13-58-16

### DIFF
--- a/src/st/st-background-effect.c
+++ b/src/st/st-background-effect.c
@@ -1,30 +1,59 @@
 #include "st-background-effect.h"
 #include "st-cogl-wrapper.h"
-#define ST_BACKGROUND_EFFECT_CLASS(klass)    (G_TYPE_CHECK_CLASS_CAST ((klass), ST_TYPE_BACKGROUND_EFFECT, StBackgroundEffectClass))
-#define ST_IS_BACKGROUND_EFFECT_CLASS(klass) (G_TYPE_CHECK_CLASS_TYPE ((klass), ST_TYPE_BACKGROUND_EFFECT))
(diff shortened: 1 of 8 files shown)


brownsr:
Moved to another PR

matrix9180:
Could you please link to the new PR...

JosephMcc:
@matrix9180 This is the merged PR: https://github.com/linuxmint/cinnamon/commit/6aeeb22fc9a2ba339b606f343ec025dce6f796d7

Issue #7195, opened 2018-01-13
Adding the Workspace-switcher applet disables clicking on the corner of the screen


Janne-d-Arc:
```
 * Cinnamon version 3.6.7
 * Distribution: Mint 18.3
 * Graphics hardware *and* driver used: GeForce 950 with Nvidia Drivers 384.111
 * 64 bit
 ```

**Issue**
Usually I don't need to click exactly on the icon of the applet at the end of the panel to activate it, I just need to click on the same corner of the screen, but if I add the Workspace-switcher applet, this function doesn't work anymore.

**Steps to reproduce**
Just adding the Workspace-switcher applet, any other applet at the end of the panel will not be activated after clicking on the same corner of the screen.

**Expected behaviour**
Clicking on the corner of the screen should activate any applet at the end of the panel even if the Workspace-switcher applet has been added.

**Other information**
Mouse pointer on the screen corner without the Workspace-switcher applet:
![panel1](https://user-images.githubusercontent.com/13596575/34902810-bd6d8bfc-f7e8-11e7-8e15-5575162fc7fd.png)

Mouse pointer on the screen corner with the Workspace-switcher applet:
![panel2](https://user-images.githubusercontent.com/13596575/34902811-bd7f8104-f7e8-11e7-8f86-f3491cf8e12f.png)

JosephMcc:
I can confirm this. Weird but I don't know why it's happening.

germanfr:
I accidentally got a fix for this in #7456.

PR #7456, opened 2018-04-11
Workspace switcher: fix blurry #panelRight's applets in visual mode

germanfr:
With simple buttons:
screenshot from 2018-04-12 12-20-47

In visual mode:
screenshot from 2018-04-12 12-21-09

After this patch:
screenshot from 2018-04-12 12-23-26


germanfr:
Also, seems to fix #7195.

PR #7465, opened 2018-04-14
Restore broken backwards compatibility in the cinnamon-json-makepot util

germanfr:
* The last argument is not always the out path (for example with -i)
 * The first argument can't be stripped because it is replaced by the executable name after calling exec.


germanfr:
Although it has been replaced by `cinnamon-xlet-makepot` it was there to maintain backwards compatibility, but it wasn't working at all.

collinss:
@germanfr I already had a fix pending for this file in #7390. I went ahead and added fixes to that pr for the other issues you mentioned here. Hope you don't mind. It was easier than waiting for one to be merged and then resolving the merge conflicts in the other.

Issue #6136, opened 2016-12-28
How to disable user switching in cinnamon?


Tomek-1:
Hi all,

I'd like to disable user switching option for all users.

In a pure gnome it only needs, to set two following keys in dconf database:

> dconf write /org/gnome/desktop/lockdown/disable-user-switching true
> dconf write /org/gnome/desktop/screensaver/user-switch-enabled false

and works as needed.

For the cinnamon I added two more keys:

> dconf write /org/cinnamon/desktop/lockdown/disable-user-switching true
> dconf write /org/cinnamon/desktop/screensaver/user-switch-enabled false

   
but it does not help - user switching is still available.
It seems that cinnamon ignores that keys.

Regards,

My system:

Debian 8.6 (Jessie)
Kernel:  4.8.11-1-bpo8+1 amd64
Cinnamon: 3.0.4-2-bpo8+1 (jessie-backports)

DVeris:
I have the same problem.
it is bug

DVeris:
System:    Host: ilya-HP-G72-Notebook-PC Kernel: 4.4.0-53-generic x86_64 (64 bit gcc: 5.4.0)
           Desktop: Cinnamon 3.2.7 (Gtk 3.18.9-1ubuntu3.2)
           Distro: Linux Mint 18.1 Serena

DVeris:
dconf write /org/cinnamon/desktop/screensaver/user-switch-enabled false
Switch users button are still available

leigh123linux:
@mtwebster screensaver seems to ignore the key.

Tomek-1:
The workaround is to rename or delete file:

/usr/bin/gdmflexiserver

Of course if you use GDM.

Regards.


2017-04-15 9:24 GMT+02:00 DVeris <notifications@github.com>:

icarter09:
@Tomek-1 can you check if this issue was fixed from the merged PR?  If not, then feel free to reopen this issue. Thanks.

PR #7476, opened 2018-04-17
key "org.cinnamon.desktop.lockdown" consideration

G3rb:
related to this feature request.
#6136
I change the code for this applet just to remove the "switch user" option when the gsettings key "org.cinnamon.desktop.lockdown" "disable-user-switching" is true. I need it in Linux Mint 18.3 so I change it in this version.


clefebvre:
Implemented at https://github.com/linuxmint/Cinnamon/commit/1e9e82a19d4cfafd043b4a683b0a74a0b9d95f29.

Please don't use patches/curl, github doesn't seem to handle them properly.. https://patch-diff.githubusercontent.com/raw/linuxmint/Cinnamon/pull/7476.patch. You're better off using proper git branches/forks.

G3rb:
Ok, I'm new on Github. I need more practice. 

thanks

PR #7482, opened 2018-04-18
Retire outdated bluetooth code.

eli-schwartz:
- HAVE_BLUETOOTH is never used after it is defined in js/mis/config.js.in
- pkg-config checks for applet_libdir fail since https://github.com/GNOME/gnome-bluetooth/commit/a6cb98e2e48db37a4b9cc3981f463cd135076d8c
  gnome-bluetooth only supports being called from the system girepository
- BLUEBERRY_LIBS are not actually linked into the final binaries, since
  they aren't included anywhere and no code calls functions from them.
- Everything seems to be moved over to blueberry anyway.


eli-schwartz:
Thanks! I initially noticed when one of my users asked me to remove the gnome-bluetooth dependency, since "it works if I manually force-remove it". So I went looking to see what uses it and if it is safe to move to optional dependencies... but then said "hey, this should be removed altogether". :smile:

(I've been looking at pruning outdated dependencies, but didn't look much further than "oh, this is asked for in configure.ac".)

PR #6993, opened 2017-11-07
Network applet: Port to libnm

lkundrak:
The libnm-glib is depreacted for a long time already.

PR #7486, opened 2018-04-20
Unified network change PRs with additional debian/ fix

Fantu:
This include https://github.com/linuxmint/Cinnamon/pull/6993 and https://github.com/linuxmint/Cinnamon/pull/7118 and require/include https://github.com/linuxmint/Cinnamon/pull/7506


Fantu:
Found another issue with circleci tests: gir1.2-nm-1.0 is missed in stretch (my tests on stretch-backport with updated network-manager from unofficial backport not helped to saw it), now I'll check for solution

Fantu:
tried today on stretch with cinnamon 3.6 debian packages backport including this PR and cinnamon crash, tried to rebuild network-manager with https://github.com/NetworkManager/NetworkManager/commit/b2af5f7ab676675a496adfa08541ecca43a9d9f8 but still crash and I was unable to debug the exact cause of issue because the backtrace seems not directly related to it but its conseguence :(

clefebvre:
OK, we're getting closer.

@Fantu can you rebase to solve the conflict on debian/control?

I found the source of the issue (crash when network applet is running):

https://git.gnome.org/browse/network-manager-applet/commit/?id=7a59d41e5f6666d0da51f1f7aae7518befdb1182

This issue is present in Debian stretch. It makes gir1.2-nma use the old libs (NMClient/glib) instead of libnm. This results in making it impossible to use NM and NMA gir at the same time.

In Cinnamon libnma is used via gir in modemManager.js.

clefebvre:
So, to summarize the state of this PR:

In Mint 19 and modern distros, everything's working fine, no problem at all.

In LMDE 3 and Debian (with NM 1.6), everything's working after applying two patches:

- https://git.gnome.org/browse/network-manager-applet/commit/?id=7a59d41e5f6666d0da51f1f7aae7518befdb1182 fixes NM and NMA conflicting when loaded over GI and Cinnamon to crashing as a result.
- https://github.com/NetworkManager/NetworkManager/commit/b2af5f7ab676675a496adfa08541ecca43a9d9f8 fixes the wireless toggle and the stdout warnings.

leigh123linux:
So Debian is broken by design?, why should we accommodate their frozen broken packages?

Fantu:
I also think is not good :(
I opened bugs about but need someone explain better:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=892998 (this I also not updated that not is the cause of crash)
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=896818

leigh123linux:
@Fantu Do debian maintainers like a challenge?, it took redhat 4 months to fix.

https://bugzilla.redhat.com/show_bug.cgi?id=1520398

Can debian do better than an enterprise distro?

clefebvre:
We just need to illustrate the issues properly and show that they're specific to Debian's versions of libnm and libnma and affect any Debian developer, not some random future version of Cinnamon which Debian doesn't ship.

I did this here for libNMA:

https://gist.github.com/clefebvre/411ab17558fc4dc438085ce2b6cc9093

and here for libNM:

https://gist.github.com/clefebvre/c957c3eb290116d07aa524e97743a79e

@Fantu can you show these explanations to Michael Biebl <biebl@debian.org>. They speak for themselves and they're isolating both issues.

Then the process is easy:

- We get these issues patched in Debian Stretch
- libNM becomes a library which works in all major distros
- we migrate to it with this PR
- we tag the move in a 3.8.x point release
- everybody's happy and everything works everywhere

maxyz:
I don't think that we have to wait for a solution of this issue in stretch, if that ever happens, it's going to take a while, it needs approval of the stable release team, and the package will only be in proposed updates till the next point release.
I understand that this might be a bit a hassle for lmde3, but it might be simpler and faster to ship network-manager package (if you use a version 1.6.2-3+~lmde3 then it would be possible for a 1.6.2-3+deb9 package to update it) with the mentioned patches there.

Fantu:
@clem Watch https://github.com/linuxmint/Cinnamon/pull/7536 instead that is better (rebased), also include a change commit I missed here and add also a fix for a bug introduced in 3.8.1 network changes

PR #7487, opened 2018-04-20
user applet: create our own switch-user-symbolic icon

NikoKrause:
the heads/circles in the system-switch-user-symbolic icon from Adwaita
are not paths and therefore don't change their color

create our own icon based on the Adwaita one

### DIFF
--- a/files/usr/share/cinnamon/applets/user@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/user@cinnamon.org/applet.js
@@ -92,7 +92,7 @@ MyApplet.prototype = {
 
             if (GLib.getenv("XDG_SEAT_PATH")) {
                 // LightDM
-                item = new PopupMenu.PopupIconMenuItem(_("Switch User"), "system-switch-user", St.IconType.SYMBOLIC);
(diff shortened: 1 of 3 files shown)


NikoKrause:
This issue wasn't as visible in dark themes, but it's obvious in brighter themes:
screenshot-cinnamon-2018-04-20-205140
screenshot-cinnamon-2018-04-20-205054

I used the user-switch, which has been used before in this applet, and added the arrow from the Adwaita icon.

NikoKrause:
@jimmac You added the `system-switch-user-symbolic` icon to the Adwaita icon theme. Can you fix this issue in Adwaita?
screenshot-cinnamon-2018-04-20-205140

JosephMcc:
I don't agree with this. Users like when there icons change to match their icon theme. You're breaking that here to essentially cover what looks like a bug in Adwaita. The better solution would be to add the one you want to our own themes. Not change to some non standard name.

NikoKrause:
@JosephMcc The icons in this applet have never been modifiable with themes, because it ships with it's own icons. I just changed this particular icon to the Adwaita icon recently.

https://github.com/linuxmint/Cinnamon/tree/master/files/usr/share/cinnamon/applets/user%40cinnamon.org/icons

PR #7503, opened 2018-04-24
Fix panel tooltips position with multiple monitors

germanfr:
Also removed redefinition of the panels location enum.

### DIFF
--- a/js/ui/tooltips.js
+++ b/js/ui/tooltips.js
@@ -1,20 +1,17 @@
 const Lang = imports.lang;
 const Mainloop = imports.mainloop;
 const St = imports.gi.St;
+const Gio = imports.gi.Gio;
 
 const Applet = imports.ui.applet;
 const Main = imports.ui.main;
 const SignalManager = imports.misc.signalManager;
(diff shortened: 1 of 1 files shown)


js/ui/tooltips.js
```diff
@@ -1,20 +1,17 @@
...
 const Mainloop = imports.mainloop;
 const St = imports.gi.St;
+const Gio = imports.gi.Gio;
 
 const Applet = imports.ui.applet;
 const Main = imports.ui.main;
 const SignalManager = imports.misc.signalManager;
 const Tweener = imports.ui.tweener;
-const Gio = imports.gi.Gio;
+const {PanelLoc} = imports.ui.panel;
```
jaszhix: CJS will complain about exported variables with lexical scoping since [PanelLoc in panel.js](https://github.com/germanfr/Cinnamon/blob/ef1e87fe642b6d9991484bc33d78d422baf4ee45/js/ui/panel.js#L69) is declared with `const`. Changing it to `var` would fix it.

germanfr: @jaszhix done :+1:

PR #7516, opened 2018-04-28
proper keybindings names

Visgean:
Hi, I think that keybinding names right now dont make sense - I have been using workspace grid and after latest update cant switch up and down - because the shortcuts are being hijacked by expo/scale. 

I have not tested this properly and I am not that familiar with the codebase to see if this is breaking something else. 

Best regards, 
Visgean.


mtwebster:
Hi, 

These names you're attempting to change match with dconf settings for the window manager (muffin.) That extension needs to change the names *it* is using to be unique.  Customarily, one might prefix the name of the extension.

For example, 'desktop-grid-switch-to-workspace-up.'

It is the extension that's attempting to do the hijacking, please request the developer change these names.

Visgean:
I know that the extension is doing the switching but it still does not make sense to hijack the constant names like this, it deserves refactoring.

PR #7521, opened 2018-04-29
[WIP] Rework notification image loading

itzexor:
This needs testing.

This reworks external notification image loading so that we can support icon names in the `image-path` and `app_icon` hints as specified by v1.2 of the specification, prevents uncaught errors loading images, and ensures notifications always have a valid icon.

Note that absolute paths aren't actually mentioned by the specification, but support for that format already existed in our code, so I've kept it.

> The "image-data" and "icon_data" hints should be a raw image data structure of signature (iiibiiay) which describes the width, height, rowstride, has alpha, bits per sample, channels and image data respectively.
> 
> The "app_icon" parameter and "image-path" hint should be either an URI (file:// is the only URI schema supported right now) or a name in a freedesktop.org-compliant icon theme (not a GTK+ stock ID).

### DIFF
--- a/js/ui/messageTray.js
+++ b/js/ui/messageTray.js
@@ -1840,8 +1840,8 @@ MessageTray.prototype = {
         if (AppletManager.get_role_provider_exists(AppletManager.Roles.NOTIFICATIONS) && !this._notificationRemoved) {
             this.emit('notify-applet-update', notification);
         } else {
-            if (notification.isTransient)
-                notification.destroy(NotificationDestroyedReason.EXPIRED);
+            // if there is no tray applet, the notification expires as soon as it is hidden
+            notification.destroy(NotificationDestroyedReason.EXPIRED);
         }
         this._notification = null;
         this._notificationRemoved = false;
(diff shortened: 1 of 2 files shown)


ernestask:
Would be nice to eventually fix the issue. Received https://bugzilla.redhat.com/show_bug.cgi?id=1737167, which is caused by it.

ernestask:
The package maintainer didn’t even bother commenting here, but a head’s-up: the changes are now being tested in Fedora.

leigh123linux:
@ernestask Why would I comment here? as I didn't use this pull request for fedora.


https://src.fedoraproject.org/rpms/cinnamon/blob/master/f/notification.patch

ernestask:
I don’t know, to collaborate with project you’re packaging? It’s ass-backwards to add random patches downstream and not even bother about the upstream.

leigh123linux:
@ernestask That wasn't the case, the patch was provided by mtwebster

itzexor:
I don't think anyone's ever going to test this, and there is now an alternate patch available. Someone can submit that one instead.

PR #7524, opened 2018-04-30
Remove exported variable warning

germanfr:


Odyseus:
Hello, @germanfr.

Just *for the record*.

The change from **let** to **var** fixes the following warning logged when calling `Util.spawn_async`.

> Cjs-WARNING **: Some code accessed the property 'subprocess_callbacks' on the module 'util'. That property was defined with 'let' or 'const' inside the module. This was previously supported, but is not correct according to the ES6 standard. Any symbols to be exported from a module must be defined with 'var'. The property access will work as previously for the time being, but please fix your code anyway.

germanfr:
Yes, thanks @Odyseus.

PR #7526, opened 2018-04-30
Xlet settings: don't show a row for "generic" settings entries

germanfr:


germanfr:
> The generic type seems to already be ignored from what I can tell, though I admit it wouldn't hurt to handle that case.

It isn't ignored this adds a row whether the type was found or not.
```python
if 'dependency' in item:
    revealer = JSONSettingsRevealer(info['settings'], item['dependency'])
    section.add_reveal_row(widget, revealer=revealer)
else:
    section.add_row(widget)
```

I'll apply what you requested, it seems like a better idea :+1:

collinss:
> It isn't ignored this adds a row whether the type was found or not.

Perhaps it's a change with a newer version of Gtk or something (I'm still using an 18.3 base), but in my testing there was nothing else added (or else it was removed as the previous widget was re-added multiple times). Either way, we should probably be handling that case anyway. Thanks!

germanfr:
It's not related to Gtk. The widget variable is reused in the loop and if it isn't re-assigned a new widget it tries to add it again, resulting in an empty row. Open one of my extensions and look closely at the bottom :sweat_smile:

collinss:
Weird, I'm not seeing that here. The inspector doesn't show any extra rows either.

germanfr:
You must be right and it must be a version thing. Here you can see what the problem is:

screenshot from 2018-04-30 23-47-16

Odyseus:
Hello, @everybody.

I can confirm your findings, g@ys:

- Gtk 3.22 (Ubuntu 18.04): I see the same warning and visual glitch as reported by @germanfr.
- Gtk 3.18 (Linux Mint 18.3): No warning and no visual glitch as reported by @collinss.

Applying the changes on this pull request fixes it for Gtk 3.22.

PR #7531, opened 2018-05-01
Notification close policy

germanfr:


clefebvre:
Quick comment about this PR:

Before this PR notifications would disappear after a timeout.
After this PR, when the timeout is reached, notifications remain visible if the mouse is on top, and disappear when the mouse leaves.

PR #7532, opened 2018-05-02
[WIP] Simplify separators

germanfr:
Now that margins are implemented there's no need to do custom painting.
This also allows the separator to extend up to the menu's full with.

This makes ~many~ theme's separators invisible ~(those who don't define a background color).~ They need to do something like this:
```diff
+.popup-menu-item.separator {
+    background-color: #555555;
+    height: 1px;
+    margin: .5em 1em;
+}
 
-.popup-separator-menu-item {
+.popup-separator-menu-item { /* Cinnamon < 4.0 */
     -gradient-height: 1px;
     -gradient-start: #555555;
     -gradient-end: #555555;
     -margin-horizontal: 1em;
     height: 1em;
 }
```
---

**EDIT:** I have changed the commit to use the `separator` class instead of the old `popup-separator-menu-item` class. The vast majority of themes needed to change anyways, and it was hard to keep compatibility with both implementations in the same theme.

### DIFF
--- a/data/theme/cinnamon.css
+++ b/data/theme/cinnamon.css
@@ -47,13 +47,7 @@ StScrollBar StButton#hhandle:hover,
 StScrollBar StButton#vhandle:hover {
     background-color: rgba(204,204,204,0.5);
 }
-.separator {
-    -gradient-height: 2px;
-    -gradient-start: rgba(85,85,85,1);
-    -gradient-end: #555555;
-    -margin-horizontal: 1.5em;
-    height: 1em;
-}
+
 .slider {
     height: 1em;
     min-width: 15em;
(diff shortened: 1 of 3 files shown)


clefebvre:
We should have a PR on the themes ready, so we can merge it all together. I'll mark this as WIP for now.

germanfr:
Ok I'll do it later. It's simple.

germanfr:
I rebased this to master to solve the conflicts and updated the themes in linuxmint/mint-themes#160

clefebvre:
Thanks German, we also need one on https://github.com/linuxmint/cinnamon-spices-themes.

germanfr:
Then that'll take more time. So I'll mark it as WIP again for now.

PR #7542, opened 2018-05-03
[RFT] Bugfix for cs_startup: Fix creation of autostart files

Cobinja:
If an autostart app is changed, and there was no autostart file in the
user directory, the file got created with invalid values, because of
trying to load the non-existing file.

This also contains a little code cleanup for GLib keyfile flags
and fixes a typo when setting an app's description.


maxyz:
Looks good to me

PR #7562, opened 2018-05-07
Cinnamon theme: window-list progress improvements

NikoKrause:
use border radius for progress
use border-radius dependent on top/right/left panel for window-list-box

grey progress (instead of red)


NikoKrause:
screenshot-cinnamon-2018-05-08-160825
screenshot-cinnamon-2018-05-08-160853

Issue #7576, opened 2018-05-13
[Feature Request] screensaver: set "Delay before locking" with manually inputted number


lxhillwind:
```
 * Cinnamon version (Cinnamon 3.8.2)

 ```
In screensaver setting, value of "Delay before locking" can only be chosen from an array (from file `/usr/share/cinnamon/cinnamon-settings/modules/cs_screensaver.py`), which is not flexible.

![old](https://user-images.githubusercontent.com/10025859/40276881-71a43c0c-5c47-11e8-8d6b-b540b61bb915.png)

This option is set with `GSettingsComboBox`; how about change it to `GSettingsSpinButton`?

![new](https://user-images.githubusercontent.com/10025859/40276882-77fbb74c-5c47-11e8-9ccc-ab531e1bda8a.png)

PR #7577, opened 2018-05-13
"Delay before locking" get value from user input.

lxhillwind:
Implements #7576


clefebvre:
Most time durations are configured like that. I don't think we need that flexibility and if someone really needed it (maybe for testing purposes) then the value is also exposed via advanced tools such as gsettings or dconf UIs.

PR #7589, opened 2018-05-20
Sound applet: fix seeking not detected for not hardcoded players

germanfr:
The getter is async but is treated as sync, so players that don't support seeking always showed the control unless they were in the hardcoded array.

For example _Gradio_. It doesn't support seeking but had the seeker.

### DIFF
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -233,6 +233,173 @@ class VolumeSlider extends PopupMenu.PopupSliderMenuItem {
     }
 }
 
+class Seeker extends Slider.Slider {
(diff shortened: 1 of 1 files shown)


germanfr:
I mark this as WIP because there is a little bug I'll fix in a couple of weeks when I have time.

germanfr:
I think this is ready. I solved a little issue that was already there but it didn't reveal until now. I've been using it and no issues. I rebased it just to keep some of the previous variable names. 

PS: I don't know why it is failing to build on lmde, but it is working fine here.

collinss:
> PS: I don't know why it is failing to build on lmde, but it is working fine here.

I'm pretty sure it was already failing before. Looking at the logs, it's getting a segfault even before it starts trying to build Cinnamon, so it's probably a problem in one of the other packages that Cinnamon depends on.

PR #7602, opened 2018-05-25
CSD System Settings

feren:
What the changes do:
- Adds a window to the UI file itself, rather than relying on the code to make a window, so that it's better integrated with its header bar
- Adds a header bar, replacing the 'top_bar' control, and merges the 'header_stack' with the center of the header bar, with 'custom title' turned on, so the top toolbar is one with the window controls in the whole of the UI. Still works 100% with the stock System Settings pages
- Adds a label to the main page toolbar, to substitute for the lost title bar text, which has a value equal to the window's title on init
- Turns on expand for the header_stack, to make it stretch all the way around the title bar from left to right.

I'm personally not expecting this to be merged, or even considered in Cinnamon, but given you made the revamped Mint Welcome a Client Side Decorations window, thought this contribution might as well fit in Mint 19.
Also, sorry if it's based on a slightly outdated version of this file, it's based on the latest version that's on the Mint 19 PPA...

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
@@ -187,11 +187,11 @@ class MainWindow:
     def __init__(self):
         self.builder = Gtk.Builder()
         self.builder.add_from_file(config.currentPath + "/cinnamon-settings.ui")
-        self.window = XApp.GtkWindow(window_position=Gtk.WindowPosition.CENTER,
-                                     default_width=800, default_height=600)
+        self.window = self.builder.get_object('MainWind')
         main_box = self.builder.get_object("main_box")
         self.window.add(main_box)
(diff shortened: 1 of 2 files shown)


feren:
I'm going to be safe and leave the XApp import in there, as even if it causes CodeFactor to detect 'an issue', you people put that import in the code before I modified it, so to be safe I'll leave it be for now...

collinss:
XApp.GtkWindow is used here intentionally because it adds additional features that a normal gtk window does not. For example, it allows cinnamon-settings to specify progress information that can be used by the window list applet (for example when refreshing the cache of 3rd party applets). I also allows the window to properly set it's icon according to which page is shown. The window was removed from the ui file for good reason, and there are several features that will break as a result of this change you are proposing. Also, since this is essentially a visual change, screenshots in the description would be nice so that we can see at a glance what it looks like before and after the change.

PR #7615, opened 2018-06-02
[WIP] Use safe JavaScript comparison operators

marcus-lundgren:
There are a number of comparison operators that are scary, which unfortunately is used in multiple JavaScript files. This PR hopes to change all of the affected files in order to create a good code base. My hope is that in the future these operators will never be used and that this is enforced in the CI environment.

I will not change any other logic in this PR.

### DIFF
--- a/js/ui/dnd.js
+++ b/js/ui/dnd.js
@@ -66,14 +66,14 @@ function addDragMonitor(monitor) {
 
 function removeDragMonitor(monitor) {
     for (let i = 0; i < dragMonitors.length; i++)
-        if (dragMonitors[i] == monitor) {
+        if (dragMonitors[i] === monitor) {
             dragMonitors.splice(i, 1);
             return;
         }
 }
 
 function isDragging() {
-    return currentDraggable != null;
+    return currentDraggable !== null;
 }
 
 var _Draggable = new Lang.Class({
@@ -131,7 +131,7 @@ var _Draggable = new Lang.Class({
         if (this.inhibit)
             return false;
 
-        if (event.get_button() != 1)
+        if (event.get_button() !== 1)
             return false;
 
         if (Tweener.getTweenCount(actor))
@@ -183,11 +183,11 @@ var _Draggable = new Lang.Class({
         // didn't start the drag, to drop the draggable in case the drag was in progress, and
         // to complete the drag and ensure that whatever happens to be under the pointer does
         // not get triggered if the drag was cancelled with Esc.
-        if (event.type() == Clutter.EventType.BUTTON_RELEASE) {
+        if (event.type() === Clutter.EventType.BUTTON_RELEASE) {
             this._buttonDown = false;
             if (this._dragInProgress) {
                 return this._dragActorDropped(event);
-            } else if (this._dragActor != null && !this._animationInProgress) {
+            } else if (this._dragActor !== null && !this._animationInProgress) {
                 // Drag must have been cancelled with Esc.
                 this._dragComplete();
                 return true;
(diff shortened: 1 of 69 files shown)


js/ui/dnd.js
```diff
@@ -183,11 +183,11 @@ var _Draggable = new Lang.Class({
         // didn't start the drag, to drop the draggable in case the drag was in progress, and
         // to complete the drag and ensure that whatever happens to be under the pointer does
         // not get triggered if the drag was cancelled with Esc.
-        if (event.type() == Clutter.EventType.BUTTON_RELEASE) {
+        if (event.type() === Clutter.EventType.BUTTON_RELEASE) {
             this._buttonDown = false;
             if (this._dragInProgress) {
                 return this._dragActorDropped(event);
-            } else if (this._dragActor != null && !this._animationInProgress) {
+            } else if (this._dragActor !== null && !this._animationInProgress) {
```
jaszhix: Changes like this might have side effects, When you do non-strict comparison against `null`, we are sometimes intentionally checking for `undefined` as well.

marcus-lundgren: Yes. This PR is scary. I haven't gotten an environment yet to test this in (waiting for LM 19 BETA). Those issues need to be found and fixed. This PR might however fix other subtle bugs introduced by not using these strict comparisons.

jaszhix: Try this http://mirrors.advancedhosters.com/linuxmint/isos/testing/linuxmint-19-cinnamon-64bit-beta.iso

mtwebster:
I really can't agree with this.

There are way too many changes here for a single PR, particularly with the potential for regressions.

I'd prefer to find instances in our code selectively, where potential for undefined behavior can be demonstrated by not using strict comparisons.  I'd also be willing to start enforcing it for new feature PR's, but not to the total exclusion of abstract comparisons.  I'm not interested in mass changes (and new fun side effects) just 'because.'

Regardless of what the prevailing wisdom and best practices currently are, what we have here works as is.  We'd need to go and change an awful lot of code to cover all our bases again (that is, once we remember what it was we were doing in that particular spot in the first place..) and I don't think the ends justify the means.

I hope this makes some sense, I'm not trying to be discouraging here, but we have practical considerations to think about, as a small group of developers.

marcus-lundgren:
@mtwebster Yeah, just got a similar thought last night that this might be too big in one go.

I guess my initial thinking was that this hopefully could be merged at the beginning of developing a new version (e.g. 4.0). Then every developer could hopefully work together ironing out the potential issues. But I agree that the diff is too large to grasp.

I agree that we shouldn't do the (mass) change just because it probably should be done, when it appears to work. I just have some follow up questions below. Communicating through text can be tricky in these cases, so please know that I just want a constructive discussion and have no hard feelings regarding this PR.

Regarding changing code in small bits. I can definitively put some time and effort in changing code that have no apparent side-effects. E.g. when checking the currently pressed mouse button or when a function call is guaranteed to return values of a determined type. I could write some summary of why this is so as well in the PR. Should be an easy enough task for me to do when the mind is tired from work. These PRs would then be small enough to grasp, at most a file at a time probably.

Could you please elaborate in which cases in your statement of not being in favour of "[...] total exclusion of abstract comparisons" would be more beneficial than to have code that states what is actually intended to happen? I can understand the use of relational abstract operators (<, <=, >, >=), as there is no strict counter-part (unfortunately), but not the == and != ones. I would argue that there will always be potentially an unwanted side-effect if one wishes to make use of the type-conversions.

If a developer chooses to make use of the abstract comparisons, then that person should have to motivate why there is no unwanted side-effect now or in the future with that code block. The issue with abstract comparisons is also that code added in the same area probably will use the abstract operators as well, due to fear of breaking something (as I've seen in a number of PRs here). In the words of the Douglas Crockford: "I am not saying that it isn't useful. I am saying that there isn't a case when it isn't confusing". And if that seems to cumbersome, then why use the operator at all if there is a good strict counter-part that will fix everything?

Phew, that was a lot of text. Hope you've had a great weekend!

marcus-lundgren:
BTW, I will try this big mass change on my own system. Will be interesting to see if it breaks or not :-)

jaszhix:
Its safer to use strict equality, I agree with you, but there are valid use cases for non-strict, like the one I mentioned earlier (I always disable [no-eq-null](https://github.com/eslint/eslint/blob/master/docs/rules/no-eq-null.md)). Take this for example:

```js
'string' == new String('string') // true
'string' === new String('string') // false
```

Some of the code is probably relying on type conversion behavior, and the GObject bindings don't always behave how you might expect in vanilla JS. I think as long as people are aware of the [exceptions](https://dorey.github.io/JavaScript-Equality-Table/), there's no need to change old code, especially without testing.

marcus-lundgren:
@jaszhix Thank you for clarifying how comparing with NULL only will be true for NULL/UNDEFINED! I am too damaged after having worked with PHP for a couple years, as that operator is **much** more broken than the JavaScript equivalent. The formal definition can be seen here btw: http://www.ecma-international.org/ecma-262/8.0/index.html#sec-abstract-equality-comparison

That will also explain your example with the string primitive and string object comparison (case 8 and 9).

Please do send me an example when GObject bindings behave differently than vanilla JS. I would like to stay away from pitfalls as much as I can.

Thank you and @mtwebster for the discussion!

PR #7654, opened 2018-06-16
st: Enable bumpmap background effect

MasonLeeBack:


brownsr:
Hi.  It's currently turned off because it doesn't work properly.  There's already an open PR of mine that works fine apart from not coping with rounded corners.  I've been busy with other things recently, but I should be able to finish this off soon after the Mint 19 release.  I appreciate that you clearly want this bit of functionality back !

Regards

Simon

PR #7655, opened 2018-06-16
Update po/README: provide a link to Launchpad

dhgutteridge:
Provide a link to Launchpad, since I see potential contributors are often using GitHub as their starting point, and trying to submit translation corrections via GitHub rather than Launchpad.


dhgutteridge:
The lmde3 CircleCI test failed with a segfault for some reason, nothing to do with my trivial commit.

PR #7683, opened 2018-07-02
cinnamon-settings-users.py: fix symlink attack vulnerability

mgerstner:
This script runs as root and allows to configure e.g. other user's icon
files. These icon files are written to the respective user's $HOME/.face
location. If an unprivileged user prepares a symlink pointing to an
arbitrary location then this location will be overwritten with the icon
content. This vulnerability thus allows to corrupt the system or other
user's files. The content is not attacker controlled, luckily.

This change introduces temporary privilege drops to the target user's
privileges before attempting to write the respective .face files. This
way the kernel can decide if permissions are okay or not.


mgerstner:
This was assigned CVE-2018-13054.

PR #7733, opened 2018-07-14
[WIP] Network Applet: Added Airplane Mode.

icarter09:
As part of the Cinnamon Roadmap, added Airplane Mode to the Network Applet. This can be accessed from Network Applet in the tray icon.

### DIFF
--- a/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
@@ -1762,6 +1762,9 @@ CinnamonNetworkApplet.prototype = {
             this.menu.addAction(_("Network Connections"), Lang.bind(this, function() {
 				Util.spawnCommandLine("nm-connection-editor");
             }));
+            this.menu.addAction(_("Airplane Mode"), Lang.bind(this, function() {
+                Util.spawnCommandLine("rfkill block all");
+            }));
 
             this.menu.connect("open-state-changed", Lang.bind(this, this._updateForMenuToggle));
 


mrmajik45:
SeemsGood

clefebvre:
Wouldn't that be better as a switch?

icarter09:
@clefebvre Yeah I can change it. Can you provide or guide me towards an example of what you have in mind for it being a switch? Thanks.

clefebvre:
It could be similar to what we have in sound-applet's right click to mute sound output. If you're going to enable that airplane mode, you'll probably want to disable it after a while.

clefebvre:
@mtwebster can you check this and see what you think about it? You worked on rfkill for bt so you might know things we don't :)

icarter09:
@clefebvre thanks for pointing me towards the sound applet. To disable airplane mode, I was just clicking on the wifi icon again. I realize now how non-intuitive that is and the switch would help bridge that gap. If @mtwebster signs off on the underlying logic for rfkill being ok, then I'll add the switch in.

mtwebster:
I've looked into not using rfkill as a wrench in this manner, but honestly I don't see an issue with it, there isn't much in the way of alternatives that I can find.  There's are a couple libraries, but they don't much address this basic control, and at least one is no longer maintained.

icarter09:
@clefebvre after looking more at the network applet, is Airplane Mode a necessary feature? Currently the 'Wireless' connection menu item is a switch, which enables/disables wireless. I feel like Airplane Mode functionality is just a repeat of the Wireless menu switch.

clefebvre:
Not really... well an airplane mode (just like rfkill) would be a little bit larger than just a wireless switch, since it would also apply to bluetooth adapters.

Should such a switch be in the network applet? Is it a required feature? I'm not sure.

clefebvre:
Closing this for now.

Issue #7480, opened 2018-04-18
cinnamon-settings info 


JohnML1:
```
systeminfo from screenfetch:

OS: Mint 18.3 sylvia
Kernel: x86_64 Linux 4.13.0-38-generic
Uptime: 10m
Packages: 2674
Shell: bash 4.3.48
Resolution: 1920x1080
DE: Cinnamon 3.6.7
WM: Muffin
WM Theme: Numix-Cinnamon-Transparent (Mint-X)
GTK Theme: Mint-X-Teal [GTK2/3]
Icon Theme: Mint-X-Blue
Font: Noto Sans 9
CPU: Intel Core i5-6400 CPU @ 3.3GHz
GPU: GeForce GTX 1050 Ti
RAM: 857MiB / 15988MiB

sources.list

deb http://mirror.netcologne.de/linuxmint/packages sylvia main upstream import backport 

deb http://ftp.fau.de/ubuntu xenial main restricted universe multiverse
deb http://ftp.fau.de/ubuntu xenial-updates main restricted universe multiverse
deb http://ftp.fau.de/ubuntu xenial-backports main restricted universe multiverse

deb http://security.ubuntu.com/ubuntu/ xenial-security main restricted universe multiverse
deb http://archive.canonical.com/ubuntu/ xenial partner

 ```

**Issue**
upload systeminfo results in error:

cinnamon-settings info
Loading Info module
gvfs-open: Error:: Fehler beim Öffnen des Ortes: Der angegebene Ort wird nicht unterstützt

translate: error opening URL

**Steps to reproduce**
click on button

**Expected behaviour**

open browser

**Other information**

hope thats no duplicate bug report

smurphos:
I get the same error with 3.6.7 on Mint 18.3 - 'gvfs-open: Error:: error opening location: The specified location is not supported'

I also get the error if I run `upload-system-info` directly so I think this issue belongs in https://github.com/linuxmint/xapps.

Anyway - the issue is resolved for me by running `gist-paste --login` in the terminal and authenticating with my Github credentials.

NikoKrause:
This used to upload as an anonymous gist to github, but this feature has been recently deprecated: https://laravel-news.com/github-deprecates-anonymous-gist-creation

JohnML1:
Thanks NikoKrause/smurphos  for clarification!

.. so pastebin doesn't work too:
[...]

PR #7742, opened 2018-07-16
cs_info: Temporarily hide the Upload system info button

JosephMcc:
This doesn't currently function and we can bring it back once we find a proper
replacement.

Closes: https://github.com/linuxmint/Cinnamon/issues/7480


clefebvre:
This should take care of it: https://github.com/linuxmint/xapps/commit/f40a5376554c2675ec9ed1e4c45d30329a8a4c32

Issue #7175, opened 2018-01-02
Power management applet does not tell user which program is inhibiting power management


CottonEaster:
```
 * Cinnamon version (cinnamon --version) 3.6.7.
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...) 18.2
 * 32 or 64 bit: 64
 ```

**Issue**

The power management applet does not tell user *which* program is inhibiting power management. Thus, when I am confronted by this

![image](https://user-images.githubusercontent.com/9998846/34504523-efc21278-f016-11e7-9d59-7fb11912b1c1.png)

I wonder whether it is the hung file transfers (see [this nemo bug report](https://github.com/linuxmint/nemo/issues/1683)) that are inihibiting the power-saving - or not. 

**Steps to reproduce**

Hard to know - which, really, is the problem.

**Expected behaviour**

It would be good were the applet to give one a clue about the blocking program.

**Other information**

**Comment by clefebvre** [comment]
I suppose we could show the appId and reason in the tooltip. I think we've access to these via dbus.

**Comment by troxor** [comment]
This has been biting me lately, so here's _one_ way to determine this:
```
% dbus-send --session --print-reply --dest=org.gnome.SessionManager \
     /org/gnome/SessionManager \
     org.gnome.SessionManager.GetInhibitors
method return time=1520625795.516989 sender=:1.3 -> destination=:1.250 serial=2360 reply_serial=2
   array [
      object path "/org/gnome/SessionManager/Inhibitor79"
   ]
```
```
% dbus-send --session --print-reply --dest=org.gnome.SessionManager \
    /org/gnome/SessionManager/Inhibitor79 \
    org.gnome.SessionManager.Inhibitor.GetAppId
method return time=1520625994.042388 sender=:1.3 -> destination=:1.253 serial=2365 reply_serial=2
   string "/usr/bin/vivaldi-stable"
```
```
% dbus-send --session --print-reply --dest=org.gnome.SessionManager \
    /org/gnome/SessionManager/Inhibitor79 \
    org.gnome.SessionManager.Inhibitor.GetReason
method return time=1520625799.584925 sender=:1.3 -> destination=:1.251 serial=2361 reply_serial=2
   string "WebRTC has active PeerConnections"
```

It turned out to be a pinned tab in the Vivaldi/Chrome browser inhibiting power management.

PR #7748, opened 2018-07-19
inhibit applet: Display list of inhibiting programs

kwest0x2a:
I needed to track down what was inhibiting my computer from suspending, came across #7175 in my searching, and followed @troxor's helpful instructions in one of the comments to do so. Figured I'd take a stab at implementing the feature so the next guy/gal doesn't have to retrace my steps. :)

Screenshot in St.Side.BOTTOM orientation. The list appears below the switches in other orientations.
inhibitorlistscreenshot


clefebvre:
Thanks @kwest0x2a.

PR #7841, opened 2018-09-10
Spices.py: Don't try to load metadata for themes

germanfr:
Themes don't have metadata.json and it was failing anyways, causing a lot of warnings in cs_themes.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
@@ -402,6 +402,9 @@ class Spice_Harvester(GObject.Object):
     def _load_metadata(self):
         self.meta_map = {}
 
+        if self.themes: # Themes don't have metadata.json
+            return
+
         for directory in self.spices_directories:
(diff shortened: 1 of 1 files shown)


collinss:
Hi @germanfr. Theme's actually do have metadata unless they're system folders or they're installed manually (i.e. not through cinnamon-settings or cinnamon-install-spice). When the install function in Spices.py is called, it creates a timestamp in metadata.json so that later it can tell if there's an update. In the case where a theme doesn't include metadata.json (they can if they want) the file is created. See https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py#L652 for reference.

I haven't tested this, but not loading metadata for themes will most likely break the updating mechanism. I'm thinking it would be better to hide the logging for themes rather than not loading the metadata at all.

germanfr:
Oh you are certainly right. But it doesn't create them for manually installed themes so it always prints the warnings. I didn't notice it wasn't just some of them because I have a lot (for testing purposes). Sorry.

Issue #7179, opened 2018-01-05
Notifications don't respect user specified timeouts


tanderson-google:
```
Cinnamon 3.4.6
Debian testing
 ```

**Issue**

Cinnamon should respect user-defined timeouts on notifications. https://developer.gnome.org/notification-spec/

A timeout of 0 should make a notification stay-on-screen, and a positive value specifies the screen-time in ms.

It looks like Cinnamon does try to respect the user timeout:
https://github.com/linuxmint/Cinnamon/blob/b1e0bc3921e9392518e92ec27b70362fcbb3549d/js/ui/notificationDaemon.js#L322
But the implementation has a bug because it doesn't seem to work.

**Steps to reproduce**

1. Launch Google Chrome
2. Enable native notifications by navigating to chrome://flags and setting #enable-native-notifications to "Enabled". Restart Chrome
3. Visit https://tests.peter.sh/notification-generator/#requireInteraction=true and click on "Display the notification"
Edit: These steps won't work any more because Chrome now sets the urgency to 2 (critical) to workaround the issue.

**Expected behaviour**

The notification should stay on the screen, but times out after ~5s.

clefebvre:
This is by design.

Issue #3896, opened 2015-02-10
Notifications do not respect notify-send -t settings


broomdodger:
I tried using the "-t" option to change how long notifications appear but nothing changes.

-t, --expire-time=TIME
Specifies the timeout in milliseconds at which to expire the notification.

broomdodger:
The original post describing `notify-send -t` failing is using Mint 17.1 Cinnamon.

Yesterday I tried the same using Mint 17.1 xfce... and there it works.

mtwebster:
in cinnamon settings, go to Notifications, check "Remove notifications after their timeout is reached"

broomdodger:
Checked or un-checked... has **NO** effect on `notify-send -t`

That checkbox **ONLY** changes how **many** items are displayed in the **tray** notification area. It has no effect on the **duration** of notification ***bubble***.

Note: both Mint 17.1
The following **displays** the notification ***bubble*** correctly for both Cinnamon and xfce.
It fails to keep the notification ***bubble*** displayed for 10 seconds in Cinnamon but works in xfce.

The effect of the `-t 10000` is to keep the notification ***bubble*** displayed for 10 seconds, rather than the default... about 5 seconds.
```
n="/usr/bin/notify-send"
i="-i /home/tech/Pictures/icons/preferences-system-time-symbolic.svg"

test="Bzz Bzz Bzz"
 * * * * * DISPLAY=:0 ${n} ${i} -t 10000 "${test}"
```

broomdodger:
More info about `notify-send`

notification-spec-0.9.txt

**Expiration Timeout**

The timeout time in milliseconds since the display of the notification at which the notification should automatically close.

If -1, the notification's expiration time is dependent on the notification server's settings, and may vary for the type of notification.

If 0, the notification never expires.

mtwebster:
Do these other DE's have a tray-type applet/panel container to catch missed notifications?  This isn't really cut and dry for us because of the tray, regardless of what the spec says.  I'm curious about their behavior in these cases.

broomdodger:
@mtwebster
> Do these other DE's have a tray-type applet/panel container to catch missed notifications?

Neither XFCE nor MATE have a tray-type applet/panel container to catch missed notifications.

Both XFCE and MATE **display** a notification **bubble** for the time specified by the `-t` option.

I now understand how it works in cinnamon:
The `-t` option specifies how long a notification will remain in the notification **tray**.

[...]

PR #7853, opened 2018-09-15
Make critical notification respect timeouts

shaicoleman:
When sending a critical notification, the timeout is currently ignored.
e.g.: `notify-send -t 1000 -u critical 'test'`
This will fix it so timeouts are respected

For critical notification without an explicit timeout, the current behavior is kept, and they will not dismiss automatically
e.g.: `notify-send -u critical 'test'`

Currently there's no way to have notifications that both work in full screen apps (e.g. calibre viewer in full screen mode) and respect the timeout.
This makes it possible.

Related issues: #3896 and #7179

### DIFF
--- a/js/ui/notificationDaemon.js
+++ b/js/ui/notificationDaemon.js
@@ -323,7 +323,7 @@ NotificationDaemon.prototype = {
 
         // Find expiration timestamp.
         let expires;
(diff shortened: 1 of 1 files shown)


shaicoleman:
Closing it for now, as it stopped working for me

PR #7866, opened 2018-09-19
dnd: Fix invalid object access errors, migrate to ES2015 classes

jaszhix:
Partially based on an [upstream patch](https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/209).


jaszhix:
https://github.com/linuxmint/Cinnamon/pull/7873

PR #7879, opened 2018-09-21
Update calculator search provider example

germanfr:
Remove unused imports and avoid creating unnecessary objects on each key press.


germanfr:
Although it is just an example, I use it day to day, and it would be great if it was merged because it makes it more efficient.

PR #7541, opened 2018-05-03
Windows overview improvements

germanfr:
* Fix placement of the close button. For some reason, the parent of the window textures can't be a `StWidget` because it doesn't adjust properly to the window bounds.
 * Add hover indicator and fix overlapping windows.
 * Replace Lang.bind with native bind and arrow functions.
 * Use `MetaButtonLayout` to get the side of the close button.
 * Optimize usage of the available space to have larger window previews.
 * Close overview when the last window is closed and show a placeholder if there are no open windows.
 * Go back to desktop on click if there are no open windows.
 * Respect animation effects settings.
 * Remove delay after closing a window.

Depends on linuxmint/muffin#318
Fixes #7876, fixes #7638, fixes #7656, fixes #446, fixes #7572.

Issue #6376, opened 2017-03-24
Workspace switcher applet too slow when using mouse wheel


agoode:
Pull #6021 changes the workspace switcher applet to make it more usable with touchpads. Unfortunately, it now makes it much less usable with discrete mouse wheels.

The delay should not apply if the event scroll source is `CLUTTER_SCROLL_SOURCE_WHEEL`.

See https://developer.gnome.org/clutter/stable/clutter-Events.html#ClutterScrollSource

germanfr:
Hi, that's a great idea. Unfortunately for some reason the `ClutterEvent` object received has no such function and it throws this error.
```
(cinnamon:2935): Cjs-WARNING **: JS ERROR: GLib.Error g-invoke-error-quark: Could not locate clutter_event_get_scroll_source: 'clutter_event_get_scroll_source': /usr/lib/libclutter-1.0.so.0: undefined symbol: clutter_event_get_scroll_source
```

It breaks even when I try to test if it exist like this.
```js
if(event.get_scroll_source) { // <-- Breaks here
    is_mouse = event.get_scroll_source() === Clutter.ScrollSource.WHEEL;
}
```

I have the latest version of Clutter installed on my system (1.26, which is where this function was added).

Also, I'd like to mention that this feature was added in order to make it **usable** for touchpads, because when a mouse wheel throws one scroll event a touchpad throws 6, and it was impossible to switch one workspace at a time. Unless you are jumping 4 workspaces I don't notice any difference with my mouse. It would be great if this could be solved with that, but unfortunately I'm not able to solve the issue I explained above.

Something as simple as this would do the trick if it worked.

```patch
+        let is_mouse = event.get_scroll_source() === Clutter.ScrollSource.WHEEL;
+       
         // Do the switch only after a ellapsed time to avoid fast
         // consecutive switches on sensible hardware, like touchpads
-        if ((now - this._last_switch) > MIN_SWITCH_INTERVAL_MS ||
+        if (is_mouse || (now - this._last_switch) > MIN_SWITCH_INTERVAL_MS ||
```

JosephMcc:
Even if you got this to work, it wouldn't work in Ubuntu Xenial or Debian Jessie based distros which don't support clutter 1.26 (aka Mint). Beside that I don't see how this is "unusable". I tested the original work when it was done and use it regularly now and it works fine with a scroll wheel.

olejorgenb:
[...]

Issue #5140, opened 2016-03-21
Workspace switchnig very slow


marben:
Just upgraded from 2.6.12 to 2.8.6 and noticed that workspace switching got much slower.
From almost instant to about 0.5s.
This is true even if i remove the workspace switcher applet from tray.

It's quite essential for me. Is there a way i could profile the switching process / try other libraries / change some default options or something like that?

Which library is responsible for workspace management?

dajusc:
This bugs me too. :-(

I noticed that workspace switching is significantly faster after a reboot, and becomes slower and slower after time...

Vahan86:
@marben, is this still an issue for you with Cinnamon 3.2?

marben:
@Vahan86 Unfortunately I don't use that computer any more. It was an old Atom processor. I don't have this issue with my current machine running 3.2, but it could well be just because of it's much superior speed.

My assumption is, that not many people use such slow processor, therefore nobody noticed.

Is there a reason, why cinnamon 3.2 should be better than 2.8? Perhaps i could dig my old machine from a drawer and boot a mint install image to check.

jonorthwash:
This is an issue for me with 3.2.  It gets slower until `cinnamon --replace` and then it's normal again—for a while.

mondalaci:
Sadly, this issue still persists after all these years, and affects Cinnamon 4.0.10. Window switching starts out very snappy for me, then it increasingly gets slower to the point that after about a week of heavy desktop use, it takes 1s for every switch which deeply affects my productivity.

bowars:
I recently switched to Cinnamon and ran into the same problem. Thanks to the comments on issue #1037 by @mondalaci, @mhagger and @oxysoft, I believe I have narrowed down the issue and found a potential fix by patching `/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js`:

```
@@ -886,13 +886,15 @@ class AppGroup {
         const {metaWindows, trigger, set} = this.groupState;
         const refWindow = metaWindows.indexOf(metaWindow);
         if (metaWindow) {
-            this.signals.connect(metaWindow, 'notify::title', (...args) => this.onWindowTitleChanged(...args));
-            this.signals.connect(metaWindow, 'notify::appears-focused', (...args) => this.onFocusWindowChange(...args));
-            this.signals.connect(metaWindow, 'notify::gtk-application-id', (w) => this.onAppChange(w));
[...]

Issue #2081, opened 2013-05-24
[Regression] Performance


ManIVIctorious:
Cinnamon 1.6 was much faster than 1.8

PR #7916, opened 2018-10-06
Improve workspace switching performance

jaszhix:
This moves all callbacks that are called on the `switch-workspace` signal to a new signal emitted from `Main.wm` (`WindowManager`) after it is done switching workspaces. Before, everything connected to `global.window_manager`'s switch-workspace signal which is emitted in C. What this PR does is add signal methods to `WindowManager.prototype`, and makes it so only it receives the C signal. It then emits its own switch-workspace signal that everything else now connects to. This allows the workspace switching to happen first, then everything else can react after Clutter paints the workspace.

- This adds a new API, `afterStageRelayout` to `Main.wm`. It is a self-contained signal handler that connects to the stage's `queue-relayout` signal with connect_after. This is used to emit `Main.wm`'s `switch-workspace` signal that other parts of Cinnamon will receive, but does so after the next stage relayout so its not blocking workspace switching.
- Removes invocation of `_queueUpdateRegions` on `switch-workspace` in layout.js/Chrome. This function is heavy and seems to be getting called very frequently via `Chrome.prototype._windowsRestacked` emission. After looking back at why it was added, and testing panel auto hide, hot corners, window, and workspace overviews, I'm not seeing any regressions from its removal.
- This is backwards compatible, as it is not changing the signal on `global.window_manager`, which is emitted from C code. Maintainers of multi-version xlets might want to consider connecting their `switch-workspace` signal to `Main.wm` instead though.

Tested with and without #7541.

Refs #5140 #6376 #2081 #1413 #1037

### DIFF
--- a/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
@@ -1005,7 +1005,7 @@ class CinnamonWindowListApplet extends Applet.Applet {
         this.signals.connect(global.screen, 'window-workspace-changed', this._onWindowWorkspaceChanged, this);
         this.signals.connect(global.screen, 'window-skip-taskbar-changed', this._onWindowSkipTaskbarChanged, this);
         this.signals.connect(global.screen, 'monitors-changed', this._updateWatchedMonitors, this);
-        this.signals.connect(global.window_manager, 'switch-workspace', this._refreshAllItems, this);
+        this.signals.connect(Main.wm, 'switch-workspace', this._refreshAllItems, this);
 
         this.actor.connect('style-changed', Lang.bind(this, this._updateSpacing));
 
--- a/files/usr/share/cinnamon/applets/workspace-switcher@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/workspace-switcher@cinnamon.org/applet.js
@@ -280,6 +280,7 @@ class CinnamonWorkspaceSwitcher extends Applet.Applet {
     }
 
     _onWorkspaceChanged(windowManager, wm, from, to) {
+        if (!this.buttons[from] || !this.buttons[to]) return;
         this.buttons[from].activate(false);
         this.buttons[to].activate(true);
     }
--- a/files/usr/share/cinnamon/applets/workspace-switcher@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/workspace-switcher@cinnamon.org/applet.js
@@ -280,8 +280,12 @@ class CinnamonWorkspaceSwitcher extends Applet.Applet {
     }
 
     _onWorkspaceChanged(windowManager, wm, from, to) {
-        if (!this.buttons[from] || !this.buttons[to]) return;
-        this.buttons[from].activate(false);
+        // Make sure we always reset everything to inactive before setting
+        // a new active item. This prevents multiple buttons from being highlighted.
+        for (let i = 0; i < this.buttons.length; i++) {
+            this.buttons[i].activate(false);
+        }
+
         this.buttons[to].activate(true);
     }
 
--- a/files/usr/share/cinnamon/applets/workspace-switcher@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/workspace-switcher@cinnamon.org/applet.js
@@ -231,7 +231,7 @@ class CinnamonWorkspaceSwitcher extends Applet.Applet {
 
         this._createButtons();
         global.screen.connect('notify::n-workspaces', Lang.bind(this, this.onNumberOfWorkspacesChanged));
-        global.window_manager.connect('switch-workspace', this._onWorkspaceChanged.bind(this));
+        Main.wm.connect('switch-workspace', this._onWorkspaceChanged.bind(this));
         global.settings.connect('changed::panel-edit-mode', Lang.bind(this, this.on_panel_edit_mode_changed));
 
         let expoMenuItem = new PopupMenu.PopupIconMenuItem(_("Manage workspaces (Expo)"), "view-grid-symbolic", St.IconType.SYMBOLIC);
@@ -279,7 +279,7 @@ class CinnamonWorkspaceSwitcher extends Applet.Applet {
         }
     }
 
-    _onWorkspaceChanged(wm, from, to) {
+    _onWorkspaceChanged(windowManager, wm, from, to) {
         this.buttons[from].activate(false);
         this.buttons[to].activate(true);
     }
--- a/js/ui/expoThumbnail.js
+++ b/js/ui/expoThumbnail.js
@@ -1098,7 +1098,7 @@ ExpoThumbnailsBox.prototype = {
 
     show: function() {
         this.switchWorkspaceNotifyId =
-            global.window_manager.connect('switch-workspace',
+            Main.wm.connect('switch-workspace',
                                           Lang.bind(this, this.activeWorkspaceChanged));
 
         this.workspaceAddedId = global.screen.connect('workspace-added', Lang.bind(this, function(screen, index) {
@@ -1206,7 +1206,7 @@ ExpoThumbnailsBox.prototype = {
     },
 
     hide: function() {
-        global.window_manager.disconnect(this.switchWorkspaceNotifyId);
+        Main.wm.disconnect(this.switchWorkspaceNotifyId);
         global.screen.disconnect(this.workspaceAddedId);
         global.screen.disconnect(this.workspaceRemovedId);
 
@@ -1683,7 +1683,8 @@ ExpoThumbnailsBox.prototype = {
         this.emit('allocated');
     },
 
-    activeWorkspaceChanged: function(wm, from, to, direction) {
+    activeWorkspaceChanged: function(windowManager, wm, from, to, direction) {
+        if (!this.thumbnails[this.kbThumbnailIndex]) return;
         this.thumbnails[this.kbThumbnailIndex].showKeyboardSelectedState(false);
         this.kbThumbnailIndex = global.screen.get_active_workspace_index();
         this.thumbnails[this.kbThumbnailIndex].showKeyboardSelectedState(true);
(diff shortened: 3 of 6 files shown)


brownsr:
This is causing a display bug for me.  Specifically the workspace switcher applet, in the visual representation mode.  Clicking on the different workspaces changes the workspaces  and windows seem to end up in the right place, and the OSD shows fine, however the highlighting does not work - the highlighting stays on the first workspace representation and does not move.

jaszhix:
@brownsr I think I see what you mean - are you moving a window to another workspace? Only reproducer for me.

brownsr:
Cjs-Message: 19:45:37.600: JS LOG: [LookingGlass/info] Resized nvidia-prime with normalized size (64x64px)

(nvidia-prime:1712): Gdk-CRITICAL **: 19:45:37.606: gdk_window_thaw_toplevel_updates: assertion 'window->update_and_descendants_freeze_count > 0' failed

(blueberry-tray.py:1813): Bluetooth-WARNING **: 19:45:39.182: Error setting property 'Powered' on interface org.bluez.Adapter1: GDBus.Error:org.bluez.Error.Blocked: Blocked through rfkill (g-io-error-quark, 36)
Cjs-Message: 19:45:39.361: JS LOG: [LookingGlass/info] Adding systray: blueberry-tray.py (48x48px)
Cjs-Message: 19:45:40.362: JS LOG: [LookingGlass/info] Resized blueberry-tray.py with normalized size (64x64px)
Cjs-Message: 19:45:40.856: JS WARNING: [/usr/share/cinnamon/js/misc/fileUtils.js line 210 > Function 283]: reference to undefined property "undefined"

(cinnamon:1689): Cjs-WARNING **: 19:45:40.856: JS ERROR: Exception in callback for signal: switch-workspace: TypeError: this.buttons[from] is undefined
_onWorkspaceChanged@/usr/share/cinnamon/applets/workspace-switcher@cinnamon.org/applet.js:283:9
_emit@resource:///org/cinnamon/cjs/modules/signals.js:142:27
WindowManager.prototype.emitSwitchWorkspace/<@/usr/share/cinnamon/js/ui/windowManager.js:736:52


(cinnamon:1689): Cjs-WARNING **: 19:45:42.079: JS ERROR: Exception in callback for signal: switch-workspace: TypeError: this.buttons[from] is undefined
_onWorkspaceChanged@/usr/share/cinnamon/applets/workspace-switcher@cinnamon.org/applet.js:283:9
_emit@resource:///org/cinnamon/cjs/modules/signals.js:142:27
WindowManager.prototype.emitSwitchWorkspace/<@/usr/share/cinnamon/js/ui/windowManager.js:736:52


(cinnamon:1689): Cjs-WARNING **: 19:45:42.607: JS ERROR: Exception in callback for signal: switch-workspace: TypeError: this.buttons[from] is undefined
_onWorkspaceChanged@/usr/share/cinnamon/applets/workspace-switcher@cinnamon.org/applet.js:283:9
_emit@resource:///org/cinnamon/cjs/modules/signals.js:142:27
WindowManager.prototype.emitSwitchWorkspace/<@/usr/share/cinnamon/js/ui/windowManager.js:736:52

Cjs-Message: 19:45:43.516: JS LOG: [LookingGlass/info] Adding systray: mintupdate.py (48x48px)
Cjs-Message: 19:45:44.517: JS LOG: [LookingGlass/info] Resized mintupdate.py with normalized size (64x64px)

(cinnamon:1689): Cjs-WARNING **: 19:45:46.127: JS ERROR: Exception in callback for signal: switch-workspace: TypeError: this.buttons[from] is undefined
_onWorkspaceChanged@/usr/share/cinnamon/applets/workspace-switcher@cinnamon.org/applet.js:283:9
_emit@resource:///org/cinnamon/cjs/modules/signals.js:142:27
WindowManager.prototype.emitSwitchWorkspace/<@/usr/share/cinnamon/js/ui/windowManager.js:736:52

jaszhix:
Thanks @brownsr. I can't reproduce this one, but took a guess at a fix. If this works for you and still highlights the buttons correctly, then the signal might be firing multiple times.

brownsr:
will try when I can.  I am having troubling building (master as well as this) at the moment.

brownsr:
The good news ... all the warning messages went away :-)
Unfortunately the display issue is still there :-(

jaszhix:
@brownsr I tried to reproduce this on my optimus laptop with no success. I see a delay for the button to be highlighted, but it does highlight correctly through expo, hot corners, clicking the switcher applet, and using keyboard shortcuts while the expo/window overviews are open. I tested this on Nvidia and Intel graphics. Does this happen on a fresh session with no windows open? Also, do you have a copy of the workspace switcher applet in your user directory? The arguments to `_onWorkspaceChanged` changed, and I could see this occurring with an old version of the applet, but not sure what else it could be.

brownsr:
Same result on a fresh session with no windows.  Nividia and intel.  vertical or horizontal panels.  No local copy. Tried different themes. Generally no error messages in log. Code looks the same as above - I just tried the change because your test guarantees both activations get skipped if one button is not defined.  Is there something subtle in the build ?  I have mint 19 base + cjs, muffin, screensaver, xapp, csd all from master, Cinnamon built from the PR.  

Just to make sure we are looking at the same thing, it's the active workspace border highlighting that I am looking at.

jaszhix:
The highlighting of the button? I've also tested both workspace switcher view modes.

image

Yeah, I don't doubt its setup correctly, just thought of a time when I had a stock applet in my user directory and it confused me for a while. It could be since its deferring, it may be having unpredictable results if the main thread is busy - or is that maybe a behavior of GLib to ignore low priority function calls in that case? I'll come back to this and see if I can find another way to do this without relying on mainloop.

jaszhix:
In windowManager.js, if you replace `emitSwitchWorkspace` with

```js
emitSwitchWorkspace: function() {
        this.emit('switch-workspace', ...arguments);
    },
```
Does it still emit junk data for you? It might also be useful to add `global.log(...arguments)` in _onWorkspaceChanged in the applet. Since I can't reproduce this, any debug help would be appreciated.

brownsr:
Sorry @jaszhix I can't spend more time looking at this.  I've no idea why you can't reproduce this when it fails every time for me.  I suggest you get someone else to try it as well.

Looking at the code I did have a couple of queries.  

The first was why you are making the change

from        this.signals.connect(global.window_manager, 'switch-workspace', this._refreshAllItems, this);
to        this.signals.connect(Main.wm, 'switch-workspace', this._refreshAllItems, this);
etc. 

Is that fundamental to the change ?  The reason for the query is that there seem to be quite a few other places using global.window_manager, so if you are changing them out for a different call then surely it should be done everywhere for consistency.  I must admit I don't really know what the use of those globals was originally intended to do.

The second was related to 
            cinnamonwm.completed_switch_workspace();
            this.emitSwitchWorkspace(...arguments);

and relates to the sequencing.  The workspace switching is marked as completed, and then all the signals are emitted ?  Not that I could find any difference reversing the order.

brownsr:
Sorry @jaszhix I can't spend more time looking at this.  I've no idea why you can't reproduce the errors when it fails every time for me.  I suggest you get someone else to try it as well.

Looking at the code I did have a couple of queries.  

The first was why you are making the change

from        this.signals.connect(global.window_manager, 'switch-workspace', this._refreshAllItems, this);
to        this.signals.connect(Main.wm, 'switch-workspace', this._refreshAllItems, this);
etc. 

Is that fundamental to the change ?  The reason for the query is that there seem to be quite a few other places using global.window_manager, so if you are changing them out for a different call then surely it should be done everywhere for consistency.  I must admit I don't really know what the use of those globals was originally intended to do.

The second was related to 
            cinnamonwm.completed_switch_workspace();
            this.emitSwitchWorkspace(...arguments);

and relates to the sequencing.  The workspace switching is marked as completed, and then all the signals are emitted ?  Not that I could find any difference reversing the order.

jaszhix:
Before, everything connected to `global.window_manager`'s switch-workspace signal which is emitted in C. What this PR does is add signal methods to `WindowManager.prototype`, and makes it so only it receives the C signal. It then emits its own switch-workspace signal that everything else now connects to. This allows the workspace switching to happen first, then everything else can react after Clutter paints the workspace. As far as I can tell, I already switched out all the signal connections on `global.window_manager` switch-workspace, and replaced them with `Main.wm`, which is what `WindowManager.prototype` is assigned to in main.js.

This changes the signal callback arguments because the signal emission makes the first argument the JS object (Main.wm), and we can't control that. This is backwards compatible, we're not changing the original `global.window_manager` signal, its still there for third party xlets.

We could do this on the aptly named switch-workspace-complete signal, but since it doesn't pass `from` and `to` arguments to the signal callback, the actors need to connect to the same signal that actually performs the switch. This causes too many things to queue repaints at once in no particular order I think, and why there is a 1-2 second delay when switching workspaces.

brownsr:
Now I understand what the PR is about !

jaszhix:
On master it looks like cinnamon is frozen for a moment when switching workspaces with animations on. If I disable animations, then its harder to compare but still seems faster on this branch. I think this issue used to be worse with animations off before we updated Clutter in Muffin. 

The workspace animation is broken for me on master, and has been for a while. I first realized it was there on my laptop some time ago, so it may be a multi-monitor issue. Multiple monitors gives Clutter a lot more things to render, and that can affect animations. This is fixed on this branch by using `Meta.later_add` with the BEFORE_REDRAW LaterType enum, which might makes things appear slower, but I think its just letting the animation fully render where as before it would get cut off prematurely or entirely. 

Since we default to animations being enabled, they should either work correctly, or not cause as much lag when they don't. The animation could also be made to be shorter now that its given a chance to fully render. As far as the user knows (with hardware similar to what I use), there's no workspace animation, just a delay between switching workspaces.

clefebvre:
I can't really tell if the animation is better than before... it's going too fast and affecting too large an area for me to really be able to tell.

However, overall this feels like a regression to me. I understand that everything happened at the same time on master and that now the switch is actually done before the effects.. but that introduces visual lag in my opinion.

On master, as I click a workspace, it gets highlighted in the switcher, it's all happening at once and without delay. On this PR, it's a two-time process and it feels laggy.

jaszhix:
OK, I'll scrap this for now. After using new-panel-layout for a while, I'm not sure there is a problem to fix still. I don't want to add more complexity to windowManager.js if its not needed.

Issue #7895, opened 2018-09-28
Ambiguous "Fonts" items in 


jeffcann:
```
 * Cinnamon version - 3.8.9
 * Distribution - Mint 19
 * Nvidia with native driver
 * 64 bit
 ```

**Issue**
I often have to change font scaling, and to do that I search for "fonts" in the Cinnamon menu. Searching for "fonts" bring up 2 identical (well, very similar) entries in the menu. Took me a while to realise that they are actually different things. One opens up a dialogue with all the installed fonts, the other opens up the control panel with font selection and settings.

**Steps to reproduce**
1. Open the Cinnamon menu (I use Win-Key)
1. Type 'fonts'
1. Notice there are 2 entries, same icons with different colours. 

**Expected behaviour**
I'd like to see some disambiguation.
Perhaps one could be called "Installed Fonts", and the other "Font Settings"

**Other information**

icarter09:
@jeffcann I put up a PR for this issue. The other font selection belongs to Gnome. I'm not for sure if that is something that can be changed through the Cinnamon repos.

PR #7920, opened 2018-10-09
Issue 7895: Changed label from "Fonts" to "Font Settings".

icarter09:
This is a PR for issue #7895. Changed the font label for the font that belongs to Cinnamon, the other belongs to Gnome.


collinss:
@icarter09 Just fyi, those .desktop files are generated automatically with https://github.com/linuxmint/Cinnamon/blob/master/generate_cs_module_desktop_files.py. The name is pulled from the cs_fonts.py module in cinnamon-settings. Any changes you make to the desktop files will be overwritten the next time the script is run. I'm not sure what the best solution is though.

icarter09:
@collinss Thanks for the heads up on that. I had a feeling that the .desktop files were being generated, but I wasn't sure how. Would changing the value of the name attribute in the Module class get picked up by the auto generator?

collinss:
> Would changing the value of the name attribute in the Module class get picked up by the auto generator?

Yeah, it should. Adding 'settings' to the entry in cinnamon-settings might be a little bit redundant and inconsistent though.

collinss:
Honestly I'm not sure what would be better. I'd almost be tempted to override the .desktop file for the other one instead. You could call it "Font Browser" or something. It would remove the ambiguity in the menu and would still work in Cinnamon Settings. Of course that would probably be a distribution-level change, but that might be better anyway. Idk though. This is probably one you want to defer to @clefebvre on.

icarter09:
@clefebvre let me know what you think would be the best course of action for this ticket.

clefebvre:
This should be requested upstream indeed, to GNOME.

MATE's font viewer describes itself as "MATE Font Viewer", ideally GNOME's should say "Font Viewer".

PR #7938, opened 2018-10-17
menu: Fix application items not rendering

jaszhix:
`_listApplications` was changed to return an array in #7395, but this was only handled in `_doSearch`.


brownsr:
Works well :-)

PR #7943, opened 2018-10-18
[WIP] Add a window list with app grouping

jaszhix:
This PR includes a modified version of Icing Task Manager that adheres to the Cinnamon style guide. There are other additions as well.

- Changed the name to "Grouped window list".
- Adds most of the utility functions found in ITM's utils.js file to misc/util.
- Adds wrappers for `Mainloop.timeout_add` to util.js with `setTimeout` and `setInterval` that the applet uses. I also decided to make them available globally. My reasoning for this is because they are ubiquitous in all browsers and NodeJS. I have been cleaning up some of the older unmaintained third party xlets recently, and have been coming across a surprising amount of xlets that do not stop looping after they are removed from Cinnamon. These abstractions make using Mainloop safer and easier.
- Adds a utility for live-reloading Cinnamon xlets on code change with Gulp. I've been developing my applets this way for a while, and figured it should be in Cinnamon. I modified the configuration to support Gulp 4, and to be able to watch any xlet through command arguments.

To use this, you will need NodeJS. I recommend [v10](https://github.com/nodesource/distributions).
- Run `npm install -g gulp@^4.0.0`
- In the Cinnamon directory run `npm install`.
- To start the task that will watch and reload on code changes, run `gulp spawn-watch --uuid="<xlet uuid>" --type="<xlet type>"`. For example, to develop the applet this PR is adding you would run `gulp spawn-watch --uuid="grouped-window-list@cinnamon.org" --type="applet"`

Currently, nothing has been removed from the code, but there are a few niche options that I am considering for removal, and don't think are appropriate with good defaults for most people. Its good to have a balance between flexibility and ease of use, and too many options can make using software tedious sometimes - hurting the user experience. Here's a few things I'm thinking about:

- [ ] Removal of Firefox bookmarks, history, and most visited. This code is heavy and I'm not confident we need to shove potentially thousands of items in a context sub-menu.
- [ ] Removal of menu item icon type toggling option.
- [ ] Simplifying the options so they are less overwhelming.
- [ ] Patch cinnamon.css with the theme support ITM needs for all of its app states, which should make most of the theme options unnecessary.
- [ ] Add better icon scaling that doesn't need manual configuration.

I am inviting anyone who is interested to suggest items to add to the to-do list, submit patches to this branch, and looking forward to seeing what ideas you guys have for this code base - whether they are technical or design related.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -31,7 +31,7 @@
     "rules": {
         "comma-dangle": 0,
         "no-cond-assign": 0,
-        "no-console": 2,
+        "no-console": 0,
         "no-constant-condition": 0,
         "no-control-regex": 0,
         "no-debugger": 0,
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -1,0 +1,1121 @@
+const Cinnamon = imports.gi.Cinnamon;
+const Meta = imports.gi.Meta;
+const Clutter = imports.gi.Clutter;
+const St = imports.gi.St;
+const Main = imports.ui.main;
+const Tweener = imports.ui.tweener;
+const DND = imports.ui.dnd;
+const Tooltips = imports.ui.tooltips;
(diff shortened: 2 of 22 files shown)


jaszhix:
Moved to an upstream branch #7944

PR #7944, opened 2018-10-18
New panel layout

clefebvre:
The idea is to change the default panel layout to make Cinnamon feel more modern out of the box.

Here's what needs to be done in terms of applets:

- [x] Add a forked version of ITM named grouped-window-list
- [x] Switch the default panel settings to replace panel-launchers/window-list with grouped-window-list
- [x] Bring back sound applet to the left of the clock

In grouped-window-list:

- [x] Make it pin Firefox, Terminal and Nemo
- [x] icon spacing 6px
- [x] close windows on middle-click
- [x] new window option in context menu
- [x] Hide instance numbers
- [ ] Render multi-instance groups differently (Windows does that well, looks like two stacked cards)
- [x] When left-click is used to cycle through windows, left-click should launch the app when there are no instances present

Panels:

- [x] ignore theme panel size
- [x] Make it possible for each panel zone (left/center/right) to have its own icon size
- [x] Increase the default panel size to 40px
- [x] Set icon size defaults to 0/0/24px
- [x] Fix text size

Features to be removed:

- [x] thumb icon
- [x] icon padding
- [x] thumb close icon size
- [x] thumb padding -> hardcode to 3*ui_scale
- [x] thumb size -> keep this feature but change to a combo to select "Small, Medium, Large".
- [x] thumb timeout -> hardcode delay to hide thumb to 50, hardcode delay between thumbs to 50, make delay on hover (to show first thumb) configurable with a combo: 50, 250, 500.
- [x] peek fade time --> keep this feature, add a fade-out as well
- [x] number display --> turn into an ON(smart)/OFF switch
- [x] app button transition duration -> let the theme handle that

Testing: 

- [ ] Review performance compared to window-list (run window mapping tests)

In Mint (outside of this PR):

- [x] Add a section in the welcome screen to easily switch back and forth between this new look and the 18.3 - Mint-X look.
- [x] Switch the default Cinnamon theme to Mint-Y-Dark
- [x] In Mint-Y-Dark, make panel applets/icons (or fg color in general) whiter/lighter.

-------------------------------------------

The second part of the PR is about the menu:

- [x] Allow distributions to set a default icon name and menu label in gsettings
- [x] Expand favorite section so it uses the available space
- [ ] Remove user applet from enabled-applets
- [ ] Add switch-user to menu, consider changing the layout a bit to feature session buttons via symbolics either at the bottom or the top of the menu.
- [ ] Expand search field so it uses the available space
- [ ] Add session buttons


jaszhix:
This PR includes a modified version of Icing Task Manager that adheres to the Cinnamon style guide. There are other additions as well.

- Changed the name to "Grouped window list".
- Adds most of the utility functions found in ITM's utils.js file to misc/util.
- Adds wrappers for `Mainloop.timeout_add` to util.js with `setTimeout` and `setInterval` that the applet uses. I also decided to make them available globally. My reasoning for this is because they are ubiquitous in all browsers and NodeJS. I have been cleaning up some of the older unmaintained third party xlets recently, and have been coming across a surprising amount of xlets that do not stop looping after they are removed from Cinnamon. These abstractions make using Mainloop safer and easier.
- Adds a utility for live-reloading Cinnamon xlets on code change with Gulp. I've been developing my applets this way for a while, and figured it should be in Cinnamon. I modified the configuration to support Gulp 4, and to be able to watch any xlet through command arguments.

To use this, you will need NodeJS. I recommend [v10](https://github.com/nodesource/distributions).
- Run `npm install -g gulp@^4.0.0`
- In the Cinnamon directory run `npm install`.
- To start the task that will watch and reload on code changes, run `gulp spawn-watch --uuid="<xlet uuid>" --type="<xlet type>"`. For example, to develop the applet this PR is adding you would run `gulp spawn-watch --uuid="grouped-window-list@cinnamon.org" --type="applet"`

Currently, nothing has been removed from the code, but there are a few niche options that I am considering for removal, and don't think are appropriate with good defaults for most people. Its good to have a balance between flexibility and ease of use, and too many options can make using software tedious sometimes - hurting the user experience. Here's a few things I'm thinking about:

- [ ] Removal of Firefox bookmarks, history, and most visited. This code is heavy and I'm not confident we need to shove potentially thousands of items in a context sub-menu.
- [ ] Removal of menu item icon type toggling option.
- [ ] Simplifying the options so they are less overwhelming.
- [ ] Patch cinnamon.css with the theme support ITM needs for all of its app states, which should make most of the theme options unnecessary.
- [ ] Add better icon scaling that doesn't need manual configuration.

I am inviting anyone who is interested to suggest items to add to the to-do list, submit patches to this branch, and looking forward to seeing what ideas you guys have for this code base - whether they are technical or design related.

js/misc/util.js
```diff
@@ -403,6 +404,240 @@ function findIndex(array, callback) {
...
+};
+
+/**
+ * filter:
+ * @array (array): Array to be iterated.
+ * @callback (function): The function to call on every iteration.
+ *
+ * Returns (array): Returns the new filtered array.
+ */
+function filter(arr, callback) {
```
germanfr: Aren't `filter`, `each` and `find` already built into cjs arrays? No need to duplicate them here.

jaszhix: They are, but on Spidermonkey there is a performance gain to using for loop alternatives instead. This is a less of an issue on V8. I ran these tests in the Firefox 52 ESR, which is the closest to our environment.

https://jsperf.com/aarontgrogg-array-filter-vs-for-loop
https://jsperf.com/loops/33
http://jsben.ch/ABI1C
http://jsben.ch/aY0kG

germanfr: Those tests compare filter vs raw code, but that is not the case here because at the end what slows it down is calling the callback function I guess. I don't know about the implementation but sure is very similar (?). I'm going to test it on Cinnamon anyway, just because I'm curious :)

jaszhix: Cool. The jsben.ch tests are ones I set up - here's one similar to the findIndex one, but with filter: http://jsben.ch/umjQN

germanfr: My results conclude that you are right, the loop implementation is faster BUT, the difference in my tests for arrays of 10M elements is a couple of ms (and it even becomes zero the more you use it). For arrays of 100 elements (if you go crazy) the difference is non existent. I think native methods should be used for consistency, or preferably normal loops which is perfectly fine, but defining those seems like not ok at all. It's just my opinion :)

[Here are my tests](https://drive.google.com/open?id=1cPWm7nbT57zC4d-glCi1XUGRau_WciLC) (it is a cinn extension).

jaszhix: Thanks for looking into this more. While I agree this won't make a dramatic difference in a lot of cases, I think these methods should stay in the PR. Take a function that iterates multiple things to get its return value and gets called relatively frequently for instance, for invocation the latency can add up - especially when we want Cinnamon to be as responsive.

These are there to provide an option, not to replace every instance of filter, findIndex, etc - as you point out the gains diminish with smaller arrays, but the gains are still there and consistent.

Here's the results I got from test-extension:

Cjs-Message: 13:45:25.762: JS LOG: [LookingGlass/info] Array.prototype.filter: 20.6ms
Cjs-Message: 13:45:25.763: JS LOG: [LookingGlass/info] Custom filter:              11.7ms

If this were a function that was meant to assist with an animation, it could mean the difference between something being jerky or smooth, since CInnamon needs to spend 16ms or less for each frame.

germanfr: I understand your point, but in those cases why not to use loops? I presume they are even faster than these functions.

PS: those results are from 1M array items and appear like the first time or so you use the functions, then they are like the same time. In an animation there would be no difference for example, assuming you have an array of <1M items.

jaszhix: I think you're seeing that because of Spidermonkey's JIT optimization- if it doesn't get GC'd its still cached as byte code, but we shouldn't rely on that. 

This allows the performance benefit of for loops, but the readability of explicitly named functions like forEach. The each function I added in particular can iterate both arrays and objects, which can be convenient. For objects it uses Object.keys and a for loop, which is a little faster than for...in.

germanfr: Sorry @jaszhix I still don't get where is exactly that critical place in this PR where you have a large array and save more than 1 ms. I think that such micro-optimisations belong to cjs, not cinnamon. But idk 😅

germanfr: Yes, of course. But why not implement them in cjs instead? If we can make native methods faster why can't we improve them instead of creating faster duplicates so that cinnamon can benefit from that in every single place? It seems like a better solution to me, but maybe there is a reason not to do that I can't see.

jaszhix: Mainly because I'd prefer to keep CJS even as possible with GJS, otherwise back-porting patches gets harder. I could upstream it, but not sure they would want them. If they do, great, but then we can't add the optimizations on our schedule.

germanfr: Oh I understand. Maybe some day! Sorry for bothering with this 😅

smurphos:
Intriguing and exciting.

Some questions/suggestions?

I don't think ITM currently supports presenting itself as a target for 'Add to panel' from the menu. It is possible to implement this in grouped-window-list - I see from the commits that the intention is not to have a panel launcher applet in the default config as grouped-window-list will come with pre-pinned apps.

I don't know if this is technically doable but will grouped-window-list be able to handle pinned custom launchers and differentiate them from the main application.  My use case is half a dozen Chrome apps (Inbox / Calendar / Contacts etc) which I have in the menu and as panel-launchers, but for obvious reasons in ICT and the regular window-list they are just another Chrome window.

Hopefully advanced-menu will retain Cinnamenu's Search Providers management GUI? I think that Search providers are probably one of Cinnamon's most undiscovered features amongst the general user base. I originally discovered them via Cinnamenu.

Looking forward to testing - not so much looking forward to the theme fixing..... ;-)

smurphos:
> Hi @smurphos, looks like the "Add to panel" functionality is working for me - I recall adding support for that a while ago. Pinning Chrome apps works for me as well. Not sure what the issue could be. If its an ITM-specific problem, feel free to open an issue on its repo.

Thanks - it's working now - I've recently moved back to ITM from CobiWL - when I tried just after removing my existing panel-launchers add to panel spawned a fresh panel-launcher. Looks like a reboot has sorted that behaviour out. Also I note the Chrome apps are distinguished from the generic Chrome window - kudos @jaszhix

claudiux:
Hi @clefebvre,

You project to "Remove user applet from enabled-applets".
Will there be a "enabled-user-applets" entry in dconf?

All these new changes will appear in Cinnamon 3.8, or in Cinnamon 4.0?

Best regards.

Claudiux

claudiux:
Thank you for the clarification, @jaszhix . Too bad that Cinnamenu does not replace the menu. When I install Linux Mint on a computer, the first thing I do is replace the menu by Cinnamenu and the taskbar by Icing Task Manager.

clefebvre:
Yes, this is getting merged but fixed hopefully today. We've looked at the transparent padding within the icons themselves. In most icon themes, there's none at 16px, and 2px at 24px.. that explains why we both thought we were wrong (we were :)).

PR #7947, opened 2018-10-20
notifications applet: new icons

NikoKrause:
these icons are from elementary icon theme


germanfr:
Hey @NikoKrause, I wonder if you could add the `success` class to the normal notification dot?

NikoKrause:
I talked with @clefebvre about this and he didn't wanted to be annoyed by the notifications.

NikoKrause:
@germanfr I think the critical icons are never used. At least I wasn't able to trigger those icons with `notify-send`.

NikoKrause:
Elementary OS 5 does animate it's panel icons now:
Elementary panel animations

germanfr:
Critical notifications stay on screen and never go to the tray that I know. I mainly said it because they all are very similar icons and so that it was easier to tell if there was a notification at first sight.

NikoKrause:
The notification applet is invisible by default. So if you see a bell, you know there's a new notification.

NikoKrause:
And it shows the number of notifications

germanfr:
True, true. It was just an idea. It doesn't really matter.

NikoKrause:
Yeah, I had the same idea with adding the success class.

PR #7967, opened 2018-10-29
Some fixes to the new panel layout

germanfr:


js/ui/panel.js
```diff
@@ -2921,33 +2921,26 @@ Panel.prototype = {
...
 
     getPanelZoneIconSize: function(locationLabel, iconType) {
         let zoneConfig = this._panelZoneIconSizes;
+        let symbolic = iconType === St.IconType.SYMBOLIC;
 
         if (!zoneConfig) {
-            let height = this.height / global.ui_scale;
             global.logError(`[Panel ${this.panelId}] Unable to find zone configuration`);
-            return toStandardIconSize(height);
+            return _calculatePanelZoneIconSize(0, symbolic);
```
jaszhix: Hi @germanfr, thanks for the fixes. It looks like `this` is missing here.

js/ui/panel.js
```diff
@@ -2825,10 +2824,11 @@ Panel.prototype = {
         if (height === this.height) return;
 
         this.height = height;
-        this.emit('size-changed', height);
 
         // In case icon sizes are responding to panel height
         this._onPanelZoneIconSizesChanged();
+
+        this.emit('size-changed', height);
```
jaszhix: Why was this moved?

js/ui/panel.js
germanfr: Oops

js/ui/panel.js
germanfr: Because it painted the previous icon size instead of the current one.

jaszhix: That's the signal for panel height - `_onPanelZoneIconSizesChanged` would be doing the signal emission for `icon-size-changed`. I recently fixed a systray bug that was causing the icon size to be behind - was that what it was? If so, can you check if this still occurs after https://github.com/linuxmint/Cinnamon/pull/7944/commits/4f657d8527094b5589f03036e978708bb4dc8d7a? It would be safer to resize the panel first.

germanfr: No, it was the whole panel, not just the systray. You can check by setting the panel height to the maximum, because just there icons increase, but they don't until you decrease -1px.

jaszhix: The intention behind breaking them up with specific signals was so we could stop letting applets control the _panel's_ state. If there is still logic in applets depending on it, we should try to fix that in applets. I currently can't reproduce the issue as you describe it, icons always scale on height change. Applets I'm using:

```
$ gsettings get org.cinnamon enabled-applets
['panel1:right:3:systray@cinnamon.org:360', 'panel1:right:8:show-desktop@cinnamon.org:362', 'panel1:right:6:notifications@cinnamon.org:366', 'panel1:right:7:calendar@cinnamon.org:372', 'panel1:right:5:sound@cinnamon.org:373', 'panel1:left:0:Cinnamenu@json:374', 'panel1:left:1:grouped-window-list@cinnamon.org:379', 'panel1:right:1:multicore-sys-monitor@ccadeptic23:393']
```

jaszhix: Ah, nevermind, I see this fixes the icons disappearing - it just makes me more curious about why though - all that does is emit a signal that now fewer applets should be using for size adjustments - `on_panel_height_changed` should essentially be replaced with `on_panel_icon_size_changed`. `on_panel_height_changed` does nothing in ui/applet.js, its there for compatibility.

jaszhix: ...and figured out why! It [this line](https://github.com/linuxmint/Cinnamon/blob/f867487d6d1978636fada7beea4f1f0ab7be3e44/js/ui/panel.js#L2915) here that's causing it. For some reason emitting the panel height signal afterwards propagates that somehow, but its the check on if the icon size change maybe being too strict. I suppose it doesn't make a difference when the panel height signal is called, but if we blanket patched it with signal re-ordering we might not get to the root cause. I'd rather not have code that leads other devs to think it behaves in ways it doesn't, so I'll look at that tomorrow.

germanfr: It was all the icons in the panel, not specific to any applet (even IconApplets). So I don't know.

jaszhix: Ok, your solution looks good. I was confused why the icons weren't updating, but then remembered `_onPanelZoneIconSizesChanged` is only supposed to be run on size change, calling it outside of that will be less likely to propagate changes as an optimization.

PR #7978, opened 2018-11-04
melange/lookingGlass: better object inspection

itzexor:
Reworks object parsing code so that we can inspect more types,
present the information more clearly, and prevent misbehavior
with types we shouldn't iterate like importers.

Notable behavior changes:
 - Boxed instance proxies such as Clutter.Color, Clutter.Rect, etc
   are now inspectable like other objects.

 - Inspecting importers no longer loads every file.

 - Errors during command eval() are now inspectable.

 - Values are currently truncated at 120 characters. When an
   inspectable object's toString value is longer you will need
   to manually call toString() on it to view the full value.
   Strings are viewable in a popup dialog.

 - The 'shortValue' key in the Inspect dbus method is no longer
   used as it is not available on all of the methods, and so
   easier to deal shortening values on the Melange side.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-looking-glass/page_inspect.py
+++ b/files/usr/share/cinnamon/cinnamon-looking-glass/page_inspect.py
@@ -54,8 +54,13 @@ class InspectView(pageutils.BaseListView):
 
     def setInspectionData(self, path, data):
         self.store.clear()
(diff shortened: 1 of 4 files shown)


itzexor:
camel case is the existing style

PR #8004, opened 2018-11-09
IndicatorManager: Clean up

jaszhix:
- Uses arrays instead of objects for storing collections.
- Uses SignalManager instead of its own signal manager.
- Uses class syntax.
- Uses faster for-loop wrapped array methods from misc/util.
- Cleans up object references when an instance is destroyed.
- Corrects some grammatical errors in comments.

### DIFF
--- a/js/ui/indicatorManager.js
+++ b/js/ui/indicatorManager.js
@@ -34,6 +34,7 @@ const Main = imports.ui.main;
 const DBusMenu = imports.ui.dbusMenu;
 const SignalManager = imports.misc.signalManager;
 const CinnamonConfig = imports.misc.config;
+const {each, map, filter, find, findIndex} = imports.misc.util;
 
 const SNICategory = {
     APPLICATION: 'ApplicationStatus',
@@ -63,7 +64,7 @@ const WATCHER_OBJECT = '/StatusNotifierWatcher';
 
 const ITEM_OBJECT = '/StatusNotifierItem';
 
-const StatusNotifierWatcherIface = 
+const StatusNotifierWatcherIface =
     '<node> \
         <interface name="org.kde.StatusNotifierWatcher"> \
             <method name="RegisterStatusNotifierItem"> \
@@ -102,14 +103,13 @@ function IndicatorManager() {
 }
 
 IndicatorManager.prototype = {
-    
+
     _init: function() {
         this._enable = false;
         this._signalStatus = 0;
         this._signalSystray = 0;
-        this._indicatorsSignals = null;
         this.statusNotifierWatcher = null;
-        this._indicators = {};
+        this._indicators = [];
         this._blackList = [];
         this.setEnabled(global.settings.get_boolean("enable-indicators"));
         this._signalSettings = global.settings.connect('changed::enable-indicators',
@@ -122,9 +122,9 @@ IndicatorManager.prototype = {
 
     _onSystrayManagerChanged: function(manager) {
         let rolesHandled = manager.getRoles();
-        for (let id in this._indicators) {
-            this._conditionalEnable(id, rolesHandled);
-        }
+        each(this._indicators, (indicator) => {
+            this._conditionalEnable(indicator.id, rolesHandled);
+        });
     },
 
     // handlers = { "signal": handler }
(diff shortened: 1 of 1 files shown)


lestcape:
@jaszhix with the last fix, you are returning now a Clutter.Actor here: https://github.com/linuxmint/Cinnamon/commit/7ad7a02162339739fa4b80d1d1536f83eda8128f#diff-05db65ea5625e393ffab82cc7a763e9cR1276

Return a Clutter.Actor is a bad idea because you can not set a theme in a Clutter.Actor  and now an indicator actor can have two different types of actors, because it also can be an St.Icon: https://github.com/linuxmint/Cinnamon/blob/master/js/ui/indicatorManager.js#L1222

So, finally the manager is producing different actors depending from where the actor is constructed and it can cause incompatibilities to whatever use the API from an extension for example.

The issue you mention in the ubuntu side have nothing to be with the problem you fixed. Probably the issue was caused with the code i introduce to resolve the incompatibility between the Clutter.Actor and the St.Icon. As you revert that code probably the problem was gone but again the API is inconsistent.

jaszhix:
Hi @lestcape, thanks for the tip. Your suggested change using `new_from_bytes` does indeed work, and I will update this PR accordingly when I rebase it.

lestcape:
In case you don' t find a good way, this code should work (as a workaround the GdkPixbuf.Pixbuf missing support), but I don't recommended perform that task in the compositor thread. Please note that depending of the size of the image the function `_changeARGBToRGBA` can freeze the shell. A recommend procedure is implement this function in C and in an asynchronous way using an auxiliary thread, just convert the pixel format before the `GdkPixbuf.Pixbuf` will be constructed.

 ```js
// This function convert a ARGB pixel format to a RGBA pixel format.
_changeARGBToRGBA: function(bytes, width, height) {
    let temp;
    for (let x = 4*(width*height-1); x>=0; x-=4) {
        temp = bytes[x+3];
        bytes[x+3] = bytes[x+0];
        bytes[x+0] = bytes[x+1];
        bytes[x+1] = bytes[x+2];
        bytes[x+2] = temp;
    }
    return bytes;
},

bytes = this._changeARGBToRGBA(bytes, width, height);
let pixbuf = GdkPixbuf.Pixbuf.new_from_bytes (bytes,
                                              GdkPixbuf.Colorspace.RGB, //Cogl.PixelFormat.ARGB_8888,
                                              true, //has_alpha
                                              8, //bits_per_sample
                                              width,
                                              height,
                                              rowstride);
 ```

jaszhix:
@lestcape The issue appears to be depending on how the icon is created, it could be St.Icon or Clutter.Actor. I think this is fine for now because we don't theme the St.Icon, and it serves a purpose of having the icon search path built into its initialization. When an icon is added to the systray applet, its treated like a Clutter.Actor, and only Clutter methods are called on it. I can't see a difference between normal systray icons and indicator ones on this branch, and since we now handle icon sizes globally, it might not make sense to theme it.

lestcape:
@jaszhix this was in that way, because the default Cinnamon theme already apply an special style to icons inside the panel. Is an effect like a light shining behind the icon. But if this was removed i don' t know, i don' t use cinnamon for a long time ago. If it's fine to you... that's ok.

jaszhix:
I pushed a commit that uses St.Icon anyway, its a sub-class of St.Widget (which sub-classes Clutter.Actor), so should be fine.

lestcape:
I think if you set the Clutter.Actor.content the theme will never be apply, because the content override any attempt, but well, maybe I'm outdated and this works.

jaszhix:
As a test, I added this rule to cinnamon.css:

```css
StIcon {
    width: 16px !important;
    height: 16px !important;
}
```

After restarting cinnamon, most of the icons were forced to 16px, except the tray icons - regardless if indicators were enabled and what the icon size setting was for its panel zone. This makes sense for normal icons, because the normal tray icons are CinnamonTrayIcon instances, and I guess for indicators its because its actually IndicatorActor being used in the systray applet, not the icon itself? In any case seems to be fine with the way it works now.

lestcape:
Make a test of the sizes when you already know they are hard-code (https://github.com/linuxmint/Cinnamon/commit/7c40ef4b960c1556583fbfbe83508fef2f1803ff) don't has sense at all to me, but i don' t know what you want to prove with that, if is fine to you that nice.

What you have right now are some indicators that can have a theme, that are `_createIconByName` and also some indicators that can **NOT** have a theme, that are `_createIconFromPixmap`. In any case non of those icons have currently an style class, so they directly can not have a theme, but they can have a theme indirectly using the parent class name and then the StIcon class.

Like the parent class https://github.com/linuxmint/Cinnamon/blob/master/js/ui/indicatorManager.js#L956 have an style class set as as a 'applet-box', this automatically apply an style to the child elements in the default theme. That style can be see for those icons that are create by `_createIconByName`, but not with those that are create with  `_createIconFromPixmap`. As how both are indicators icons, i think just for consistency should have not difference between them regardless if they should have or not themes.

But to be honest is like you want it be. Also in the case that i will use cinnamon, i don' t care how the indicators are show or if they have consistency or not.

jaszhix:
We added icon size configuration per-panel zone (https://github.com/linuxmint/Cinnamon/commit/0d6c75f5b2ebe390813f4b65e25a62ff5df5d472). They can be set between 16px-64px, the first commit you referenced just changes the defaults.

I didn't notice any change with the applet-box style class applied to that actor, so I removed it. This may have been different when you added indicator support originally.

lestcape:
> This may have been different when you added indicator support originally.

Yes, probably is like you say, i don't know how is right now. Originally have theme support was a requirement and the main incentive to port it to Cinnamon because this was the main difference with status icons, as indicators are shell friendly and the status icons are Cuttler embeded GtkWidgets only. If that change some how or now this is not important is ok.

PR #8006, opened 2018-11-09
systray-applet: more spacing between icons

NikoKrause:
right now the icons look pretty squashed, especially with small icon
sizes


smurphos:
Couldn't the spacing be varied depending on the icon / panel size that has been set? Isn't increasing the spacing to account for small icons just going to create an issue for large icon sizes? Or add a simple configuration GUI to the systray applet to allow the user to set a spacing value.

clefebvre:
@NikoKrause ^

PR #8008, opened 2018-11-09
gwl: Simplify thumbnail layout

germanfr:
* Remove unnecessary layout elements inside elements.
 * Add a close clone to the left to properly center the title.
 * Remove close button blurriness.

### DIFF
--- a/data/theme/cinnamon.css
+++ b/data/theme/cinnamon.css
@@ -1312,9 +1312,15 @@ StScrollBar StButton#vhandle:hover {
 /* ===================================================================
  * Grouped window list (grouped-window-list@cinnamon.org)
  * ===================================================================*/
-
-.grouped-window-list-thumbnail-label {
-    padding-left: 4px;
+.grouped-window-list-thumbnail-title {
+    spacing: 2px; /* Separate title from close button */
+}
+.grouped-window-list-thumbnail-close {
+    color: #ccc;
+}
+.grouped-window-list-thumbnail-close:hover {
+    color: #fff;
+    text-shadow: white 0px 0px 5px;
 }
 .grouped-window-list-number-label {
     z-index: 99;
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -905,7 +905,7 @@ class AppGroup {
 
         each(this.hoverMenu.appThumbnails, (thumbnail) => {
             if (thumbnail.metaWindow === metaWindow) {
-                thumbnail.labelContainer.child.set_text(metaWindow.title);
+                thumbnail.label.set_text(metaWindow.title);
                 return false;
             }
         });
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/applet.js
@@ -250,8 +250,7 @@ class GroupedWindowListApplet extends Applet.Applet {
             lastCycled: null,
             lastTitleDisplay: null,
             scrollActive: false,
-            thumbnailMenuOpen: false,
-            thumbnailCloseButtonOffset: global.ui_scale > 1 ? -10 : 0
+            thumbnailMenuOpen: false
         });
 
         // key-function pairs of actions that can be triggered from the store's callback queue. This allows the
@@ -1003,7 +1002,6 @@ class GroupedWindowListApplet extends Applet.Applet {
     }
 
     onUIScaleChange() {
-        this.state.set({thumbnailCloseButtonOffset: global.ui_scale > 1 ? -10 : 0});
         this.refreshAllAppLists();
     }
 
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/constants.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/constants.js
@@ -1,10 +1,5 @@
-const CLOSE_BTN_SIZE = 22;
 const constants = {
-    CLOSE_BTN_SIZE,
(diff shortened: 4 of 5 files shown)


jaszhix:
This PR will conflict with #7986, which also changes styles and includes bug fixes. 

I also think this is not as sane of a default as some styling. Without any default theming, we have to make sure every theme doesn't look like this:

image

The thumbnail styles were originally inlined because the close button size could be customized in ITM, and some scaling trouble.

We decided that the current close button size worked well, and too small of a button could pose usability issues. Theme authors have the power to make the UI less usable, and often times due to abandonment, or upgrades, they inadvertently do. I think this change makes it more likely users will be looking at a broken interface, and makes theme maintenance more difficult.

This change also breaks thumbnail scaling and makes thumbnails clip out of the monitor without constraint - which is another good reason to not give theme authors complete control of the UI. Its easier to make thumbnails not clip the monitor when we can predict some things, instead of calling expensive `peek_theme_node` calls.  

Tested by cherry picking into #7986, resolving conflicts.

https://github.com/jaszhix/Cinnamon/commit/4061b1fc88e58ec1e63e5795d32c9fab2af78caf

germanfr:
I understand your point with the icon size. That is something totally unnecessary to this PR. I'm not sure why is the text in your image so small in proportion to the rest of the UI though.

About the overflow. Is that with any special styling or is it always? I'm not able to reproduce right now. In fact, the width is applied to the outermost container like it was before. It only turns out it is a box layout instead of a st.box with a box layout inside.

germanfr:
I discovered why is the overflow, and it's because of the auto-merging forgot to also remove [these lines](https://github.com/linuxmint/Cinnamon/pull/8008/commits/258c324be015afe08bebb72b914778da618bf91c#diff-81b3056bae7441aa2fdaca2ba3d863c6L759) that set the width of the label (no longer needed). Fixed that, I'll revert the customizable size thing and there are no more problems I can see.

germanfr:
I'll mark this as WIP until #7986 is merged into master, and then I'll rebase.

germanfr:
@jaszhix ready, thoughts?

jaszhix:
This is still conflicting with my bug fix PR, I thought you were going to wait?

jaszhix:
Hi @germanfr, I was able to test this again. It looks like this regresses thumbnail scaling again - for instance, if you open gnome-calculator repeatedly, the thumbnail menu should fill the entire width of the monitor, but there's a large gap between the edge, and the thumbnails scale too soon. When it switches to vertical orientation from too many thumbnails, text from the thumbnail title goes missing.

Is it necessary to theme the label container? We need to take that into consideration and call style_length on the label container in refreshThumbnails. These are not cheap calls, and I added the padding/margin checks out of necessity, so I'm not thrilled with seeing more classes added to make this more complicated. Window thumbnails are the most performance sensitive part of GWL as they have the most constructor instances.

I'm not a fan of the symbolic close button, and this PR doesn't make it obvious that is being changed. The hover pseudo class added to the button in cinnamon.css doesn't have any apparent effect. I think using the `window-close` style class is better and makes it look more like a button in most themes. Themes can still use SVGs with it.

germanfr:
Actually @jaszhix, I think the calculation was already broken before this PR, because if you look closely at the example of the calculators, the last `item-box` is outside of its container.

Now the title box has the same width assigned to the thumbnail container (they have to be the same width because they are above each other). Before the PR the titlebox didn't have a width but the label had (width - 16px) resulting in a slightly larger width than the width assigned to the thumbnail actor, leading to wider item-boxes and an appearance of fitting, but it wasn't really ok. 

In summary, it's the width given to the thumbnail actor the one that was wrong and have to be fixed :)

jaszhix:
The 16px offset can probably go, but I don't think we need to throw away everything. There's still a lot of code left over from ITM, where it was easier to make these assumptions.

> leading to wider item-boxes and an appearance of fitting, but it wasn't really ok.

Did you have a solution to this? It currently works fine on master. It may be in need for improvement, but its only broken on your branch.

germanfr:
> Is it necessary to theme the label container?

Yes, it is. From a theme's point of view, it is easier to define the spacing between the title and the thumbnail, and you can also set the font from there. There is no need to target the label specifically. In my opinion the label class should be removed.

> I'm not a fan of the symbolic close button

This is why icons exist, the theme doesn't have (and doesn't want) to maintain images for each action in the interface (for example popup menu icons). The notification has the same close icon, at least this adds some consistency. This can still be styled like a button with CSS and background images because StIcon is a themable element.

> When it switches to vertical orientation from too many thumbnails, text from the thumbnail title goes missing.

I can't reproduce at all.

> We need to take that into consideration and call style_length on the label container in refreshThumbnails

We already did that, but to the label container instead of the title container, whose width we have already calculated (it's the same as the thumbnail container), and then its children auto fit.

germanfr:
> It currently works fine on master. It may be in need for improvement, but its only broken on your branch.

It is already broken on master, look:
screenshot from 2018-11-13 20-31-07

germanfr:
> Did you have a solution to this?

I am working on it. Though I think it'd be better not t o try to fit the size exactly, because it is hard to calculate the width of elements inside elements.

jaszhix:
Ok, I understand the change in the hierarchy, and it is simpler. But I think that's where the simplicity ends here.

- You're cloning the close button. It looks like a hack.
- Actors and relayouts are performant. Theming every actor because you can isn't.
- You're breaking the scaling method I implemented _before this PR was open that's been refined in ITM for nearly 2 years_, and saying I should fix it.

This is _not_ happening on master:

thumb2
thumb1

> This is why icons exist, the theme doesn't have (and doesn't want) to maintain images for each action in the interface (for example popup menu icons). The notification has the same close icon, at least this adds some consistency. This can still be styled like a button with CSS and background images because StIcon is a themable element.

Yes, but by who? We have tons of themes that don't support all this new stuff, and the close button is a design decision that should be in a separate PR and scrutinized differently from a refactor one.

germanfr:
@jaszhix why are you always right? :)

PR #8021, opened 2018-11-13
menu-applet: several fixes

NikoKrause:
This PR fixes several issues, which were introduced by changing the favoritesBox to a scrollable Box


NikoKrause:
@germanfr Which weird theming issues?

germanfr:
See for example Railway. If you use that class to define a background/padding or anything it gets now applied there, which is not the applications scrollbox.

NikoKrause:
Changed

PR #8032, opened 2018-11-17
GWL: Upscale thumbnail size options and add Largest option

smurphos:
Thumbnail sizes, other than the Large option were in my opinion too small to be useful.

This commit adds a 'Largest' option to match the regular window-list applets 4 options and changes the range to 6 (small) to 15 (largest) with 9(medium) as the default.

These are still not as large as the equivalent window-list applet thumbnails, to avoid unnecessary resizing when grouped thumbnails exceed the monitor height or width.

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/settings-schema.json
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/settings-schema.json
@@ -245,12 +245,13 @@
   },
   "thumbnail-size": {
     "type": "combobox",
-    "default": 6,
+    "default": 9,
     "description": "Thumbnail size",
     "options": {
-      "Small": 3,
-      "Medium": 6,
-      "Large": 9
+      "Small": 6,
+      "Medium": 9,
+      "Large": 12,
+      "Largest": 15
     },
     "tooltip": "Controls the size of the thumbnails."
   },


jaszhix:
Hi @smurphos, what resolution are you testing on? 

I think the thumbnails are only supposed to provide a vague representation of the windows - enough to be distinguished. Hover peek is there for more useful previewing. The size option to me is more about how many window selections does one want to fit on the screen - on lower resolutions, this might make the thumbnail menu collapse into the vertical layout too easily.

This looks a little big by default on my not-so-common resolution of 3440x1440, and bigger than the window list, which is already smaller than the current GWL default. The bigger the thumbnails are, the longer the cursor needs to move to make a selection. I would recommend adding a larger option, but preserving the current settings and default.

smurphos:
I'm on 1366 x 768. I guess I should have checked on an external display if the actual onscreen display size is resolution dependent. On this display the window-list largest is still quite a big bigger than my modified largest. So you think that 3, 6, 9 and 12 with 6 as default is a good middle ground?

jaszhix:
Yeah, the thumbnail size is a multiplier in the scaling logic, so it will vary depending on the resolution - I think if this isn't working as well on lower resolutions, it should be addressed in the scaling code. The default of 6 works well on my 1080p laptop - that's the smallest resolution I use normally. I did work in 1280x1024 for a while recently when addressing some scaling issues in #7986, and don't remember the thumbnails looking too small.

smurphos:
This is what i get at default medium settings for both applets (prior to this PR) at 1366 x 768 - it's a pretty common resolution if this is is taken as indication -  https://www.ubuntu.com/desktop/statistics So I guess the scaling code is where this can be looked at. I'll close this and re-raise as an issue for the time being.

screenshot-screen-2018-11-17-081008
screenshot-screen-2018-11-17-080956

Small is super tiny for GWL

screenshot-screen-2018-11-17-080913
screenshot-screen-2018-11-17-080902

NikoKrause:
I see the same as on the screenshots. "Small" is unusable/unhelpful

PR #8042, opened 2018-11-20
cs_themes.py: speed up loading

mtwebster:
Defer loading the download page and chooser entries until after the page is loaded.

The spices/downloader stuff takes a long time to load, and ends up freezing
the ui a bit when the Themes page is selected.  Load it after any animations
have completed.

The chooser buttons are trying to all load at once.  Instead load them
sequentially and display a spinner until loading is complete.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/ChooserButtonWidgets.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/ChooserButtonWidgets.py
@@ -97,41 +97,6 @@ class PictureChooserButton(BaseChooserButton):
         else:
             self.button_image.set_size_request(-1, button_picture_size)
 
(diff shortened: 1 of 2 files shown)


brownsr:
Sorry, but this is terribly laggy for me.  I am seeing some odd images shown in place of the icons when first loaded, look like black and white circles in circles.  Second time through performance seems normal, if just going backwards and forwards in the settings.  But if the settings window is closed and reloaded then the load times are back to laggy again.  At the moment the solution seems much worse than the original problem.  Sorry !

mtwebster:
Np this is why I didn't just push it, I was afraid it would be inconsistent with different hardware, number of themes, etc.. just a quick effort to improve the response time.

And those circles within circles are gtk spinners :)

PR #8050, opened 2018-11-21
extension: Prevent loading xlets that use imports.gi.NMClient that have no 4.0 directory

jaszhix:
Refs
https://github.com/linuxmint/cinnamon-spices-applets/issues/2097
https://github.com/linuxmint/cinnamon-spices-applets/pull/2094

These xlets are affected:

```js
var knownCinnamon4Conflicts = [
    // Applets
    'turbonote@iksws.com.b',
    'vnstat@linuxmint.com',
    'netusagemonitor@pdcurtis',
    'multicore-sys-monitor@ccadeptic23',
    // Desklets
    'netusage@30yavash.com',
    'simple-system-monitor@ariel'
];
```

CC @pdcurtis @arielandrade

### DIFF
--- a/js/ui/extension.js
+++ b/js/ui/extension.js
@@ -22,6 +22,19 @@ var State = {
     OUT_OF_DATE: 3
 };
 
+// Xlets using imports.gi.NMClient. This should be removed in Cinnamon 4.2+,
+// after these applets have been updated on Spices.
+var knownCinnamon4Conflicts = [
+    // Applets
+    'turbonote@iksws.com.b',
+    'vnstat@linuxmint.com',
+    'netusagemonitor@pdcurtis',
+    'multicore-sys-monitor@ccadeptic23',
+    // Desklets
+    'netusage@30yavash.com',
+    'simple-system-monitor@ariel'
+];
+
 // macro for creating extension types
 function _createExtensionType(name, folder, manager, overrides){
     let type = {
@@ -186,6 +199,8 @@ Extension.prototype = {
         this.iconDirectory = null;
         this.meta = createMetaDummy(uuid, dir.get_path(), State.INITIALIZING);
 
+        let isPotentialNMClientConflict = knownCinnamon4Conflicts.indexOf(uuid) > -1;
+
         const finishLoad = () => {
             // Many xlets still use appletMeta/deskletMeta to get the path
             type.legacyMeta[uuid] = {path: this.meta.path};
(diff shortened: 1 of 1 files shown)


pdcurtis:
@jashix What is the problem which requires a different version for Cinnamon 4. I thought we had wrapped the nm options in a  try-catch in  to so it would work with both versions to support distributions such as fedora in  vnstat@linuxmint.com ,  netusagemonitor@pdcurtis and netusage@30yavash.com .
Is there a tested fix I can patch into these and does there really need to be multiple versions?

jaszhix:
@pdcurtis I think it has to do with the build configuration - see https://github.com/linuxmint/Cinnamon/commit/fbb12f3d381c4ad5c40f039120db54745613340d. Any use of the old API is a no-op now. For the time being, applets using them will be ignored before they're able to segfault Cinnamon.

I should have copied the files, made a progress commit, then applied a separate commit for the fix so I could point to a diff to show what change is needed, but I'll explain. If you look [here](https://github.com/linuxmint/cinnamon-spices-applets/blob/f5ba871afc589a75abf0825373ef3516356c3cac/multicore-sys-monitor%40ccadeptic23/files/multicore-sys-monitor%40ccadeptic23/4.0/DataProviders.js#L15), I basically made the fail-safe (inside the catch block) the only import - using `imports.gi.NM`. Then pass `null` to NM.Client.new(), and make any other conditional behavior default to the new API. That's the only change that needs to be made to made from what I could tell. Since this would break older environments expecting the old NM API, I forked it into a 4.0 directory.

I fixed `netusage@30yavash.com`, and removed `turbonote@iksws.com.b` (broken). The remaining ones are yours or @arielandrade - I was going to wait a bit on them since they have maintainers.

pdcurtis:
@jazhix Thanks. I will first run meld on the changes to netusage@30yavash.com into the fork and see what you actually changed there as I did the last set of modifications.  
I was going to make some further changes to  netusage@30yavash.com in any case as it currently fills the error log if there is no connection present. 
I will come back to you if I need to and I hope you will be prepared to test any changes as I do not have a Cinnamon 4 system. 
It may take couple of days before I have the time. When is the Beta or something I can easily use for test coming out.

pdcurtis:
@jaszhix I have had a look at  `netusage@30yavash.com` and one thought has occurred to me: We originally used a try-catch so it would work with either form of NM because there was a split of usage  between the other distributions that 'bundled' cinnamon in some way. Do all the other distributions now use the new NM or should one be changing the try-catch to avoid the segfault instead of just using a multi-version solution?

PR #8057, opened 2018-11-23
gwl: Improve thumbnail menu input and button label handling

jaszhix:
This makes it so hovering over an app button doesn't immediately give the opening thumbnail menu exclusive input focus. Until the user decides to interact with the menu, we should assume they are typing in another window.

When a menu is opened via Super + space toggling, it will always grab input.

In addition, other fixes in this PR includes:

- Button labels randomly showing the window title when the app title is set.
- Button labels not always showing the focused window label when the fourth option is set.
- Focus style on app buttons not showing when the applet first loads.
- Incorrect key navigation on vertical thumbnail menus when the panel is oriented horizontally.

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -78,7 +78,8 @@ class AppGroup {
         this.groupState.connect({
             isFavoriteApp: () => this.handleFavorite(true),
             getActor: () => this.actor,
-            launchNewInstance: () => this.launchNewInstance()
+            launchNewInstance: () => this.launchNewInstance(),
+            checkFocusStyle: () => this.checkFocusStyle()
         });
 
         this.signals = new SignalManager(null);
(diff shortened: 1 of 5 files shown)


smurphos:
This causes a regression.

Hover over a focused app, highlight the thumbnail, then move the mouse away. GWL reverts to active while the app remains focused in the foreground.

jaszhix:
@smurphos Thanks for catching that - I pushed a fix.

smurphos:
I'm seeing invisible/blank labels on launching a pinned app right now. Switching the label mode or refreshing the theme brings them back. Unpinned apps are OK.

jaszhix:
@smurphos Thanks, looks like its because of the animated width transition to 1px when hidden. This worked fine before, but with the allocation changes it looks like the label width needs to be set explicitly when shown.

jaszhix:
@smurphos Pushed another fix. Does that work?

smurphos:
It fixes the label appearing. Now there's an issue with the width of a pinned app not reverting to the normal closed width if there is another open app in the applet. It does revert as soon as the mouse hovers over the applet. 

When there isn't another app open it does revert but the animation is a bit glitchy on my machine.

smurphos:
All looking good now ;-)

ovari:
Does this work correctly when a user defines Super+Space to cycle though keyboard layouts? Thank you

jaszhix:
If there's another keyboard shortcut using that combination it may not. You can change the binding in the applet's settings.

PR #8068, opened 2018-11-25
[Next] [WIP] Introduce a wrapper container for reducing signal usage

jaszhix:
This is loosely based on the idea of this [upstream MR](https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/153). 

>  Signals are not specially performant. One of the reasons is that they acquire a global lock for signal handlers lookup.

Since it could be compatibility breaking to do away with `GenericContainer` in a lot of public-facing classes (e.g., `set_skip_paint`), this adds a wrapper that exposes GObject class vfuncs to normal JS classes. This way we can benefit from the performance gains without needing to change a lot of code.

Todo:

- [x] Replace remaining `Cinnamon.GenericContainer`s
- [x] Consider adding other performance sensitive vfuncs
- [ ] Compare performance with master and optimize
- [ ] Fix checking vfunc props causing the process to crash ([upstream MR](https://gitlab.gnome.org/GNOME/gjs/merge_requests/259))


jaszhix:
Closing in favor of #8230.

PR #8077, opened 2018-11-28
windowManager: Add an effects blacklist

jaszhix:
This overrides the mapping effect for xpad because it [toggles the visibility](https://bazaar.launchpad.net/~xpad-team/xpad/trunk/view/head:/src/xpad-pad.c#L531) of its windows after re-decorating them, which causes artifacts in Muffin.

Ref https://github.com/linuxmint/muffin/issues/388


jaszhix:
https://github.com/linuxmint/muffin/pull/390

PR #8094, opened 2018-12-02
gwl: Fix panel buttons allocation

germanfr:
* Don't use buttons margin to also separate the label from the icon
 * Don't reuse the badge allocation box, otherwise the label starts in the middle of the icon (this was fixed with an aproximation leading to huge spacings).
 * Use the CSS spacing property to separate the icon from the label, like the normal window list does.
 * Properly center icons in vertical panels.

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -252,7 +252,16 @@ class AppGroup {
 
         this.setIcon();
         this.updateIconBoxClip();
+        this.setIconPadding();
         this.setMargin();
+    }
+
+    setIconPadding() {
+        if (this.state.isHorizontal) {
+            this.actor.style = 'padding-right: 0px; padding-left: 0px;';
+        } else {
+            this.actor.style = 'padding-bottom: 0px; padding-bottom: 0px;';
+        }
     }
 
     setMargin() {
@@ -350,9 +359,10 @@ class AppGroup {
     getPreferredWidth(actor, forHeight, alloc) {
         let [iconWidth] = this.iconBox.get_preferred_width(forHeight);
 
-        alloc.min_size = iconWidth;
+        // alloc.min_size = iconWidth; // future: use same allocation as the rest of the applets
+        alloc.min_size = this.state.trigger('getPanelHeight');
         if (this.isButtonIconOnly()) {
-            alloc.natural_size = iconWidth;
+            alloc.natural_size = alloc.min_size;
         } else {
             let [labelWidth] = this.label.get_preferred_width(forHeight);
             alloc.natural_size = Math.min(MAX_BUTTON_WIDTH, iconWidth + this.spacing + labelWidth);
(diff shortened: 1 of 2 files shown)


germanfr:
This will probably conflict with #8091 and #8084. They address the same issue partially but don't really fix it.

files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
```diff
@@ -103,6 +102,10 @@ class AppGroup {
         });
         this.actor._delegate = this;
 
+        this.spacing = 0;
+        this.actor.connect('style-changed',
+            () => this.spacing = this.actor.peek_theme_node().get_length('spacing'));
```
jaszhix: I think this is negatively affecting performance. I added a log here, and this signal fires on every allocation change. Input latency improves when I move this line:

```js
this.spacing = this.actor.peek_theme_node().get_length('spacing')
```

to `onIconBoxStyleChanged` - we already have `iconBox` connecting to `style-changed`, and its one too many IMHO - we have three other allocation signals, and they are firing at the same frequency.

We don't need to worry about theme changing because the `theme-set` signal callback does a destroy/create cycle.  You could add this to _init, and it should take care of it.

```js
setTimeout(() => this.actor.style_length('spacing'), 0);
```

`get_length` is a heavy call that starts a cascade of C function calls, and we shouldn't call it on every allocation. I've been trying to optimize their usage to mixed results. St.Widget.prototype.style_length from environment.js guards against finalized/unrealized actors, and in #8092 it caches the theme node. I think the best way forward is to only use theme introspection we need - #8091 and #8084 demonstrates none of this is necessary, at least right now.

I can't see a difference when I comment out spacing entirely - I'm looking for the things that I fixed in the PRs, and I'm not sure its worth the additional complexity here.

files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
```diff
@@ -353,35 +358,32 @@ class AppGroup {
...
-        alloc.min_size = iconNaturalSize + iconSpacing;
-        if (this.state.orientation === St.Side.TOP || this.state.orientation === St.Side.BOTTOM) {
-            let max = this.labelVisible && this.groupState.metaWindows.length > 0 ?
-                labelNaturalSize + iconNaturalSize + iconSpacing : 0;
-            alloc.natural_size = Math.min(iconNaturalSize + Math.max(max, labelNaturalSize), MAX_BUTTON_WIDTH * global.ui_scale);
+        let [iconWidth] = this.iconBox.get_preferred_width(forHeight);
+
+        // alloc.min_size = iconWidth; // future: use same allocation as the rest of the applets
+        alloc.min_size = this.state.trigger('getPanelHeight');
+        if (this.isButtonIconOnly()) {
```
jaszhix: This is calling a function to check three conditions that are already in state, in an allocation function that gets called rapidly depending on Clutter's frame rate. If we are going to check state in this function, it would be more optimal if it were a single state property, like `orientation` was.

files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
```diff
@@ -466,16 +472,14 @@ class AppGroup {
         if (this.label.text == null) {
             this.label.set_text('');
         }
-        // TODO: This should be set by the theme.
-        this.label.set_style('padding-right: 4px;');
 
         if (!animate) {
             this.label.show();
-            this.label.width = width;
+            this.actor.set_width(width);
```
jaszhix: Minor nitpick - in GI these properties have setters which call the same functions - I think its cleaner and more composable as prop assignment.

PR #8096, opened 2018-12-03
gwl: apply style class to the thumbnail menu

brownsr:
The style class was not being set, leading to the need for a subsequent workaround.  This change uses the function in the popup menu code designed to set a specific style class on a popup menu, and so means the intended dedicated CSS is now applied.

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/menus.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/menus.js
@@ -886,6 +886,7 @@ class AppThumbnailHoverMenu extends PopupMenu.PopupMenu {
         });
 
         this.appThumbnails = [];
+        this.setCustomStyleClass("grouped-window-list-thumbnail-menu");
     }
 
     onButtonPress() {
@@ -1052,7 +1053,6 @@ class AppThumbnailHoverMenu extends PopupMenu.PopupMenu {
             this.destroyThumbnails();
         }
         this.addWindowThumbnails(this.groupState.metaWindows);
-        this.setStyleOptions(false);
     }
 
     destroyThumbnails() {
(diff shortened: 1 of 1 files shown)


brownsr:
Will need some minor associated CSS changes to the core themes, e.g. Linux Mint has some beefy padding in grouped-window-list-thumbnail-menu which was being ignored.

files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/menus.js
```diff
@@ -1175,7 +1152,6 @@ class AppThumbnailHoverMenu extends PopupMenu.PopupMenu {
                 this.appThumbnails[i].thumbnailActor.realize();
```
jaszhix: This realize() call was a work-around for `setStyleOptions`, and could be removed. 

Thumbnail scaling is incorrect now - if you open enough thumbnails they clip the monitor area, probably will need to check padding in `refreshThumbnails`. That will result in more `get_length` calls, unless we figure out a way to do it once and not in every thumbnail.

brownsr:
Sounds like we leave this to be picked up for the next version then.

Regards

Simon

clefebvre:
@jaszhix @brownsr what's the status on this one?

jaszhix:
As far as I'm aware, this hinges on a `grouped-window-list-thumbnail-menu` class being added to themes. When I last tested this, padding was blown up on the thumbnail menu.

brownsr:
@clefebvre @jaszhix The CSS needed is already in the set that was introduced for gwl, that particular bit of CSS  was just not used, and over-written with defaults.  So if this change is pulled then it will have an effect dependent on what is in the current CSS for the theme.  Some themes (Mint-Y in particular) have quite a bit of padding there, which will want to be reduced.  That's pretty easy.  I've already modified quite a few of the spices themes to be well behaved when this change happens.  Our own core themes will want some minor changes, I'm rather expecting that will come my way ... As to what design effect this enables, well at the moment the thumbnails have a hard-coded frame around them which looks fine for themes with light menu borders and minimal menu corner rounding, but cumbersome for others.  I've tended to use the new CSS to get rid of the surrounding frame entirely for these themes, which looks better.  Themes not modified for the gwl CSS work fine.

Issue #8111, opened 2018-12-06
Workspace labels cut off in expo view when font scaling is more than 1.0


olfek:
```
 * Cinnamon version - 4.0.2
 * Distribution - Fedora 29
 * Arch - 64 bit
 ```

**Issue**
Workspace labels cut off in expo view when font scaling is more than `1.0`.

**Steps to reproduce**

1. Set font scaling to `1.5`
2. Go into expo view for workspaces

**Expected behaviour**
Workspace labels not cut off in expo view

**Other details**
Workspace labels also known as `OSD` in workspace system settings.

![screenshot from 2018-12-05 23-50-36](https://user-images.githubusercontent.com/27283110/49552102-05508980-f8ea-11e8-8eca-e341353d7143.png)
![screenshot from 2018-12-05 23-49-29](https://user-images.githubusercontent.com/27283110/49552103-05508980-f8ea-11e8-86aa-9798cc2fe4ea.png)

smurphos:
What theme are you using and where did you get it from?

olfek:
It's most likely because those text labels haven't been set to wrap content, they've been given a fixed height.

olfek:
It'd be nice to have this fixed before the next Cinnamon release, can't imagine it would be a big change.

smurphos:
Shouldn't be - it's a one liner to change the hard-coded height in px to em. Mint X suffers the same bug.

PR #8142, opened 2018-12-15
Stock theme - fix expo workspace entry not expanding to fit scaled fonts

smurphos:
Fixes https://github.com/linuxmint/Cinnamon/issues/8111

Also see https://github.com/linuxmint/mint-themes/pull/175


clefebvre:
@smurphos wouldn't 1em work better? It seems to fix the problem just as well, and without making the entry too big compared to the text.

smurphos:
Yep - you are right.

PR #8150, opened 2018-12-17
cs_calendar.py Port the remaining date/time settings to python

collinss:
Notable changes:
* The time and date widgets were all moved to their own settingswidgets, so the timezone map is now free to use the whole width. This also means that the window is no longer forced to expand horizontally.
* The lock button was removed. This is not a problem since the csd dbus backend prompts for a password as soon as the user trys to change something that needs elevated privileges, and holds on to those credentials for several minutes. The sytemd/timedated backend doesn't require elevated priviledges to work.
* The widgets for setting the date and time are now inside a revealer, so they are hidden when ntp is enabled, rather than just insensitive as they were before.
* A new TimeChooser widget was added for setting the time. The widget was also made available as part of the xlet settings api. The settings example applet, and the settings api reference documentation were updated to reflect this change.

note: the timezone map widget was replaced with the one from libtimezonemap1 - which is almost identical, but with introspection added so that we can use it in python.

image


clefebvre:
Looking at it in Mint, in debian/control, the dependency should be on gir1.2-timezonemap-1.0. I can't find that lib in Debian Stretch though... 

@leigh123linux do you have it in Fedora?

leigh123linux:
@clefebvre  I have

```
rpm -qf /usr/lib64/girepository-1.0/TimezoneMap-1.0.typelib
libtimezonemap-0.4.5.1-4.fc29.x86_64
```

clefebvre:
hmm I can't find it in Debian..

collinss:
> hmm I can't find it in Debian..

It looks like the one we have comes from Ubuntu repositories, so it might not be in Debian. I guess if we're going to use this we'd have to add it to our LMDE3 repository. On the upside, I don't think the package ever changes much, so there shouldn't be too much work involved...

clefebvre:


```diff

Issues
======
+ Solved 7
- Added 58
           

Complexity increasing per file
==============================
- files/usr/share/cinnamon/cinnamon-settings/modules/cs_calendar.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/Cinnamon/Cinnamon/pullRequest?prid=2743236&bid=10303860)

PR #8158, opened 2018-12-19
[WIP] Recorder - Improve declarations, avoid double using a function

brownsr:
This does not solve the current crash, but corrects some things that don't look right.


brownsr:
Closing, the other PR fixes the problem

PR #8161, opened 2018-12-20
gwl: Average the progress of all windows belonging to an app group button

jaszhix:
The progress code was originally based on the Cinnamon window list's code, but it didn't account for things GWL needs to account for, such as multiple windows with progress sharing a single button.

This also splits the progress allocation into a stand-alone function so we don't have to call `queue_relayout` when just the progress is changing, and fixes the progress not showing after the applet or Cinnamon is restarted.

This was tested by using the [xapp-gtk-window](https://github.com/linuxmint/xapps/blob/master/test-scripts/xapp-gtk-window) script in the xapps repository, clicking "Create shortcut" in GWL's context menu, and opening the script from the generated desktop file so it can be pinned/grouped.

Depends on #8160


clefebvre:
@jaszhix please rebase.

PR #8180, opened 2018-12-21
window-list: Prevent _addWindow from getting called before windows finish mapping

jaszhix:
This fixes up #8123. Some applications cause `setIcon` to recurse too much. 

The `window-monitor-changed` signal is emitted before CinnamonWindowTracker has been notified via workspace signals, and as a result the CinnamonApp object doesn't yet exist for the window at this point.

Tested per-monitor window support on three panels on different monitors, and not seeing regressions. This includes calling `move_to_monitor` on a minimized window.

C/O @itzexor


buzo-ffm:
See also: https://bugs.archlinux.org/task/61292

PR #8192, opened 2018-12-21
gwl: Fix incorrect window accounting in multi-workspace cases

### LINKED ISSUE #8190: GWL - moved apps don't appear as open when moved to another workspace
```
 * Cinnamon version 4.08
 * Distribution - Mint 19.1
 * nvidia gtx 760 driver 390.77
 * 64 bit

[xsession-errors.zip](https://github.com/linuxmint/Cinnamon/files/2703800/xsession-errors.zip)


 
 * Attach /home/<username>/.xsession-errors, or /var/log/syslog
 ```

**Issue**
If any application is moved to another workspace than the workspace it is spawned in, then I'm not able to get a preview of that application, when hovering on applications icon in taskbar in other work spaces than the workspace it has been moved to.
It doesn't appear to be active either - unless switching to the actual workspace it has been moved to.

**Steps to reproduce**
open any program - move program to another work space, and watch the active indicator for that program disappears in the taskbar, and hovering doesn't produce a miniature.

**Expected behaviour**
Like other programs in several work spaces: To show miniature when mouse pointer is hovering.
When the function "show windows from all work spaces" is on that is.

**Other information**
**Same behaviour is confirmed on a live cd Linux Mint 19.1 Cinnamon 64 on another pc.**

Also some programs for example Libreoffice Writer icon will stick to task bar even after program is closed. Only way to remove sticky unpinned icon is to restart Cinnamon.

Some programs when moved will show up in taskbar as active in some work spaces  and not in others.

Resetting Cinnamon clears all issues!

### LINKED ISSUE #8252: GWL - Previes grafics destroyed (Calc - After resume)
```
 * Cinnamon 4.0.8
 * Mint 19.1
 * Graphics hardware *and* driver used
    Device-2: NVIDIA GP107 [GeForce GTX 1050 Ti] driver: nvidia v: 390.87
 * 64 bit
 * Attach /home/<username>/.xsession-errors, or /var/log/syslog
 ```

**Issue**
On hovering the icon the shown grafic is destroyed. Other programs show the right preview

**Steps to reproduce**
Not sure. Propabli has to do with Suspend to RAM and Resume.

**Expected behaviour**
Should always work as expected

**Other information**
menu_20190103_09 40 52

jaszhix:
- Fixes windows not being removed from all workspaces when "Show windows from all workspaces" is enabled.
- Fixes the 'insensitive' workspace option in the context menu not being based on the workspace the last focused window is actually on.
- Removes property assignment on the MetaWindow object - state props should stay localized to the applet.
- Fixes windows being removed from the active workspace when a window's
"Visible on all workspaces" option is toggled while "Show windows from all workspaces" is enabled.
- Fixes null MetaWindowActor TypeError in `AppMenuButtonRightClickMenu.prototype.populateMenu`.

Closes #8190 
Closes #8252

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appList.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appList.js
@@ -366,8 +366,11 @@ class AppList {
         if (!this.state) return;
 
         if ((metaWindow.is_on_all_workspaces() || this.state.settings.showAllWorkspaces)
-            && !metaWindow.__gwlFinalize__) {
-            metaWindow.__gwlFinalize__ = true;
+            && !this.state.removingWindowFromWorkspaces) {
+            // Abort the remove if the window is just changing workspaces, window
+            // should always remain indexed on all workspaces while its mapped.
+            if (!metaWindow.showing_on_its_workspace()) return;
+            this.state.removingWindowFromWorkspaces = true;
             this.state.trigger('removeWindowFromAllWorkspaces', metaWindow);
             return;
         }
@@ -387,6 +390,7 @@ class AppList {
                 return false;
             }
         });
+
         if (refApp > -1) {
             this.appList[refApp].windowRemoved(metaWorkspace, metaWindow, refWindow, (appId, isFavoriteApp) => {
                 if (this.state.settings.titleDisplay > 1) {
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appList.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appList.js
@@ -239,8 +239,11 @@ class AppList {
 
         if (metaWindow && !this.shouldWindowBeAdded(metaWindow)) return;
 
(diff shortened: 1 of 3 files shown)


fusentast:
Does "In Progress" mean that it's an ongoing process that  will resolve the issues here getting solved?

I'm asking because I can see that the corresponding files have been changed several times, and the latest Cinnamon update of today version 4.0.10 doesn't resolve some of the issues here (#8190), providing the use of multiple work spaces and "show windows from all work spaces" are ticked.

jaszhix:
The fix is on master (will be in 4.2), but isn't on the maintenance branch. I would ping @clefebvre or @mtwebster since they usually handle those.

fusentast:
> The fix is on master (will be in 4.2), but isn't on the maintenance branch. I would ping @clefebvre or @mtwebster since they usually handle those.

Thank you for your reply jaszhix.
Not being impatient, I was wondering because I've just started to get involved in  reporting bugs at Github, and I really try my best to be constructive - so no criticism from my point of view.

The hole Mint team is doing a tremendous job in developing and maintaining Linux Mint.

Issue #8166, opened 2018-12-20
Menu applet - search: name field prioritization not working or too limited


s44i7h-sam:
```
 * Cinnamon version 4.0.8
 * Distribution - Mint 19.1
 ```

**Issue**
Menu applet does not prioritize applications whose name contains in some way the searched pattern or word. Instead it threats the `Name=` field the same as any other searched field like `Keywords=`.

**Steps to reproduce**
Open the menu applet and watch the results as you type *`s` `t` `a` `r` `t`*. "LibreOffice" is shown at the top of the list instead of the more obvious "Startup Applications" whose name actually matches the regex pattern `\bstart`.
Another example: type *`s` `c` `r` `e` `e` `n` `s`* and watch the results as you type - "Screensaver" and "Screenshot" remain at the bottom of the list.
The issue is the same even if the whole word is typed out. I have a custom .desktop file which includes the whole word "def" in its `Name=` field. When typing *`d` `e` `f`* "Preferred Applications" is shown first because its `Keywords=` field contains "default". To reproduce this add the word "def" to some .desktop file's `Name=` field.

**Expected behaviour**
Those applications which contain the searched pattern in the name itself should be shown first. The applet should perform (on the `Name=` field) a regex search (not regex match!) for `\b${pattern}` (start of *any* word in the name).

**Other information**
This issue has to do with the `_listApplications` function in the `/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js` file.
The LM 18.3 version of this applet seemed to work better in this regard, so a look at recent changes of the `_listApplications` function could be helpful.

s44i7h-sam:
After playing a bit with the code in the `_listApplications` function I realised that this function itself already works somewhat according to the expected behavior (see above). The name results are added to the `res` array first, then other fields are searched for, going through each application again, with the effect that `res` can contain duplicates (in case both name and some other field of the same application contain the pattern).

---

Here's the updated report (steps to reproduce are the same as above):

**Issue**
The problem is rather in the `_displayButtons` function where the order of the `res` array (now named `apps`) is not maintained because the function merely goes through the whole application list in the default (alphabetical) order and calls show() and hide() for each application button based on whether that application is in the `apps` array.

**Expected behaviour**
[...]

Issue #8070, opened 2018-11-26
Menu search returns unrelated results [Regression]


LingMan:
```
 * Cinnamon version: 4.0.1
 * Distribution: Manjaro
 * Graphics hardware *and* driver used: R9 390 with Kernel 4.18 and Mesa 18.2.5
 * 64 bit
 ```

**Issue**
Menu search lists unexpected items.

**Steps to reproduce**
Search for "cal".

**Expected behaviour**
The results are
- Calculator
- LibreOffice Calc
- maybe other relevant elements

**Actual behaviour**
The results are
- Calculator
- Color
- Date & Time
- Hot Corners
- LibreOffice Calc
...

**Other information**
This is a regression since Cinnamon 3.8. It feels like whatever fixed https://github.com/linuxmint/Cinnamon/issues/4053 got removed again.

Xeronel:
I can confirm that I am also experiencing this issue. It is particularly bad for me when I type "ter" and get things like Skype, Online Accounts, Network, etc and my expected items like "Terminal" are way down at the bottom of all the unrelated items.

Curly060:
I can confirm this, too. Commit a849090d94aa6104b638aba5150f5c317ded5198 is the culprit. That commit totally negates my original pull request https://github.com/linuxmint/Cinnamon/pull/6929.
@clefebvre Could you please have a look at this? I know this search thing is a looong ongoing story, but now we are really back to square one.

jarq6c:
Just installed LM 19.1 and this problem persists. Search worked perfectly in 18.3. I expect to see Firefox and Firewall as the top search results if I type "f" or "fi", but this is not the case until "fir" is typed. If you type "f" or "fi", Archive Manager is the top search result.

icarter09:
This issue is directly related to #8166, which explains why the search is actually returning correct results, but not in the intended affect that a search should be or how a user would anticipate the search to work.

oles:
Cinnamon 4.0.8

![screenshot from 2019-01-02 10-41-11](https://user-images.githubusercontent.com/7634050/50586728-22419380-0e7b-11e9-914e-2cc82b70231d.png)

:confused:

Curly060:
I have made another attempt at fixing this. This is still not perfect, but hopefully it respects all wishes I have seen so far with the menu search:
- by default the search only matches application names and only at word boundaries
[...]

PR #8199, opened 2018-12-23
Menu Applet: Fixed issue of menu search not matching expected results

icarter09:
This PR is for issue #8070 and #8166. 

I removed logic that was attempting to match the search on values that are not apparent to a user. If anything needs to be changed or updated, please let me know. 

If this PR is accepted, shout out to @s44i7h-sam who helped troubleshoot the issue and show me where the issue was occurring.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -3261,10 +3261,7 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
             if (!exactMatch) {
                 for (let i in this._applicationsButtons) {
                     let app = this._applicationsButtons[i].app;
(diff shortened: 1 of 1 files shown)


s44i7h-sam:
I'm glad for the help but I think you've misunderstood what the issue is about. The search *does* return the expected results. But it displays them in an alphabetical order instead of putting name matches at the top.
This patch merely limits the second search loop to the name field. I would consider this to be a regression since users rely on searching the other fields as well. For example, if I type "screen" it is expected that "Power Management" should be included in the results. However "Screensaver" and "Screenshot" should be displayed at the top of the results.

icarter09:
@s44i7h-sam Not sure about it being a regression since the only thing being displayed to the user is the application name. If the menu search is going to be searching on more than the name, then maybe the menu search should inform the user of keyword matches v/s name matches. Not informing a user that the search is going beyond the name makes it appear that the search is broken.

PR #8230, opened 2018-12-30
Replace allocation signals with callbacks

jaszhix:
This replaces #8068. This is loosely based on the idea of this [upstream MR](https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/153). 

>  Signals are not specially performant. One of the reasons is that they acquire a global lock for signal handlers lookup.

After attempting to expose vfunc callbacks in #8068, the main issues I discovered had to do with CJS/GJS handling of vfuncs. On the mozjs60 branch random crashing occurred that didn't exist on cjs master, and there's a [long-standing bug](https://gitlab.gnome.org/GNOME/gjs/issues/89) that causes inherited classes' replaced vfuncs to replace the base prototype's. Another issue is doing this effectively moves the performance sensitive base allocation logic from C to the JS context where it can't perform as well, and as a result there is worse performance.

Instead of exposing GObject classes in JS, what this does is keeps everything where it is, but adds callback support to CinnamonGenericContainer in C. When a callback is set for an allocation signal, it replaces the signal's emission. I've found this solution to perform better so far in my casual testing.

### DIFF
--- a/files/usr/share/cinnamon/applets/calendar@cinnamon.org/calendar.js
+++ b/files/usr/share/cinnamon/applets/calendar@cinnamon.org/calendar.js
@@ -214,7 +214,7 @@ class Calendar {
                        {row: 0, col: 0, col_span: offsetCols + 3});
         }
 
-        this.actor.connect('style-changed', Lang.bind(this, this._onStyleChange));
+        this.actor.set_style_changed_callback(() => this._onStyleChange());
 
         let back = new St.Button({ style_class: 'calendar-change-month-back' });
         this._topBoxMonth.add(back);
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -95,6 +95,10 @@ class AppGroup {
             can_focus: true,
             track_hover: true
         });
+        this.actor.set_allocation_callback((b, f) => this.allocate(b, f))
+        this.actor.set_preferred_width_callback((w) => this.getPreferredWidth(w))
+        this.actor.set_preferred_height_callback((h) => this.getPreferredHeight(h));
+
         this.actor._delegate = this;
 
         this.progressOverlay = new St.Widget({
@@ -149,9 +153,6 @@ class AppGroup {
 
         this._draggable = new DND._Draggable(this.actor);
 
-        this.signals.connect(this.actor, 'get-preferred-width', (...args) => this.getPreferredWidth(...args));
-        this.signals.connect(this.actor, 'get-preferred-height', (...args) => this.getPreferredHeight(...args));
-        this.signals.connect(this.actor, 'allocate', (...args) => this.allocate(...args));
         this.signals.connect(this.actor, 'enter-event', (...args) => this.onEnter(...args));
         this.signals.connect(this.actor, 'leave-event', (...args) => this.onLeave(...args));
         this.signals.connect(this.actor, 'button-release-event', (...args) => this.onAppButtonRelease(...args));
@@ -338,9 +339,10 @@ class AppGroup {
         }
     }
 
-    getPreferredWidth(actor, forHeight, alloc) {
-        let [iconMinSize, iconNaturalSize] = this.iconBox.get_preferred_width(forHeight);
-        let [labelMinSize, labelNaturalSize] = this.label.get_preferred_width(forHeight);
+    getPreferredWidth(alloc) {
+        let {for_size} = alloc;
+        let [iconMinSize, iconNaturalSize] = this.iconBox.get_preferred_width(for_size);
+        let [labelMinSize, labelNaturalSize] = this.label.get_preferred_width(for_size);
         let {iconSpacing} = this.state.settings;
         // The label text starts in the center of the icon, so we should allocate the space
         // needed for the icon plus the space needed for(label - icon/2)
@@ -354,14 +356,15 @@ class AppGroup {
         }
     }
 
-    getPreferredHeight(actor, forWidth, alloc) {
-        let [iconMinSize, iconNaturalSize] = this.iconBox.get_preferred_height(forWidth);
-        let [labelMinSize, labelNaturalSize] = this.label.get_preferred_height(forWidth);
+    getPreferredHeight(alloc) {
+        let {for_size} = alloc;
+        let [iconMinSize, iconNaturalSize] = this.iconBox.get_preferred_height(for_size);
+        let [labelMinSize, labelNaturalSize] = this.label.get_preferred_height(for_size);
         alloc.min_size = Math.min(iconMinSize, labelMinSize);
         alloc.natural_size = Math.max(iconNaturalSize, labelNaturalSize);
     }
 
-    allocate(actor, box, flags) {
+    allocate(box, flags) {
         let allocWidth = box.x2 - box.x1;
         let allocHeight = box.y2 - box.y1;
         let childBox = new Clutter.ActorBox();
--- a/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
@@ -242,7 +242,7 @@ class AppMenuButton {
         });
         this.actor.set_allocation_callback((b, f) => this._allocate(b, f))
         this.actor.set_preferred_width_callback((a) => this._getPreferredWidth(a))
-        this.actor.set_preferred_height_callback((a) => this._getPreferredWidth(a));
+        this.actor.set_preferred_height_callback((a) => this._getPreferredHeight(a));
 
         this._applet = applet;
         this.metaWindow = metaWindow;
--- a/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
@@ -238,7 +238,11 @@ class AppMenuButton {
             style_class: 'window-list-item-box',
             reactive: true,
             can_focus: true,
-            track_hover: true });
+            track_hover: true
+        });
+        this.actor.set_allocation_callback((b, f) => this._allocate(b, f))
+        this.actor.set_preferred_width_callback((a) => this._getPreferredWidth(a))
+        this.actor.set_preferred_height_callback((a) => this._getPreferredWidth(a));
 
         this._applet = applet;
         this.metaWindow = metaWindow;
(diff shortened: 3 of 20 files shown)


lestcape:
@jaszhix Hi. I don't know what you understand really, but in the case you are lost with the beautiful ideas about the performance point in the [upstream MR](https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/153), please forget it, because this issue is about remove an oldest patch and not about performance.

The GenericContainer was introduced just because the impossibility of to have a good inheritance in Gjs classes and when that issue was fixed in gjs, they want then to remove also the patch that they introduce before (the GenericContainer) in concordance. So, any solution that do not remove the usage of the GenericContainer, is not in the same direction of the **upstream MR**. 

Also please note that, remove the GenericContainer is not the only important point of this **upstream MR**, because the idea is also remove the way of how you prototypes classes. Actually is something like that:

```classWapper {  actor: ClutterActor }```

And the idea is remove the wrapper of the actor and make the actor also the class:

```classActor extends ClutterActor (...}```

So yes, this is a big change in the conception of how we can prototype classes and make inheritance, but the performance is not the point of this at all, just probably a side effect that you need to test.

Please see all this need a lot of changes internally and yes it can produce a lot of new crashes if that big change will not come all at the same time in all areas. Nothing of this means you are wrong in something, just be carefully because is really big and is more conceptual than practical.

**Comment by jaszhix** [comment]
> So, any solution that do not remove the usage of the GenericContainer, is not in the same direction of the upstream MR.

I said it is "loosely" based on it. The one area this overlaps is reducing usage of signals for allocation. I am seeing less micro-stutter on this branch than on master, so I think it's worth it.

There is no point in switching all of our classes to GObject classes, it will introduce more overhead because each time a class is constructed the monkey patching override functions in gjs will be invoked. It makes the prototype more complex and in most cases isn't necessary. From what I've seen working with GI in JS, bugs seem to occur more when you start blurring the lines between JS objects and GObjects, so it seems safer to keep those types separate. 

There are also implementation bugs that need to be resolved before vfuncs can be used in JS without side effects, see the GJS issue I linked to in the original PR message. I understand why GNOME went in this direction, it's their binding library, and it's the way it's supposed to be used in any language using GI. In my opinion however, this results in worse performance, and for the reasons mentioned in my PR message - it is moving allocation code from GenericContainer to JS, which is slower.

**Comment by jaszhix** [comment]
I don't think it's a conspiracy - I tried this approach as a wrapper solution in #8068 and it didn't work out. I am discussing what makes sense for _Cinnamon_ from a JS developer's perspective. For me, this is a performance issue, and I'm not interested in what GNOME thinks is the right way - especially when it doesn't follow ECMAScript standards. There are caveats that I see affecting any approach that sub-classes GObjects in JS with GJS in its current state, and they can't be ignored when talking about this.

**Comment by jaszhix** [comment]
> The author of the pull is not a gnome developer

His name appears on the [members page](https://www.gnome.org/foundation/membership/), and he has push access. 

What I meant was, replacing  CinnamonGenericContainer means tolerating [this bug](https://gitlab.gnome.org/GNOME/gjs/issues/89), which has performance ramifications for all actors, because it causes all actors' vfuncs to be overridden. This causes all allocation to occur in JS. That is not how JS prototype inheritance works.

**Comment by lestcape** [comment]
> What I meant was, replacing CinnamonGenericContainer means tolerating this bug.

This bug also occurs in Gtk and have a workaround, i think is worse the continuation of the usage of the GenericContainer, like also the gnome developers think, because the pull was merged.

What is an important bug that can give you the reason, is the fact that there are not way to unregistered a GObject class (https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/153#note_307911). So, load and reload a js module (extensions), will required a restart always. That is in my opinion really problematic. What you mention is not a critical bug to me, but if you think it is, that ok.

**Comment by jaszhix** [comment]
> This bug also occurs in Gtk and have a workaround

This affects all bindings, and I don't think the workaround makes much sense in terms of JS classes.

> So, load and reload a js module (extensions), will required a restart always

Cinnamon has its own file loader for extensions, so this isn't an issue.

CinnamonGenericContainer is just another Clutter actor with some hooks for allocation. There's nothing inherently wrong with it. If GNOME wants to get rid of it, fine. We don't necessarily have to, that's why Cinnamon is a fork. We can take a step back and look at some of these decisions more objectively.

**Comment by lestcape** [comment]
> I don't think the workaround makes much sense in terms of JS classes

> CinnamonGenericContainer is just another Clutter actor with some hooks for allocation. There's nothing inherently wrong with it

In my opinion, the GenericContainer workaround it's worse because it's a shell class and that mean you can not port a code when it use a GenericContainer to outside a shell context. This is in my opinion a very bad thing, because in the end  you are making code that can only be use in a pretty specific context and the interoperability is just fiction in that case. Gjs works outside any shell, also Clutter and you can make St work too.

> Cinnamon has its own file loader for extensions, so this isn't an issue

Read the link again, a loader can not avoid a GObject issue. In fact a loader have nothing to be with how the introspection works.

**Comment by lestcape** [comment]
> I don't want to debate abstract design preferences over a practical performance patch.

Well the upstream issue is all about this. Sorry if this is not important to you.

**Comment by brownsr** [comment]
I am finding a lot of issues, but as far as I can tell they are there on master too.  However the log is showing a number of callback related issues, which as far as I can tell are just related to this. I've removed some familiar caribou messages.

```
(cinnamon:1777): Cjs-WARNING **: 19:44:55.290: JS ERROR: Exception in callback for signal: open-state-changed: TypeError: this._applicationsButtons[i].actor is undefined
_onOpenStateChanged@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:1142:17
_emit@resource:///org/cinnamon/cjs/modules/signals.js:142:27
open@/usr/share/cinnamon/js/ui/popupMenu.js:2315:9
toggle_with_options@/usr/share/cinnamon/js/ui/popupMenu.js:2064:13
on_applet_clicked@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:1107:9
_onButtonPressEvent@/usr/share/cinnamon/js/ui/applet.js:278:13


(cinnamon:1777): Clutter-CRITICAL **: 19:44:55.305: clutter_layout_manager_allocate: assertion 'CLUTTER_IS_LAYOUT_MANAGER (manager)' failed

(cinnamon:1777): Clutter-CRITICAL **: 19:44:55.320: clutter_layout_manager_allocate: assertion 'CLUTTER_IS_LAYOUT_MANAGER (manager)' failed


Cjs-Message: 19:50:01.679: JS LOG: [LookingGlass/info] this.actor is undefined
Cjs-Message: 19:50:01.680: JS LOG: [LookingGlass/trace] 
<----------------
destroy@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:238:13
_refreshApps@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:2109:17
_refreshAll@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:1001:13
anonymous/onAppSysChanged/<@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:995:51
---------------->
Cjs-Message: 19:50:03.570: JS LOG: [LookingGlass/info] this.actor is undefined
Cjs-Message: 19:50:03.570: JS LOG: [LookingGlass/trace] 
<----------------
destroy@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:238:13
_refreshApps@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:2109:17
_refreshAll@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:1001:13
anonymous/onAppSysChanged/<@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:995:51
---------------->
Cjs-Message: 19:50:05.569: JS LOG: [LookingGlass/info] this.actor is undefined
Cjs-Message: 19:50:05.570: JS LOG: [LookingGlass/trace] 
<----------------
destroy@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:238:13
_refreshApps@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:2109:17
_refreshAll@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:1001:13
anonymous/onAppSysChanged/<@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:995:51
---------------->
Cjs-Message: 19:50:10.570: JS LOG: [LookingGlass/info] this.actor is undefined
Cjs-Message: 19:50:10.571: JS LOG: [LookingGlass/trace] 
<----------------
destroy@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:238:13
_refreshApps@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:2109:17
_refreshAll@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:1001:13
anonymous/onAppSysChanged/<@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:995:51
---------------->
Cjs-Message: 19:50:18.577: JS LOG: [LookingGlass/info] this.actor is undefined
Cjs-Message: 19:50:18.578: JS LOG: [LookingGlass/trace] 
<----------------
destroy@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:238:13
_refreshApps@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:2109:17
_refreshAll@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:1001:13
anonymous/onAppSysChanged/<@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:995:51
---------------->
Cjs-Message: 19:50:21.569: JS LOG: [LookingGlass/info] this.actor is undefined
Cjs-Message: 19:50:21.570: JS LOG: [LookingGlass/trace] 
<----------------
destroy@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:238:13
_refreshApps@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:2109:17
_refreshAll@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:1001:13
anonymous/onAppSysChanged/<@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:995:51
---------------->

(cinnamon:1777): Cjs-WARNING **: 19:50:24.044: JS ERROR: Exception in callback for signal: changed: TypeError: this.actor is undefined
destroy@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:238:13
_refreshRecent@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:2043:13
_emit@resource:///org/cinnamon/cjs/modules/signals.js:142:27
DocManager.prototype._reload@/usr/share/cinnamon/js/misc/docInfo.js:84:9


(cinnamon:1777): Cjs-WARNING **: 19:50:25.959: JS ERROR: Exception in callback for signal: changed: TypeError: this.actor is undefined
destroy@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:238:13
_refreshRecent@/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js:2043:13
_emit@resource:///org/cinnamon/cjs/modules/signals.js:142:27
DocManager.prototype._reload@/usr/share/cinnamon/js/misc/docInfo.js:84:9

Cinnamon warning: Log level 16: invalid (NULL) pointer instance
Cinnamon warning: Log level 8: g_signal_emit_valist: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed
Cinnamon warning: CurrentTime used to choose focus window; focus window may not be correct.
Cinnamon warning: Got a request to focus the no_focus_window with a timestamp of 0.  This shouldn't happen!
Cinnamon warning: Log level 8: Source ID 247 was not found when attempting to remove it
```

jaszhix:
Those callbacks are not changed in this PR. I can fix them up, but it might be better in a separate PR. This PR converts `style-changed`, `allocate`, `get-preferred-width`, and `get-preferred-height`. In fact, I'm not seeing anywhere in the menu applet affected by these because it doesn't use `Cinnamon.GenericContainer`.

PR #8235, opened 2018-12-30
[WIP] st-private: upstream shadow changes

brownsr:
3 upstream commits.  First a simple test and skip over an error condition.  I've also pushed this separately as it is a simple bugfix suppressing a frequent warning.  Then there is a change to ensure that the right sized shadow is used.  This appeared to have a side-effect, as it was later changed to avoid using the allocation box.


brownsr:
Needs further checking out.

PR #8236, opened 2018-12-30
[WIP] main: Disable unnecessary Clutter events while dragging windows

jaszhix:
This uses `MetaDisplay`'s grab-op signals to disable most actors' reactivity during dragging. The overlay actors remain reactive so the tiling/snapping OSD continues to work. The `position-changed` signal is unaffected - the workspace switcher's window graph and panel auto-hiding modes are tested and working correctly.

`disableReactivityRecursive` was intentionally left in the `start` function's scope so it remains private.

### DIFF
--- a/js/ui/main.js
+++ b/js/ui/main.js
@@ -541,6 +541,7 @@ function start() {
         });
         global.display.connect('grab-op-end', function() {
             Util.each(reactive, function(actor) {
+                if (actor.is_finalized()) return;
                 actor.reactive = true;
             });
             reactive = [];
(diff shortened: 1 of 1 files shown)


clefebvre:
I get a performance loss with this PR.

When I drag windows, I like to move the cursor before I actually click the button, so I catch the window on the way if that makes sense.

With that change I get a skipped frame or some kind of lag as I'm getting the window. Going back to master and it's smooth again.

jaszhix:
I think I know what you're talking about, but I can't reproduce it reliably. When I go back to master windows don't move as quickly 100% of the time. Maybe deferring the actor reactivity, or reducing how many actors we switch reactive could work, though I don't see this problem being very noticeable unless you're looking for it, if we're talking about the same thing.

jaszhix:
After more comparison, there does seem to be an issue here with skipped frames. Closing in favor of making MetaWindowActor non-reactive by default in a later PR.

PR #8245, opened 2019-01-01
Improve property access performance

jaszhix:
This is an optimization intended for V8, but it [also works on Spidermonkey](http://jsben.ch/1DVog). It modifies the properties of GObject instances so the engine uses its optimized path for property access. [More info](https://stackoverflow.com/questions/24987896/how-does-bluebirds-util-tofastproperties-function-make-an-objects-properties).

I got the idea to use this when I was trying to optimize #8068 and to address a performance regression with [CJS built with mozjs60](https://github.com/linuxmint/cjs/pull/70). In any case, it has a noticeable performance improvement for me, most notably with the responsiveness of electron apps like VSCode.

### DIFF
--- a/js/misc/util.js
+++ b/js/misc/util.js
@@ -655,7 +655,7 @@ const FastObject = function(o) {
     return new FastObject;
 }
 FastObject();
-function toFastproperties(obj) {
+function toFastProperties(obj) {
     each(obj, function(value) {
         if (value && !Array.isArray(value)) FastObject(value);
     });
--- a/js/misc/util.js
+++ b/js/misc/util.js
@@ -642,6 +642,25 @@ function unref(object, reserved = []) {
     }, 0);
 };
 
+// MIT © Petka Antonov, Benjamin Gruenbaum, John-David Dalton, Sindre Sorhus
+// https://github.com/sindresorhus/to-fast-properties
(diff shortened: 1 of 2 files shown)


clefebvre:
I can't see any difference here, how do you test the perf increase?

clefebvre:
OK, and empirically, I mean, where/how do you see a difference, yourself when using Cinnamon?

jaszhix:
Mainly with VSCode, an easy way for me to tell how far window rendering is behind input is by highlighting text quickly. The cursor should be aligned with highlighted text, but most of the time it isn't. It's not very scientific this way, but it stays perfectly in sync in metacity, which is a baseline I use.

What this patch is doing is iterating GObject property keys set in JS and applying the work-around after the class is initialized - Object.keys() will not return any native GI properties - that is where more attention is needed. We often will reference props to a GObject for state tracking, and remove them later. This makes object access slow, and slows down everything it is being used in, which often times can span every JS module in Cinnamon depending on where they're used.

The improvement was more apparent for me when using #8010 with the [mozjs60 CJS PR](https://github.com/linuxmint/cjs/pull/70). For reasons unknown to me still (and I've tested with GJS + Cinnamon as well), there's a performance hit, and this patch helps a little.

PR #8251, opened 2019-01-03
fix regression #8070 (menu search results)

Curly060:
This commit tries to remedy the most pressing points
with the menu search for everyone. The changes are:
- by default only search in application names at word boundaries
- added a config switch to also search in keywords (keyword startsrWith)
- added a config switch to also search in description (at word
boundaries)
This is still not the desired behaviour, but certainly the best
compromise so far.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -3041,27 +3041,25 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
     _listApplications(pattern){
         let res = [];
         let exactMatch = null;
-        if (pattern){
-            res = [];
+        if (pattern) {
             let regexpPattern = new RegExp("\\b"+pattern);
-            for (let i in this._applicationsButtons) {
-                let app = this._applicationsButtons[i].app;
+            Util.each(this._applicationsButtons, (button) => {
+                let app = button.app;
                 let latinisedLowerName = Util.latinise(app.get_name().toLowerCase());
                 if (latinisedLowerName.match(regexpPattern) !== null) {
                     res.push(app.get_id());
                     if (!exactMatch && latinisedLowerName === pattern)
                         exactMatch = app.get_id();
                 }
-            }
+            });
             if (!exactMatch) {
-                for (let i in this._applicationsButtons) {
-                    let app = this._applicationsButtons[i].app;
-                    if (Util.latinise(app.get_name().toLowerCase()).indexOf(pattern)!==-1 ||
-                        (app.get_keywords() && Util.latinise(app.get_keywords().toLowerCase()).indexOf(pattern)!==-1) ||
-                        (app.get_description() && Util.latinise(app.get_description().toLowerCase()).indexOf(pattern)!==-1) ||
-                        (app.get_id() && Util.latinise(app.get_id().slice(0, -8).toLowerCase()).indexOf(pattern)!==-1))
+                Util.each(this._applicationsButtons, (button) => {
+                    let app = button.app;
+                    if ((app.get_keywords() && Util.latinise(app.get_keywords().toLowerCase()).split(';').some(keyword => keyword.startsWith(pattern))) ||
+                        (app.get_description() && Util.latinise(app.get_description().toLowerCase()).match(regexpPattern) !== null)) {
                         res.push(app.get_id());
-                }
+                    }
+                });
             }
         }
         return [res, exactMatch];


files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -3072,18 +3075,17 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
                 });
             }
         } else if (apps) {
-            for (let i = 0; i < this._applicationsButtons.length; i++) {
-                let button = this._applicationsButtons[i];
+            this._applicationsButtons.forEach(button => {
                 let appId = button.app.get_id();
-                if (apps.indexOf(appId) !== -1) {
+                if (apps.has(appId)) {
```
jaszhix: There is no Array.prototype.has, you probably meant to use `includes`.

`Cjs-Message: 19:38:16.555: JS WARNING: [/usr/share/cinnamon/js/ui/lookingGlass.js line 127 > eval 1]: reference to undefined property "has"`

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -3072,18 +3075,17 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
                 });
             }
         } else if (apps) {
-            for (let i = 0; i < this._applicationsButtons.length; i++) {
-                let button = this._applicationsButtons[i];
+            this._applicationsButtons.forEach(button => {
```
jaszhix: `Array.prototype.forEach` is slower than for loops.

You can try [these](https://jsperf.com/for-vs-foreach/75) [tests](https://jsperf.com/loops/33) in Firefox, which uses the same JS engine as Cinnamon.

An alternative is the the `each` function in misc/util, which wraps a for loop.

```js
Util.each(this._applicationsButtons, (button) => {...})
```

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -3238,37 +3240,29 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
...
-            if (!exactMatch) {
-                for (let i in this._applicationsButtons) {
-                    let app = this._applicationsButtons[i].app;
-                    if (Util.latinise(app.get_name().toLowerCase()).indexOf(pattern)!==-1 ||
-                        (app.get_keywords() && Util.latinise(app.get_keywords().toLowerCase()).indexOf(pattern)!==-1) ||
-                        (app.get_description() && Util.latinise(app.get_description().toLowerCase()).indexOf(pattern)!==-1) ||
-                        (app.get_id() && Util.latinise(app.get_id().slice(0, -8).toLowerCase()).indexOf(pattern)!==-1))
-                        res.push(app.get_id());
-                }
+            if (this.searchInKeywords && app.get_keywords() && Util.latinise(app.get_keywords().toLowerCase()).split(';').some(keyword => keyword.startsWith(pattern))) {
```
jaszhix: This could probably be broken down into multiple lines, it's not easy to read this way.

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
Curly060: Huh? Why would it think this is an Array? I have changed the type of apps from Array to Set so I don't understand this warning.

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
Curly060: forEach is used all over the place in this file. We are talking about the menu with, say, max. 1000 entries (and that would be a lot). So IMHO it is fine to use forEach here.

NikoKrause:
This commit adds new strings, which need to be translated, so IF it would be merged, it won't be before Mint 19.2.
And for Mint 19.2 I'd like to see the search feature from @germanfr implemented into the menu applet.
It's pretty great.

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
jaszhix: Ah, I see - still doesn't seem necessary - arrays are faster.

Curly060:
> This commit adds new strings, which need to be translated, so IF it would be merged, it won't be before Mint 19.2.
> And for Mint 19.2 I'd like to see the search feature from @germanfr implemented into the menu applet.
> It's pretty great.

Where can I find this search feature? I would like to give it a try. And I am more than happy if a better solution than my patch exists. I am now running a patched version which fixes theissue for me, so I can wait.

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
Curly060: @LingMan The current logic only allows to show/hide an item, so ordering is not possible at all. In that case, a Set felt semantically the most correct choice. Also, I just assumed that a Set is naturally the fastest in looking up items. That's not really the case in JS...

Curly060:
Even though I know there seems to be something much better on the way in possibly the next major release, I need the search to be usable now, so I have made one last attempt to fix it. I have updated my PR accordingly.
No new settings and strings, no translations needed and IMHO now it is quite usable again and it should fit almost everyone's wishes (within the constraints of the current implementation).

The main change is that I have removed any kind of fulltext search inside app names, keywords or description. Instead, the pattern now must always match at a word boundary (either with regex match or .startsWith).

=> This is usable for me now. Please have a look if that can be applied.

Curly060:
I have updated my PR against current master.

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -3037,27 +3037,25 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
     _listApplications(pattern){
         let res = [];
         let exactMatch = null;
-        if (pattern){
-            res = [];
+        if(pattern) {
```
LingMan: Nitpick: Missing space between `if` and `(pattern)`.

LingMan: Nitpick: Missing space between `if` and `(pattern)`.

Curly060: fixed

clefebvre:
I think @itzexor has something big in store which should supersede this, where all matches are visible but sorted according to their type of match. It's not in a PR though atm, James can you confirm?

Curly060:
I am getting a little frustrated with this PR. It fixes a rather annoying regression. The change is really tiny and most likely won't cause any problems with whatever is planned/in stock for the menu. There has been a minor release in the meantime, which would have been a perfect chance to pull this really small change, IMHO. Unfortunately, that was not the case.
So, are there any plans for another minor release which might include this? If not, I will close this PR and not waste any more of my time&energy with this.

itzexor:
Sorry I missed the initial inquiry from clem here. I have done some work on the search, but it's nothing ready to merge. It depends too heavily on running 100s or thousands of regex matches per query (for every character typed too) and I'm worried it's going to perform badly.

clefebvre:
Until we get proper hierarchy in place (i.e. the ability to sort results by type of match), I've taken the recommendations made in the issues and in this PR and implemented some of them with https://github.com/linuxmint/cinnamon/commit/d40ba4f900da90f09ecda78f3b139f7bef0fd4cf.

Issue #7191, opened 2018-01-11
Scrolling the 'All Applications' menu is slow (and causes high CPU load)


crahan:
```
 * Cinnamon 3.6.7
 * Mint 18.3
 * Intel 2nd Generation Core Processor Family Integrated Graphics Controller (X11 driver: intel)
 * 64 bit
 ```

**Issue**
When hovering the mouse over an application entry in the `All Applications` category of the Cinnamon menu and scrolling with 2 fingers on a trackpad or scrolling using a trackball, the scroll animation is choppy and results in high CPU load (100%).

However, when the mouse pointer hovers over the scrollbar for the `All Applications` category the scrolling is smooth and CPU load is much lower (half). The only difference is the app title and description lookup that appears near the bottom of the Cinnamon menu. When the title and description is showing scrolling is slow. When the information does not show scrolling is smooth.

Also, this only happens in the `All Applications` category and not with any of the other application categories (possibly due to the amount of entries?).

**Steps to reproduce**
* Open Cinnamon menu
* Go to the `All Applications` category
* Hover the mouse over `Account details`
* Scroll using 2 fingers on a trackpad or using a trackball device
* Scroll animation is choppy and skips over entries

**Expected behaviour**
Scrolling up and down the `All Application` category should be smooth.

**Other information**
None

xenopeek:
I could reproduce the high CPU load using scrollwheel on a regular mouse.

- Go to the All Applications category in the menu and use the scrollwheel while hovering the mouse cursor over the list of applications. In System Monitor I see CPU load spike to 100%.
- If I instead hover the mouse cursor over the scrollbar and then use the scrollwheel, CPU load is around 50%.

When the mouse cursor is hovering over the list of applications, as you scroll, the name and description of each application you scroll over is displayed at the bottom of the menu. When hovering over the scrollbar that doesn't happen. Could this looking up of the application name and description cause the slow down?

pyroxar:
I confirm this bug too on my laptop with Linux Mint 18.3.
![cpu](https://user-images.githubusercontent.com/23406555/34915769-d776f760-f92b-11e7-8ebe-f5809e56e1e9.png)

bambamal:
+1 I noticed the same issue. Scrolling with a touchpad or a mousewheel is very slow and sluggish. Scrolling with the scrollbar works just fine.

Bodai-76:
Hello guys.
[...]

PR #8259, opened 2019-01-04
Menu applet: Disable button reactivity while scrolling through app list to improve performance

stratts:
I'm attempting to solve #7191 by disabling the 'reactive' property of app list buttons when the mouse is scrolled, and re-enabling it when the pointer is moved or a button is clicked. This prevents the large number of enter and leave events from being triggered, which seems to be the primary cause of lag.

On my laptop this improves scrolling performance significantly, but buttons are no longer highlighted during scrolling. I feel like this is a hack around a different root cause, but I'm not familiar enough with Clutter or Cinnamon applets to figure out what that is. Even just moving the pointer around on the app list causes high CPU loads.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1137,6 +1137,7 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
         this._recentButtons = [];
         this._categoryButtons = [];
         this._searchProviderButtons = [];
+        this._appButtonsEnabled = true;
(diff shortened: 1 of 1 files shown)


crahan:
I've applied the suggested patch on my Mint install and the improvement in speed and performance is significant. Thank you @stratts!

clefebvre:
Thanks @stratts and sorry this didn't go in. We had a revamp pending on the menu and it's now superseeded by https://github.com/linuxmint/cinnamon/pull/8314 which got merged.

@itzexor, scrolling is now smooth on master after 8314 is merged, but just bringing this PR to your attention, in case you think there's still room for it.

Thanks everyone.

PR #8278, opened 2019-01-09
Simplify DocSystem/DocInfo

clefebvre:
This PR simplifies DocSystem/DocInfo (which role is to handle the list of recent documents in Cinnamon), to optimize performance, memory usage and prevent memory leaks (by removing complicated code).

It:

- Removes a lot of unused code.
- Simplifies the listing of recent docs in the menu@cinnamon.org and recent@cinnamon.org applets.
- Makes everybody use the same unique instance of DocInfo.
- Frees recent objects which aren't used.
- Moves the limit of 20 recent items from JS to C (Gtk.RecentManager can't sort or clamp result, this is done in C in DocSystem.. it used to be sorted in C but clamped in JS before this PR).
- Remove the number of objects in memory (for instance recent applet no longer keeps a handle on recent objects, only their URI)

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -18,12 +18,12 @@ const FileUtils = imports.misc.fileUtils;
 const Util = imports.misc.util;
 const DND = imports.ui.dnd;
 const Meta = imports.gi.Meta;
+const DocInfo = imports.misc.docInfo;
 const GLib = imports.gi.GLib;
 const Settings = imports.ui.settings;
 const Pango = imports.gi.Pango;
 const AccountsService = imports.gi.AccountsService;
 const SearchProviderManager = imports.ui.searchProviderManager;
-const DocInfo = imports.misc.docInfo;
 
 const MAX_FAV_ICON_SIZE = 32;
 const CATEGORY_ICON_SIZE = 22;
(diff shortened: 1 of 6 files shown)


clefebvre:
This should be tested on a computer with thousands or tens of thousands of recent items.

The number of recents can be obtained from the command line:
`python3 -c "import gi; gi.require_version('Gtk', '3.0'); from gi.repository import Gtk;r=Gtk.RecentManager.get_default();print(len(r.get_items()));"`

Compared to master, with thousands of recents, we're hoping to see a reduction in how long Cinnamon takes to start (especially on hard drives after a cold reboot), and the amount of RAM it uses. It could also have an impact on CPU usage right after opening files (when the list of recents is rebuilt).

mint-dev-tool's cinnamon-stats-tracker should help comparing stats.

clefebvre:
So, it turns out GTK's recent manager has a limit at 1000:

https://github.com/GNOME/gtk/blob/master/gtk/gtkrecentmanager.c#L116

I hit that limit by brute loading text files:

`find / -name "*.txt" -exec xdg-open {} \;`

clefebvre:
So, it turns out GTK's recent manager has a limit at 1000:

https://github.com/GNOME/gtk/blob/master/gtk/gtkrecentmanager.c#L116

I hit that limit by brute loading text files:

find / -name "*.txt" -exec xdg-open {} \;

clefebvre:
Testing the changes with 1000 recent items on a HDD with no cache (after a cold reboot) unfortunately showed no signs of performance improvements. Cinnamon loads in 4.6s and takes 60MB RAM, both on master and on this branch.

The code itself is much simpler than on master though and that itself is an improvement.

PR #8295, opened 2019-01-14
Backport appsys changes

clefebvre:
This PR backports commits made to appsys upstream.

In particular, it removes the dependency on cinnamon-menus and makes AppSys standalone.

Note though that some of the components in Cinnamon still use gir1.2-cmenus:

- cinnamon menu editor
- menu applets (core and spices)

Appsys is embedded in Cinnamon and does not support categories, so it can't replace cmenus in its current form.


clefebvre:
Backported up to:

https://github.com/GNOME/gnome-shell/commits/master/src/shell-app-system.c
May 28th 2014.

https://github.com/GNOME/gnome-shell/commits/master/src/shell-app.c
Nov 14, 2013

clefebvre:
https://github.com/GNOME/gnome-shell/commits/master/src/shell-app-system.c
Nov 17th 2018.

clefebvre:
Performance looks largely unchanged without HDD cache:

- master: 57MB RAM, 4.1s startup (1.6s appsys, 700ms menu)
- PR: 58MB RAM, 4.0s startup (1.3s appsys, 950ms menu)

With HDD cache, we're seeing a slightly faster init:

- master: 52ms appsys + 500 menu
- PR: 33ms appsys + 510 menu

TLDR: Performance isn't much different.

clefebvre:
The main pros of this PR are:

- It updates some really old code (some elements in appsys go all the way back to 2012)
- It makes appsys similar to upstream (that will ease future backports)
- It completely separates cmenu from appsys, and that makes it simpler for us to optimize the menu applet.

clefebvre:
The following functions are removed:

```
appsys.lookup_setting()
appsys.lookup_settings_app()
appsys.lookup_settings_app_by_tree_entry()
appsys.search_settings()

```
These functions return NULL since Cinnamon 1.8 (Mint 14), so we can remove usage safely.

Also removed are the unused:

```
appsys.lookup_app_for_path()
appsys.initial_search()
appsys.subsearch()
app.get_faded_icon()
```

Also removed is:

`apps.sys.get_tree()`

which can be replaced by (that code doesn't need the changes in this PR): 

```
tree = new CMenu.Tree({ menu_basename: "cinnamon-applications.menu" });
tree.load_sync();
```

And:

`appsys.lookup_app_by_tree_entry(entry)`

replaced with:

`appsys.lookup_app(entry.get_desktop_file_id())`

^^ this can be changed in applets without this PR.

Finally, the following changes are brought by this PR:

`app.get_keywords()` no longer returns a string, but an array of stings.. so it should be replaced by `app.get_keywords().join(" ")`.

appInfo.should_show() should be tested to know if an app should be visible:

`if (appInfo && !appInfo.get_nodisplay()) ` --> `if (appInfo && appInfo.should_show() && !appInfo.get_nodisplay())`

clefebvre:
Almost everything is ready on master and spices for this to be merged.

app.get_keywords() is used in Cinnamenu@json. @jaszhix what do you think? Do you want to make its code compatible with both old and new (i.e. string and string array), or start a 4.2 version for it?

appinfo.get_nodisplay() is used in:

- Cinnamenu@json
- tracker@glebihan
- CinnVIIStarkMenu@NikoKrause
- officeCenter@scollins
- graphicsCenter@scollins

@glebihan @NikoKrause @collinss @jaszhix, this isn't critical but it might be a good idea to review whether or not appInfo.should_show() should be used in combination with appInfo.get_nodisplay(), not just for this PR, but for current Cinnamon/spices.

jaszhix:
@clefebvre I would be open to starting a 4.2 version for it.

After looking at the glib source, it appears they both do the same thing. `g_desktop_app_info_get_nodisplay`is a little simpler and might perform better.

clefebvre:
Ah cool, I'll try to get this merged tomorrow.

collinss:
@clefebvre I'm not sure what should_show() does differently than get_nodisplay() (without looking at the code), but it seems to me that get_nodisplay() already gives me everything I need there, so I don't think I'll make any additional changes. I also refactored Office Center and Graphics Center the other day so that they don't use appsys anymore, and rely solely on CMenu, so I don't have a problem with this being merged on my end.

PR #8297, opened 2019-01-15
Various menu improvements

itzexor:
- Fixes very minor memory 'leak' where we added the application id string to a list of known apps every time it was launched, instead of once the first time the app became known.
- Fixes transient and search provider buttons possibly being left around forever.
- All the rest is code simplification, duplication reduction, and removal of unneeded code


itzexor:
rebased on master w/ appsys changes

clefebvre:
I see a small loading improvement 680ms -> 650ms or so.

PR #8311, opened 2019-01-18
xlet settings list widget: some fixes and improvements to the tree

### LINKED ISSUE #8009: xlet settings-schema: The TRUE and FALSE words in a list containing a boolean are not translated
```
 * Cinnamon version: 3.8.9
 * Distribution: Mint 19
 ```

**Issue**
Example in the Settings of the vpnLookOut@claudiux applet, in French language: the words TRUE and FALSE are not translated.
vpnLookOut@claudiux applet Settings capture

**Steps to reproduce**
Install the applet vpnLookOut@claudiux (or any other applet that have, in the Settings, a list containing data with at least one boolean field).
Switch to a language other than English.
Open Settings... (by right-clicking on the applet icon).

**Expected behaviour**
That the words TRUE and FALSE (`"type": "boolean"`) become translatable by default or, preferentially, by entries in the .pot file. (Because, in this case, TRUE and FALSE can also be translated into French as OUI (YES) and NON (NO).)

**Other information**
Part of `settings-schema.json`:
```
    "manageClients": {
        "type": "checkbox",
        "indent": true,
        "default": false,
        "description": "Manage VPN-related Apps",
        "tooltip": "This option allows this applet to manage the VPN-related applications you listed above."
    },
    "clientsList": {
        "type": "list",
        "description": "Managed applications",
        "tooltip": "Name: What you want.\nCommand: The command to run the application, without its path.\nRestart: Set to TRUE if you want to restart this application as soon as the VPN link becomes active again.\nShutdown: Set to TRUE if you want to close the application as soon as the VPN link is stopped.\nVPN only: Set to TRUE if you want to quickly stop this application if it is started while the VPN is idle.",
        "columns": [
            {
                "id": "name",
                "title": "Name",
                "type": "string"
            },
            {
                "id": "command",
                "title": "Command",
                "type": "string"
            },
            {
                "id": "restart",
                "title": "Restart",
                "type": "boolean"
            },
            {
                "id": "shutdown",
                "title": "Shut down",
                "type": "boolean"
            },
            {
                "id": "vpnOnly",
                "title": "VPN only",
                "type": "boolean"
            }
        ],
        "default": [
            {
                "name": "Transmission",
                "command": "transmission-gtk",
                "restart": false,
                "shutdown": true,
                "vpnOnly": false
            }
        ],
        "dependency": "manageClients"
    }

```

### PR DESCRIPTION
**Pull request by collinss**
This improves how data is displayed in the tree of an xlet settings list widget in several ways:

1.) Boolean values are now displayed as a checkbox rather than the values TRUE and FALSE.
2.) Icon type values are now displayed as an icon rather than text.
3.) Columns with a key-value options list now display the key rather than the value. In most cases this makes more sense and it brings it more in line with the combo box widget which is used in the add/edit dialog.
4.) A new property 'align' was added to allow the alignment of the content to be controlled. The settings example applet and the docs were updated to reflect this change.

Fixes #8009

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/TreeListWidgets.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/TreeListWidgets.py
@@ -131,6 +131,12 @@ class List(SettingsWidget):
 
             if render_type == 'boolean':
                 renderer = Gtk.CellRendererToggle()
+
+                def toggle_checkbox(renderer, path, column):
+                    self.model[path][column] = not self.model[path][column]
+                    self.list_changed()
+
+                renderer.connect('toggled', toggle_checkbox, i)
                 prop_name = 'active'
             elif render_type == 'icon':
                 renderer = Gtk.CellRendererPixbuf()
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/TreeListWidgets.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/TreeListWidgets.py
@@ -123,9 +123,38 @@ class List(SettingsWidget):
         types = []
         tv_columns = []
         for i in range(len(columns)):
-            types.append(VARIABLE_TYPE_MAP[columns[i]["type"]])
-            renderer = Gtk.CellRendererText()
-            column = Gtk.TreeViewColumn(columns[i]["title"], renderer, text=i)
+            column_def = columns[i]
+            types.append(VARIABLE_TYPE_MAP[column_def['type']])
+
+            has_option_map = 'options' in column_def and isinstance(column_def['options'], dict)
+            render_type = 'string' if has_option_map else column_def['type']
+
+            if render_type == 'boolean':
+                renderer = Gtk.CellRendererToggle()
+                prop_name = 'active'
+            elif render_type == 'icon':
+                renderer = Gtk.CellRendererPixbuf()
+                prop_name = 'icon_name'
+            else:
+                renderer = Gtk.CellRendererText()
+                prop_name = 'text'
+
+            column = Gtk.TreeViewColumn(column_def['title'], renderer)
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by claudiux** [comment]
Great work @collinss ! Thank you very much.

File: files/usr/share/cinnamon/cinnamon-settings/bin/TreeListWidgets.py [hunk 1]
```diff
@@ -123,9 +123,38 @@ def __init__(self, label=None, columns=None, height=200, size_group=None, dep_ke
...
-            renderer = Gtk.CellRendererText()
-            column = Gtk.TreeViewColumn(columns[i]["title"], renderer, text=i)
+            column_def = columns[i]
+            types.append(VARIABLE_TYPE_MAP[column_def['type']])
+
+            has_option_map = 'options' in column_def and isinstance(column_def['options'], dict)
+            render_type = 'string' if has_option_map else column_def['type']
+
+            if render_type == 'boolean':
+                renderer = Gtk.CellRendererToggle()
```
Odyseus: I connected the **toggled** event to the **Gtk.CellRendererToggle**.

```suggestion
                renderer = Gtk.CellRendererToggle()
                renderer.connect("toggled", self.checkbox_toggled_cb, i)
```

And this is the **checkbox_toggled_cb** function:

```python
# col is the i index passed to the renderer connection.
def checkbox_toggled_cb(self, cell, path, col):
    if path is not None:
        iter = self.model.get_iter(path)
        self.model[iter][col] = not self.model[iter][col]

        self.list_changed()
```

collinss:
@Odyseus In-line editing of boolean values has been added. I did it a little differently than the code you suggested, but the same idea.

claudiux:
@collinss Great! It works like a charm. Many thanks to you!

PR #8314, opened 2019-01-19
Menu applet: Various improvements

itzexor:
This seems to work great on my end, but it may break stuff so it needs testing.

- buttons significantly faster by not using PopupBaseMenuItem
- buttons all have a 'name' and 'description' property to simplify enter/leave events
- removed some classes that just added an icon or label
- combined all enter/leave events to a single handler to consolidate all the code into one location and reduce the amount of duplicated code
- the applications box now uses only one context menu item. this is how recent buttons already worked, but there was 1 menu per application button and on my system it was about 110 menu item actors
- scrollToButton() now resets vscroll adjustment to 0 when no button is passed. avoids passing first item and having it calculate when we just want to scroll to the top.
- a lot of other random cleanups

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -786,14 +786,6 @@ class RecentButton extends SimpleMenuItem {
     }
 }
 
-class NoRecentDocsButton extends SimpleMenuItem {
-    constructor(label) {
-        super(false, false);
-        this.actor.set_style_class_name('menu-application-button');
-        this.addLabel(label, 'menu-application-button-label');
-    }
-}
-
 class RecentClearButton extends SimpleMenuItem {
     constructor(applet) {
         super();
@@ -2316,7 +2308,9 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
                 this._recentButtons.push(recent_clear_button);
                 this.applicationsBox.add_child(recent_clear_button.actor);
             } else {
-                let no_recent_button = new NoRecentDocsButton(_("No recent documents"), null, false, null);
+                let no_recent_button = new SimpleMenuItem(false, false);
+                no_recent_button.actor.set_style_class_name('menu-application-button');
+                no_recent_button.addLabel(_("No recent documents"), 'menu-application-button-label');
                 this._recentButtons.push(no_recent_button);
                 this.applicationsBox.add_child(no_recent_button.actor);
             }
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -22,7 +22,6 @@ const DocInfo = imports.misc.docInfo;
 const GLib = imports.gi.GLib;
 const Settings = imports.ui.settings;
 const Pango = imports.gi.Pango;
-const AccountsService = imports.gi.AccountsService;
 const SearchProviderManager = imports.ui.searchProviderManager;
 const SignalManager = imports.misc.signalManager;
 const Params = imports.misc.params;
@@ -896,7 +895,7 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
 
         this.settings = new Settings.AppletSettings(this, "menu@cinnamon.org", instance_id);
 
-        this.settings.bind("show-places", "showPlaces", () => this._refreshBelowApps);
+        this.settings.bind("show-places", "showPlaces", this._refreshBelowApps);
 
         this._appletEnterEventId = 0;
         this._appletLeaveEventId = 0;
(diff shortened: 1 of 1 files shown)


files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -24,12 +24,15 @@ const Settings = imports.ui.settings;
 const Pango = imports.gi.Pango;
 const AccountsService = imports.gi.AccountsService;
```
Odyseus:
```suggestion
```

Since you are cleaning code, you could remove this import since it isn't used anymore.

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -1093,7 +885,7 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
 
         this.settings = new Settings.AppletSettings(this, "menu@cinnamon.org", instance_id);
 
-        this.settings.bind("show-places", "showPlaces", this._refreshBelowApps);
+        this.settings.bind("show-places", "showPlaces", () => this._refreshBelowApps);
```
Odyseus:
```suggestion
        this.settings.bind("show-places", "showPlaces", () => this._refreshBelowApps());
```

This callback isn't called. Missing parentheses.

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -1190,7 +980,7 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
     onAppSysChanged() {
         if (this.refreshing == false) {
             this.refreshing = true;
-            Mainloop.timeout_add_seconds(1, Lang.bind(this, this._refreshAll));
+            Mainloop.timeout_add_seconds(1, this._refreshAll.bind(this));
```
Odyseus:
```suggestion
            Mainloop.timeout_add_seconds(1, () => this._refreshAll());
```

This could also be an arrow function.

itzexor:
@Odyseus thanks for the review, i'll look into those. This branch is still heavy WIP right now.

clefebvre:
Please rebase.

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -1412,39 +1225,48 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
...
+        if (!button.withMenu)
+            return;
+
+        if (!this.contextMenu) {
+            let menu = new PopupMenu.PopupSubMenu(null); // hack: creating without actor
+            menu.actor.set_style_class_name('menu-context-menu');
+            menu.connect('open-state-changed', Lang.bind(this, this._contextMenuOpenStateChanged));
+            this.contextMenu = menu;
+            this.applicationsBox.add_actor(menu.actor);
+        }
```
Odyseus: Hello, @itzexor.

A little suggestion to improve the context menu logic.

#### Current behaviour

Attempting to open the context menu of any button with right click while another context menu is opened requires two clicks on a button. One to close the previously opened menu, and another to open the desired menu.

#### Proposed behaviour

Attempting to open the context menu of any button while another context menu is opened should close the opened menu and at the same time open the context menu of the clicked button.

I managed to achieve the proposed behaviour with the following changes:

```suggestion
        }

        if (this.contextMenu.sourceActor !== button.actor &&
            this.contextMenu.isOpen) {
            this.contextMenu.close();
        }

```

Odyseus:
```suggestion
        }

        if (this.contextMenu.sourceActor !== button.actor &&
            this.contextMenu.isOpen) {
            this.contextMenu.close();
        }

```

itzexor:
I'm going to close this for now because some things have come up, and i don't have the time to work on it right now. Thanks for the pointers, i will look into them when i come back to this soon.

brownsr:
I've been trying this PR a number of times.  I can't find any issues with it, and the performance is a definite improvement

Regards

Simon

itzexor:
- rebased on master
- fixed indentation issue
- moved some more experimental commits to a new branch https://github.com/itzexor/Cinnamon/commits/menu-various-3 which is branched off of this one, so most of those commits are just these ones.

I think this may be ready for testing and merge if it looks good to everyone. There are a lot of changes here that touch a lot of different parts of the menu so I've probably missed a few things here and there.

The loop warning (which is not new) is fixed on a different branch now, opened as PR #8396

clefebvre:
Quick black box testing:

- It feels fast!
- Mouse wheel scrolling the all applications category is much smoother than on master!
- Small regression described below

When moving the mouse back and forth between the categories box and the apps box, the apps move 1px to the right/left. This was fixed before, it had to do with a +1 somewhere in the vector box if I remember right.

clefebvre:
Another small regression:

- Right-click an app, choose uninstall. The pkexec window appears but the menu doesn't close.

itzexor:
- rebased on master
- fix menu closing for context menu item activation
- hopefully fix vectorBox size/position errors
- added retry timeout loop to 'add to panel' context menu item. sometimes this would fail because the applet instance can take a bit of time to actually be loaded.

PR #8317, opened 2019-01-19
Revert "st-icon: create shadow pipeline only at paint (#8238)"

brownsr:
This reverts commit 6d4ae223069566360b4edd0514003422c67bbd66.
Jason measured a small worsening in latency, and as the upstream changes were to avoid segfaults that I don't think we have, there seems no reason to keep this.


jaszhix:
After re-testing on master, I'm not sure this revert is necessary. It was in a sub-ms range of latency with Typometer, and I think it's not noticeable. If this patch does prevent a segfault (perhaps one that could be possible that we just don't know about), then the patch should stay applied.

PR #8340, opened 2019-01-25
Add workspace names to workspace switcher

firejump:


locvio:
Hi Bartek,
As a user, I am very interested in this change.  Ie having workspace names appear in the switcher instead of visual representations or the numbers ("1 2 3 ") that are currently implemented.   So "myWorkspace_1 myWorkspace_2 myWorkspace_3"..  Is that what this ticket was for?  I am wondering if you know the history of this work, and why this branch was deleted without the change being merged.  thanks so much! 
Alex

firejump:
Hi Alex.
This is indeed what this pull request is for, however right now it's a ugly hack that can't be merged upstream. I added this as a PR to the upstream repo rather than my private fork by mistake, that's why it's removed. I unfortunately don't have resources to develop it further at this point.

locvio:
Bartek,

gotcha..  What would the the best way to register  a "+1" interest in this feature with the cinnamon team?  My company migrated to cinnamon and a few of us really miss having the workspace indicator!

Thanks again for your help!

Alex

JosephMcc:
There are some problems I'd see with this idea. The main one being that the workspace switcher would take up a huge amount of space on the panel. Even with just a couple workspaces.

locvio:
hey Joseph,
From that respect this was no problem at all.. was using for years with six workspaces in xfce each with descriptive names. Granted, on a pretty high res!

Alex

PR #8341, opened 2019-01-25
appsys: Adjust for performance

jaszhix:
After f8fadd5be74196e3197323124b8eca657d9770c6, I've been noticing increased lag on my relatively high-res display setup. This is an attempt to regain some of the performance lost. Synthetic tests do slightly hint towards there being less performance with the new app system changes. After discussions on slack, I seem to be the only one, but haven't heard from anyone who also is using multiple high-resolution displays (2K+), so if anyone is able to test this along with the muffin patch it would be appreciated.

One other thing of note with my computer is it has low clocks at 2GHz, and I think this makes CPU usage on cinnamon, being a single threaded application, more noticeable.

**Requires** https://github.com/linuxmint/muffin/pull/418

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -57,21 +57,6 @@ fi
 
 AM_CONDITIONAL(BUILD_RECORDER, $build_recorder)
 
-AC_ARG_ENABLE([systemd],
-              AS_HELP_STRING([--enable-systemd], [Use systemd]),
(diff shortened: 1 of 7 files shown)


jaszhix:
https://github.com/linuxmint/Cinnamon/pull/8346

PR #8346, opened 2019-01-29
App system refinements

jaszhix:
This PR removes and refactors code that either can't be used in Cinnamon, or causes stability/performance issues.

**Requires** linuxmint/muffin#418


jaszhix:
Honestly, I'm livid I'm spending all this work on this when we just need to revert gnome's technical debt in #8295.

PR #8348, opened 2019-01-29
Backport appsys changes

clefebvre:
2nd try, following up on #8295.


clefebvre:
I'm going to reduce the number of commits first and squash some of them together (other than my self inflicted typos, some commits upstream in GNOME don't make sense or compile on their own). 90 commits is a lot and it makes it hard to review the changes.

PR #8351, opened 2019-01-30
General optimizations

itzexor:
Simple profiling with the GJS/CJS profiler revealed:
1. paths being called relatively frequently when the features they support weren't enabled:
-magnifier cursor update
-desklet mouse tracker
2. window restack connections are always a top consumer on general non-cinnamon-ui usage
3. <s>uiGroup is always allocating, and the allocate does nothing beyond default</s>
4. notification daemon is always checking the WindowTracker to clear notifications, even when there are none

### DIFF
--- a/js/ui/deskletManager.js
+++ b/js/ui/deskletManager.js
@@ -77,7 +77,7 @@ function init(){
         global.settings.connect('changed::' + DESKLET_SNAP_INTERVAL_KEY, _onDeskletSnapChanged);
 
         deskletsLoaded = true;
-        enableMouseTracking(true);
+        updateMouseTracking();
         global.log(`DeskletManager started in ${new Date().getTime() - startTime} ms`);
     });
 }
@@ -86,7 +86,8 @@ function getDeskletDefinition(definition) {
     return queryCollection(definitions, definition);
 }
 
-function enableMouseTracking(enable) {
+function updateMouseTracking() {
+    let enable = definitions.length > 0;
     if (enable && !mouseTrackTimoutId) {
         mouseTrackTimoutId = Mainloop.timeout_add(500, checkMouseTracking);
     } else if (!enable && mouseTrackTimoutId) {
@@ -252,7 +253,7 @@ function _onEnabledDeskletsChanged() {
 
     // Make sure all desklet extensions are loaded.
     // Once loaded, the desklets will add themselves via finishExtensionLoad
-    initEnabledDesklets();
+    initEnabledDesklets().then(updateMouseTracking);
 }
 
 function _unloadDesklet(deskletDefinition, deleteConfig) {
--- a/js/ui/main.js
+++ b/js/ui/main.js
@@ -464,7 +464,6 @@ function start() {
 
     global.screen.connect('window-entered-monitor', _windowEnteredMonitor);
     global.screen.connect('window-left-monitor', _windowLeftMonitor);
-    global.screen.connect('restacked', _windowsRestacked);
 
     global.display.connect('gl-video-memory-purged', loadTheme);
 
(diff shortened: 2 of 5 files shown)


js/ui/main.js
```diff
@@ -350,23 +350,8 @@ function start() {
...
-                        let width = global.stage.width;
-                        [alloc.min_size, alloc.natural_size] = [width, width];
-                    });
-    uiGroup.connect('get-preferred-height',
-                    function(actor, forWidth, alloc) {
-                        let height = global.stage.height;
-                        [alloc.min_size, alloc.natural_size] = [height, height];
-                    });
+    uiGroup = new Clutter.Actor({ name: 'uiGroup' });
+    global.stage.connect('notify::size', () => { uiGroup.size = global.stage.size });
```
itzexor: fixme: replace with bind constraint like the one used for stage_bg actor below

jaszhix:
Just for reference, this was closed after discussions on Slack. I did not intend for this to be closed. There were two problems I pointed out:

- 885582d switches uiGroup to an StWidget actor, but this loses the skip_paint optimization in CinnamonGenericContainer, and seemed less responsive. I tried to [benchmark](https://openbenchmarking.org/result/1902033-SK-VARIOUSJA09&amp;obr_sor=y) this with gputest, which might at least show the impact Cinnamon has on the GPU.
- 2dd1532 slows the cursor movement. I'm not sure why. It seems like a muffin or xorg bug which had less impact while Cinnamon was polling XQueryPointer.

itzexor:
- rebased on master
- dropped uiGroup commit to another branch for future investigation
- fixed up magnifier commit to cover hopefully all cases where it can be first called.

i think the desklet one definitely wastes resources if you're not using desklets. i'd like to keep that one in if possible.

js/ui/deskletManager.js
```diff
@@ -252,7 +253,7 @@ function _onEnabledDeskletsChanged() {
 
     // Make sure all desklet extensions are loaded.
     // Once loaded, the desklets will add themselves via finishExtensionLoad
-    initEnabledDesklets();
+    initEnabledDesklets().then(updateMouseTracking);
```
jaszhix: The definitions array handling is synchronous and is already determined before we start loading desklets, so we don't need to call this through the promise prototype. The promise is only needed if we want to have something run after every desklet is loaded from disk. The cursor movement improves when I call updateMouseTracking before initEnabledDesklets. Maybe a CJS/GLib bug (since the promise itself is implemented as an idle source)?

js/ui/deskletManager.js
```diff
@@ -77,7 +77,7 @@ function init(){
         global.settings.connect('changed::' + DESKLET_SNAP_INTERVAL_KEY, _onDeskletSnapChanged);
 
         deskletsLoaded = true;
-        enableMouseTracking(true);
+        updateMouseTracking();
```
jaszhix: This could be moved to line 73 after `definitions = getDefinitions()`. The array won't change after that point until a desklet is added or removed.

itzexor: The mouse tracker source calls methods on the desklet instances, so I figured after the promise was the proper place to put it to ensure the function could only ever be called after the first instance was fully loaded. Granted it would have to take > 500ms for the instance to load for it to cause an issue, is my logic incorrect here? I don't really know the internals of the extensions system.

OTOH if definitions[i].desklet is falsy before it fully loads, calling it immediately should be ok.

js/ui/deskletManager.js
itzexor: I have no idea why it would have that effect, so I can't really say

js/ui/deskletManager.js
jaszhix: I didn't see that it was checking the desklet itself, that makes sense to me.

itzexor:
- rebased on master
- fixed mistake in magnifier.js where I replaced `new Object(roi)` with `Object.Clone(roi)` (non existent function) instead of `Object.assign({}, roi)` as I meant to. This gets rid of a "use an object literal" warning.

Issue #8319, opened 2019-01-19
Select monitor for each workspace


brunoribeiro94:
**Issue**

I have two monitor 17", I'd like know if is possible make this settings.

Ex:
Monitor 1 | Show WORKSPACE 1 Monitor 2 | Show WORKSPACE 2

I tryed in display settings, I not found this option that I desire.

Maybe some software can do this ?

My OS: Linux Mint 18.1 Cinnamon 64-bit

pyroxar:
Good idea.

pyroxar:
Hi @collinss 
You work now about workspace? Did you see it? What you think?

icarter09:
Many thanks for submitting your feature request for Cinnamon. Your feature request has been documented and will be reviewed by the Linux Mint team.

For more information on our workflow and feature requests, read https://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/faq.html.

PR #8352, opened 2019-01-30
Workspace switcher applet: various improvements to the graph style

collinss:
This improves the workspace switcher visual style in the following ways:

- The clutter.actor with layout manager was removed and buttons are now added directly to the actor box. Part of this change means that the spacing is no longer hard-coded and can be conrolled by the theme.
- Events that trigger the creation of new buttons now queue said creation in an idle callback rather than recreating them directly. This potentially reduces the number of times they are recreated (for example at startup there could be several events that trigger them to be regenerated, but they really only need to be regenerated once, after all the signals are fired).
- An issue is now fixed where the creation of struts didn't trigger the regeneration of the buttons, so the wrong workarea was often used in size caluculations, which in turn caused incorrect placement of (for example) fullscreen windows. This was particularly an issue when cinnamon first started.
- The visual layout is now suppressed based on estimated dimensions (anything less than 20px) rather than on the size ratio between width and height. This seems more appropriate than some aribitrary ratio, and it makes more sense given our new model of panel with easily editable heights and per-zone sizes.
- Size calculations were improved - and in many cases simplified. They no longer make assumptions about padding and border width (which could easily be wrong for a given theme) but rather calculates these correctly. Also final size and position values are now rounded to the nearest integer to ensure they are pixel-perfect.

### DIFF
--- a/files/usr/share/cinnamon/applets/workspace-switcher@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/workspace-switcher@cinnamon.org/applet.js
@@ -3,6 +3,7 @@ const Clutter = imports.gi.Clutter;
 const Lang = imports.lang;
 const Applet = imports.ui.applet;
 const Main = imports.ui.main;
+const Mainloop = imports.mainloop;
 const Meta = imports.gi.Meta;
 const PopupMenu = imports.ui.popupMenu;
 const SignalManager = imports.misc.signalManager;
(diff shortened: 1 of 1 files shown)


pyroxar:
Hi @collinss 
You work now on workspace? Did you see it:
https://github.com/linuxmint/Cinnamon/issues/8319

brownsr:
Great work, looks good to me.

collinss:
Hi @pyroxar that's very much outside the scope of my pull request. It would also require a lot of work with some major structural changes to our window manager, and would affect a number of other components as well. I'm not saying that it will never happen, but it's not terribly likely.

PR #8368, opened 2019-02-04
Correctly handle icon/theme directories based on XDG directories

eli-schwartz:
In commit b4338cb6c0fe4e3dc3d0befbff9b1bd3fa93844f partial support was added for using the XDG directories in order to locate these; however, not all the code was wired up to actually use this.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
@@ -29,6 +29,7 @@ except ImportError:
     import simplejson as json
 
 home = os.path.expanduser("~")
+xdg_data_home = os.getenv("XDG_DATA_HOME", os.path.expanduser("~/.local/share"))
 locale_inst = '%s/.local/share/locale' % home
 settings_dir = '%s/.cinnamon/configs/' % home
 
@@ -166,8 +167,8 @@ class Spice_Harvester(GObject.Object):
         self.settings.connect('changed::%s' % self.enabled_key, self._update_status)
 
         if self.themes:
-            self.install_folder = '%s/.themes/' % (home)
-            self.spices_directories = (self.install_folder, )
+            self.install_folder = os.path.join(xdg_data_home, 'themes')
+            self.spices_directories = (self.install_folder)
         else:
             self.install_folder = '%s/.local/share/cinnamon/%ss/' % (home, self.collection_type)
             self.spices_directories = ('/usr/share/cinnamon/%ss/' % self.collection_type, self.install_folder)
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
@@ -28,6 +28,8 @@ try:
 except ImportError:
     import simplejson as json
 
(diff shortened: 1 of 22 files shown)


files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
```diff
@@ -165,8 +166,8 @@ def __init__(self, collection_type, window=None):
         self.settings.connect('changed::%s' % self.enabled_key, self._update_status)
 
         if self.themes:
-            self.install_folder = '%s/.themes/' % (home)
-            self.spices_directories = (self.install_folder, '%s/.themes/' % (home))
+            self.install_folder = os.path.join(xdg_data_home, 'themes')
+            self.spices_directories = (self.install_folder, self.install_folder)
```
eli-schwartz: Why is spices_directories a tuple of the same thing twice?

Cobinja:
i think it would be a better approach to use the corresponding GLib functions to get the XDG dirs:
[g_get_user_data_dir ()](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-get-user-data-dir) and
[g_get_home_dir ()](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-get-home-dir)
Then os.path.expand() from there.
They are available in python with GLib.get_user_data_dir() and GLib.get_home_dir()

eli-schwartz:
Maybe if I were doing this from javascript, I'd use what seems to be the preferred method. But I don't see the point in using magic GLib bindings. Why go through the gobject type system just to read an environment variable?

Anyway the existing code right there already uses expanduser rather than GLib.get_home_dir(), and it is a one-line environment variable lookup in a pretty readable methodology that works for any python code anywhere. Using GLib doesn't even simplify it.

clefebvre:
I don't think we need to change this. We support reading from XDG data dirs but that doesn't mean we need to install there, ~/.icons and ~/themes are OK as well. We also don't need to listen/monitor to more places than we already do, that has a cost, we even see it fail.

Last but not least, I don't think we should change the location of the default icon file (for cursor link).

eli-schwartz:
Well, indeed we shouldn't try to create the legacy folder if the PR changes to prefer the XDG version, and then set up a new file watch on that. Amended.

So currently, this should prefer installing in the XDG directory, prefer setting up a watch on that directory but set up an additional one if the one directly in ~ exists as well, and yes, move the cursor link. IMO it makes sense to use XDG everywhere, and in the long run, we won't need to check ~/.icons and ~/themes because they'll never exist.

files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
eli-schwartz: Answer: it shouldn't be, now fixed via #8828 e2690547e0baeeedfd3ce4add9645d86d1cb66a0

Hmm...

clefebvre:


```diff

Issues
======
+ Solved 7
- Added 5
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/Cinnamon/Cinnamon/pullRequest?prid=3022516&bid=11124968)

PR #8370, opened 2019-02-06
cinnamon-app: Simplify icon retrieval for window-backed apps

jaszhix:
Since the pixmap is cached in muffin, and icon-changed receivers destroy their icons on emission, there's no need for a complex caching solution dedicated to window-backed app icons.

Requires https://github.com/linuxmint/muffin/pull/422

### DIFF
--- a/src/cinnamon-app.c
+++ b/src/cinnamon-app.c
@@ -156,16 +156,24 @@ get_actor_for_icon_name (CinnamonApp *app,
 }
 
 static ClutterActor *
+get_failsafe_icon (int size)
+{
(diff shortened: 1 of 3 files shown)


brownsr:
Looking fine here.  I was initially apprehensive that the hacking into texture-cache might have broken background or border textures, but I can't see any problems, so well done figuring out that all of that code was just for icons.  The function names hardly advertised that ;-)  I was also a bit wary of the equal width and height of icons, as I know some icons can be a few % off square, but I've seen no problems in practice.  So looking good to me.

Simon

PR #8396, opened 2019-02-17
menu applet: fix last eslint warning

itzexor:
Avoid creating functions inside loop warning by adding a new callback
params argument to RecentContextMenuItem class so we don't have to
capture scope within the loop.


clefebvre:
@itzexor please rebase.

itzexor:
Rebased on master.

Note: I forgot to include in the commit message that this changes the `g_app_info_launch` function call  to remove the invalid `null` at the end. In C this function takes a `GError` out parameter, but that is not supported in the bindings, so it takes only 2 parameters: `([command line arguments], GAppLaunchContext)`

PR #8397, opened 2019-02-17
overview: Fix incorrect thumbnail sizes after focused window resizing

### LINKED ISSUE #8304: [EDIT]"Show all windows": the position of a window is drawn outside its shape - found the steps to reproduce!!! 👍
* Cinnamon version: 4.0.9
 * Distribution - Mint 19.1
 * Graphics hardware *and* driver used: Intel Haswell-ULT Integrated Graphics 
 * 32 or 64 bit: 64 bit

Issue:

Everytime windows minimized in the system tray (after the 3 steps) are drawn outside their boxes launching the overview "Show all windows" callable with ctrl + alt + down shortcut. 

Steps to reproduce: 

1. Open any application with a GUI.
2. Press the maximize button for two times and reduce the window in the system tray.
3. Press Ctrl + Alt + Down to see the issue.

Frequency:
This happens every time I do all these operations. 

Graphics:
  Device-1: Intel Haswell-ULT Integrated Graphics vendor: Hewlett-Packard 
  driver: i915 v: kernel bus ID: 00:02.0 
  Display: x11 server: X.Org 1.19.6 driver: intel resolution: 1366x768~60Hz 
  OpenGL: renderer: Mesa DRI Intel Haswell Mobile v: 4.5 Mesa 18.0.5 
  direct render: Yes 

This is a screenshot. The problem, in this case, is that Firefox is outside the shape in which you press the cross to close its window. 

schermata del 2019-01-10 19-00-54

schermata del 2019-01-16 16-28-12
immagine

I never had this bug in previous versions, so I think something happened with Cinnamon 4.0.9

jaszhix:
Closes #8304

### DIFF
--- a/js/ui/workspace.js
+++ b/js/ui/workspace.js
@@ -91,7 +91,7 @@ WindowClone.prototype = {
     refreshClone: function(withTransients) {
         this.actor.destroy_all_children();
 
-        let {x, y, width, height} = this.metaWindow.get_outer_rect();
+        let {x, y, width, height} = this.realWindow;
         let clones = WindowUtils.createWindowClone(this.metaWindow, 0, 0, withTransients);
         let leftGap, topGap;
         for (let clone of clones) {
@@ -676,7 +676,7 @@ WorkspaceMonitor.prototype = {
      */
     _computeWindowLayout: function(metaWindow, slot) {
         let [x, y, width, height] = this._getSlotGeometry(slot);
-        let rect = metaWindow.get_outer_rect();
+        let rect = metaWindow.get_input_rect();
         let topBorder = 0, bottomBorder = 0, leftBorder = 0, rightBorder = 0;
 
         if (this._windows.length) {


germanfr:
@jaszhix borders no longer adjust to windows. They adjust the shadow of the window.

germanfr:
If changing from `get_outer_rect` to `get_input_rect` really solves the issue, then I guess the issue is with muffin.

jaszhix:
What do you suggest then? I did some bisecting in this bug has been around for a while so it has nothing to do with recent changes.

jaszhix:
The main change here is we're basing the clone on MetaWindowActor - that is the way it should have been done, and is how we handle window actors everywhere else in Cinnamon. I was tempted to get rid of that overly abstracted WindowClone class and just use Clutter.Clone, but kept the actual fix as simple as I could. We need to use the input region when dealing with window actor coordinates.

germanfr:
You have to have unmaximized windows or CSD windows to see what I say.

jaszhix:
I can't spend anymore time on this. Feel free to open a better PR and I'll happily close this.

clefebvre:
Fixed in https://github.com/linuxmint/muffin/pull/410

jaszhix:
I think I see the confusion on why this was closed. In the referenced issue I mention it being related to linuxmint/muffin#405, but that was before I was able to reproduce it. This PR was opened after that comment. Feel free to close again if I'm mistaken.

jaszhix:
What do you mean? It looks like it's changing two lines.

PR #8399, opened 2019-02-17
WindowManager: JS code optimizations

clefebvre:
Reduce the number of signal listeners

### DIFF
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -471,30 +471,12 @@ var WindowManager = class WindowManager {
         const settingsState = {
             'desktop-effects-on-dialogs': global.settings.get_boolean('desktop-effects-on-dialogs'),
             'desktop-effects-on-menus': global.settings.get_boolean('desktop-effects-on-menus'),
-            'show-snap-osd': global.settings.get_boolean('show-snap-osd'),
-            'show-tile-hud': global.settings.get_boolean('show-tile-hud'),
-            'workspace-osd-visible': global.settings.get_boolean('workspace-osd-visible'),
-            'workspace-osd-x': global.settings.get_int('workspace-osd-x'),
-            'workspace-osd-y': global.settings.get_int('workspace-osd-y'),
-            'workspace-osd-duration': global.settings.get_int('workspace-osd-duration'),
-            'alttab-switcher-style': global.settings.get_string('alttab-switcher-style'),
             'desktop-effects': this.settings.get_boolean('desktop-effects'),
-            'workspaces-only-on-primary': this.settings.get_boolean('workspaces-only-on-primary'),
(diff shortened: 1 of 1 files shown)


js/ui/windowManager.js
```diff
@@ -468,34 +468,22 @@ var WindowManager = class WindowManager {
...
-            'workspace-osd-visible': global.settings.get_boolean('workspace-osd-visible'),
-            'workspace-osd-x': global.settings.get_int('workspace-osd-x'),
-            'workspace-osd-y': global.settings.get_int('workspace-osd-y'),
-            'workspace-osd-duration': global.settings.get_int('workspace-osd-duration'),
-            'alttab-switcher-style': global.settings.get_string('alttab-switcher-style'),
-            'desktop-effects': this.settings.get_boolean('desktop-effects'),
-            'workspaces-only-on-primary': this.settings.get_boolean('workspaces-only-on-primary'),
-            'snap-modifier': this.settings.get_string('snap-modifier'),
-        };
+        this.desktop_effects_enabled = this.settings.get_boolean('desktop-effects');
```
jaszhix: https://jsperf.com/dot-notation-vs-square-bracket-notation

Bracket notation is faster in Spidermonkey and V8. I think this part of the PR should be reverted.

js/ui/windowManager.js
```diff
@@ -565,14 +553,7 @@ var WindowManager = class WindowManager {
...
     onSettingsChanged(settings, key, type) {
-        switch (settings.schema) {
-            case 'org.cinnamon':
-                this.settingsState[key] = global.settings[type](key);
-                break;
-            case 'org.cinnamon.muffin':
-                this.settingsState[key] = this.settings[type](key);
-                break;
-        }
+        this.settingsState[key] = global.settings[type](key);
```
jaszhix: Since bracket notation is faster, I think the previous approach would be preferred where we dynamically look up properties for all settings. It's a little extra code, but I don't think it's hard to follow here

js/ui/windowManager.js
```diff
@@ -468,34 +468,22 @@ var WindowManager = class WindowManager {
 
         this.settings = new Settings({schema_id: 'org.cinnamon.muffin'});
 
-        const settingsState = {
```
jaszhix: I think it's cleaner if we separate state from the base prototype instance. It makes it possible to make property access faster. Take for example:

```js
let something1 = this.propA;
let something2 = this.propB;
// We're asking the engine to look up multiple properties on `this`, it has to look up a key on a complex object that is possibly inheriting other prototypes, including functions, references, and primitives.
```
```
let {state} = this; // We're asking the engine to look up a key on the complex object once.
let something1 = state.propA;
let something2 = state.propB;
// We're asking the engine to look up multiple properties on a much less complex object.
```

PR #8402, opened 2019-02-18
Melange/lookingGlass: better introspected object inspection

itzexor:
- Removes utility function `color(pixel)` from the lookingGlass eval template as this clutter function isn't actually introspected(anymore?) and thus won't work.
- Massive object inspector rework. Description from commit:

Since GObject properties are lazily loaded into their JS wrapper
objects, we can't easily inspect them and get all of the available
properties, functions, etc. This adds a pretty hacky method to
obtain them by looking up all of the introspection data, then looking
up each property on the target object.

This should only ever be used with melange for debugging purposes. Since
this causes _every_ property to be cached into JS, some of which are
additional GObjects/GBoxed/etc, it can and probably does cause higher
memory usage and possibly other weird side effects.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-looking-glass/page_inspect.py
+++ b/files/usr/share/cinnamon/cinnamon-looking-glass/page_inspect.py
@@ -3,6 +3,19 @@
 import pageutils
 from gi.repository import Gtk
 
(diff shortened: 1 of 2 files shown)


itzexor:
i'm thinking about redoing some of this

PR #8424, opened 2019-02-26
Remove JS prototype bloat

jaszhix:
It is extra costly to lookup properties on a prototype instance, so instead of modifying the prototype (which runs the risk of the engine not being able to optimize our code as much), we can make them pure functions that take simple inputs and return simple outputs. Place them on the window object for equivalent convenience. Though most of these were either unused, or their override had a better alternative. 

One such case is maybeGet.

`Cjs-Message: 23:23:45.899: JS WARNING: [/usr/share/cinnamon/js/ui/popupMenu.js 1965]: reference to undefined property "_delegate"`

This was mostly needed to shut up Spidermonkey warnings, but in most cases Spidermonkey was being extra verbose because CJS/GJS configures the environment that way. It is completely valid to just use a falsey check before using an instanceof comparison, but even that is unnecessary because `undefined` instanceof {any object} will always be false. In this case Spidermonkey's extra (forced) verbosity is harmful and creating confusion.

- [x] Turn GObject.prototype.is_finalized into a pure function
- [x] Remove overrides that were accepted by upstream GJS
- [x] Update spices
- [x] Review other GObject prototype overrides
- [x] Reduce finalized GObject warnings in CJS (https://github.com/linuxmint/cjs/pull/76)


js/misc/history.js
```diff
@@ -67,12 +67,13 @@ HistoryManager.prototype = {
...
-        if (this._historyIndex != this._history.length) {
-            this._historyIndex = this._history.length;
+        let {length} = this._history;
+        if (this._historyIndex != length) {
+            this._historyIndex = length;
             this._indexChanged();
         }
 
-        return this._historyIndex.maybeGet(this._history.length);
+        return this._history[this._history.length - 1];
```
itzexor: should this be `return this._history[length - 1];` ?

.eslintrc.json
```diff
@@ -26,7 +26,8 @@
         "get_locale_string": false,
         "__dirname": false,
         "__filename": false,
-        "__meta": false
+        "__meta": false,
+        "capitalize": false
```
itzexor: do we need `"clamp": false` here as well since that was added?

jaszhix: yes, good catch

jaszhix:
@itzexor Agreed, the is_finalized method is more of a workaround and hopefully just a tool to help us transition to better gobject lifecycle state tracking, though it's relatively inexpensive to grab and compare the string representation from CJS. I mainly added it because some C gobjects (previously MetaWindowActor, not sure if it's the case now in linuxmint/muffin#410) would cause Cinnamon to segfault.

PR #8428, opened 2019-02-27
background effects - rework and turn on

brownsr:
This reworks, enables, and extends the previous background effect facility that was turned off in the past as it was broken.

There are two possible background effects - bumpmap and blur. Both are turned on by CSS, and can be layered on each other if desired.
example CSS

```
background-bumpmap: url("/usr/share/cinnamon/bumpmaps/hex.png");
background-blur: 1px;
```
Note that the size of the blur is treated as an iteration count. So background-blur: 1px is the basic effect, while background-blur: 2px applies the effect again on top of itself, and so on. The basic blur is relatively mild, twice is stronger, anything above 3 is overkill IMO.
Speeds seem fine with no noticeable speed penalty unless a large blur count is put in for test purposes.

The blur effect is probably the more useful. The bumpmap effect looks dramatic on a background, but rubbish over something like a text file being edited.

The effect works by snapshotting the screen underneath an actor, applying the effect to it, and then drawing it out as the lowest layer in the process of drawing an actor. There are no restrictions as to what can be layered on it, though clearly you won't want to put something without transparency over it or the effect would be hidden.

In the screenshots below the bumpmap is shown with nothing layered on it. The blur is shown with a semi-transparent background-gradient on top of it.

screenshot from 2018-04-08 13-59-20

screenshot from 2018-04-08 13-58-16


src/st/st-widget.c
```diff
@@ -382,22 +398,20 @@ st_widget_paint_background (StWidget *widget)
...
-  //   }
-
-  // const char *bumpmap_path = st_theme_node_get_background_bumpmap(theme_node);
-
-  // g_object_set (effect,
-  //               "bumpmap",
-  //               bumpmap_path,
-  //               NULL);
+    {
+      if ((theme_node->background_blur > 0 || theme_node->background_bumpmap)
```
jaszhix: Might be good to do these logical comparisons when the change in CSS is detected, instead of on every paint.

src/st/st-widget.c
```diff
@@ -278,6 +282,18 @@ st_widget_dispose (GObject *gobject)
   st_widget_remove_transition (actor);
 
   g_clear_pointer (&priv->label_actor, g_object_unref);
+  if (priv->background_blur_effect)
+    {
+      g_object_run_dispose (G_OBJECT (priv->background_blur_effect));
+      g_object_unref (priv->background_blur_effect);
+      priv->background_blur_effect = NULL;
+    }
+      if (priv->background_bumpmap_effect)
```
jaszhix: Indentation is off.

src/st/st-widget.c
```diff
@@ -1617,7 +1695,7 @@ st_widget_ensure_style (StWidget *widget)
   g_return_if_fail (ST_IS_WIDGET (widget));
 
   if (widget->priv->is_style_dirty)
-    st_widget_recompute_style (widget, NULL);
+      st_widget_recompute_style (widget, NULL);
```
jaszhix: We use two spaces after a bracket-less if statement.

src/st/st-widget.c
```diff
@@ -486,8 +500,15 @@ st_widget_style_changed (StWidget *widget)
...
-  if (clutter_actor_is_mapped (CLUTTER_ACTOR (widget)))
-    st_widget_recompute_style (widget, old_theme_node);
+  if (clutter_actor_is_mapped (CLUTTER_ACTOR (widget))
+  || (old_theme_node != NULL &&
+      (old_theme_node->background_blur > 0
+       || old_theme_node->background_bumpmap != NULL)))
+    {
+      st_widget_recompute_style (widget, old_theme_node);
+
+      st_widget_add_background_effects(widget, old_theme_node);
```
jaszhix: Maybe we can check if the theme node is eligible before calling `st_widget_add_background_effects`? We should know after `st_widget_recompute_style`.

src/st/st-widget.c
```diff
@@ -1540,6 +1563,61 @@ on_transition_completed (StThemeNodeTransition *transition,
   st_widget_remove_transition (widget);
 }
 
+static void st_widget_add_background_effects (StWidget    *widget,
+                                              StThemeNode *old_theme_node )
+{
+StThemeNode *new_theme_node = st_widget_get_theme_node (widget);
```
jaszhix: We could probably make `st_widget_recompute_style` return the `new_theme_node` passed as an argument to this function and avoid a function invocation in the paint cycle.

src/st/st-widget.c
```diff
@@ -1540,6 +1563,61 @@ on_transition_completed (StThemeNodeTransition *transition,
...
+      if (new_theme_node->background_bumpmap != NULL)
+        {
+          if (widget->priv->background_bumpmap_effect == NULL)
+            {
+              const char *bumpmap_path;
+
+              widget->priv->background_bumpmap_effect = (StBackgroundBumpmapEffect *) st_background_bumpmap_effect_new ();
+              bumpmap_path = st_theme_node_get_background_bumpmap(new_theme_node);
+              widget->priv->background_bumpmap_effect->bumpmap_path = strdup (bumpmap_path);
+              for (int i=0;i<4;i++)
```
jaszhix: Spaces here will make the code easier to read.

src/st/st-widget.c
```diff
@@ -1540,6 +1563,61 @@ on_transition_completed (StThemeNodeTransition *transition,
...
+{
+StThemeNode *new_theme_node = st_widget_get_theme_node (widget);
+
+  if (old_theme_node)
+    {
+      if (old_theme_node->background_blur > 0)
+        {
+          if (widget->priv->background_blur_effect != NULL)
+            {
+              g_object_run_dispose (G_OBJECT (widget->priv->background_blur_effect));
```
jaszhix: In `st_widget_recompute_style`, we know if the style has _changed_, so it might be good to use that boolean state instead of assuming any processing is needed because there is an old theme node.

src/st/st-background-effect.c
```diff
@@ -55,377 +77,283 @@ static const gchar *box_blur_glsl_shader =
...
-      return TRUE;
-    }
-  else
-    {
-      return FALSE;
-    }
+static gboolean
+st_background_blur_effect_pre_paint (ClutterEffect *effect)
+{
+  return TRUE;
```
jaszhix: What is the purpose of this function?

brownsr:
@jaszhix Thanks for the review. It may take a while for me to work through
it as I have a heavy workload the next two days and then I am away at the
weekend.  your detailed comments are much appreciated :-)

clefebvre:
Hi guys, I'm prefixing it as WIP for the time being.

brownsr:
Just working through some of these now. Yes, it does show some odd things if you put it on .popup-menu-content.  Not quite sure where .popup-menu-content fits in the processing for popups, but putting it here does not shapshot the screen background, perhaps something to do with the processing of the various layers in the menu.  It also does not use the radiused corners of the popup as a whole, presumably because this element of CSS does not need to have matching radiused corners defined generally.    It works fine in .menu though.

src/st/st-widget.c
brownsr: OK

src/st/st-widget.c
brownsr: OK

src/st/st-widget.c
brownsr: OK

src/st/st-background-effect.c
brownsr: Doesn't appear to be necessary, removed.

src/st/st-widget.c
brownsr: Perhaps, if I can work out how to do that, but there's only a couple of very fast and cheap tests here, so I can't imagine any practicable difference.

brownsr: i.e. the work to set up the effects is only done the first time through.

src/st/st-widget.c
brownsr: I think that will just move the tests in the function to be done preceding it.

src/st/st-widget.c
brownsr: Yes, but this function is also called without being preceded by recompute_style, so that won't work in that instance.

src/st/st-widget.c
brownsr: Yes, I can see that might be possible in that case.  But this function is also called when there is no style change.  Could separate out into two functions, perhaps i.e. have one dedicated to just the case where there is a style change.  I can't see there would be a speed improvement, but it could possibly make the code a tiny bit easier to follow.

src/st/st-widget.c
jaszhix: I see an impact on performance immediately upon using Cinnamon with no CSS changes. You are adding a _lot_ of code to St's paint function. Please try testing in higher resolutions.

src/st/st-widget.c
jaszhix: After #8230 is merged, you could piggy-back off of the change condition I added, to give some idea. I find it hard to believe you can say there will be no improvement on something you haven't tried yet.

src/st/st-widget.c
jaszhix: Sounds like a lot of overhead for a feature maybe 25% of Cinnamon users will take advantage of (optimistically).

src/st/st-widget.c
jaszhix: We have a private struct used by this class, so we can check if the style has changed, and guard this more. This is a paint function, and it _will_ impact latency, even when the user is not interacting with the Cinnamon UI - it is part of the stage.

src/st/st-widget.c
brownsr: Yes, but remember I am only re-implementing something that was already there and had to be turned off last time because it was broken.  There's no hurry as far as I am concerned.  With all the major performance / latency stuff that you've done, the last thing I want to do is to compromise that.  So right now I am waiting for the dust to settle, maybe next release, before looking at this PR again and seeing how it might be improved.

brownsr:
@clefebvre I'm not planning further work on this, as far as I am concerned it's been ready to pull since the last changes.

brownsr:
Segfault during testing, so back to WIP.

HiroshiAhsan:
Brownsr, any progress on this PR?

KAMI911:
What an interesting feature. It is sad to see that the development stalled.

TPO100Cofficial:
if this isn't being worked on by @brownsr anymore, would it be fair for someone else to come in, take the proposed changes from this, and finish the implementation work?

Alex-Sour:
I would love to see this implemented into Cinnamon! Windows like the terminal would look so much better with transparency and blur.

echuber2:
> Segfault during testing, so back to WIP.

Is there information about _what_ exactly had segfaulted? Is there a test case that's still reproducible? Does this need to be started over from scratch given how much time has passed?

rdlf4:
> > Segfault during testing, so back to WIP.
> 
> Is there information about _what_ exactly had segfaulted? Is there a test case that's still reproducible? Does this need to be started over from scratch given how much time has passed?

We're left with no answers. No way to reproduce, no info on repository, let alone what is required and what dependencies need to be taken care of, cinnamon compilation instructions, additional files or github links. Nada.
It's almost as if the foundation of Cinnamon were so sensitive that they don't feel like sharing anything about it or letting anyone in. And when we look for those answers ourselves, we typically end up with a NOT FOUND page or a project lost in time, which was started by a random person (not the linuxmint team, of course) and then "real life called, and they had to abandon it".
And that's exactly how your linux mint 21 experience is going to be like, just as the previous version.

TPO100Cofficial:
god, i can't believe linux mint dev team bad¹
[1] don't have enough resources or at _worst_ are putting them in the wrong places and thus doing work that has somewhat subpar results

which, is a shame!! because cinnamon is a really cool desktop environment imo - it's a bit buggy, but it gets way less love than it deserves (themes that work for cinnamon are...not that diverse in distinct styles and extensions are not very plentiful!!) and xapps are like...the only competition to gnome apps i've found to be comfortable at all ²
> [2] clarification: _(kde is in my experience buggy and has ui design choices i really don't like, gnome is oversimplified and some app developers have started saying "don't theme my app", to which... screw off!)_

i really hope what happened to firefox/mozilla doesn't happen to linux mint, where the once gold standard loses its focus from bad management but noone has the resources to fork it well and now they're making....5 built-in themes with bad text contrast, for _a limited time?_ with all that donation money (are last update's mint-y tweaks comparable to that?)

@rdlf4 so like, what do you suggest we (not me or anyone specific, people somewhat critical of mint in general) do about it?

rdlf4:
> god, i can't believe linux mint dev team bad¹ [1] don't have enough resources or at _worst_ are putting them in the wrong places and thus doing work that has somewhat subpar results
> 
> which, is a shame!! because cinnamon is a really cool desktop environment imo - it's a bit buggy, but it gets way less love than it deserves and xapps are like...the only competition to gnome apps i've found to be comfortable at all
> 
> ( i really hope what happened to firefox/mozilla doesn't happen to linux mint, where the once gold standard loses its focus from bad management but noone has the resources to fork it properly and now they're making....5 built-in themes with bad text contrast, for _a limited time?_ with all that donation money ( are last update's mint-y tweaks comparable to colorways? ) )
> 
> @rdlf4 so like, what do you suggest we (not me or anyone specific, people somewhat critical of mint in general) do about it?

Alright, you're bringing a lot to the table here, so I feel like focusing on what's being discussed - in no way do I wish the Mint team to lose focus on supporting Cinnamon -- matter of fact, by restyling Cinnamon means supporting it even further, but on the visual style department.
And it's not like the community wants a KDE 2.0 (gosh why did you even drag KDE into this?), No, it's not that! It's just that blatant solid color that's been there for all these years, that has to change. And BECAUSE TRANSPARENT CINNAMON THEMES are very popular - and therefore a common topic, by adding Blur effect on top of it, system elements/controls such as Menu, panel, applets, etc would benefit by looking more modern and offering a fresh redesign.
And lets not kid ourselves here, we know that Blur does impact the system if implemented poorly or with the wrong settings, which could lead to GPU meltdown (a little bit of exaggeration there, but you get the picture) - that's what Cinnamon tweak settings should be for.

So give us Blur, and let us play around with it. Those with a beefier apu from a desktop computer can make it look prettier than those with a mobile apu, so having options to suit our needs and system settings is important.
And for those who are like "Oh what's the problem looking the same after all these years?" all I can say is fine. You don't want it, that's on you. But nothing remains looking the same over time, and offering Cinnamon a more modern look shows that the team is keeping up with the time and with the community's request. (We've been trying this by ourselves after all, so why turn a blind eye on it?)

Don't get it twisted, I'm not saying "do it today! Or tomorrow! Or for the next release!" No, I'm asking the team to bring their attention to this topic and feature request, tell us how do-able it is with the current version of Cinnamon (does it allow for such a thing without compromising the Cinnamon team's agenda? Could it be done in an official way, meaning further supported for future Linux Mint releases?) - so some clarification from the team on this topic would be a nice place to start from.

I'm sorry for the long post.

TPO100Cofficial:
@rdlf4 oh! yeah
if they can't do it now, that's alright, but it feels really odd that they haven't done it with how many cinnamon themes use transparency
which got me suspecting maybe there could be mismanagement, but that's just a gut feeling i took way too far

PR #8447, opened 2019-03-08
melange/all: huge style refactor

itzexor:
This is about as good as I can get it. It makes pylint warnings at least somewhat usable, from ~400 warnings to like 60 or less....

pylintrc:
- add unused-argument and wrong import posistion to ignored warnings since
  we trigger these often with gi callback signatures and the way we have
  to specify gi lib versions before import via a function call.
- ignore some common size/width vars from naming restrictions

melange:
- convert everything except dbus methods to snake_case
- various other style cleanups here and there
- remove unused memory page, as this is no longer supported by cjs
  and i haven't seen any replacement functionality.


files/usr/share/cinnamon/cinnamon-looking-glass/cinnamon-looking-glass.py
```diff
@@ -161,30 +166,32 @@ def __init__(self, parent):
...
+                    combo.set_active_iter(self.store.insert(1, ["user", new_file]))
                 else:
                     combo.set_active(-1)
             return False
 
-    def isValid(self):
-        return self.entry.get_text() != "" and self.filename != None and os.path.isfile(os.path.expanduser(self.filename))
+    def is_valid(self):
+        return (self.entry.get_text() != "" and
+                self.filename is None and
```
itzexor: should be `is not None`

files/usr/share/cinnamon/cinnamon-looking-glass/cinnamon-looking-glass.py
```diff
@@ -209,28 +216,28 @@ def selectFile(self):
         return result
 
 class FileWatchHandler(pyinotify.ProcessEvent):
-    def __init__(self, view):
+    def myinit(self, view):
```
itzexor: this is the correct way to override ProcessEvent. you aren't supposed to override __init__

Issue #8453, opened 2019-03-11
How to install search providers?


Fabian42:
```
 * Cinnamon version 3.2.7
 * Distribution - debian 9.8
 * 64 bit
 ```

**Issue**

The cinnamenu settings say "No search provider extensions found", but I don't see how I would install the search providers. They're not in Synaptic, dragging from here to the settings window doesn't work (would be awesome), they're not in the extensions list, I see no readme anywhere here, …

I assume I have to put the files somewhere, but where?

(I'm talking about [these](https://github.com/linuxmint/cinnamon/tree/master/docs/search-providers-examples),)

**Comment by Odyseus** [comment]
Hello, @Fabian42.

You can install search providers into `~/.local/share/cinnamon/search_providers`. The folder structure should look something like this:

```
[~]
└──.local
    └──share
        └──cinnamon
            └──search_providers
                └──apt@cinnamon.org
                └──calc@cinnamon.org
                └──chromium-history@cinnamon.org
                └──trackerprovider@cinnamon.org
                └──wikipedia@cinnamon.org
                └──yahoo@cinnamon.org
```

After installing search providers, Cinnamon needs to be restarted or the Cinnamenu applet reloaded for the search providers to appear in Cinnamenu's settings window.


The default Cinnamon menu applet also supports search providers, but they need to be enabled globally for them to be used by the default menu.

You can use the command line to enable search providers globally:

```shell
gsettings set org.cinnamon enabled-search-providers [\"calc@cinnamon.org\"]
```

Or dconf Editor (GUI):

![Screenshot-area-2019-03-11-104422](https://user-images.githubusercontent.com/3822556/54128430-d0713880-43ea-11e9-851b-d43d5ac93930.png)

Fabian42:
Great, works! This should be in some kind of readme or, even better, in a help text in the Cinnamenu settings.

My findings so far:
* The Wikipedia search results have no icon.
* The apt-get search results have a "forbidden" icon
* The calculator changes the results screen to only show the calculation and result, but there are still tooltips of other results and they can be navigated and selected with the keyboard.
* The Chromium history search provider does not work with Chrome (as expected, but it could have).
* I don't know what the "Tracker search" provider does. I haven't managed to find a search result from it.
* I didn't try Yahoo.

Should these be reported? The folder is called "example search providers", that sounds like they aren't meant as full products. But it can't hurt to fix bugs, right?

Odyseus:
> The Wikipedia search results have no icon.

[...]

PR #8473, opened 2019-03-26
Implement simple calculator in Cinnamon Menu

obaqueiro:
A simple but straightforward way to implement a calculator in the Cinnamon Menu. 

I liked how in OSX Spotlight search you have the ability to write some math and it evaluates it for you. 

With very minimal google-fu to understand how Cinammon applets work, I managed to scratch my own itch and implement something similar.

Even though using  'eval' is fugly, the regex in the back makes it somewhat safer... so that I don't have to implement a math parser.  

Added to my repo for posterity. Feel free to modify and improve as you see fit.


NikoKrause:
You can try the calculator search provider: https://github.com/linuxmint/cinnamon/issues/8453

clefebvre:
Same here, I think we're better off using a search engine for this and not using eval like that.

PR #8477, opened 2019-03-28
main: Misc. clean up

jaszhix:
https://github.com/linuxmint/cinnamon/commit/7ae1c33c6d403959e8a7bd0d11e13ebef8699b90

- Removes eval usage.
- Removes `_addXletDirectoriesToSearchPath`. The imports[xlet] prefix is no longer used in spices.
- Uses length caching for loops.

Includes #8230 to avoid a conflict.


brownsr:
I've given this a quick run through and could not see any obvious problems. I found a zoom in issue, but this appears to be coming from muffin 410, so have logged this on the muffin tracking issue.

PR #8487, opened 2019-04-01
cinnamon-doc-system: tidyup of recents

brownsr:
slists and glists were mixed up, there was a compile warning with the macro,
and I took the opportunity to put in the normal prepend/reverse trick


clefebvre:
@brownsr why use append/reverse? is it faster?

brownsr:
Yes. Appending to the end means a full list scan each time. Putting it at
the start is much faster. So adding everything at the start and then
reversing the list is faster overall. You can find that trick in lots of
places in our code Base. As this list is just twenty entries long the
actual savings here are miniscule,  but the principle is good.

PR #8488, opened 2019-04-02
st-texture-cache: use a slightly more efficient linked list method

brownsr:
difficult to test not knowing how many sliced images are flying around.  A batch of menu load tests came up around 20ms faster afterwards on average, though with high variability in the test results both before and after so I might be kidding myself.  No regressions in evidence, not that I would expect that as this is a well tried approach.  https://developer.gnome.org/glib/stable/glib-Doubly-Linked-Lists.html#g-list-append

### DIFF
--- a/src/st/st-texture-cache.c
+++ b/src/st/st-texture-cache.c
@@ -1019,9 +1019,10 @@ load_sliced_image (GTask        *result,
         {
           GdkPixbuf *pixbuf = gdk_pixbuf_new_subpixbuf (pix, x, y, data->grid_width, data->grid_height);
           g_assert (pixbuf != NULL);
-          res = g_list_append (res, pixbuf);
+          res = g_list_prepend (res, pixbuf);
         }
     }
+  res = g_list_reverse (res);
   /* We don't need the original pixbuf anymore, though the subpixbufs
      will hold a reference. */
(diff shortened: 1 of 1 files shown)


jaszhix:
It would be good, going forward, to test overall performance when proposing these kind of patches. On my machine this patch along with #8487 causes the mouse cursor to move fast and slow at random times. I think the variability makes more of an impact at 60FPS, and any high variability is a bad thing when we are dealing with asynchronous frame timings with Xorg, and changing functions that affect how quickly Clutter gets through its redraw queue. 

I'll try to test these with gputest in a bit, though it's not really as good as just looking at _where_ in Cinnamon this code is used - it's nowhere near just the menu, but that's all that was tested here.

As an aside, it would be nice if C patches were open for more than one day, so there's an opportunity to review.

brownsr:
I don't see how this can be causing variability.  It's a full list load up of the same number of elements in either case before or after.  This really is a widely used method, there are quite a few instances of it in muffin and clutter.

jaszhix:
Sorry, I was testing this and #8487 at the same time. After some benchmarking and manual testing, #8487 appears to be the cause.

image

https://openbenchmarking.org/result/1904083-SP-APRIL848836

Issue #8285, opened 2019-01-12
LMDE3 Cinnamon 4.0.9 cinnamon-settings power crash and power applet unable to get battery state.


smurphos:
```
 * Cinnamon version 4.0.9
 * Distribution - LMDE3
 * Graphics hardware *and* driver used
Graphics:  Card: InnoTek Systemberatung VirtualBox Graphics Adapter
           bus-ID: 00:02.0 chip-ID: 80ee:beef
           Display Server: X.Org 1.19.2 drivers: modesetting (unloaded: fbdev,vesa)
           Resolution: 1360x768@59.80hz
           GLX Renderer: Chromium
           GLX Version: 2.1 Chromium 1.9 Direct Rendering: Yes
 * 64 bit
 * Attach /home/<username>/.xsession-errors, or /var/log/syslog
 ```

**Issue**
The power settings module is crashing in LMDE3 following update to Cinnamon 4.0.9. I've reproduced this issue in a Virtualbox install - however a couple of users on bare-metal have also reported it at the Mint forum - see https://forums.linuxmint.com/viewtopic.php?f=247&t=285476. On the same hardware in another virtualbox install main edition 19.1 is not showing the problem. The power applet is also unable to detect a battery state. Both worked fine on Cinnamon 3.8.6.

LMDE
```
UPower client version 0.99.4
UPower daemon version 0.99.4
```
Mint 19.1
```
UPower client version 0.99.7
UPower daemon version 0.99.7
```

Please see pastebin link for output of cinnamon-settings power, upower -e, upower -d, xsession-errors and syslog. I've got no mintreport in LMDE to get a crash log

https://pastebin.com/4eQqrKK2

**Steps to reproduce**

Launch power settings from menu, power applet or via cinnamon-settings power

**Expected behaviour**

No crash.

**Other information**

smurphos:
Can be resolved by manually installing the following from the buster repos.

libupower-glib3_0.99.9-2
gir1.2-upowerglib-1.0_0.99.9-2

smurphos:
upower_0.99.9-2 is also needed to allow the power applet to display the power state in its pop-up menu.

mhd-malek:
> upower_0.99.9-2 is also needed to allow the power applet to display the power state in its pop-up menu.

The 3 packages already installed and same problem

pioruns:
Confirming. Using real hardware, fresh install LMDE Cindy.

[...]

PR #8493, opened 2019-04-04
Fix settings power module in LMDE 3

clefebvre:
Debian's version of UPowerGlib doesn't have DeviceLevel.FULL/HIGH/NORMAL...

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py
@@ -471,7 +471,14 @@ class Module:
         self.battery_label_size_group.add_widget(label_box)
         hbox.pack_start(label_box, False, False, 0)
 
-        if battery_level == UPowerGlib.DeviceLevel.NONE:
+        battery_level_detection = True
+        try:
+            UPowerGlib.DeviceLevel.FULL
+        except:
+            print("UPowerGlib doesn't support UPowerGlib.DeviceLevel.FULL, looks like an old version. Skipping battery level detection.")
+            battery_level_detection = False
+
+        if battery_level == UPowerGlib.DeviceLevel.NONE or not battery_level_detection:
             label = Gtk.Label()
             label.set_markup("%d%%" % int(percentage))
             label.set_size_request(30, -1)


clefebvre:
This is an attempt to fix https://github.com/linuxmint/cinnamon/issues/8285... it probably needs more work, because we seem to get an int when calling set_device_battery_primary(). See the traceback:

```
$ cinnamon-settings power
True
Using pam module (python3-pampy)
Loading Power module
  File "/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py", line 619, in <module>
    window = MainWindow()
  File "/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py", line 329, in __init__
    self.go_to_sidepage(cat, path, user_action=False)
  File "/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py", line 133, in go_to_sidepage
    sidePage.build()
  File "/usr/share/cinnamon/cinnamon-settings/bin/SettingsWidgets.py", line 123, in build
  File "/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py", line 186, in on_module_selected
    self.build_battery_page()
  File "/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py", line 291, in build_battery_page
    primary_settings.add_row(self.set_device_battery_primary(device))
  File "/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py", line 379, in set_device_battery_primary
    details = UPowerGlib.Device.state_to_string(state)
TypeError: Expected a UPowerGlib.DeviceState, but got int
```

My commit above doesn't fix that.

While trying to reproduce the issue, I bumped into the fact that LMDE's version of the lib didn't support DeviceLevel.FULL/HIGH/NORMAL, so I fixed that.

Another fix needed is to do something similar in the applet.

@mtwebster can you have a look, see if you like my fix and if we go that way? Don't hesitate to add/revert as you see fit.

mtwebster:
I changed the c-s-d power module to prevent this from happening in the first place: https://github.com/linuxmint/cinnamon-settings-daemon/commit/4842c2d8013577ff323732f0a384815a4a150afa

It seemed the easier fix (and there was still a runtime warning about the missing property that would have needed to be fixed anyhow).

mtwebster:
This should be unnecessary now.

PR #8502, opened 2019-04-09
cinnamon-doc-system: Switch back to GSList

jaszhix:
GSList usage seems to perform better in CinnamonDocSystem. This switches back to it while preserving the warning reductions with appropriate casting, and the prepend + reverse trick. Not seeing any GCC warnings from cinnamon-doc-system.c.

Measurements:

image

https://openbenchmarking.org/result/1904086-SP-APRIL848805&obr_sor=y

### DIFF
--- a/src/cinnamon-doc-system.c
+++ b/src/cinnamon-doc-system.c
@@ -23,7 +23,7 @@ static guint signals[LAST_SIGNAL] = { 0 };
 
 struct _CinnamonDocSystemPrivate {
   GtkRecentManager *manager;
-  GList *infos_by_timestamp;
+  GSList *infos_by_timestamp;
 
   guint idle_recent_changed_id;
 };
@@ -45,7 +45,7 @@ static const int MAX_RECENT_ITEMS = 20;
  *
  * Returns: (transfer none) (element-type GtkRecentInfo): Cached recent file infos
  */
-GList *
+GSList *
 cinnamon_doc_system_get_all (CinnamonDocSystem    *self)
 {
   return self->priv->infos_by_timestamp;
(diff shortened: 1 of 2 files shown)


collinss:
Not that I'm doubting you, but did you by chance test this with both a big and a small list? It's always possible with this sort of thing that one method is better for small lists while another is better for large ones (especially when sorting is involved like it is here). If that's the case here then we should consider which one has the more significant gains before deciding which method to use.

Also, in this case I would think logging the run time of these functions give us a better picture of what's going on in this case. (I'm not an expert in benchmarking, mind you, but I am a scientist, and the general rule of thumb for getting the most accurate data is to eliminate as many sources of noise as possible.)

jaszhix:
I think the biggest problem is that this is getting called when any files change, and so it's keeping the thread busy when a lot of file I/O is happening. This can explain why some people experience latency from the original change, and others do not. Since this runs on the same thread as the compositor, the increased latency is visible. It might be a good idea to split this into another process somehow, and then have Cinnamon lazy load it from cache as needed.

brownsr:
This is actually really interesting.  I like to get rid of warning
messages,  they clutter up build files and occasionally indicate something
that is a real problem. That was the case here, glist and glist structures
were being mixed up. That could well segfault in a different architecture,
compiler, compiler version, change to the data structures or perhaps to
changes to glib internals. It's not safe practice unless designed for (like
in gobject structures I presume). No way a change to an in memory list of
twenty elements should be having anything more than a miniscule performance
effect, for something that is intended to track recently modified files. If
you are seeing an effect then that would seem to imply that the routine is
being called vastly more than it should be, eg on any file access rather
than on changes. Perhaps.  I am on mobile so can't go digging into it now.
If so then investigating that might be a profitable thing to look at.

brownsr:
Sorry, autocorrect strikes again. 'glist and glist' should be 'glist and
gslist '

jaszhix:
The structs are compatible and use the same free functions if you dig into glib deep enough. While I agree this is potentially an issue, up until now we haven't had any bugs related to this as far as I can tell. I generally prefer to look at a problem and then try to find a solution, but when doing that in reverse and touching up warnings without a known issue associated, in old stable code, it makes me a little nervous.

> No way a change to an in memory list of twenty elements should be having anything more than a miniscule performance effect, for something that is intended to track recently modified files. If you are seeing an effect then that would seem to imply that the routine is being called vastly more than it should be

That's outside of the scope of this PR. I agree, but until we can figure that part out this restores performance to where it was before. I would spend more time benchmarking but I'm out of time. I run docker, resilio sync, and have a ton of file IO in the background, this function is constantly getting called. Please don't take this partial revert personally, I'm just trying to make cinnamon usable for me.

brownsr:
Oh I don't take it personally at all, and I don't have issues with it. I
just find it interesting that it has probably exposed a significant
underlying issue.  Which can be addressed later ;-)

Issue #8298, opened 2019-01-15
Can't overwrite older .face file after security fix was applied


dilinger:
```
 * Debian
 * 64 bit
 ```

**Issue**
A security hole was fixed in https://github.com/linuxmint/Cinnamon/pull/7683. However, this fix is incomplete - users who previously had a .face file owned as root are unable to overwrite it, as the new code drops privileges. Here's the error I get:

```
Traceback (most recent call last):
  File "/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py", line 709, in _on_face_menuitem_activated
    shutil.copy(path, os.path.join(user.get_home_dir(), ".face"))
  File "/usr/lib/python2.7/shutil.py", line 119, in copy
    copyfile(src, dst)
  File "/usr/lib/python2.7/shutil.py", line 83, in copyfile
    with open(dst, 'wb') as fdst:
IOError: [Errno 13] Permission denied: '/home/dilinger/.face'
```

**Steps to reproduce**

Run an older unfixed version; in my case, I'm running the version that's currently in Debian stable:
ii  cinnamon       3.2.7-4      amd64        Innovative and comfortable deskto

Run cinnamon-settings-users (which runs as root using gksu)

Click on a user, click on the picture, and choose a new picture for that user.

At this point, there should be a .face file in that user's home directory that is owned by the root user

Close the cinnamon-settings-users window.

Now apply the security update, or upgrade to a newer version of Cinnamon.  In my case, I applied the security fix.

Open up cinnamon-settings-users again.

Click on a user, click on a picture, and choose a new picture for that user.

Instead of overwriting the .face file, it won't be able to update it.  The old root-owned .face file remains.

**Expected behaviour**

The .face file should always be overwritten by a new one that is owned by the user (rather than root).

**Other information**

**Comment by dilinger** [comment]
Here's a patch that fixes this issue.  It's pretty simple; just delete the old .face file before dropping privileges.

```
--- /usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py.orig2019-01-14 17:24:37.799003654 -0800
+++ /usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py	2019-01-14 17:38:06.248595816 -0800
@@ -675,6 +675,7 @@
                 image.thumbnail((96, 96), Image.ANTIALIAS)
                 face_path = os.path.join(user.get_home_dir(), ".face")
                 try:
+                    os.remove(face_path)
[...]

PR #8515, opened 2019-04-15
cinnamon-settings-users.py: update symlink vulnerability fix

dilinger:
A security hole was fixed in #7683, where a user's .face file being
modified with root privileges allowed a symlink attack to overwrite
root-owned data.

Unfortunately, this fix is incomplete - simply dropping privileges
means that users who previously had an older .face file would be
unable to overwrite it. The old code left .face owned by root.

This fix first removes the old .face file as root, then drops
privileges and writes the new .face file.

This bug is explained further in #8298


dilinger:
If Cinnamon's only aiming for Python 3, I could change that try/except to the following instead:

`
with contextlib.suppress(FileNotFoundError):
    os.remove(face_path)
`

PR #8547, opened 2019-05-06
theme settings: Add capability to override the active theme's scrollbar size and click behavior

mtwebster:
Pretty much done.  Only remaining concerns:

- A 1.5 default multiplier is used for gtk2 programs.  The size there is applied to the entire widget's size, not just the 'handle bar'.  In gtk3, the value applies just to the handle.  1.5 looks good for Mint-Y, not so much in Mint-X.
- The multiplier can be changed via a hidden gsettings key.  We could use baked-in multipliers for known themes?
- hidpi - gtk2 is not scaled, so given a 1.5 multiplier for normal dpi, you'd need a 3.0 multiplier for hidpi mode so gtk2 programs look correct.  For now nothing is done here - should there be an extra switch to 'compensate for hidpi', or, when the settings is changed in cinnamon-settings, check for hidpi then and compensate.  (This would only 'fix' things once the setting is altered)
- should this affect cinnamon?

image


smurphos:
Would it make sense to also move the Overlay scroll bars toggle from Effects to the Themes GUI so all the scrollbar related settings are in one place?

mtwebster:
Thanks, you're right.

Got a little more carried away:
Screenshot from 2019-05-06 21-11-08

The little test widget updates when the settings are changed, to reflect whatever is active.  In the end the user will have to either log out and back in, or restart all programs.  There will need to be a note to that effect I suppose.

I may look into gtk2 scrollbar width adjustment, programs using that toolkit are less and less common, but are still around.

mtwebster:
Added a restart recommendation:
Screenshot from 2019-05-18 23-09-50

Not entirely sure I'm happy with it though.

lestcape:
>  all the more important to be able to show a demo widget?

If you don't find how to show a real demo widget, probably one way you have is draw the result instead of create a real widget. I think in python draw Gtk widget to a Cairo context will work well (in gjs is not working yet). The idea of make it in that way was taken from the Firefox code and also from how Mutter is render the title bar. So, the credit of this idea is from this projects.

As an example, this attempt to display the window buttons in that way was wrote in gjs, but can easily be ported to Python:
https://stackoverflow.com/questions/50313744/how-to-darwing-the-window-buttons-from-a-gtk-context-in-gjs
https://gitlab.gnome.org/GNOME/gjs/issues/167

mtwebster:
Hi @lestcape I just ended up using a GtkPlug/Socket for now (files/usr/share/cinnamon/cinnamon-settings/bin/scrollbar-test-widget.py) - it's not the greatest (there's a flicker when it reloads itself that's sort of unavoidable) but it does the job and was fairly simple.

It's not wayland compatible I'm sure but that's not a bridge I need to cross right now :)

lestcape:
Ah okay, it's good that you've already found a way... I only provided some ideas but i can not ensure it really will be functional at all as is not tested by me specifically for what you want to do.

**Note:** I do not know who get the idea about to control the style of Gtk in real time, but I think this it's a very good idea. Mainly, because things like [that](https://stopthemingmy.app/) happens.  Thankfully here there is a counterpart that gives to the user the freedom to modify a Gtk theme. To me it's very nice to see this pull. So, please make it happens...

Have a good day ahead...

PR #8550, opened 2019-05-09
Create unique names for desktop apps with duplicate display names.

mtwebster:
need to:

- ~~prettify exec or flatpak names (capitalize, strip paths)~~
- ~~handle OnlyShowIn/NotShowIn cases with identical executables (i.e. nautilus has two - one for gnome/unity, another for *not* gnome/unity.  Cinnamon-menus accepts both as valid for a cinnamon session.~~
- ~~implement a hardcoded list if necessary for corner cases~~
- optimize

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -2276,10 +2276,9 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
         while ((nextType = iter.next()) != CMenu.TreeItemType.INVALID) {
             if (nextType == CMenu.TreeItemType.ENTRY) {
                 let entry = iter.get_entry();
-                let appInfo = entry.get_app_info();
-                if (appInfo && !appInfo.get_nodisplay()) {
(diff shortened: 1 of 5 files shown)


mtwebster:
Some sample behavior:

Screenshot from 2019-05-18 22-01-50
Screenshot from 2019-05-18 22-00-31
Screenshot from 2019-05-18 22-00-47

- Some flatpak apps distinguish themselves without the need for our interference (nautilus 3.28)
- We never show the executable name for flatpaks for now, only 'Flatpak'.  This is mainly because
  the normal executable name we'd get from their desktop file will always be 'flatpak' (the launcher).
  We could try to tease out the actual name from the Exec string, but I'm not sure how reliable it would
  be...
- unrelated, flatpaks mask normal apps if they share the same desktop file name (regardless of their
  full path).  This could be fixed in cinnamon-menus, but I'm not sure really how large a problem it
  will be, to be worth dealing with.
- For now, it's assumed there will only be a single flatpak version of anything.  It's quite possible for a
  user to have added other repos (like nightly or dev channels) that would make duplicates of flatpaks.
  Still looking into this a bit..

PR #8555, opened 2019-05-11
Workspace switcher: shade labels of empty workspaces in simple button mode

pwfrank:
- Darkens labels of empty workspaces in "simple buttons" mode. Workspaces with only minimized windows are considered "not empty"

This PR includes a workspace-button pseudoclass in cinnamon.css to define the darkened label. I'm not sure if this is acceptable for compatibility reasons, but I'm not sure of another way to accomplish this

### DIFF
--- a/data/theme/cinnamon.css
+++ b/data/theme/cinnamon.css
@@ -1537,6 +1537,10 @@ StScrollBar StButton#vhandle:hover {
     box-shadow: inset 0px 0px 8px 1px rgba(255,255,255, 0.7);
 }
 
+.workspace-button:shaded {
+    color: #13191c;
+}
+
 /* Controls the style when using the "Visual representation" option */
 .workspace-graph {
     padding: 3px;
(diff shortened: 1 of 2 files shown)


germanfr:
I'm not sure I understand what this does or why this is neccessary, but this piece of code:

```js
update() {
        let empty = true;
        let windows = this.workspace.list_windows();
        windows = windows.filter( Main.isInteresting );
        if (windows.length !== 0) {
            empty = false;
        }
        this.shade(empty);
    }
```

 can be simplified to:

```js
update() {
        let windows = this.workspace.list_windows();
        let empty = windows.some(Main.isInteresting);
        this.shade(empty);
    }
```

which doesn't duplicate the windows array each time :)

pwfrank:
Thanks for the code suggestion. I don't do JS much, so I appreciate the input

> I'm not sure I understand what this does or why this is neccessary

Right now, the workspace-switcher applet's simple-button mode is much less usable than its "visual representation" (workspaceGraph) mode simply because there's no way to tell if a workspace contains windows just by looking at the applet. For example, if I want to start working in an empty workspace, I have to "move to next workspace" repeatedly until I find one. In "visual representation" mode, you simply have to look at the applet to find a workspace with no windows, to which you can then switch using a "switch to workspace X" keyboard shortcut. 

This PR has the applet display "simple button" labels of unoccupied, unfocused workspaces in a darker color to make it clear what workspaces are unused at any time (shade(used) ff18dea, using attributes defined by .workspace-button:shaded in cinnamon.css 22741c8)

PR #8556, opened 2019-05-12
[WIP] Fix window effects clones animating above all windows

jaszhix:
This still causes the clone to map above all windows for a split second. Might need to initialize the clone with `show_on_set_parent` set to false, and delay hiding the window actor in muffin a little longer.

### DIFF
--- a/js/ui/windowEffects.js
+++ b/js/ui/windowEffects.js
@@ -24,8 +24,9 @@ class Effect {
             y: source.y,
         });
 
-        global.overlay_group.add_child(this.actor);
-        global.overlay_group.set_child_above_sibling(this.actor, null);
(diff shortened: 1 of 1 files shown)


mtwebster:
This is still pretty noticeable on the laptop I'm testing on (not the fastest but not the slowest either).  I think once it was effective, then on all subsequent closes you could still see the clone on top.

jaszhix:
I worked on this over the weekend, and my suggested change worked but the shadows are still layered on top. While this does need some more thought and consideration, I don't think it should block anything for 4.2.

jaszhix:
Fixed in https://github.com/linuxmint/cinnamon/pull/8596

PR #8575, opened 2019-05-19
Dev/virtual keyboard icon

joequant:
Change virtual keyboard icon to something that is in the adwaita icon theme


clefebvre:
Implemented by https://github.com/linuxmint/cinnamon/commit/f0c926cdc65f35c213e10d79f1bf3fa1150407df (it has to be done in the generate script.. the file you modified here is a generated one).

PR #8576, opened 2019-05-20
st: Move processing after early returns where possible

brownsr:
saves a little processing when functions have early returns. Most of the examples are just
cheap initializations, but there are a few that are a little bigger.


src/st/st-scroll-view.c
```diff
@@ -550,8 +551,8 @@ st_scroll_view_get_preferred_height (ClutterActor *actor,
                                      gfloat       *natural_height_p)
 {
   StScrollViewPrivate *priv = ST_SCROLL_VIEW (actor)->priv;
-  StThemeNode *theme_node = st_widget_get_theme_node (ST_WIDGET (actor));
-  gboolean account_for_hscrollbar = FALSE;
+  StThemeNode *theme_node;
+  gboolean account_for_hscrollbar;
```
jaszhix: This will be initialized as `0` implicitly, which is the `FALSE` value. I think explicit declaration here is cleaner.

src/st/st-box-layout.c
```diff
@@ -380,7 +380,7 @@ st_box_layout_paint (ClutterActor *actor)
 {
   StBoxLayout *self = ST_BOX_LAYOUT (actor);
   StBoxLayoutPrivate *priv = self->priv;
-  StThemeNode *theme_node = st_widget_get_theme_node (ST_WIDGET (actor));
+  StThemeNode *theme_node;
```
jaszhix: I would initialize this with NULL, so we're not introducing overhead from referencing random memory and not using it.

src/st/st-theme-node.c
```diff
@@ -2644,23 +2653,35 @@ const PangoFontDescription *
...
-  char *family = NULL;
+  PangoStyle font_style;
+  gboolean font_style_set;
+  PangoVariant variant;
+  gboolean variant_set;
+  PangoWeight weight;
+  gboolean weight_absolute;
+  gboolean weight_set;
+  double size;
+  gboolean size_set;
```
jaszhix: What is wrong with just defining these after the return? We're sacrificing readability for less noisy compilation.

src/st/st-scroll-view.c
brownsr: OK

src/st/st-box-layout.c
brownsr: OK

src/st/st-theme-node.c
brownsr: OK.  could rework as if ..else with the declarations in the else, but the extra indention won't help readability

jaszhix: I really think we should be basing these decision off of profiling the code. We're making shots in the dark here. Do we know that the early returns are the path most traveled? Otherwise, if the early return conditions are true very often, we're adding an extra step in these functions.

brownsr: sorry, don't get it.  What extra step ? I am not making any assumptions about early return path frequency.

jaszhix: You're mass applying the same optimization everywhere in St. I think you are making quite an assumption because I don't think you've profiled all of these code paths. Now multiply "cheap" by 60 frames per second, that's how often these functions can get called.

brownsr: No , I haven't profiled all the paths, quite true.  That would probably take an absolute eternity, and I'm not sure I could even readily separate them all out.  I could do each change one at a time, benchmark, and see which ones make a difference.  As the code changes are so small, that would take a long time.  And I don't expect any to be negative by the nature of the change, so that would seem an awful lot of work.  The improvement is modest, about +0.9% on a quick check, and that's more than I expected as the same thing in muffin was just 0.3%. I'm quite happy grinding out this small stuff.  And I usually spot things along the way doing it, your comment elsewhere about the multiple draw_framebuffer calls has led me to another small optimisation of similar scale, so that's PRd now as well.

Issue #7945, opened 2018-10-18
Unable to download applets, themes, desklets and extensions when using network proxy


jesperloehr:
* Cinnamon 3.8.9
 * Mint 19
 * 64 bit

**Issue**
After upgrading from Cinnamon 3.8.8 to 3.8.9 it is not possible to refresh cache or download when using Applets, Themes, Desklets or Extensions in System Settings. It gives the error: "An error occured while trying to access the server. Please try again in a little while. Details: timed out."

**Steps to reproduce**
I only have internet access through a proxy server. Proxy is configured in network settings. 

**Expected behaviour**
The System Settings application should use the configured proxy.

**Other information**
I identified the problem is from a change in /usr/share/cinnamon/cinnamon-settings/bin/Spices.py installed from the cinnamon-common package.
In 3.8.9 it has been changed to use http.client.HTTPSConnection to fetch updates and there is no code that checks for the proxy configuration.
```
def _url_retrieve(self, url, outfd, reporthook, binary):
    #Like the one in urllib. Unlike urllib.retrieve url_retrieve
    #can be interrupted. KeyboardInterrupt exception is raised when
    #interrupted.
    count = 0
    blockSize = 1024 * 8
    parsed_url = urlparse(url)
    host = parsed_url.netloc
    try:
        connection = HTTPSConnection(host, timeout=15)
        headers = { "Accept-Encoding": "identity", "Host": host, "User-Agent": "Python/3" }
        connection.request("GET", parsed_url.path, headers=headers)
        urlobj = connection.getresponse()
        assert urlobj.getcode() == 200
```

In 3.8.8 it used urllib.request.urlopen. 
```
def _url_retrieve(self, url, outfd, reporthook, binary):
    #Like the one in urllib. Unlike urllib.retrieve url_retrieve
    #can be interrupted. KeyboardInterrupt exception is raised when
    #interrupted.
    count = 0
    blockSize = 1024 * 8
    try:
        with urlopen(url, timeout=15) as urlobj:
            assert urlobj.getcode() == 200
```
Proxy configuration is configured for urllib.request in /usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py.
```
if __name__ == "__main__":
    import signal

    ps = proxygsettings.get_proxy_settings()
    if ps:
        proxy = urllib.ProxyHandler(ps)
    else:
        proxy = urllib.ProxyHandler()
    urllib.install_opener(urllib.build_opener(proxy))
```
[...]

PR #8589, opened 2019-05-28
fix to use proxy settings when available on cinnamon-settings

gabmed:
This PR is related to this issue: https://github.com/linuxmint/Cinnamon/issues/7945

Currently, cinnamon-settings (applets,desklets,etc) is ignoring system-wide proxy settings.
This fix uses proxy settings (when available) to connect to spices hosts.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
@@ -15,6 +15,7 @@ try:
     import dbus
     from PIL import Image
     import datetime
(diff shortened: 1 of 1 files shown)


gabmed:
Use of HTTPSConnection is already in master code, not added by this PR :/

collinss:
I wouldn't worry too much about that. I doubt anyone is using <= python 3.4.3 anyway. Codacy and CodeFactor aren't perfect either, so failing a check from them doesn't mean this wont get merged. It just helps us catch little issues here and there.

PR #8617, opened 2019-06-08
Make the libnm library optional

filakhtov:
:wave: 

Coming from Gentoo system, I have been maintaining patches to keep Cinnamon NetworkManager-free for quite a while now (since 2.8) and finally got some time to properly handle build files to:

- Introduce `--disable-networkmanager` configure flag;
- Make sure to not check for `libnm` when this flag is given;
- Make sure that `NM-1.0.gir` file is not referenced if the flag is given;
- Replace `Makefile.in` with `Makefile.am` in the `files` subdirectory and create separate `Makefile.am` files for `etc`, `usr/share` and `usr/bin`;
- Remove dependency on `nm-applet` from both `cinnamon.session` and `cinnamon2d.session` if NM is disabled;
- Skip `network` plugin installation;

Cheers! :beers:


clefebvre:
Thanks @filakhtov.

I'd prefer to keep files dynamic though (i.e. not list its items within the build system). That's the reason we use files/ in the first place, so we can easily drop files in there and mimic the filesystem.

Maybe you can remove the network applet during the build instead?

collinss:
If we move to meson like we're hoping, it will be a simple matter because the function to install a subdirectory has the option to exclude files and directories. I'm not sure if autotools has the equivalent functionality or not.

clefebvre:
OK let's address this for 4.4.

filakhtov:
Hey @clefebvre addressed your feedback. I have used `bindir`, `datadir` and `sysconfdir` variables for installing files into their appropriate locations, but otherwise you can still have the same drop-in experience as we have today.

Fantu:
Hi, if minimal changes are needed probably debian patches about it can help:
https://salsa.debian.org/cinnamon-team/cinnamon/blob/experimental/debian/patches/make_nm_optional
https://salsa.debian.org/cinnamon-team/cinnamon/blob/experimental/debian/patches/Drop-nm-applet-from-the-session-files.patch
Was already used in some version in latest years, I not used archs that use them but I suppose are working correctly.

filakhtov:
@Fantu thanks for sharing. These patches are small indeed, but they introduce an automagic dependency on network manager, which is not something always desired. Additionally, second patch unconditionally removes nm-applet from the session start, meaning that it won't be automatically started at login.

Fantu:
nm-applet as write in patch description is autostarted anyway, I can confirm it because this patch it affects every architecture, on my Bionic and Sid it turns out to be started

filakhtov:
Hmm, thanks for an update. Possibly it is done via an XDG autostart, but nonetheless automagic dependency is less than ideal, because it doesn't give users control over how they want their dependencies wired.

Fantu:
What is the problem? If nm is in build-deps (and/or already installed) have it in cinnamon build, instead if it is not there it could not compile correctly. If you don't want it, not have in build-deps and/or your build system. If is compiled on production system instead and you have nm, why not have support of it in cinnamon? (it would be the only case in which an additional choice would make sense but I suppose is very rare or near null case)
Sorry if my english is too bad.

filakhtov:
Your English is fine, no need to worry about that :smile: 

Coming from Gentoo system, where user have complete control of how their systems are built and which dependencies are used, I might want to have NetworkManager installed and use it through CLI, but I don't want it to be dragged into Cinnamon, nor I care about nm-applet. I understand that for binary distros it is perhaps not a big deal, but for projects like LFS for example it actually is!

I might have NetworkManager libraries and dev files installed only because I develop against it, but not using it as a network stack of my system and don't want my actual system libraries to be linked against it, etc.

It is generally a good rule of thumb to provide control over optional features and let users decide what they want and what they don't, even if by default it goes all enabled, there is a switch to configure things the desired way.

eli-schwartz:
Random thought: what actually needs NM installed? Why does /usr/lib/cinnamon/Cinnamon-0.1.typelib link to /usr/lib/girepository-1.0/NM-1.0.typelib at all? ISTR only the js code in the network@cinnamon.org applet actually knows or cares that NM is available, and it directly uses imports.gi.NM, so... couldn't this just be made a runtime-optional thing? That would be much better than configure switches and templating all the code.

If you are going to fully disable it, then you should make said applet not be installed, BTW.

filakhtov:
@eli-schwartz good catch, I will ensure that applet is also not installed, but having it to pick things up at runtime won't be ideal, because same reasons I described above apply.

eli-schwartz:
I feel like everything I said went right over your head. :/

Your current PR is an attempt to install the applet completely as-is, yet remove the typelib dependency and not autostart nm-applet. That would seem to be quite "optional at runtime" already. Is there something intrinsically evil about runtime detection of available features? Have you never seen code that does a try/except **in an interpreted scripting language** to import an optional feature but gracefully fall back on disabling it?

Your statements about automagic dependencies don't matter if the dependency is not automagic, but nonexistent. I am proposing what Debian calls "suggests", what Arch Linux calls [optdepends](https://wiki.archlinux.org/index.php/PKGBUILD#optdepends), and what Gentoo does not seem to support but I guess recommends simply mentioning in elog (for example https://bugs.gentoo.org/511122).

Potentially the only thing that should matter at all is whether cinnamon-session is responsible for autostarting the nm-applet program, but as @Fantu already pointed out, nm-applet installs an xdg-autostart for nm-applet and therefore it doesn't matter whether cinnamon tries to autostart it or not.

IMO the correct way to handle this is to simply remove it from the session files altogether, remove NM-1.0 from the typelib *always*, and make sure that the network@cinnamon.org applet doesn't crash or do anything else crazy when NM is not available on the system. Then cinnamon would automatically make the applet available *iff* NM is installed, and otherwise do nothing, which should hopefully solve everything for everyone except people who dislike when 132KB of useless files are installed on their system.

eli-schwartz:
This would actually have a significant impact on me in Arch Linux, because requiring a compile time option to disable networkmanager means I'm not going to cripple cinnamon for a handful of networkmanager haters, but my suggestion means I could ship a single package that optdepends on networkmanager instead of having it as a hard requirement, and then users who like networkmanager can install it and cinnamon automatically picks it up and uses it, while users who do not like networkmanager can simply not install it, and cinnamon will continue to work just fine without it.

filakhtov:
Okay, let me try and better explain things:

> Is there something intrinsically evil about runtime detection of available features?
In my opinion - yes, there is. Imagine there is an application "A" that can use a library "B". There is a system that provides library "B", but this library used by an application "C". Now if I install application "A" on that same system - it will automatically leverage library "B" at runtime. This gives no control on what dependencies are used by which application whatsoever and creates a dependency management nightmare.

> Have you never seen code that does a try/except in an interpreted scripting language to import an optional feature but gracefully fall back on disabling it?
I have seen, and I have to work with these things often times too - it is not fun at all. It sometimes leads to incompatibility issues and dependency resolution hell as well. If you put this all into SELinux or AppArmor context, such dynamic dependency detection might yield unexpected results and lead to quite a bit of a fun debugging journey. I would like things to be explicitly wired and controlled as such.

eli-schwartz:
> This gives no control on what dependencies are used by which application whatsoever

True.

> and creates a dependency management nightmare.

Completely false. Take for example https://www.archlinux.org/packages/extra/x86_64/vim/ which is built with support for various script interpreters. Vim is designed to support this via dlopen, so distros can build binary artifacts which work everywhere (no need to build 40 different distro packages for each USE flag permutation) while at the same time not forcing every vim user in all of Arch Linux from installing lua, perl, ruby, tcl, and Python both 2.x and 3.x

(Arch Linux, Fedora, and Linux Mint are all binary distros, and would no doubt be equally interested in engineering a new feature in a manner which is usable by a binary distro.)

If we're talking philosophy, failure to support graceful runtime detection of features is exactly what leads to dependency management nightmares and exponentially increasing repository sizes.

> It sometimes leads to incompatibility issues and dependency resolution hell as well. If you put this all into SELinux or AppArmor context, such dynamic dependency detection might yield unexpected results and lead to quite a bit of a fun debugging journey. I would like things to be explicitly wired and controlled as such.

If your SELinux rules are forbidding cinnamon from opening up libnm's GObject typelib *but only when cinnamon is built without the networkmanager USE flag* then maybe those rules suck and should be better designed to not be ifdef'ed based on a USE flag. Either way I would expect said SELinux rule would prevent the typelib from being opened, thus preventing networkmanager from being used, thus being no different from building without networkmanager support at all.

eli-schwartz:
I don't see how this matters anyway. All I said was that I think it may be possible to make cinnamon not use networkmanager at all, except for the applet itself, which would mean that your USE flag would only have to delete the applet, rather than also generating different Cinnamon typelibs and session files. By extension, it should also be possible to separately ship cinnamon and the applet, and make the applet hard depend on networkmanager but only when it's installed separately.

I don't understand why your Appeal To Ideological Purity considers this a bad thing.

filakhtov:
I understand what you are saying, just disagree with that.

> There is absolutely no downside to this. If you don't have these script interpreters installed, there is no dependency and surely cannot be a dependency management anything (let alone a nightmare).

There are downsides to this, everything has its trade-offs. Provided VIM example opens up a potential for more security issues, but VIM provides configure time options to disable all interpreters, I suspect because its authors understand implications.

Regardless, I think if we look at this problem differently and apply modular design to this problem then, probably, network management features, including applet and relevant typelibs, should be a separate package and have a hard dependency on NetworkManager and ModemManager libraries, because they aren't simply working without these libs.

If we go even further, and apply a bit of Unix philosophy, then Cinnamon should provide an API for networking stack and bridges should be written to support multiple backend providers, such as NetworkManager, systemd-networkd, and Arch's own netctl.

But, I would like to step out of this argument and say something that will likely help both of us get what we want. Two steps:

1. Provide configure time option, such as `--disable-networking` that completely disables networking applet and typelib dependency - that solves my problem. Happy to extend my PR to drop applet completely.
2. Refactor network applet in such a way, that if aforementioned flag is not given and applet is installed, then Cinnamon will rely on dynamic detection and handle things gracefully - solving your use case. PR is welcome :wink: 

And I hope we both will be happy. Thoughts?

eli-schwartz:
That sounds very complicated, so I will tell you what actually does make me happy. #8688 makes me happy, so I've gone ahead and done it. :)

eli-schwartz:
@filakhtov Given that #8688 is now merged, this should be simpler to implement (and also needs rebasing).

P.S. Have you by any chance experimented to see what the applet's behavior is like, when NM is not installed?

filakhtov:
@eli-schwartz it doesn't work at all, because https://github.com/linuxmint/cinnamon/blob/master/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js#L5 import will fail. This applet makes no sense without NetworkManager, because it uses GObject bindings provided by NM to query information and perform any changes to network devices.

filakhtov:
As to the rebasing this and moving forward, I haven't heard from the owners of this repository in a while and am not sure if they are still willing to accept this change or what the state is. ping @clefebvre @collinss

eli-schwartz:
> it doesn't work at all, because https://github.com/linuxmint/cinnamon/blob/master/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js#L5 import will fail. This applet makes no sense without NetworkManager, because it uses GObject bindings provided by NM to query information and perform any changes to network devices.

I don't know why you're responding like I'm unaware of that fact? I did mention earlier that I know the applet imports NM, "but I think it would be nice if it fails gracefully".

Does the applet gracefully fail, or does it segfault the desktop? Does it incur noticeable runtime penalties due to not loading, and/or continuing to try to load even though it can't?

I am curious about the answers to these questions, so I thought I'd ask if you happened to check. If you haven't, then that's fine too. Removing the applet entirely via a configure switch is definitely another valid way to solve the feature request, so, go for it. :)

As for hearing back from the maintainers, I think they were just busy on other things for a while :) so I would guess that updating the PR would be well worth it so they have something to look at.

filakhtov:
It produces interpreter errors like this:
```
error t=2019-11-15T10:33:00Z 
[network@cinnamon.org]: Requiring NM, version none: Typelib file for namespace 'NM' (any version) not found
[network@cinnamon.org]: Error importing applet.js from network@cinnamon.org
```
and fails to initialize applet altogether. This happens every time panel related settings are modified/refreshed.

clefebvre:
Hi @filakhtov sorry for the delay. We're getting close to 4.4. so we'll review all remaining PRs in the coming week.

filakhtov:
Rebased to resolve conflicts and is ready to go ahead.

src/Makefile.am
```diff
@@ -240,6 +240,9 @@ libcinnamon_la_CPPFLAGS = $(MUFFIN_CFLAGS) $(cinnamon_cflags)
 
 Cinnamon-0.1.gir: libcinnamon.la St-1.0.gir
 Cinnamon_0_1_gir_INCLUDES = Clutter-0 ClutterX11-0 CoglPango-0 Cogl-0 Meta-Muffin.0 Soup-2.4 CMenu-3.0
+if HAVE_NETWORKMANAGER
+Cinnamon_0_1_gir_INCLUDES += NM-1.0
+endif HAVE_NETWORKMANAGER
```
eli-schwartz: Why are you adding this?

filakhtov: When building Cinnamon with `--enable-introspection` flag it will produce `Cinnamon-0.1.gir` (which is XML with metadata description of the exported symbols) and `Cinnamon-0.1.typelib` (which is respective binary code) those are part of GObject Introspection bindings that allow using various features provided by the `libcinnamon` from any non-C compatible language that can consume GObject stuff, like for example JS. 

I know you have removed it in your #8688, but that might be a backwards incompatible change, because there might be software that is already dependent on NM being exposed through `Cinnamon-0.1.gir`. I'm not saying that is a good way to go about it, I'm just saying that we don't know what things exist out there (outside of Cinnamon itself) that might be broken after we removed this. I haven't seen any discussion about this on your PR nor anyone pointed out why this dependency existed in the first place. So I would be curious to find this out.

From my investigation, there is no any single applet or extension in Cinnamon itself that is using NM via Cinnamon, only directly. There were some references to it in https://github.com/linuxmint/cinnamon/blob/master/js/ui/extension.js#L25-L26, but according to this comment this is no longer the case after 4.2 release.

I'm comfortable taking this out, but I would like for maintainers to have a look and verify that we are comfortable with this change and all this makes sense. Let's see what they have to say once this reviewed and I will action this appropriately.

eli-schwartz: That is a direct import of imports.gi.NMClient, not an import of imports.gi.Cinnamon using the utility functions from imports.gi.NM

-_-

It is okay to be "backwards incompatible" during the move from 4.2 to 4.4 with something that was never documented, *always* wrong, and not even proven to exist anywhere, which I still don't believe does exist anywhere because I can't think why anyone would think to do it.

PR #8619, opened 2019-06-08
Fix: update the input regions when going fullscreen

Cobinja:
This fixes a regression where input regions where not updated anymore when a window goes fullscreen.


Cobinja:
This was already cherry-picked, thus closed

PR #8620, opened 2019-06-10
[WIP] menu-applet: use displayButtons for initial item visibility

itzexor:
So that recents and places are properly hidden if they were rebuilt since
the menu was last opened.

Fixes: https://github.com/linuxmint/alpha-testing/issues/12

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -30,7 +30,7 @@ const MAX_FAV_ICON_SIZE = 32;
 const CATEGORY_ICON_SIZE = 22;
 const APPLICATION_ICON_SIZE = 22;
 
-const INITIAL_BUTTON_LOAD = 30;
+const INITIAL_BUTTON_LOAD = 15;
 const NUM_SYSTEM_BUTTONS = 3;
 const MAX_BUTTON_WIDTH = "max-width: 20em;";
 
@@ -1143,7 +1143,7 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
             }
             this._allAppsCategoryButton.actor.style_class = "menu-category-button-selected";
 
-            Mainloop.idle_add(Lang.bind(this, this._initial_cat_selection, n));
+            Mainloop.idle_add(() => { this._displayButtons(-1) });
         } else {
             this.actor.remove_style_pseudo_class('active');
             if (this.searchActive) {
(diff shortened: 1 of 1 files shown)


itzexor:
The initial load is lowered to 15 because we're re-looping the whole list after open instead of resuming at INITIAL_BUTTON_LOAD.

itzexor:
doing some more work with this to improve the menu opening times

lestcape:
> doing some more work with this to improve the menu opening times

The advantage of split the processing is that it can be perceived by the user as faster, because items in the view-port are loaded inline, while the non visible items are loaded on idle after that. So, in the way they was, the menu was then presented more quickly to the user and the rest of processing was occurred after that, when the user is deciding what thing to do. Now to presented the menu, the user need to wait for **all** items to be loaded.

**How to detect the difference:** Create a VM with a virtual reduced hardware and install a lot of programs in it. Then you can perceived the difference between one behaviour and the another pretty easy. Please note that in a faster computer is pretty difficult if not impossible to detect the difference. Also note that a big number of Linux users have an old machine.

Probably you want to have the @mtwebster review of this, as this is a regression of his code and also he experimented a lot with the menu to be opened faster. For example: one thing that he detected is that preload the menu outside the monitor coordinates speed up the process a lot.

itzexor:
I understand the reasoning behind the split loading and it definitely does improve the user experience, especially on lower end machines. I tried to preserve that in my initial commit on this branch, but reduced the number of initial load items to 15 because the _displayButtons() call is going to loop back over those 15 or how many ever again, unlike the old method that resumed where we left off.

The improvements I was attempting are a bit different. The main issue I have with both the existing behavior and any prototype changes I've tested is that when we have 50+ actors redrawing or laying out at once we get noticeable dropped frames, even on high end computers. It's not very noticeable using the menu alone, but if you have any other animations, videos, or whatever running, you will get a very noticeable stutter or 2 when the menu is opened.

I greatly reduced this in a previous branch by removing some in-javascript allocate code, but the frame times still jump up to 30 or 40ms causing a few dropped frames.

itzexor:
I even tried replacing _displayButtons()'s loops with this in order to attempt to reduce the number of actors changing per frame, but it didn't work as well as i'd hoped.

```function throttledLoop(items, callback, n = 0) {
    let i = 0;
    while (n < items.length && i < THROTTLED_LOOP_N_PER_IDLE) {
        callback(items[n]);
        n++;
        i++;
    }
    if (n < items.length)
        Meta.later_add(Meta.LaterType.IDLE, () => throttledLoop(items, callback, n));
}```

Also tried with `Meta.LaterType.BEFORE_REDRAW`

lestcape:
Probably one thing that will help the shell in general to prevent frames drops is a thing like that: https://gitlab.gnome.org/GNOME/mutter/merge_requests/520 Also reduce the code that can delayed the execution will help in general. Sure if you reduced the memory allocation, this was a help then because the memory allocation is one of that delayed procedure.

I think reduce the number of actors that are instantiate with a GenericContainer can also help. In gnome shell they used normal St.BoxLayout actors as a replace of GenericContainer.

I also notice that there are a difference if the items are hidden/removed from the container when the menu is closed.

itzexor:
I have a different fix for this in my menu-various-3 branch. It's not final yet but it doesn't change the open code at all.

PR #8636, opened 2019-06-13
Xlets removal confirmation dialogs

Odyseus:
This PR adds confirmation dialogs when trying to remove applets/desklets from their *Remove* context menu items. If the <kbd>Control</kbd> key is pressed when trying to remove an xlet, the xlet will be removed without confirmation.

### Notes

- I added the confirmation dialog to the desklets and not just to the applets to keep the behaivor of both types of xlets somewhat homogeneous.
- I added the <kbd>Control</kbd> key behavior just to avoid a possible annoyance. If one wants to remove several xlets, one can simply press the <kbd>Control</kbd> key to remove the rest of the xlets and avoid the confirmation dialog.

### DIFF
--- a/js/ui/applet.js
+++ b/js/ui/applet.js
@@ -15,6 +15,7 @@ const Flashspot = imports.ui.flashspot;
 const ModalDialog = imports.ui.modalDialog;
 const Signals = imports.signals;
 const Gettext = imports.gettext;
+const Cinnamon = imports.gi.Cinnamon;
 
 var AllowedLayout = {  // the panel layout that an applet is suitable for
     VERTICAL: 'vertical',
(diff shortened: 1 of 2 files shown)


pizzadude:
good idea.

clefebvre:
Very nice, thanks for implementing this @Odyseus. 

Here's my review:

- It works well.
- The label should express a confirmation, typically, "Are you sure you want to remove %s?".
- The ctrl label to skip confirmation shouldn't be in the dialog. I understand you want it to be discoverable, but this isn't good practice. See what happens when you delete a file in nemo. You can skip the confirmation with a key combination, for comfort (and it's kinda standard anyway), but it's not something you see written everywhere.

Issue #8662, opened 2019-06-24
Search in system settings doesn't work


vushe:
```
 * Cinnamon version (cinnamon --version)
   - 4.0.10, not from daily builds.
 * Distribution - Mint >16
 * Graphics hardware *and* driver used -- multiple systems
 * 64 bit
 ```

**Issue**
Issue is old. I face it since I started using Mint in 2016. Sorry if it's a duplicate, didn't find other similar reports.

**Steps to reproduce**
When you start searching in settings menu, if interface is in Russian, search is not working (even if you try searching in English). If interface is in English, search in English works fine. Maybe something's wrong with encoding and issue is not only for Russian.

**Expected behaviour**
Search should work as it works when interface is in English.

**Other information**
In example I added complete phrase in search field. I expected to see only first element, however, filtering didn't happen.

![image](https://user-images.githubusercontent.com/15868217/60047828-5764cb00-96d3-11e9-8f35-2c6e7aaddde0.png)

PR #8668, opened 2019-06-26
Use utf-8 for search encoding

vushe:
Described in detail [here](https://github.com/linuxmint/cinnamon/issues/8662)

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
@@ -371,7 +371,7 @@ class MainWindow:
             # unicode is default in Python 3
             pass
         text = unicodedata.normalize('NFD', text)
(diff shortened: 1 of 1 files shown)


clefebvre:
This creates the following regression..

strip_accents() is used so that unaccentuated strings match accentuated results.

For instance, if I search for "themes":

- Without this PR, "thèmes" is found (which is the expected behaviour)
- With this PR, "thèmes" is not found (that's a regression)

What locale are you using? Maybe we should only call strip_accents() for latin languages?

I'll close this PR, but the issue remains. We need a different solution.

vushe:
That's ok to close PR if regression is found.
In Cyrillic search doesn't work **at all**.
Well because there is no Cyrillic in ASCII.
I'm pretty sure in the big scale `utf-8` is better. But yes, regression shouldn't happen.

vushe:
Most probably I don't understand something but...
image

PR #8674, opened 2019-06-29
Cinnamon Settings: Fix search for locales with a non-Latin character set

LingMan:
This is an alternative solution to https://github.com/linuxmint/cinnamon/pull/8668 for fixing https://github.com/linuxmint/cinnamon/issues/8662. It should work on all locales but I'm no localization expert. Please test.

> Currently all characters in a search string which can not be represented in the
> ASCII character set filtered from the search string. This happens to work for
> Latin based locales because the strings first get normalized to NFD, which
> represents e.g. 'é' as the base character 'e' followed by the combining
> character '´'. Only the combining characters get filtered and the base
> characters remain.
> 
> However, in locales, which do not use the Latin character set, the base
> characters get filtered as well and the search string ends up being empty.
> 
> Fix this by not converting to ASCII and instead filtering the combining
> characters explicitly.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
@@ -473,15 +473,8 @@ class MainWindow:
             self.search_entry.set_text("")
 
     def strip_accents(self, text):
-        try:
-            text = unicode(text, 'utf-8')
-        except NameError:
-            # unicode is default in Python 3
-            pass
-        text = unicodedata.normalize('NFD', text)
-        text = text.encode('ascii', 'ignore')
-        text = text.decode("utf-8")
-        return str(text)
+        text = unicodedata.normalize('NFKD', text)
+        return ''.join([c for c in text if not unicodedata.combining(c)])
 
     def filter_visible_function(self, model, iter, user_data = None):
         sidePage = model.get_value(iter, 2)


clefebvre:
Thanks @LingMan that works much better! Tested in French here.

vushe:
Thanks for PR @LingMan!
It is interesting to know though why previous solution didn't work for French... 

Also, is French the only testing language?

LingMan:
@vushe, your solution didn't work because it turned the entire function into a noop. In your [screenshot](https://github.com/linuxmint/cinnamon/pull/8668#issuecomment-506844658) you searched for `thè` and of course that finds `Thèmes` just fine. However, the purpose of this function is that `the` (note the missing accent on the e) also works.

I believe clefebvre is French which is why he tests his locale and issues with it are immediately apparent to him. Ideally we'd have a broad range of users with different locales confirming it works for them.

vushe:
Thanks @LingMan. Yup, I think I've missed this functionality from the function. And even now this functionality doesn't work with Cyrillic. I get something completely incorrect (yet I get result) when typing latin letters on Russian locale.

In this particular case Clefebvre had issue reporter to confirm before merging... 

Anyway, glad it works now! As reported, issue existed for years.

LingMan:
Could you give some example strings on what you type, what you expect to be found, and what you actually find? Also some explanations of the rules would be nice, since I have zero experience with Cyrillic.

To be honest, I'm not surprised this doesn't work correctly yet. The old solution was a horrible hack and this PR replaced it with a strictly better one - but still a hack. Sadly, localization is a nightmare from the design perspective.

vushe:
Exactly, localization is a nightmare. Nice video!

Cyrillic (I basically mean Russian) consists of differently written letters when compared to Latin languages, e.g. `с` is `s`. Some letters represent combinations of letters, e.g. `ш` is `sh`. But I don't think we should go into that.

My suggestion would be to return nothing when nothing is found.
Alternatively, I'd expect that, when I search for `themes`, I get what respects to this word in English (iirc, this way they've made search in settings in Windows 10).

Currently I'm not sure why I type `the` and get `Network`.

I don't find it anyhow critical though.

By the way, would be good to have a fallback text when nothing is found?

Here is how it works now:
image

PR #8688, opened 2019-07-04
Cinnamon-0.1.typelib: do not link to libnm typelib

eli-schwartz:
Nothing in libcinnamon uses libnm, and the typelib does not need to either. In fact, the only part of cinnamon that uses libnm is the applet network@cinnamon.org and that directly imports.gi.NM, so it doesn't need to get it via exposure through the GObject bindings for cinnamon itself.

Removing references to NM from the Cinnamon typelib and rebuilding cinnamon results in a fully working Cinnamon desktop and a fully working network applet, so there is no downside to just dropping it.

Motivated by #8617 except I want to get it right -- there is no reason this cannot be a fully generic solution. We don't need a configure switch for a purely simple case of overlinking.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -239,7 +239,7 @@ libcinnamon_la_LIBADD = $(CINNAMON_LIBS) $(MUFFIN_LIBS) $(libcinnamon_libadd)
 libcinnamon_la_CPPFLAGS = $(MUFFIN_CFLAGS) $(cinnamon_cflags)
 
 Cinnamon-0.1.gir: libcinnamon.la St-1.0.gir
-Cinnamon_0_1_gir_INCLUDES = Clutter-0 ClutterX11-0 CoglPango-0 Cogl-0 Meta-Muffin.0 Soup-2.4 CMenu-3.0 NM-1.0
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Note that NetworkManager is still a fully required part of the build, we check for libnm via pkg-config and distro packages must hard-depend on it until and unless followup PRs attempt to resolve unrelated issues like "should the applet be installed if users just dislike NetworkManager".

mtwebster:
By this logic, there's no need for libnm to be part of the build at all, correct?  It only needs to be part of runtime dependencies (for now).  Can you remove it from configure.ac as well?

eli-schwartz:
Is that actually a good idea? How do you, for example, check that the installed system *will* have a new enough version of libnm installed at runtime? It's not uncommon for software to do build-time checks for the availability of runtime-only dependencies.

That being said, depending on how it's desired to resolve #8617 w.r.t. the applet itself, it might be that it doesn't need to be a runtime dependency either. In which case it would make sense to remove the configure check too.

Hence I would like to focus on a single change that I know is good, and leave the rest for a followup. It might be desired to make this a configure switch, or it might be desirable to do runtime detection, and I simply don't know as I have not as of now tested the applet's behavior if libnm is not installed.

mtwebster:
ok

PR #8701, opened 2019-07-08
[WIP] cinnamon-window-tracker.c: Add support for desktop files from flatpaks

Cobinja:
This adds support for associating windows from flatpaks with their respective .desktop-file.
This is needed for pinning them to window lists like our builtin Grouped Window List.

Depends on https://github.com/linuxmint/muffin/pull/527

### DIFF
--- a/src/cinnamon-app-system.c
+++ b/src/cinnamon-app-system.c
@@ -61,6 +61,7 @@ struct _CinnamonAppSystemPrivate {
 static void cinnamon_app_system_finalize (GObject *object);
 static void on_apps_tree_changed_cb (GMenuTree *tree, gpointer user_data);
 CinnamonApp * lookup_heuristic_basename (CinnamonAppSystem *system, const char *name);
+CinnamonApp * lookup_heuristic_basename_for_flatpak_app (CinnamonAppSystem *system, const char *name);
 gchar *strip_extension (gchar *wm_class);
 gboolean case_insensitive_search (const char *key,
                                   const char *value,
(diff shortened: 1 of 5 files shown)


Cobinja:
Force pushed a version with proper flatpak support, which takes into account that flatpak .desktop-files can be renamed from their original upstream ones.

src/cinnamon-app-system.c
```diff
@@ -897,6 +1013,26 @@ cinnamon_app_system_lookup_startup_wmclass (CinnamonAppSystem *system,
...
 
+/**
+ * cinnamon_app_system_lookup_startup_wmclass_for_flatpak_app:
+ * @system: a #CinnamonAppSystem
+ * @wmclass: (nullable): A WM_CLASS value
+ *
+ * Find a valid flatpak application whose .desktop file contains a
+ * StartupWMClass entry matching @wmclass.
+ *
+ * Returns: (transfer none): A #CinnamonApp for @wmclass
```
itzexor: This should be annotated (nullable) as g_hash_table_lookup can return NULL, as well as the explicit return NULL....unless these are considered error conditions

src/cinnamon-app-system.c
```diff
@@ -897,6 +1013,26 @@ cinnamon_app_system_lookup_startup_wmclass (CinnamonAppSystem *system,
   return g_hash_table_lookup (system->priv->startup_wm_class_to_app, wmclass);
 }
 
+/**
+ * cinnamon_app_system_lookup_startup_wmclass_for_flatpak_app:
+ * @system: a #CinnamonAppSystem
+ * @wmclass: (nullable): A WM_CLASS value
```
itzexor: this probably shouldn't be labeled nullable (or optional) if we just return NULL when it is NULL

src/cinnamon-app-system.c
```diff
@@ -658,18 +677,63 @@ on_apps_tree_changed_cb (GMenuTree *tree,
...
+        GList *flatpak_renamed_from, *l;
+        GHashTableIter flatpakIter;
+        gpointer flatpakKey, flatpakValue;
+
+        flatpak_renamed_from = cinnamon_app_get_flatpak_renamed_from (app);
+        l = flatpak_renamed_from;
+        while (l)
+          {
+            g_hash_table_replace (self->priv->desktop_filename_to_flatpak_app,
+                                  (gchar*)(l->data), app);
```
itzexor: possible lost ref on app?

Cobinja: Resolved with https://github.com/linuxmint/cinnamon/pull/8701/commits/e58b0f94d29643876782beeec9210809cae2b53d

Cobinja:
Closed in favor of https://github.com/linuxmint/cinnamon/pull/9705

PR #8719, opened 2019-07-14
Don't show uninstall option for flatpaks or snaps

RavetcoFX:
cinnamon-remove-application can't handle them


RavetcoFX:
Whoops, we do support flatpak

PR #8726, opened 2019-07-18
Auto Theme Switcher

chromer030:
Please implement a internal theme switcher , to switch system and gtk theme from light to dark and visa versa based on daytime.
this feature is fully present and working on macOS Catalina.


JosephMcc:
Totally lost on this PR. Seems like you meant to make some sort of feature request and I'm pretty sure we already have one for this somewhere.

PR #8729, opened 2019-07-19
Prevent uninstalling from the menu of certain package names

RavetcoFX:
Prevent uninstalling of mint* and cinnamon* packages from the menu as to prevent damaging the desktop.

Prevent uninstalling snap and appimage as cinnamon can't handle those yet


clefebvre:
I don't think we should do that here.

If we want to shield users on deletion of certain packages we should do that in the mint remover script directly. It's a bit weird to have cinnamon refer to mint packages...

As for snap and appimage, whether or not you can uninstall them, you definitely might want to remove the menu item at the very least.

PR #8734, opened 2019-07-22
Change xlet about dialog to external python/gtk in it's own process

collinss:


JosephMcc:
Don't forget you can also remove the theming for the dialog from the default cinnamon.css

files/usr/bin/xlet-about-dialog
```diff
@@ -0,0 +1,122 @@
...
+    def close(self, *args):
+        Gtk.main_quit()
+
+    def set_icon(self):
+        if 'icon' in self.metadata:
+            self.set_logo_icon_name(self.metadata['icon'])
+        else:
+            icon_path = os.path.join(self.metadata['path'], 'icon.png')
+            if os.path.exists(icon_path):
+                GdkPixbuf.Pixbuf.new_from_file_at_size(icon_path, 128, 128)
```
Odyseus: Hello, everybody.

@collinss: Here are two problems:

1. The icon is not set (missing call to `set_logo`). `self.set_logo(GdkPixbuf.Pixbuf.new_from_file_at_size(icon_path, 48, 48))`
2. Icon sizes for xlets (the icon.png file) are limited to 48x48 pixels (I don't remember why). Setting it here to 128x128 pixels results in blurred images ([128x128](https://user-images.githubusercontent.com/3822556/61641946-6dd05900-ac76-11e9-8635-4fda9abe7e45.png) [48x48](https://user-images.githubusercontent.com/3822556/61641948-6dd05900-ac76-11e9-8836-d822800c5f0b.png)).

May I also suggest that you add a call to `self.set_default_icon_name()` at the end of the `set_icon` method? As it is, when the dialog is open, its icon in the panel (the grouped/window list applets) will use the **gtk-missing-icon** (or **image-missing** icon). With the following addition, the icon for the xlet type will be displayed on the panel.

```python
self.set_default_icon_name('cs-%s' % self.metadata['type'])
```

collinss: Well, I guess I should know better than to submit a pull request at 2am. I remember seeing that I some of the icons were broken, and thought I'd fixed it, but clearly I did not.

collinss: One issue with setting it the pixel size to 48 is that those icons are much smaller than the ones rendered via icon name, and there doesn't seem to be a way to fix that. It's probably still better than having blurry icons, but I don't know.

collinss:
@JosephMcc Thanks! I forgot about that one. It's fixed now in the default theme. I'll submit prs for mint-x/y later. I've got to run to work for a couple hours right now.

files/usr/bin/xlet-about-dialog
collinss: everything should be working now

Issue #8737, opened 2019-07-26
Cinnamon-desktop-editor Translations Missing


kelebek333:
Hi,

Translations of cinnamon-desktop-editor is missing. Following files don't have any translations.

directory-editor.ui
launcher-editor.ui

Cinnamon version; any

![cinnamon-desktop-editor-1](https://user-images.githubusercontent.com/1300133/61982750-61495900-b006-11e9-9bad-efd37ad7aa7c.png)

![cinnamon-desktop-editor-2](https://user-images.githubusercontent.com/1300133/61982754-65757680-b006-11e9-9815-b8e7b82bfb20.png)

claudiux:
And `/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.ui` too.

![Capture du 2019-07-27 05-59-21](https://user-images.githubusercontent.com/33965039/61989708-b9e31b00-b033-11e9-971f-3b0d32a95b89.png)

PR #8739, opened 2019-07-28
Fix translation issues

claudiux:
~~Second part to fix~~ Fixes entirely https://github.com/linuxmint/mint19.2-beta/issues/66 and https://github.com/linuxmint/cinnamon/issues/8737.


claudiux:
I have rebased to reword the commits.

PR #8746, opened 2019-07-31
Update Spyces.py - Bypass all caches downloading json and zip files

claudiux:
Simple addition of a question mark at the end of their URL to download, bypassing all caches, the latest version of json and zip files.

An alternative to this PR could be to clear the server cache each time a new version of a Spice has arrived.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
@@ -34,10 +34,10 @@ settings_dir = '%s/.cinnamon/configs/' % home
 
 URL_SPICES_HOME = "https://cinnamon-spices.linuxmint.com"
 URL_MAP = {
-    'applet': URL_SPICES_HOME + "/json/applets.json",
-    'theme': URL_SPICES_HOME + "/json/themes.json",
-    'desklet': URL_SPICES_HOME + "/json/desklets.json",
-    'extension': URL_SPICES_HOME + "/json/extensions.json"
+    'applet': URL_SPICES_HOME + "/json/applets.json?",
+    'theme': URL_SPICES_HOME + "/json/themes.json?",
+    'desklet': URL_SPICES_HOME + "/json/desklets.json?",
+    'extension': URL_SPICES_HOME + "/json/extensions.json?"
 }
 
 ABORT_NONE = 0
@@ -607,7 +607,7 @@ class Spice_Harvester(GObject.Object):
     def _install(self, job):
         uuid = job['uuid']
 
-        download_url = URL_SPICES_HOME + self.index_cache[uuid]['file']
+        download_url = URL_SPICES_HOME + self.index_cache[uuid]['file'] + "?"
         self.current_uuid = uuid
 
         fd, ziptempfile = tempfile.mkstemp()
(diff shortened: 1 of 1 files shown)


claudiux:
This PR should fix https://github.com/linuxmint/cinnamon-spices-themes/issues/580.

brownsr:
HI @claudiux there's been a security change on the spices site, I think things are working fine now.

claudiux:
Hi @brownsr,

Sorry but the problem is not solved.

When I try to update the latest version of SpicesUpdate@claudiux, from Cinnamon Settings, the old version of the zip file is installed. However the "last-edited" field in metadata.json was updated: 1563823305. The user obtains an old version of this applet, declared as the latest version.

When I try to download the zip file using button on the site, I again obtain its old version.

Since your security change, adding a question mark at the end of the download url seems useless: I obtain also the old version of the zip file.

claudiux:
Using a random word after the question mark does the job! The latest version of zip file is downloaded.

I will change my code.

claudiux:
Now, a random query is added to the URL of json and zip files.

I used uuid4() to create this random query.

So, the request for `/files/applets/SpicesUpdate@claudiux.zip` becomes like a request for `/files/applets/SpicesUpdate@claudiux.zip?02472d96-43b4-4a4f-acf8-f164c0e65288`.

This works like a charm with Cinnamon Settings. But the problem of the Download button of the Spices web site is not solved.

Bundy01:
Hi,
I confirm that the last PR solves the problem with Cinnamon's settings.

clefebvre:
This was fixed on the server side. It's an elegant solution but I don't think we need it anymore.

claudiux:
@clefebvre Glad to know that this has been corrected server side. But this solution also bypass all other caches: proxy-cache, browser cache, etc. It's not cool to get an old version of a file when we hope to download its most recent version.

claudiux:
@clefebvre This method could be used for the Download button links of the Spices website, replacing the uuid by the 'last-edited' parameter of each Spice.

claudiux:
@clefebvre The problem is still here: https://github.com/linuxmint/cinnamon/issues/8747#issuecomment-525093214

I have seen it on Cinnamon of all versions. People think they have downloaded the latest version and end up with an older version.

On 4.2, if I replace the original Spyces.py file by the mine, then all works well.

Regards.

claudiux:
@clefebvre This issue is still here in Cinnamon 4.2.4. I just updated (using Cinnamon Settings) the `placesCenter@scollins` applet, which I made the latest version, and got the previous version!

collinss:
It can take a while to update on the site. I don't remember the specifics of how it works, but I want to say that there's a delay because the site only checks every so often for a newer version, so it may take an hour or so to update. I just checked myself, and it's using the version from a few days ago. This means it's almost certainly not a cache issue because I haven't been to that page in months. If it were loading a cached version, it probably wouldn't have shown the version it did.

claudiux:
@collinss Please test yourself this link: [https://cinnamon-spices.linuxmint.com/files/applets/placesCenter@scollins.zip](https://cinnamon-spices.linuxmint.com/files/applets/placesCenter@scollins.zip). You will obtain this:
Capture du 2019-09-04 20-07-12


 Then test this link: [https://cinnamon-spices.linuxmint.com/files/applets/placesCenter@scollins.zip?cf6c9eee-1d29-42f1-8e52-91db49195f2b](https://cinnamon-spices.linuxmint.com/files/applets/placesCenter@scollins.zip?cf6c9eee-1d29-42f1-8e52-91db49195f2b). You will obtain this:
Capture du 2019-09-04 20-08-40
The zip file is already on the server. But we obtain a cached version by the default link. (Where is this cache? I don't know and it is not very important). The solution to this issue is so simple and is in this PR.

PR #8775, opened 2019-08-06
Adds two Spices-developer-only menu entries to view or reload source code

claudiux:
I propose two optional menu entries in the context menu of Spices applets and desklets (not for system applets/desklets), allowing to:

1. open the directory containing their source code;
2. reload their source code.

EDIT:

These menu entries appear only if the ~~`~/.cinnamon/SPICESDEV` file exists~~ "I am a Cinnamon Spices developer" option is checked in Settings -> General.

Screenshot from 2019-08-11 04-55-41

So, a Spices developer can get the entries "View source code of %s" and "Reload code of %s" (where %s is the name of the Spice) in the context menus. ~~, just doing a `touch ~/.cinnamon/SPICESDEV` and restarting Cinnamon.~~

Spices-developer-menu-entries

~~Removing this file~~ Unchecking this option and restarting Cinnamon removes these menu entries.

### DIFF
--- a/data/org.cinnamon.gschema.xml.in
+++ b/data/org.cinnamon.gschema.xml.in
@@ -11,13 +11,22 @@
         using the Alt-F2 dialog.
       </_description>
     </key>
+    <key name="spices-developer" type="b">
+      <default>false</default>
+      <summary>
+        Enable internal tools/options useful for Cinnamon Spices developers (Requires Cinnamon restart)
+      </summary>
+      <description>
+        Adds entries in applets and desklets context menus, to view source and reload code.
+      </description>
+    </key>
     <key name="enabled-extensions" type="as">
       <default>[]</default>
       <_summary>Uuids of extensions to enable</_summary>
       <_description>
         Cinnamon extensions have a uuid property; this key lists extensions
         which should be loaded.  disabled-extensions overrides this setting for
-        extensions that appear in both lists. Append ! in front of uuid to 
+        extensions that appear in both lists. Append ! in front of uuid to
         override version check
       </_description>
     </key>
@@ -53,13 +62,13 @@
       <summary>The next applet id to give the next added applet</summary>
       <description>Do not change manually.</description>
     </key>
-    
+
     <key name="next-desklet-id" type="i">
       <default>0</default>
       <_summary>The next desklet id to give the next added desklet</_summary>
       <_description>Do not change manually.</_description>
     </key>
-    
+
     <key name="enabled-desklets" type="as">
       <default>[]</default>
       <_summary>Uuids of desklets to enable</_summary>
@@ -113,13 +122,13 @@
         Whether the panel autohides or not.
       </_description>
     </key>
-    
+
     <key name="panels-show-delay" type="as">
       <default>['1:0']</default>
       <summary>Duration of the delay before a hidden panel is shown</summary>
       <description>Duration of the delay (in milliseconds)</description>
     </key>
-    
+
     <key name="panels-hide-delay" type="as">
       <default>['1:0']</default>
       <summary>Duration of the delay before a shown panel is hidden</summary>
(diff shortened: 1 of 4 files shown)


claudiux:
I have nothing more to add.

Odyseus:
Hello, @claudiux.

Wouldn't it be better to just add an option to toggle the visibility of those menu entries (in **General** > **Miscellaneous Options** for example, and flagged with the *Requires Cinnamon restart* notice)? The way you implemented it, without any documentation about its existence, it has no *discoverability* other than by reading this PR or Cinnamon's source code.

Maybe adding a generic key called **developer-tools** that can then be used as `global.settings.get_boolean("developer-tools")`. The **developer-tools** key could then be reused if in the future are added more features that assist in the development of xlets or Cinnamon itself. Just an idea.

Additionally, using a *gsetting* eliminates the need to [initialize a file](https://github.com/linuxmint/cinnamon/pull/8775/files#diff-f166c2bedbb381ede2230178e39b8bc1R21) and [check it's existence](https://github.com/linuxmint/cinnamon/pull/8775/files#diff-f166c2bedbb381ede2230178e39b8bc1R22) twice (for desklets [in here](https://github.com/linuxmint/cinnamon/pull/8775/files#diff-0307fb0e70771fa929d29b5f6f5936b3R27) and [here](https://github.com/linuxmint/cinnamon/pull/8775/files#diff-0307fb0e70771fa929d29b5f6f5936b3R28)).

claudiux:
Hello @Odyseus,

Very good idea! I will try to implement it.

Just, maybe Spices developers should be distinguished from Cinnamon/Mint developers. The risks are not the same.

Spices developers do not require special permissions.

Cinnamon/Mint developers require write permission on certain folders, which must be listed.

On my system, I'm in the group named `root` and I have _chmoded_ `g+w` the `/usr/share/cinnamon*` folders and some others. Is it a good idea to do that when an user checks the box "I'm a Cinnamon/Mint developer"? This is to be discussed, and I can not take this initiative alone.

Odyseus:
> Just, maybe Spices developers should be distinguished from Cinnamon/Mint developers. The risks are not the same.

Are you referring to the name of the **developer-tools** key? If so, it was just a quick suggestion. You could call it anything else.

But you should first think on the necessity of the features that you are adding. Because upon closeer inspection, I kind of find them redundant, since you can access those same shortcuts on Melange (**View source** and **Reload code**). And they are available not just for applets/desklets but also for extensions. Additionally, you can't access the context menu of an applet/desklet that it is loaded but not shown on the panel/desktop, so you don't have access to the **View source** and **Reload code** items from its context menu, but you do have access to them from Melange.

About permissions, I don't think that we need the hand holding. If someone opens a location to **/usr/share/whatever** and doesn't know that a normal user doesn't have write permissions...well...that system is just doomed to die. Furthermore, an xlet developer working on his/her xlet directly inside the xlets install location is just *suicidal* (LOL) (imagine that you made huge changes to one of your xlets and then you use one of the xlets managers to update your xlets...Pufff, changes gone). Such behavior simply shouldn't be promoted.

claudiux:
The **org.cinnamon.developer-tools** key already exists; it is used to allow the user to use some Alt-F2 commands, as:
 `r` to restart Cinnamon;
 `rt` to reload the theme;
 `lg` to run Looking Glass.

It is set to true by default.

claudiux:
I edited the first message.

The Codacy errors are minor. I used the _() function as used for other messages. My longest line of code is shorter than another already present.

clefebvre:
Why not just add a Reload option in the burger menu of the applet configuration screen? It would be there for everybody.. no need to configure whether you're a developer or not. As for the source code shortcut, I don't see the point.. a nemo bookmark while you're working on it does the trick.

collinss:
You can already reload an applet in melange. I think anywhere else would be redundant. As for the source code shortcut, I agree with @clefebvre. It's extremely niche. Even among applet devs I think a lot of them would prefer to use their git folder rather than the one in .local, and that could be anywhere.

Issue #8796, opened 2019-08-15
Grouped Window List: window thumbnails are hard to use because they disappear immediately when the pointer leaves them


LinuxOnTheDesktop:
```
 * Cinnamon version (cinnamon --version) 4.2.3
   - Please specify if you are using the daily builds PPA (https://launchpad.net/~linuxmint-daily-build-team/+archive/ubuntu/daily-builds). No.
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...) Mint 19.2
 * Graphics hardware *and* driver used: Intel UHD Graphics 620; Mesa
 * 32 or 64 bit: 64
 * Attach /home/<username>/.xsession-errors, or /var/log/syslog: not relevant
 ```

**Issue**

As soon as one's pointer goes just a tiny amount beyond a window thumbnail, the window thumbnail disappears (well, unless one is moving the pointer into another thumbnail). This makes the previews very hard to use - as [this video shows](https://app.box.com/s/nqhse33jtncl3053sgbm2t7eajxu35tf). Compare the behaviour on Windows 10, where the preview disappears only when the pointer has moved some (further) distance away from the thumbnail. [Here is a video](https://app.box.com/s/bi30wgs16xhml7gobwzjes84er1ktlz2) of the Windows behaviour. It seems to me that here Windows is very much getting things right.

**Steps to reproduce**

Please see the first of the two videos (the one showing Mint).

**Expected behaviour**

Please see the second of the two videos (the one showing Windows).

**Other information**

I looked at the applet's code to see whether I could make the required change myself, but it proved beyond me.

LinuxOnTheDesktop:
Here are reproducible steps that allow one to appreciate the severity of the problem.

1. Have two windows from one application - e.g. from XFCE4-terminal - open.
2. Have Grouped Window List installed on a panel.
3. Use the aforementioned panel to try to switch to one particular window of the aforementioned application. That is: try to get one window rather than the other window focused.
4. If you try 3 a few times, then you will find the following. It is all too easy to end up with no windows focused, or end up with the wrong window from the application focused, or to end up with an application from a different application focused.

If you tire of my bug reports about usability, then please ask yourself following (even though I know that sometimes I have been overly keen and that sometimes I have made other mistakes in my bug reports). Should basic usability in your flagship DE really take a second place to, say, adding a feature to the xapps?

smurphos:
[...]

PR #8814, opened 2019-08-22
[GWL] Added thumbnail-close-timeout option to fix #8741 and #8796

### LINKED ISSUE #8741: Grouped Window List: thumbnails disappearing on gap
```
 * Cinnamon 4.0.10
   - No daily build
 * Mint 19.1
 * Radeon HD 4850, "radeon" driver
 * 64 bit
 ```

**Issue**
When you move the cursor between the panel and the thumbnail, the thumbnail disappears.
Sometimes you need to show the window by moving the cursor from the panel to the thumbnail, but this works only when you move the cursor very fast. It's frustrating.
thumbnail

**Steps to reproduce**
Enable thumbnails, open a program, move the cursor on the panel icon to show the thumbnail, move the cursor on the gap between the panel and the thumbnail.

**Expected behaviour**
The thumbnail does not disappear when the cursor is on the gap.

claudiux:
Fixes #8741 and #8796.

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/applet.js
@@ -360,6 +360,7 @@ class GroupedWindowListApplet extends Applet.Applet {
             {key: 'hover-peek-time-in', value: 'peekTimeIn', cb: null},
             {key: 'hover-peek-time-out', value: 'peekTimeOut', cb: null},
             {key: 'thumbnail-timeout', value: 'thumbTimeout', cb: null},
+            {key: 'thumbnail-close-timeout', value: 'thumbCloseTimeout', cb: null},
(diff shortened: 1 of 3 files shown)


clefebvre:
Hi @claudiux, we're actually planning to remove quite a few options in GWL. We shouldn't add new ones especially for things which we can reach consensus on. We don't want to provide a DE people can tune in every little detail one can imagine. We're talking about an animation speed here, we should be able to just code it right. If it's too fast, let's slow it down. I don't think this should be configurable.

clefebvre:
500 is definitely too big, it makes the experience feel really sluggish.

Another thing to take into consideration, the value used in onMenuEnter() should match the one used in onMenuLeave(), otherwise you end up with multiple thumbs shown concurrently when browsing back and forth between apps in the GWL, and that really gives a feeling of lag and instability.

claudiux:
OK @clefebvre, I understand your arguments. This PR was just made to fix 2 issues. I close it waiting for the next version of GWL.

PR #8818, opened 2019-08-26
xlet-settings.py - Adding option to reload the xlet in the burger menu of its configuration screen

claudiux:
as suggested by @clefebvre in https://github.com/linuxmint/cinnamon/pull/8775#issuecomment-522583675


clefebvre:
ah cool, thanks for doing this @claudiux !

PR #8836, opened 2019-09-03
xapp-status applet: Update to utilize XAppStatusIconMonitor.

mtwebster:
(see xapp pr for details))

issues with this applet:
- currently doesn't drop its dbus name when removed from the panel (which interferes with XAppStatusIcons deciding on whether or not to use a GtkStatusIcon.


clefebvre:
> currently doesn't drop its dbus name when removed from the panel (which interferes with XAppStatusIcons deciding on whether or not to use a GtkStatusIcon.

The current applet (the one on master) also has that problem.

PR #8841, opened 2019-09-06
I fixed the "Backup" option in cinnamon center. 

pyroxar:
Now you can open timeshift from the cinnamon center.


collinss:
Hi @pyroxar some of those external programs are for other distros besides Mint (eg. fedora, arch, etc) You might want to make sure that one isn't being used by any other distro before you remove it.

pyroxar:
If the command is not available to run, cinnamon center doesn't show the icon.

I changed command form "pkexec timeshift-gtk" to "vlc", when i remove vlc, option doesn't still available. 

I tested it with the timeshift too, you can try my code, when you remove the timeshift, option will be removed too.

On the screen I show you the original code which use this "tricks".
image

collinss:
Oh yes, I'm not saying you shouldn't add timeshift, but your pr is also removing `systemd-manager-pkexec` and that may be used by other distros.

pyroxar:
Good point.

clefebvre:
The name should match the name in the application menu, and ideally the window title. Here's it's simply "Timeshift".

Because Timeshift brands itself (i.e. uses a name rather than describing itself as a functionality.. like "System Snapshots" or something..), I'm not sure it should be added to the settings.

I'll close this PR for this reason.

PR #8871, opened 2019-09-25
Move the 'show' applet to right corner, similar to Windows systems.

pyroxar:
In my opinion (I know it from experience), it is very convenient for new users (for example) from Windows.

obraz

### DIFF
--- a/data/org.cinnamon.gschema.xml.in
+++ b/data/org.cinnamon.gschema.xml.in
@@ -31,7 +31,7 @@
     </key>
 
     <key name="enabled-applets" type="as">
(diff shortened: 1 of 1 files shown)


smurphos:
Just a heads up that you would need to patch https://github.com/linuxmint/mintwelcome/blob/master/usr/lib/linuxmint/mintwelcome/mintwelcome.py
as-well so that your proposed new layout is respected when Mint users use mintwelcome to switch between 'Modern' and 'Traditional' layouts.

pyroxar:
I think that move the "show" applet to right corner is good proposition for all layouts.

smurphos:
I've got no preference - it's one of the first applets I get rid off on a new install. I do use the bottom right hot corner for my show desktop function though so I'm with you there ;-)

PR #8873, opened 2019-09-26
option to restart without user extensions

norbusan:
A user might have installed broken extensions/applets/etc that make cinnamon fail to start. At the moment the `cinnamon-launcher` falls back to starting a fallback panel, but the registered three fallback panels might not be installed, which would leave the user with an empty desktop without any feedback.

In case starting cinnamon fails, a dialog is shown that allows to restart cinnamon. A checkbutton is added that allows disabling applets/extensions etc from user's home directory. If the checkbutton is checked, the env var CINNAMON_TROUBLESHOOT=1 is set.

When searching for extension's directory, only search user directories if this variable is not set.

### DIFF
--- a/files/usr/bin/cinnamon-launcher
+++ b/files/usr/bin/cinnamon-launcher
@@ -32,15 +32,22 @@ elif os.path.exists("/usr/bin/tint2"):
 
 def confirm_restart():
     d = Gtk.Dialog(title=None, parent=None, flags=0)
+    d.set_default_size(400,-1)
     d.add_buttons(Gtk.STOCK_NO, Gtk.ResponseType.NO,
                   Gtk.STOCK_YES, Gtk.ResponseType.YES)
     d.set_keep_above(True)
-    d.set_position(Gtk.WindowPosition.CENTER_ALWAYS)
+    d.set_position(Gtk.WindowPosition.CENTER)
     box = d.get_content_area()
     c = Gtk.CheckButton(label="Disable local extensions")
     label = Gtk.Label()
     label.set_line_wrap(True)
-    label.set_markup("<span size='large'><b>%s</b></span>\n\n%s\n\n" % (("Cinnamon just crashed. You are currently running in Fallback Mode."), ("Do you want to restart Cinnamon?")))
+    label.set_markup("<span size='large'><b>%s</b></span>\n\n%s\n\n%s\n\n" %
+            (
+                ("Cinnamon just crashed. You are currently running in Fallback Mode."),
+                ("Do you want to restart Cinnamon?"),
+                ("If you suspect the source of the crashes might be local extensions "
+                 "or applets, you can disable all user supplied extensions using the checkbox below.")
+            ))
     box.add(label)
     box.add(c)
     label.show_all()
@@ -52,7 +59,6 @@ def confirm_restart():
             os.environ["CINNAMON_TROUBLESHOOT"] = "1"
         return(True)
     return(False)
-
 
 if __name__ == "__main__":
     setproctitle("cinnamon-launcher")
--- a/files/usr/bin/cinnamon-launcher
+++ b/files/usr/bin/cinnamon-launcher
@@ -31,14 +31,27 @@ elif os.path.exists("/usr/bin/tint2"):
 
 
 def confirm_restart():
-    d = Gtk.MessageDialog(parent=None, flags=0, message_type=Gtk.MessageType.WARNING, buttons=Gtk.ButtonsType.YES_NO)
+    d = Gtk.Dialog(title=None, parent=None, flags=0)
+    d.add_buttons(Gtk.STOCK_NO, Gtk.ResponseType.NO,
+                  Gtk.STOCK_YES, Gtk.ResponseType.YES)
     d.set_keep_above(True)
     d.set_position(Gtk.WindowPosition.CENTER_ALWAYS)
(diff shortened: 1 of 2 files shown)


clefebvre:
Hi @norbusan 

I don't think we need to pass any info. In cinnamon-launcher we currently have two buttons "Yes" and "No". We should simply change that to a three buttons dialog, with one of the button starting cinnamon with no local spices.

The way that can work is simply by using an environment variable. Say CINNAMON_TROUBLESHOOT=1 or whatever.. we simply read that in global or in spices managers and when it's set we don't load anything from ~.

norbusan:
Hi @clefebvre
Thanks for your comments, indeed, that is the better idea. I will work on a psych and send a new PR, shouldn't be too difficult.

Best

norbusan:
Hi @clefebvre 
here is a first shot at it. The dialog is not beautiful, and I have opted for a checkbox instead of three buttons. If the checkbox is set, export an env var that is checked extension.js when search for the directory of an extension.

Is this somehow what you intended to?

files/usr/bin/cinnamon-launcher
```diff
@@ -31,14 +31,27 @@ elif os.path.exists("/usr/bin/tint2"):
 
 
 def confirm_restart():
-    d = Gtk.MessageDialog(parent=None, flags=0, message_type=Gtk.MessageType.WARNING, buttons=Gtk.ButtonsType.YES_NO)
+    d = Gtk.Dialog(title=None, parent=None, flags=0)
+    d.add_buttons(Gtk.STOCK_NO, Gtk.ResponseType.NO,
+                  Gtk.STOCK_YES, Gtk.ResponseType.YES)
```
JosephMcc: It would be better not to use GtkStock in new code.

norbusan: Hi @JosephMcc 
thanks for your comment - Can you explain why using Gtk.STOCK_X is deprecated? I checked the library documentation and couldn't find anything related to it being deprecated.

OTOH, if not, would you prefer a simpl `_("Yes")` and `_("No")`?

Thanks

norbusan:
BTW, can anyone here give me a hint *how* I can actually trigger cinnamon crashing so that I can test the loading?

Thanks

collinss:
> BTW, can anyone here give me a hint how I can actually trigger cinnamon crashing so that I can test the loading?

Looking glass has that functionality - super+L, then choose actions->Crash Cinnamon in the bottom right of the window that pops up.

norbusan:
> Looking glass has that functionality - super+L, then choose actions->Crash Cinnamon in the bottom right of the window that pops up.

Cooool, thanks, didn't know that!

norbusan:
Just confirmed that my personal applets (in `~`) are not loaded.

norbusan:
Any further comments to this from @clefebvre or @JosephMcc ? I am planning to upload new packages of Cinnamon 4.2 to Debian/unstable, and I either wait until this is fixed upstream, or patch it into the Debian packages. I would of course prefer upstream inclusion to keep changes minimal.

Thanks for your consideration

clefebvre:
Hi @norbusan sorry for the delay, I'll try to review this during the week.

norbusan:
> Hi @norbusan sorry for the delay, I'll try to review this during the week.

No problem! I have pushed one more commit to beautify the dialog a bit. This is what we currently ship in Debian, but if you include something similar, we will of course use your solution.

All the best

Norbert

PS: cinnamon 4.2 has entered Debian/testing, finally we are more or less uptodate!

files/usr/bin/cinnamon-launcher
Odyseus: Hello, everybody.

@norbusan: For an up-to-date documentation of Gnome libraries (Gtk, Clutter, GLib, Gio, etc.), exists [this on-line documentation](https://lazka.github.io/pgi-docs/) that can also be [downloaded for off-line use](https://github.com/lazka/pgi-docs) as a static HTML site, and also can be downloaded as [Devhelp books](https://github.com/pygobject/pgi-docs-devhelp). These documentations are extracted directly from the Debian testing/unstable source code of said libraries. Each documentation details when a function, constant, class, etc. has been introduced and/or deprecated.

norbusan: Thanks @Odyseus 
so I see now that the STOCK stuff is deprecated. The explanation for replacement is minimal, so would the following be what you have in mind?
```
    d.add_buttons("_No", Gtk.ResponseType.NO,
                  "_Yes", Gtk.ResponseType.YES)
```
?

Odyseus: Oh, yes. Completely disregard my comments on `Gtk.MessageDialog`. I didn't see that you added more widgets to the dialog further down the code.

In regards to `gettext` on Python, it's pretty much the same as in any other language. You just use a call to `_()` with the strings that you want to be exposed for localization. The call to `gettext.install("cinnamon", "/usr/share/locale")` at the beginning of this file installs the function `_()` in Python’s built-ins namespace ([gettext module docs](https://docs.python.org/3/library/gettext.html#module-gettext)).

By the way, you also need to add the calls to `_()` to the strings at around line **43** of this file.

norbusan: Ok, thanks. I have changed to `_("Yes")` and `_("No")`, and added the same translation call to the other strings. Pushed to this PR.

clefebvre:
Great, there's some minor cosmetic changes to do to the dialog and the msgids but it works very well. I'll merge this and tackle the cosmetics on master.

norbusan:
Thanks @clefebvre for merging, and all your work on cinnamon!

PR #8884, opened 2019-10-05
Make volume step in applet follow the volume media keys value set.

deskwizard:
Requires modification to schema, see the following commit.
https://github.com/deskwizard/cinnamon-desktop/commit/53658157be1cb82b91df4a55e82802ac1bf228a9

https://github.com/deskwizard/cinnamon-settings-daemon
https://github.com/deskwizard/cinnamon-desktop

### DIFF
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -35,7 +35,7 @@ const ICON_SIZE = 28;
 const CINNAMON_DESKTOP_SOUNDS = "org.cinnamon.desktop.sound";
 const MAXIMUM_VOLUME_KEY = "maximum-volume";
 
-const VOLUME_STEP_DIR = "org.cinnamon.desktop.keybindings.media-keys";
+const VOLUME_STEP_DIR = "org.cinnamon.desktop.sound";
 const VOLUME_STEP_KEY = "volume-step";
 
 class ControlButton {
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -30,12 +30,13 @@ let x = _("Playing");
 x = _("Paused");
 x = _("Stopped");
 
-const VOLUME_ADJUSTMENT_STEP = 0.05; /* Volume adjustment step in % */
-
 const ICON_SIZE = 28;
 
 const CINNAMON_DESKTOP_SOUNDS = "org.cinnamon.desktop.sound";
 const MAXIMUM_VOLUME_KEY = "maximum-volume";
+
+const VOLUME_STEP_DIR = "org.cinnamon.desktop.keybindings.media-keys";
+const VOLUME_STEP_KEY = "volume-step";
 
 class ControlButton {
     constructor(icon, tooltip, callback, small = false) {
@@ -141,7 +142,7 @@ class VolumeSlider extends PopupMenu.PopupSliderMenuItem {
         } else {
             muted = false;
             //100% is magnetic:
-            if (volume != this.applet._volumeNorm && volume > this.applet._volumeNorm*(1-VOLUME_ADJUSTMENT_STEP/2) && volume < this.applet._volumeNorm*(1+VOLUME_ADJUSTMENT_STEP/2))
+            if (volume != this.applet._volumeNorm && volume > this.applet._volumeNorm*(1-this.applet._mediakeysStep/2) && volume < this.applet._volumeNorm*(1+this.applet._mediakeysStep/2))
                 volume = this.applet._volumeNorm;
         }
         this.stream.volume = volume;
@@ -158,10 +159,10 @@ class VolumeSlider extends PopupMenu.PopupSliderMenuItem {
         let direction = event.get_scroll_direction();
 
         if (direction == Clutter.ScrollDirection.DOWN) {
-            this._value = Math.max(0, this._value - VOLUME_ADJUSTMENT_STEP/this.applet._volumeMax*this.applet._volumeNorm);
+            this._value = Math.max(0, this._value - this.applet._mediakeysStep/this.applet._volumeMax*this.applet._volumeNorm);
         }
         else if (direction == Clutter.ScrollDirection.UP) {
-            this._value = Math.min(1, this._value + VOLUME_ADJUSTMENT_STEP/this.applet._volumeMax*this.applet._volumeNorm);
+            this._value = Math.min(1, this._value + this.applet._mediakeysStep/this.applet._volumeMax*this.applet._volumeNorm);
         }
 
         this._slider.queue_repaint();
@@ -171,7 +172,7 @@ class VolumeSlider extends PopupMenu.PopupSliderMenuItem {
     _onKeyPressEvent(actor, event) {
         let key = event.get_key_symbol();
         if (key == Clutter.KEY_Right || key == Clutter.KEY_Left) {
-            let delta = key == Clutter.KEY_Right ? VOLUME_ADJUSTMENT_STEP : -VOLUME_ADJUSTMENT_STEP;
+            let delta = key == Clutter.KEY_Right ? this.applet._mediakeysStep : -this.applet._mediakeysStep;
             this._value = Math.max(0, Math.min(this._value + delta/this.applet._volumeMax*this.applet._volumeNorm, 1));
             this._slider.queue_repaint();
             this.emit('value-changed', this._value);
@@ -187,7 +188,7 @@ class VolumeSlider extends PopupMenu.PopupSliderMenuItem {
         // visible_value: percentage of volume_norm (shown to the user)
         // these only differ for the output, and only when the user changes the maximum volume
         let volume = (!this.stream || this.stream.is_muted) ? 0 : this.stream.volume;
-        let value, visible_value, delta = VOLUME_ADJUSTMENT_STEP * this.applet._volumeMax / this.applet._volumeNorm;
+        let value, visible_value, delta = this.applet._mediakeysStep * this.applet._volumeMax / this.applet._volumeNorm;
 
         if (this.isOutputSink) {
             value = volume / this.applet._volumeMax;
@@ -965,6 +966,9 @@ class CinnamonSoundApplet extends Applet.TextIconApplet {
         this._volumeMax = this._sound_settings.get_int(MAXIMUM_VOLUME_KEY) / 100 * this._control.get_vol_max_norm();
         this._volumeNorm = this._control.get_vol_max_norm();
 
+        this._mediakeysStep_settings = new Gio.Settings({ schema_id: VOLUME_STEP_DIR });
+        this._mediakeysStep = this._mediakeysStep_settings.get_int(VOLUME_STEP_KEY) / 100;
+
         this._streams = [];
         this._devices = [];
         this._recordingAppsNum = 0;
@@ -1093,22 +1097,22 @@ class CinnamonSoundApplet extends Applet.TextIconApplet {
 
         if (direction == Clutter.ScrollDirection.DOWN) {
             let prev_muted = this._output.is_muted;
-            this._output.volume = Math.max(0, currentVolume - this._volumeNorm * VOLUME_ADJUSTMENT_STEP);
+            this._output.volume = Math.max(0, currentVolume - this._volumeNorm * this.applet._mediakeysStep);
             if (this._output.volume < 1) {
                 this._output.volume = 0;
                 if (!prev_muted)
                     this._output.change_is_muted(true);
             } else {
                 // 100% is magnetic:
-                if (this._output.volume!=this._volumeNorm && this._output.volume>this._volumeNorm*(1-VOLUME_ADJUSTMENT_STEP/2) && this._output.volume<this._volumeNorm*(1+VOLUME_ADJUSTMENT_STEP/2))
+                if (this._output.volume!=this._volumeNorm && this._output.volume>this._volumeNorm*(1-this.applet._mediakeysStep/2) && this._output.volume<this._volumeNorm*(1+this.applet._mediakeysStep/2))
                     this._output.volume=this._volumeNorm;
             }
             this._output.push_volume();
         }
         else if (direction == Clutter.ScrollDirection.UP) {
-            this._output.volume = Math.min(this._volumeMax, currentVolume + this._volumeNorm * VOLUME_ADJUSTMENT_STEP);
+            this._output.volume = Math.min(this._volumeMax, currentVolume + this._volumeNorm * this.applet._mediakeysStep);
             // 100% is magnetic:
-            if (this._output.volume!=this._volumeNorm && this._output.volume>this._volumeNorm*(1-VOLUME_ADJUSTMENT_STEP/2) && this._output.volume<this._volumeNorm*(1+VOLUME_ADJUSTMENT_STEP/2))
+            if (this._output.volume!=this._volumeNorm && this._output.volume>this._volumeNorm*(1-this.applet._mediakeysStep/2) && this._output.volume<this._volumeNorm*(1+this.applet._mediakeysStep/2))
                 this._output.volume=this._volumeNorm;
             this._output.push_volume();
             this._output.change_is_muted(false);


mtwebster:
Hi, this looks fine, except for one thing - can you add the settings key to `org.cinnamon.desktop.sound` instead of media-keys?

We'd also want a setting to expose this to users, but I'm still thinking about how to display it, so don't worry about it for now.

Thanks

deskwizard:
Hi,
I have updated the 3 repository with the change requested.
Let me know if it looks alright to you.

Thanks

files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
```diff
@@ -1093,22 +1097,22 @@ class CinnamonSoundApplet extends Applet.TextIconApplet {
 
         if (direction == Clutter.ScrollDirection.DOWN) {
             let prev_muted = this._output.is_muted;
-            this._output.volume = Math.max(0, currentVolume - this._volumeNorm * VOLUME_ADJUSTMENT_STEP);
+            this._output.volume = Math.max(0, currentVolume - this._volumeNorm * this.applet._mediakeysStep);
```
mtwebster: These in this section should just be `this._mediakeysStep` (6 occurences)

files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
```diff
@@ -965,6 +966,9 @@ class CinnamonSoundApplet extends Applet.TextIconApplet {
         this._volumeMax = this._sound_settings.get_int(MAXIMUM_VOLUME_KEY) / 100 * this._control.get_vol_max_norm();
         this._volumeNorm = this._control.get_vol_max_norm();
 
+        this._mediakeysStep_settings = new Gio.Settings({ schema_id: VOLUME_STEP_DIR });
+        this._mediakeysStep = this._mediakeysStep_settings.get_int(VOLUME_STEP_KEY) / 100;
+
```
mtwebster: Add a listener similar to: https://github.com/linuxmint/cinnamon/blob/6560b3be37f783a4c015254a8c9b84e1ff8457ab/files/usr/share/cinnamon/applets/sound%40cinnamon.org/applet.js#L1034

You could even use the same callback, just tack on something to update `this._mediakeysStep`

files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
```diff
@@ -965,6 +966,9 @@ class CinnamonSoundApplet extends Applet.TextIconApplet {
         this._volumeMax = this._sound_settings.get_int(MAXIMUM_VOLUME_KEY) / 100 * this._control.get_vol_max_norm();
         this._volumeNorm = this._control.get_vol_max_norm();
 
+        this._mediakeysStep_settings = new Gio.Settings({ schema_id: VOLUME_STEP_DIR });
```
mtwebster: There is already a settings object for this schema - `this._sound_settings` - use it instead

mtwebster:
A couple things - there are errors when you mouse-scroll on the applet itself in the panel (not the volume slider in the menu) - you're referring to `this.applet._mediakeysStep` while in the applet - so here it should just be `this._mediakeysStep` - this is in `_onScrollEvent` in the applet class.

Another issue is there's no listener for changes to that setting.  You don't need one in cinnamon-settings-daemon because you're fetching the value each time (that's bad too, but not important for now.) - you need to know when the setting changes and update the value when it does.

Also, a settings object for `org.cinnamon.desktop.sound` already exists, please use that.

I'll do some inline comments to show what I mean.

clefebvre:
Hi @deskwizard,

First of all I want to apologize for the time it took me to get involved and provide feedback. And I also want to thank you for contributing to Cinnamon and for figuring it how this all worked. As you found out there are multiple components involved, it's not all obvious and we need to acknowledge that.

I'm tempted to merge this to thank you for your work and to encourage you rather than discourage you from contributing, but that would be wrong, because that's judging your work, not the feature you developed.

I think it's overkill. We're looking at an arbitrary value which isn't significant. The amount of increment we get on our media knob.. we're not going to go as far as asking the user to tune that. The 5% value is arbitrary. We can debate its value and that would be tuning. Asking users to tune that themselves is going way too far.

We're removing similar features elsewhere in Cinnamon, for instance asking the user to fine tune the number of millisecond a particular animation effect takes.

I'm sorry to close this. I'm more interested in debating the 5% than making this configurable though. I suppose you planned on using this yourself, so you're probably against that 5% in the first place. I'd be interested to hear your thoughts on this. Is it too big? is it too small? does it cause an annoyance? how/why?

deskwizard:
Hi @clefebvre 

Don't be sorry, I understand your decision. My idea was more to make it possible to change it, not necessarily have it exposed in a menu or applet for the reason you mentioned.
It was indeed more for my own use than anything else, I just figured maybe somebody else could be interested.
In my case, at the time I had no other volume control besides the volume keys to control the volume of my vintage audio receiver (I was using an IR remote sending the volume keys to the OS, let's not get into that mess) and 5% was too big a step.
Thankfully I changed receiver and have proper volume control now, so it isn't an issue for me anymore.

Thanks for taking the time to comment, it is appreciated a lot.

I also want to thank @mtwebster for the work he's done going over my code, no matter how bad a copy/paste job I actually did. I hope your neck doesn't hurt too much shaking your head looking at what a mess I've done. Your comments is a great way for me to learn more about all that stuff. Thanks again.

Have a nice weekend you two.

clefebvre:
You too @deskwizard. Thanks for explaining this.

MichaelLawton:
I'm disappointed to see this was not merged. When using a fixed gain external amplifier (like many popular cheap Class D amps) 5% steps are often too large. It's still nice to have the amp's maximum volume available on request so simply turning down the master softvol instead is not a great solution.

Ghibranalj:
I very disappointed that this feature was not merged. Even gnome has this feature implemented.
> The 5% value is arbitrary.

yes its arbitrary. why not let the user decide the arbitrary value for themselves?

kimimaru4000:
I agree that there's a need for this in Mint. I prefer my volume step at 2 and have to recompile Cinnamon with the step changed each time.

> I think it's overkill. We're looking at an arbitrary value which isn't significant. The amount of increment we get on our media knob.. we're not going to go as far as asking the user to tune that. The 5% value is arbitrary. We can debate its value and that would be tuning. Asking users to tune that themselves is going way too far.

I would like to know your thoughts on why this is overkill. Users do not have to use it, and there is plenty of space in the Settings tab of `cinnamon-settings sound`. Users are already able to customize lots of other options, including creating custom keyboard shortcuts, which are more involved than changing the volume step. Based on the reply I don't see how this hinders the UX in any way.

> I'm sorry to close this. I'm more interested in debating the 5% than making this configurable though. I suppose you planned on using this yourself, so you're probably against that 5% in the first place. I'd be interested to hear your thoughts on this. Is it too big? is it too small? does it cause an annoyance? how/why?

This is exactly why I feel it should be configurable. It's too large for users, including myself, right now. If you reduce it, it'd be too small for others. Let the user decide what's comfortable for them.

Secret-chest:
> Hi @deskwizard,
> 
> First of all I want to apologize for the time it took me to get involved and provide feedback. And I also want to thank you for contributing to Cinnamon and for figuring it how this all worked. As you found out there are multiple components involved, it's not all obvious and we need to acknowledge that.
> 
> I'm tempted to merge this to thank you for your work and to encourage you rather than discourage you from contributing, but that would be wrong, because that's judging your work, not the feature you developed.
> 
> I think it's overkill. We're looking at an arbitrary value which isn't significant. The amount of increment we get on our media knob.. we're not going to go as far as asking the user to tune that. The 5% value is arbitrary. We can debate its value and that would be tuning. Asking users to tune that themselves is going way too far.
> 
> We're removing similar features elsewhere in Cinnamon, for instance asking the user to fine tune the number of millisecond a particular animation effect takes.
> 
> I'm sorry to close this. I'm more interested in debating the 5% than making this configurable though. I suppose you planned on using this yourself, so you're probably against that 5% in the first place. I'd be interested to hear your thoughts on this. Is it too big? is it too small? does it cause an annoyance? how/why?
> 

@clefebvre What??? Some animations are too long or too short for me. And, users don't have to change it.

s-m-e:
Hey, just another vote for **PLEASE MERGE THIS**! When my computer is attached to an external amp or radio, the 5% jumps are ... insane. Here, 5% up can mean going from "perfectly normal" to "walls are shaking". Even something as low as 1% is very clearly audible. It is a very useful option to have. The fact that so many people are compiling this by hand (or use other workarounds) should be a pointer to not follow the Fruit-philosophy of not giving the user options ... I am a user and I am not confused by options :)

ssmid:
If 5% is too much, turn down your amplifier. Higher soft volume, lower amp volume should give a higher dynamic anyway, or am I wrong?

mankinskin:
@ssimd what if you are using headphones?

ssmid:
Isn't the output of headphone jacks standardized? 5% / 20 steps should be sufficient, considering Android has even less steps (14 if I remember correctly).

mankinskin:
@ssmid that is honestly missing the point kind of. There is no reason to lock users into a fixed configuration. I want to change my volume by scrolling on the audio applet. There is no reason I would need to submit huge loudness jumps when I could have fine grained volume control of at least 1% per step (which is by the way what Windows does). Currently the sensitivity is 5 times of that.

Again: why not provide 5% as a default and allow users to override this step size in a configuration file? What could possibly be the cost of that? This is a pull request for exactly that. All maintainers need to do is test it and merge it and it is an improvement to this library.

ssmid:
It's a feature that has to be maintained, so it's up to the maintainers to decide whether it is worth it.

mankinskin:
Fair, but I don't really see the maintanance burden with this feature. Also what is the point of this project when improvements like this are seen as "maintanance burdens". Then don't be a software maintainer.

Secret-chest:
> Hi @deskwizard,
> 
> First of all I want to apologize for the time it took me to get involved and provide feedback. And I also want to thank you for contributing to Cinnamon and for figuring it how this all worked. As you found out there are multiple components involved, it's not all obvious and we need to acknowledge that.
> 
> I'm tempted to merge this to thank you for your work and to encourage you rather than discourage you from contributing, but that would be wrong, because that's judging your work, not the feature you developed.
> 
> I think it's overkill. We're looking at an arbitrary value which isn't significant. The amount of increment we get on our media knob.. we're not going to go as far as asking the user to tune that. The 5% value is arbitrary. We can debate its value and that would be tuning. Asking users to tune that themselves is going way too far.
> 
> We're removing similar features elsewhere in Cinnamon, for instance asking the user to fine tune the number of millisecond a particular animation effect takes.
> 
> I'm sorry to close this. I'm more interested in debating the 5% than making this configurable though. I suppose you planned on using this yourself, so you're probably against that 5% in the first place. I'd be interested to hear your thoughts on this. Is it too big? is it too small? does it cause an annoyance? how/why?

We're not asking them. 2% would be too little for me. But we're not asking them. They can just change it _**when they want, if they want.**_

mankinskin:
For anyone looking for a fix, you can try changing the volume step in the sound applet JS code. For me, the sound applet is located at `/usr/cinnamon/applets/sound@cinnamon.org`. In `applet.js` there is a variable
```javascript
const VOLUME_ADJUSTMENT_STEP = 0.05; /* Volume adjustment step in % */
```
The applet can also be in `~/.local/share/cinnamon`

Possibly this is something that should be configurable in the "Configure..." options of the applet. There is obviously no tradeoff here (except for developer resources).

RedSoxFan04:
I just upgraded from onboard audio to a Sound Blaster AE-5 sound card. With this sound card, 5% increments are way too large. I need 2% increments. Please merge this!

Secret-chest:
PLEASE RECONSIDER. Yes it's not changing your whole desktop look. But it's changing something you use daily. So it's impactful enough.

claudiux:
Also, you can try the sound150@claudiux applet: https://cinnamon-spices.linuxmint.com/applets/view/306

Secret-chest:
> Also, you can try the sound150@claudiux applet: https://cinnamon-spices.linuxmint.com/applets/view/306

Does that work with media keys as well?

claudiux:
> 
> Does that work with media keys as well?
Yes, it works with media keys.

Lanchon:
hi @clefebvre,

i was needing to reduce the volume step and wanted to implement the change, then i found that a solution had already been rejected.

i am willing to do a new PR. but before i invest the time, the question is: would you merge it?

### motivation for this change

the effective dynamic range of an audio setup is the ratio between the power of the loudest and the weakest sounds you can hear. the latter depend on the acuteness of you hearing as well the as background ambient noise. this lower bound can generally be approximated with a value that will be useful for most use cases. the upper bound, however, is much more variable: it can be limited by the power, size, and quality of the audio equipment involved, and also by the pain and damage volume thresholds of the human listener (this will be important later).

(thus, the dynamic range of different setups varies widely: you cannot compare the sound output you get from shoddy laptop speakers to what you get from my particular setup, which happens to be 12-way speakers including four 10-inch woofers that comfortably handle 400W RMS continuous power and 800W peaks.)

note that an effective discreet volume control has to step through the whole of the system's dynamic range using steps that are neither too small to notice, nor too large to be inconvenient or uncomfortable. because the loudness step size is mostly a given constant for humans, and the dynamic range of setups can vary so widely, there is but one inescapable conclusion: **the number of steps cannot be a predefined constant and work well for different setups.**

cinnamon media keys only support a constant number of steps for the whole volume range (20 steps of 5% increments in a 100% whole), which might be optimal for the dynamic range of laptop speakers, but totally inadequate for more powerful setups. if you review the comments above, almost all complaints come from people who -just like me- connected powerful amplifiers to their computers. (and guessing from your comment of step size not needed to be configurable, i can imagine you use laptop speakers or small bookshelf speakers.)

**side note:** human hearing perceives loudness in a logarithmic scale. ie, if the ratio the power of a pair of sounds is constant, we perceive the loudness difference between them as a specific constant "difference value", regardless of the actual absolute power level of the sounds themselves. this is why dynamic ranges are typically expressed in dB, a logarithmic scale calculated as `10*log10(ratio of powers)`, or equivalently `20*log10(ratio of amplitudes)`. this is also why discreet volume controls offer logarithmic rather than linear steps: typically 1dB difference between steps, with 2dB steps generally frowned upon as too large (3dB = twice the power, 6dB = twice the amplitude).

the underlying mixer controlling volumes under cinnamon seems to be doing logarithmic steps: each fixed percent change seems to correspond to a specific perceived loudness change. not surprising, as i am sure that the developers and engineers handling the audio backend, drivers, and hardware are experts in their fields.

because cinnamon uses percent values for audio volume, it is implicitly limited to a maximum 100 steps (without big overhauls). this is not too bad; 1dB steps are generally accepted to be ok, and the dynamic range of the human ear is limited:

> Damage Due to Sound Exposure
>
> - 85 dB SPL: OSHA (Occupational Safety and Health Administration) recommends limiting exposure to 8 hours per day.
> - 100 dB SPL: Safe exposure time drops to around 15 minutes per day.
> - 110 dB SPL: Only a few minutes of exposure can cause damage.
> - 120 dB SPL: Immediate risk of damage, even for short exposures.

so 100 logarithmic steps are more than enough for non-professional setups.

### accessibility concerns

people with middle ear or cochlear (inner ear) damage can suffer from hyperacusis, a condition in which mildly loud sounds can cause uncomfort and/or pain. the volume range in which those affected can hear well can be severely restricted, making the ability to precisely control the output volume an accessibility concern. also, many people in the autistic spectrum find themselves in a similar situation -unable to tolerate mildly loud sounds- but stemming from a completely different cause.

### implementation and UI

for implementation simplicity, we can accept volume step sizes that are an integer percent value of the whole volume range. IMHO, a step size integer (in units of 1%) should be the underlying value stored in dconf (just as it was on this rejected PR).

but given the constraint of integer percent step sizes, there are few resulting useful total number of volume steps the system can have: 100 (1% steps), 50 (2% steps), 33 (3%), 25 (4%), 20 (5%), etc, with steps larger than 5% unlikely to be useful.

the useful range of step sizes to choose from is limited, so i propose this UI:
- do not show an integer percent step size to the user
- instead show a drop-down box of textual options for "volume control adjustment":
  - coarse/standard/normal (5%)
  - fine (2%)
  - super/ultra/very fine (1%)

(i do not think that 10% steps are useful, but they can also be included by splitting coarse and standard/normal as two different settings.)

### what i am willing to do

the dconf schema and the media key handler changes.

### what would remain as TODOs

the UI and the applet changes.

---

thank you for your time!

OleksandrChornyi2010:
I still think that this should be implemented in the system, nobody is forcing the user to change this if they don't want to.

Lanchon:
@clefebvre @mtwebster 

i'm writing this with good intentions, as a way to help you gather information on your project. after 15 years of linux mint and cinnamon, i've decided to switch to arch-base kde. the biggest part of the decision has been:

- this rejected PR, the various ignored issues about it, and my [message](https://github.com/linuxmint/cinnamon/pull/8884#issuecomment-2728229275) going unanswered for over a year. essentially a plea from many users needing a way to control something (not a UI, we just needed a setting) ignored for no reason.

- the really long fight over not allowing "suspend" as the critical battery action (again, not a UI, we just needed a setting) on the grounds of it being dangerous (even though "do nothing" was allowed, which is astronomically more dangerous given that suspend neatly flushes all buffers before sleeping). essentially again ignoring the needs of users for no good reason.

and there were other similar small incidents i can't quite remember now. i guess the message is that the paper cuts do add up. and while my leaving mint is inconsequential, maybe other users are silently leaving for similar reasons... or maybe not, i really don't know.

**in any case, please consider this message as an enormous thank you note for all the effort you put in this amazing project, which fueled my years of learning linux. and consider my writing of this message not as a troll but as me being genuinely indebted at least to give an explanation to why i'm leaving, instead of the easier route of doing it silently.**

for all the hand holding mint gave me when i needed it, i may be leaving because mint now grabs my hand too hard. please accommodate users: remember being told "no you can't do that because!" is a reason why many users left windows.

thanks again.

---

PS.
for the rest, know that both [GNOME](https://wiki.archlinux.org/title/GNOME/Tips_and_tricks#Adjust_volume_in_smaller_steps) and [KDE](https://docs.kde.org/trunk_kf6/en/kmix/kmix/Settings.html) support configurable volume steps.

PR #8905, opened 2019-10-16
Move the 'show' applet to right corner, similar to Windows systems (NEW PR).

pyroxar:
In my opinion (I know it from experience), it is very convenient for new users (for example) from Windows.

I messed up, I must create new PR:
https://github.com/linuxmint/cinnamon/pull/8871

### DIFF
--- a/data/org.cinnamon.gschema.xml.in
+++ b/data/org.cinnamon.gschema.xml.in
@@ -31,7 +31,7 @@
     </key>
 
     <key name="enabled-applets" type="as">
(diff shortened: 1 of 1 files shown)


clefebvre:
I've had to use Windows a little bit and I found it really weird to have to go to the right end of the panel just to show the desktop. I guess that shows it really depends on what you're used to. Our long term users are used to having it close to the menu, that's where it's been since 2006.

I agree on new users coming from Windows, though they'll have to adapt in many other ways when adopting Linux.

PR #8919, opened 2019-10-26
settings.js: allow distributions to override individual settings keys

collinss:
The override file has the same format as the settings-schema.json file, except that only the overridden settings should be included. The file must be placed in the applet folder, and have the name `settings-override.json`. Just as with settings-schema.json, changes to (including creation or deletion of) this file will automatically trigger an upgrade of the settings.

Some additional improvements to the code were implemented for consistency and better error handling and detection.


clefebvre:
@collinss you're missing a </para> close tags in the doc.

CI told me :)

collinss:
Oops.

...Should be fixed now.

collinss:
@clefebvre btw, feel free to squash this one (or let me know and I can do it). The only reason I didn't do that myself is that I wasn't sure if you'd already checked out the branch for testing and so I didn't want to force push and cause you more trouble ;).

PR #8933, opened 2019-10-30
Panel launchers rework

itzexor:
I'm not sure if the hit testing will be accurate or not in hidpi since I am using direct actor size values and not transformed size. I am unsure which is the correct method to use since we get pre-transformed mouse coordinates from the DND code. It is easy to fix if it's incorrect now, but if it works the extra transformation overhead is best avoided.

I completely removed behavior that caused multiple full reloads. A comment said it was overkill but easier to get the scaled size correct that way. I haven't noticed any icon scaling issues, and have no idea what case this code was designed for(maybe hidpi?). It's extremely inefficient and also can cause extra visual disruptions. If we don't need this, it is also best left out.

This reworks:

- Applets sync state when changed from cinnamon itself (not from settings) without excess reloading
- DND works differently for local vs remote drags now (rearrange vs dragging in a launcher from other applet) to simplify index tracking. local objects rearrange in real time (reset if drag cancelled), remote objects still use a placeholder
- DND methods split from applet object onto the actual launcher container actor so we don't have to transform from applet-relative to container-relative coordinates
- Don't use multiple names for the same object.

fixes: multiple issues on master with DND incorrectly dropping and losing launchers (settings corruption), specific positions that cause excessive placeholders to be added, etc

### DIFF
--- a/files/usr/share/cinnamon/applets/panel-launchers@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/panel-launchers@cinnamon.org/applet.js
@@ -61,16 +61,16 @@ class PanelAppLauncherMenu extends Applet.AppletPopupMenu {
         this.addMenuItem(subMenu);
 
         item = new PopupMenu.PopupIconMenuItem(_("About..."), "dialog-question", St.IconType.SYMBOLIC);
-        this._signals.connect(item, 'activate', Lang.bind(this._launcher._applet, this._launcher._applet.openAbout));
+        this._signals.connect(item, 'activate', Lang.bind(this._launcher.launchersBox, this._launcher.launchersBox.openAbout));
         subMenu.menu.addMenuItem(item);
 
         item = new PopupMenu.PopupIconMenuItem(_("Configure..."), "system-run", St.IconType.SYMBOLIC);
-        this._signals.connect(item, 'activate', Lang.bind(this._launcher._applet, this._launcher._applet.configureApplet));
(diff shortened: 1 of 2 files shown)


itzexor:
Note: I want to squash the last 2 DND commits into one before merge, but in case something is broken somewhere I want them split for easier partial/full reversions if necessary

clefebvre:
Hi @itzexor ,

I tested it in HiDPI, here are my findings:

- DND looks solid
- I can't see any issues related to HiDPI

However I found the following regressions:

- Hovering launchers no longer highlights the surrounding zone in the panel
- I can DND from the menu into the launcher applet to create a new launcher, but using the Add dialog no longer works.

itzexor:
>     * Hovering launchers no longer highlights the surrounding zone in the panel
I forgot to move the StBoxLayout init params to the new class. I messed with DND so much I guess I became style-blind. It was missing a style-class completely.
>     * I can DND from the menu into the launcher applet to create a new launcher, but using the Add dialog no longer works.
Nice catch on this one, I missed a param on the settings rebind in the function where we unbind in order to do a settings writeback from the applet's cache. This was probably introducing subtle bugs in my DND tests for a while...


I amended those fixes to the proper commits and squashed the 2 separate DND commits into one with a proper commit message.

itzexor:
I want to rework this so the launcherbox class isn't a gobject. I'm not sure why i decided to do it that way because it implements only pure-js interfaces.

mtwebster:
works nice

PR #8937, opened 2019-11-04
Fix: Handle GFileIcon file changes

Cobinja:
This monitors file changes for GFileIcon and updates the StIcon accordingly


Cobinja:
While https://github.com/linuxmint/cinnamon/pull/8931 does fix the specific issue mentioned there,
I think this is a better way to fix it, since this fixes non-refreshing StIcons that use GFileIcons wherever they are used, not just in the sound applet.

RafaelKr:
Awesome! This is a much better solution of course, thanks.

PR #8940, opened 2019-11-05
menu-applet: don't show user-hidden categories

itzexor:
I inadvertently removed this feature in the refactoring. User-
hidden directories are marked nodisplay and so that needs to be
checked on the gmenu object either at load or display. This commit
does it at load, so consequently any unhidden category will show
all unique apps as newly installed/highlighted upon unhiding after
starting the session or applet with it hidden.

Thanks goes to @collinss for finding both the bug and the fix.

fixes: 0ed2544d7398d033a9


itzexor:
To be more clear... We check the directory before loading it, so that's why I've done it there. Hidden apps aren't loaded into JS at all.

PR #8960, opened 2019-11-16
Dialog: Treat Escape key as Gtk.Response.NO

### LINKED ISSUE #4310: when setting an already used keyboard shortcurt, ESC doubles its usage
Hey.

When setting a keyboard shortcut for, say "switch to workspace 6" in the control centre and using a shortcut that's already used (e.g. Alt+F7 which is IIRC used for "move window" per default), then a pop up comes up that askes no/yes whether to move the shortcut.

But when escape is pressed the popup goes away and both, old and new action have the same shortcut... but it works only for one of them.

Cheers,
Chris.

OneQuid:
Other actions instead of pressing the OK button should be considered as cancellation
Closes #4310


OneQuid:
Thank you. Glad that I helped a tiny bit.

PR #8959, opened 2019-11-16
Closes #8956 - cinnamon-settings.py and ExtensionCore.py: now compatibles with Python 3.8

claudiux:
(Python 3.8 is used on Arch with Cinnamon 4.2.4.)
Also replaces `self.window.set_wmclass(wm_class, wm_class)` by `self.window.set_role(wm_class)` to avoid the warning message "Gtk.Window.set_wmclass is deprecated" (https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-set-wmclass).

Closes #8956.

PR #8964, opened 2019-11-17
fix regression in ExtensionCore not working when updates are available

eli-schwartz:
In commit 6ac295a62556f22b755b17dee29d17d6721fa361, Settings>Applets>Download gained the ability to display the number of available updates, using an ngettext-translated message. However, it has never worked since ngettext was never imported in the file in question, so it simply raised a traceback.

In the same commit, `gettext.install()` was taught to declare `names="ngettext"`, but as per the python documentation, the `names=` argument requires a sequence (of one or more aliases to declare in the global namespace), not a string. So it tried to check if it could install the gettext functions: `{'x', 't', 'e', 'g'}`, which was wrong.

The fix is to either import `ngettext` directly, so it can be used as-is, or globally install `ngettext` by passing it as a one-item list to the `gettext.install()` function. I've done the latter.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
@@ -27,7 +27,7 @@ import proxygsettings
 import SettingsWidgets
 
 # i18n
-gettext.install("cinnamon", "/usr/share/locale", names="ngettext")
+gettext.install("cinnamon", "/usr/share/locale", names=["ngettext"])
 
 # Standard setting pages... this can be expanded to include applet dirs maybe?
 mod_files = glob.glob(config.currentPath + "/modules/*.py")


eli-schwartz:
ref: https://bugs.archlinux.org/task/64543

clefebvre:
Already done in https://github.com/linuxmint/cinnamon/pull/8959.

eli-schwartz:
I... don't even understand that other PR. It seems to be re-doing 058e3a1e7dd26eca4bacb37b7c20c1e9d430de6a, fixing some gtk deprecation, and switching one incorrect use of gettext.install() for another, none of which has anything to do with the commit message (that claims this is for python 3.8).

eli-schwartz:
@clefebvre 

When the ngettext call was initially added in 6ac295a62556f22b755b17dee29d17d6721fa361, I think the person who added it tried to use `gettext.install(..., names="gettext")`, but that's totally incorrect. Please reopen this PR. I can either leave it to continue explicitly importing ngettext from gettext (it works for me on python 3.8) or switch the gettext.install() call to use `names=["gettext"]` as it is required to be a sequence and not a string (well, a string is a sequence too, but you don't exactly get the right answer when you use `set('ngettext')` instead of `set(['ngettext'])`).

I initially used from gettext import ngettext, because I wasn't sure under what circumstances the gettext.install() was guaranteed to have run, BTW. But I'm totally okay with either solution.

eli-schwartz:
I've switched to using gettext.install() with the correct sequence.

Issue #7974, opened 2018-11-01
Consistent Flatpak pinning support in ITM when debian version of application is present.


jaszhix:
Moved from jaszhix/icingtaskmanager#229, @smurphos wrote: 
#
 - Cinnamon 3.8.9
 - Extension version 6.3.13
 - Linux distribution Mint 19

Hi jaszhiz,

There a are a few inconsistencies in the way ITM handles flatpak app pinning when there is also a debian packaged version of the app installed which I hope can be resolved and also included in the fork ADW for Cinnamon 4. This issue is also partly prompted by this 18.3 users experience with snap package pinning - https://github.com/jaszhix/icingtaskmanager/issues/228, which was also raised on the Mint forum. I'm not sure if the same underlying issues cause the inconsistency with snap.

For this testing I've used VLC

Preinstalled version is 3.03 - desktop file is usr/share/applications/vlc.desktop

Flatpak version is 3.04 - desktop file is /var/lib/flatpak/app/org.videolan.VLC/current/active/export/share/applications/org.videolan.VLC.desktop
There is also a symlink in /var/lib/flatpak/exports/share/applications/org.videolan.VLC.desktop

Test 1 - Launch VLC flatpack from the menu, right click on icon and select pin to panel.
Result - vlc.desktop is pinned
Preferred result - org.videolan.VLC.desktop should be pinned.

Test 2 - Right click Flatpak VLC in menu and select add to panel.
Result - org.videolan.VLC.desktop is pinned
Launching from the pinned icon launches the correct version of VLC but another VLC icon is added to ITM, it doesn't group/count the window under the pinned org.videolan.VLC.desktop
Preferred result - Correct grouping of the window.

Test 3 - Use the menu editor to rename the Flatpack VLC entry to VLC Flatpak. This creates a new desktop file - _/.local/share/applications/org.videolan.VLC.desktop
Repeat test 1 - same result
Repeat test 2 - same result

Test 4 - Use the menu-editor to create a new user added menu entry targeting the flatpak version.
Repeat test 1 - same result
Repeat test 2 - same result

Test 5 - Use a different flatpak without a corresponding debian installed version. I used Oomox for testing - desktop file is /var/lib/flatpak/app/com.github.themix_project.Oomox/current/active/export/share/applications/com.github.themix_project.Oomox.desktop with a symlink in /var/lib/flatpak/exports/share/applications/com.github.themix_project.Oomox.desktop

Retried test 1 - works fine with no unexpected results.
[...]

PR #9029, opened 2019-12-19
[WIP] Gwl ungrouped fixes

mtwebster:
Some fixes for ungrouped mode

See commit messages.  I'm uncertain about the last commit (DND zones).

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -63,6 +63,7 @@ class AppGroup {
             windowCount: params.metaWindows ? params.metaWindows.length : 0,
             lastFocused: params.metaWindow || null,
             isFavoriteApp: !params.metaWindow ? true : params.isFavoriteApp === true,
+            transientFavorite: params.transientFavorite === true,
             autoStartIndex: findIndex(this.state.autoStartApps, (app) => app.id === params.appId),
             willUnmount: false,
             tooltip: null,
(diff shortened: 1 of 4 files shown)


mtwebster:
This is for fixing the issue brought up here: https://github.com/linuxmint/mint19.3-beta/issues/23

Along with its related complications.

@smurphos @MenthaSuaveolens:

If you could try this out it would be great.. I'm sort of inventing behavior here for ungrouped mode, it was a bit undefined.  I'd appreciate any feedback.

MenthaSuaveolens:
Fresh install of LM19.3. I can't reproduce the issue. 

I discovered another issue with the GIMP-icon when Gimp is installed by Flatpak. Starting GIMP creates another temp icon on the panel. It looks that the flatpak-app is considered as another app. The same happens with LibreOffice anyhow (writer-, calc- and impress-icons are created temporarily besides the LO-icon - which is ok, of course). 
FlatpakGimp

smurphos:
@mtwebster - this fixes the presenting issue. The DND tweak makes sense to me and seems to work fine.

smurphos:
@MenthaSuaveolens - there is an open issue for problems in app grouping with flatpak apps installed alongside repo version - https://github.com/linuxmint/cinnamon/issues/7974

landon912:
Any updates or ETA on this? 

It would be a huge QOL improvement.

GeorchW:
What's the issue with the PR? I can't look into the CI. I'd love to see this fixed.

dhunter:
Obviously can't talk to the CircleCI test failure with lmde3, but for others who are using Cinnamon on the regular version of Mint 20, and are tired of having to re-add icons and/or restart Cinnamon multiple times a day, I went to /usr/share/cinnamon/applets/grouped-window-list@cinnamon.org, (made a copy, just in case) and manually added the bits from this pull request, and for more than a month now, it's worked without a hitch.  Has been a huge improvement, so would strongly suggest manually adding until whatever lmde3 test can be corrected, (again, after copying the unedited version, just in case).

Ammako:
@dhunter if you need to make local edits, it's probably better to put the edited applet in ~/.local/share/cinnamon/applets rather than editing the files in usr.

PR #9112, opened 2020-01-12
modalDialog: Quiet an ES6 warning

JosephMcc:
We now access OPEN_AND_CLOSE_TIME outside of modalDialog.js so we need to
declare this with var instead of const


mtwebster:
...

Issue #8539, opened 2019-04-28
[FR] Please to add application spread function/option for modern look


pyroxar:
**Issue**
If I have open a multiple windows' one application, for example nemo, when I clicked to taskbar then maximize the last window only. If I want choose another windows than the last I must go mouse to taskbar and select another window.
![obraz](https://user-images.githubusercontent.com/23406555/56865843-cf17c180-69d2-11e9-8c71-9d1be441aac8.png)
On my laptop screen (1920×1080) this way is very uncomfortable for my because my laptop is small but screen have big resolution.

In my opinion good new feature to add will be application spread function.
Example of this, this was worked great for me.
https://youtu.be/NxD_kWK8A5M?t=78
![obraz](https://user-images.githubusercontent.com/23406555/56866042-61b96000-69d5-11e9-9d79-5129c3ec5f54.png)
Like this but only for the selected application:
![obraz](https://user-images.githubusercontent.com/23406555/56866079-b4931780-69d5-11e9-83f2-c8277d137b14.png)

@jaszhix what you think about this?

**Other information**
In the with nemo example when I use nemo can't see any details in windows to select. It is very uncomfortable.
![syf](https://user-images.githubusercontent.com/23406555/56865998-b8726a00-69d4-11e9-984d-4b3d15798b12.png)

mtwebster:
One existing option, alt-~ opens an alt-tab switcher for all windows of a particular application only.

pyroxar:
Are you sure?
![obraz](https://user-images.githubusercontent.com/23406555/56869509-fe91f280-6a01-11e9-8f50-9ba9686fdf92.png)

JosephMcc:
The hotkey shown by mtwebster works fine here on git master. Brings up only other windows for the currently active application window.

pyroxar:
In my opinion application spread like in the Unity's Ubuntu will be very great function for everone.
In Unity in the Ubuntu 14.04 and the 16.04, was better for me
![obraz](https://user-images.githubusercontent.com/23406555/56869889-4404ee80-6a07-11e9-8483-fd4c2f978af3.png)

than Ubuntu 19.04+ with GNOME:
![obraz](https://user-images.githubusercontent.com/23406555/56869879-1c158b00-6a07-11e9-9a7a-3ef69be603bf.png)

Preview is too small for me.

pyroxar:
When i click first time then last open window is restored. When i click second time then show all windows of one type. This is very helpful.
[...]

PR #9133, opened 2020-01-23
Add `Activate, then cycle` to grouped window list

Epskampie:
Hi, I feel that the `cycle` option in the `grouped window list` applet should activate the application (if it doesn't have focus) on first click, then cycle the windows on subsequent clicks. However, since some people may depend on the current workflow, I've added it as an option: `Activate, then cycle`.

My rationale for this mode is as follows.  The current `switch windows` works like this:
* Say I've got two Firefox windows, 1 and 2, and a Terminal window. 
* I'm working in Firefox window __1__, then I need to do something in terminal, so I click its icon.
* Terminal activates, all is well. I need to go back to firefox, so I click its icon.
* Firefox __2__ comes to the front, and I'm confused, why am I seeing a different window suddenly?

For this reason I feel that the first click should active, but then you can quickly cycle windows by clicking again.

I realise some people may feel strongly about this, so please let me know your thoughts!

Edit: you can quickly try this by using the renamed extension I made: https://github.com/Epskampie/grouped-window-list-epskampie

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -750,7 +750,24 @@ class AppGroup {
                 return;
             }
             if (this.state.settings.leftClickAction === 3 && nWindows > 1) {
-                this.state.trigger('cycleWindows', null, this.actor._delegate);
+                let foundActive = false;
+                for (let i = 0, len = nWindows; i < len; i++) {
+                    if (
+                        this.groupState.lastFocused &&
+                        this.groupState.metaWindows[i] === this.groupState.lastFocused
+                    ) {
+                        if (this.groupState.metaWindows[i].appears_focused) {
+                            this.state.trigger("cycleWindows", null, this.actor._delegate);
+                        } else {
+                            handleMinimizeToggle(this.groupState.metaWindows[i]);
+                        }
+                        foundActive = true;
+                        break;
+                    }
+                }
+                if (!foundActive) {
+                    handleMinimizeToggle(this.groupState.metaWindows[0]);
+                }
                 return;
             }
             if (this.hoverMenu) this.hoverMenu.shouldOpen = false;


clefebvre:


```diff

Complexity increasing per file
==============================
- files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js  7
         

Clones added
============
- files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=4872794&bid=16153089)

Epskampie:
Is there anything I can do to have this move forward?

pyroxar:
@Epskampie what you think about this?
https://github.com/linuxmint/cinnamon/issues/8539

Epskampie:
@pyroxar you can quickly try this by using the renamed extension I made: https://github.com/Epskampie/grouped-window-list-epskampie

To install: [Download the zip](https://github.com/Epskampie/grouped-window-list-epskampie/archive/master.zip) and decompress it in `~/.local/share/cinnamon/applets/grouped-window-list@epskampie`. Then rightclick the panel and select `applets` to add it.
Don't forget to configure the applet. Change `left click action` to `active, then cycle`.

pyroxar:
I can't configure: 
obraz
Should I off the original applet?

Epskampie:
Right click:
Selection_009

pyroxar:
I can't run your applet, I have only originally cinnamon applet
obraz

mtwebster:
I'm confused -

Take a look at settings, this is already an option, from what I can see - under general, look at 'left-click-action'.  If I set it to cycle windows, it behaves (I think) exactly how you want it to.

The normal option - toggle activation of last focused window, sounds like what you expect the current behavior should be.  If what you describe is happening, then that's something that needs to be fixed.

Epskampie:
For me, the default `cycle` will always cycle windows, even if the program wasn't selected. This makes it extremely annoying to switch between two programs a lot if at least one of them has multiple windows.

The `toggle activation of last focused window` doesn't give you an easy way to switch windows. With my tweak, you can just click the taskbar icon more then once.

mtwebster:
Ah sorry - I didn't try the workflow you described with that cycle option selected. I can reproduce the issue of it always choosing another window in the group when you go back to it.  I can't imagine this was intended - it should behave the way your PR is intended to work.

So:

- toggle activation... should do just that, click it, last-used window should focus.  Click again, it should go away.
- cycle windows - should do exactly like the first option, except subsequent clicks should cycle thru the window group.  It should also remember the last window that was focused, and return to that, but this is not happening - instead it's continuing the cycle even though you were just working in an entirely different application.

I'm not sure why 'none' is an option, that's just silly, but that's another discussion.

So, I'd prefer to either a) Fix the existing option (which should *already* be working in the manner you want here), or b) replace the code related to that option with yours (I'm totally ok with this route, if it works reliably).

Epskampie:
I agree, I feel like the current "cycle" functionality is broken a bit, and should be fixed. However, I do not know people's workflow, so maybe I'm wrong, in that case adding an extra option would be fine too.

However, no maintainer seems to be active here, so this whole PR seems to be an execise in futility. :-)

mtwebster:
I'm sorry if you think so, I gave some feedback, I have no problem using this pr but I'd rather replace the existing option which I think is broken.  If someone complains later I'll deal with it then.

I know it can be frustrating waiting for responses but unfortunately other things take priority and sometimes there just isn't time to dedicate to this stuff.

Epskampie:
Ah, my bad, I didn't see you are one of the contributors to this repo, I have to get used to the way that gihub shows that.

I'll change this PR so that it modifies the `cycle` option.

clefebvre:


```diff

Clones added
============
- files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=4872794&bid=16153089)

Epskampie:
I've changed this pull request to change the `cycle` option. As you can see the code change is pretty short, and most of the new code was taken from the existing `toggle activation` option.

mtwebster:
Looks good, thanks!

PR #9163, opened 2020-02-14
Update xlet-settings: Add the width and height optional properties to a layout

claudiux:
Example of `settings-schema.json` file:
```
{
  "layout1": {
    "type": "layout",
    "height": 800,
    "width": 1000,
    "pages": [
      "page1",
      "page2"
    ],
[...]
```
Sometimes the size of the layout is not large enough, or too large, depending on what it contains.
The `height` and `width` properties are optional.

```
| Property | Min | Default |      Max      |
|  "height"| 200 |   600   | screen height |
|  "width" | 600 |   800   | screen width  |
```
If the "height" value is lesser than 200, then it is set to 200.
If it is greater than the screen height, then it is set to the screen height.
if the "height" value is missing, or incorrect, then it is set to the default value.

Ditto for the width value.

EDIT: The errors do not concern these changes.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py
@@ -95,7 +95,7 @@ class MainWindow(object):
         ##   -t <n> or --tab=<n>, where <n> is the tab index (starting at 0).
         ##   -i <id> or --id=<id>, where <id> is the id of the instance.
         ## Examples, supposing there are two instances of Cinnamenu@json applet, with ids '210' and
-        ## '235' (uncomment line 144 containing print("self.instance_info =", self.instance_info)
+        ## '235' (uncomment line 147 containing print("self.instance_info =", self.instance_info)
         ## to know all instances ids):
         ## (Please note that cinnamon-settings is the one offered in #8333)
         ## cinnamon-settings applets Cinnamenu@json         # opens first tab in first instance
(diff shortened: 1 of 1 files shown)


mtwebster:
Does the widget system not allocate enough room for the full layout?  Is it something we could address in a specific widget that always causes it?

claudiux:
It would be ideal.
We would have to find a way to fix the size of the window once the size of all widgets is fixed. And often translations change the size of widgets.

mtwebster:
Hi, can you give this a try?
```
diff --git a/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py b/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py
index 4be766a29..0302feb3e 100755
--- a/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py
@@ -247,7 +247,7 @@ class MainWindow(object):
         self.menu_button.set_popup(menu)
 
         scw = Gtk.ScrolledWindow()
-        scw.set_policy(Gtk.PolicyType.NEVER, Gtk.PolicyType.AUTOMATIC)
+        scw.set_policy(Gtk.PolicyType.NEVER, Gtk.PolicyType.NEVER)
         main_box.pack_start(scw, True, True, 0)
         self.instance_stack = Gtk.Stack()
         scw.add(self.instance_stack)
@@ -260,7 +260,17 @@ class MainWindow(object):
                 self.window.set_icon_from_file(icon_path)
         self.window.set_title(translate(self.uuid, self.xlet_meta["name"]))
 
+        def check_sizing(widget, data=None):
+            minreq, natreq = self.window.get_preferred_size()
+            monitor = Gdk.Display.get_default().get_monitor_at_window(self.window.get_window())
+
+            height = monitor.get_workarea().height
+            if natreq.height > height - 100:
+                self.window.resize(800, 600)
+                scw.set_policy(Gtk.PolicyType.NEVER, Gtk.PolicyType.AUTOMATIC)
+
         self.window.connect("destroy", self.quit)
+        self.window.connect("realize", check_sizing)
         self.prev_button.connect("clicked", self.previous_instance)
         self.next_button.connect("clicked", self.next_instance)

```

This keeps the current size under normal circumstances (800x600) but will expand if required to accommodate a taller request.  However, *if* that request ends up being bigger than the monitor's workarea (monitor size - panels) than it will revert back to the original size and turn on scrolling.

I tested this a bit with the grouped window list settings with German (only applet I know that has a lot of settings) but maybe you can play around with it and see if you can break it.  I'd rather do something like this than hardcode a size - unless you test every language you won't know if you've allowed enough.

mtwebster:
I merged the changes I posted above, let's see how that works for now, I think it's definitely preferable to guessing what might be enough.

https://github.com/linuxmint/cinnamon/commit/9a3ae39adc83473d55ba63b52ace146994588079

If it doesn't work out we can revisit this.

Issue #9196, opened 2020-03-05
Full date format in the calendar applet


sidorsett:
**Issue**
The default format string of "%A %B %-e, %Y" for the full date on top of the calendar applet is hard-coded, but can be weird in some localization.

**Expected behaviour**
The format string shall be configurable.

sidorsett:
This is already fixed in #9042

Issue #9042, opened 2019-12-21
Wrong date format on non-English locales


mraandtux:
```
 * Cinnamon version: 4.4.5
 * Distribution: Mint 19.3 (updated from 19.2)
 * Graphics hardware *and* driver used: just Intel
 * 64 bit
 ```

**Issue**
Wrong date format, not as per LC_TIME=zh_CN.UTF-8 
![2019-12-21 18-06-14屏幕截图](https://user-images.githubusercontent.com/2002796/71306793-ff0f4b80-241f-11ea-90ac-8ae8a2b86ae2.png)

**Steps to reproduce**

*Change locale other than English (US)
*Check the calendar applet or uncheck "Internet time" box on Date & Time setting to see the date format

**Expected behaviour**

The previous Cinnamon version that shipped with 19.2 DOES NOT occur with this bug.

clefebvre:
can you paste the output of `locale` and apt version `libcinnamon-desktop4`?

clefebvre:
Also, write down in Chinese the expected date and explain its components (month, week, day..) so we can troubleshoot what's happening.

mraandtux:
My libcinnamon-desktop4 version is 4.4.1, the locale command shows zh_CN.UTF-8 is on strings except LANGUAGE (zh_CN) and LC_ALL (blank) .

The expected date in Chinese with Ruby date syntax is: %Y年%b%d日 %A (a la YYYY/MM/DD *day)

cybernaut-one:
I can confirm the bug for locale LC_TIME=sr_RS@latin. I have the interface language set to en_US UTF-8 and the Region and Time Format set to sr_RS@latin. I have the same version of libcinnamon-desktop4 (4.4.1). The expected date format in Serbian (Latin) is correct in /usr/share/i18n/locales/sr_RS@latin (%A, %d. %B %Y.), but in Calendar applet (and tooltip) it shows the en_US date format (%A %d %b %Y) literally translated into Serbian (Latin) language

claudiux:
Similar bug for `LC_TIME=fr_FR.UTF-8`. The calendar shows `Dimanche 22 décembre, 2019` (title and tooltip) instead of `Dimanche 22 décembre 2019` as expected.

Karpfly2822:
[...]

PR #9197, opened 2020-03-05
calendar@cinnamon.org: patch for issue #9196

sidorsett:
Patch for the issue #9196

### DIFF
--- a/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
@@ -64,7 +64,7 @@ class CinnamonCalendarApplet extends Applet.TextApplet {
 
             this.menu.addMenuItem(item);
 
-            this._dateFormatFull = CinnamonDesktop.WallClock.lctime_format("cinnamon", "%A, %B %-e, %Y");
+            this.settings.bind("custom-date-format", "_dateFormatFull", this._onSettingsChanged);
 
             this.settings.bind("use-custom-format", "use_custom_format", this._onSettingsChanged);
             this.settings.bind("custom-format", "custom_format", this._onSettingsChanged);
--- a/files/usr/share/cinnamon/applets/calendar@cinnamon.org/settings-schema.json
+++ b/files/usr/share/cinnamon/applets/calendar@cinnamon.org/settings-schema.json
@@ -18,16 +18,20 @@
     "custom-format" : {
         "type" : "entry",
         "default" : "%A, %B %e, %H:%M",
-        "description" : "Date format",
+        "description" : "System tray label",
         "indent": true,
         "dependency" : "use-custom-format",
-        "tooltip" : "Set your custom format here."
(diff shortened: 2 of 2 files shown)


mtwebster:
I don't think there's a need for an additional setting - If you prefer it a certain way on the panel why not use that same format in the popup?  This is the only thing that needs fixing I think.

sidorsett:
Hello, Michael.

Thanks for sharing your considerations. From my side, it is an attempt to probe the ground and to learn how the open SW community works. This is my first code contribution ever, and I have some more ideas ;) Therefore, I really appreciate your response!

> I don't think there's a need for an additional setting - If you prefer it a certain way on the panel why not use that same format in the popup? This is the only thing that needs fixing I think.

Indeed, I and probably many others prefer to have them in different ways:
- Time only in the panel.
- Date in the popup.

**Rationale:** Date takes a lot of space in the panel, it doesn't change often, and you don't need to reference it often. Hence, it is fine to "hide" the date into the popup, where it can get as much space, as necessary. The time on the other hand takes the moderate amount of space on the panel; the time changes often; this is what you reference frequently and want to have visible without additional mouse clicks.

Time is what all versions of Windows have on the panel. The full date is in the popup.
"Windows" does not sound as a good authority? Well, this calendar applet itself has defaults of time for the panel and of date for the popup. Though, the applet gives you an option to configure label in the panel, but doesn't give you an option to set up format of the date in the popup. I am absolutely fine with the default setting for the panel (i.e. time). I rather don't need the configuration for the that (i.e. for the panel). But I also want to have the correct format of the date in the popup. And I am not alone. You can find people facing the similar issue for years. See [this](https://forums.linuxmint.com/viewtopic.php?t=244125) for example. Is this the right option for them to change the code? No!

It would be ideal to use the localization settings for the date in the popup. I don't know, how to do that. If you give me the hint, I can make the changes myself.

Cheers!

mtwebster:
Ok fair enough - did a little more investigating on the history of this applet and my suggestion wouldn't have been well-received :)

We need to have the default (current) value continue to be translated thru CinnamonDesktop.WallClock.lctime_format().  That is the default string that gets translated (so it has to be exactly that).

So I would say add another switch specifically for the new field.  So we'd have two switches with two hidden custom fields (which I'm not crazy about but otherwise if someone wants just to change the panel label, they'd lose the date translation we currently perform).

Also, let's use these labels:

switch: Use a custom format for the panel
label: Panel format
tooltip: Set the custom format for the panel

and

switch: Use a custom date format for the calendar
label: Calendar date format
tooltip: Set a custom date format for the calendar

We'll see how that looks..

sidorsett:
Hello, Michael.

Thanks for the feedback. Trying to understand how lctime_format() works, I encountered the issue #9042, which seems to fix the problem in Cinnamon 4.4.8. Hence, I am thinking that the fix proposed by me is not necessary. Though, I would like to play a bit with my older Cinnamon version, before recalling the pull request.

I'll be back.

Cheers!

PR #9230, opened 2020-03-31
Added new keywords

pyroxar:
Added new keywords "microphone" and "headphones" for sound.


mtwebster:
Can we just do "headphone" (no s)

pyroxar:
I made mistake because "Słuchawki" in polish are plural.

mtwebster:
It is in english also, but for keywords it's best to keep singular forms so the search matches sooner.

PR #9239, opened 2020-04-07
If statement appears to serve no purpose

StepDub:
This is my first attempt to make a change to someone else's code so feel free to give me feedback.
These 3 lines don't seem to have a purpose that I can see and I've checked at least 10 popular themes and they all work fine without them. Also, they appear to cause a bug: If the style value "height:" is set in the node of the css theme file, it makes no difference anyway, the height value is set to 30 regardless (I'm not sure way this is.) So If the text size (set in the css file) is too large, it disappears of the bottom - this is how I discovered the bug.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -2466,11 +2466,6 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
         this.mainBox._delegate = null;
 
         this.selectedAppBox = new St.BoxLayout({ style_class: 'menu-selected-app-box', vertical: true });
-
-        if (this.selectedAppBox.peek_theme_node() == null ||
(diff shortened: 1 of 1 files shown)


StepDub:
Sorry, I need to investigate why these lines were here in the first place.

clefebvre:
Hi,

Thanks for contributing. Don't hesitate to use git blame to see why lines are the way they are.

https://github.com/linuxmint/cinnamon/blame/master/files/usr/share/cinnamon/applets/menu%40cinnamon.org/applet.js

That shows you the commits which introduced them, and if you're lucky you might get a rationale in the commit message or in the comments.

https://github.com/linuxmint/cinnamon/commit/08742aa68838f95c1de419c146f096d4c1a387d3
https://github.com/linuxmint/cinnamon/commit/595f2013af4b9b2ef027d2c6b75bd5621f3eb26e

Worst case scenario you get no explanations, but at the very least you get to know who wrote the lines :)

fredcw:
Nice, thanks.

PR #9244, opened 2020-04-11
[WIP] calendar@cinnamon.org: full time (with seconds) on top of the calendar

sidorsett:
Hi,

It might be more efficient from space utilization to display time without seconds in the task bar. Though, it is occasionally useful to have a full time with seconds. For comparison, all versions of Windows have these: older versions had an analog clock with the second hand, and the new versions have a digital clock with seconds.

This pull request addresses this need adding the full time on top of the calendar applet.

Cheers!

### DIFF
--- a/data/theme/cinnamon.css
+++ b/data/theme/cinnamon.css
@@ -586,6 +586,10 @@ StScrollBar StButton#vhandle:hover {
     color: #cccccc;
     font-weight: bold;
 }
+.datemenu-time-label {
+  font-size: 200%;
+  padding-bottom: 0;
+}
 .calendar-day-base {
     text-align: center;
     width: 2.4em;
--- a/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
@@ -145,7 +145,7 @@ class CinnamonCalendarApplet extends Applet.TextApplet {
         if (this.use_custom_format) {
             if (!this.clock.get_clock_for_format(this.custom_format)) {
                 global.logError("Calendar applet: bad time format string - check your string.");
-                this.clock.set_format_string("~CLOCK FORMAT ERROR~ %l:%M %p");
+                this._labelFormat = "~CLOCK FORMAT ERROR~ %l:%M %p";
             } else {
                 this._labelFormat = this.custom_format;
             }
--- a/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
@@ -128,17 +128,17 @@ class CinnamonCalendarApplet extends Applet.TextApplet {
         let show_seconds = this.desktop_settings.get_boolean("clock-show-seconds");
 
         if (use_24h) {
-           this.clock.set_format_string(_("%H:%M:%S"));
-           this._labelFormat = "%H:%M";
+            this.clock.set_format_string(_("%H:%M:%S"));
+            this._labelFormat = "%H:%M";
         } else {
-           this.clock.set_format_string(_("%l:%M:%S"));
-           this._labelFormat = "%l:%M";
+            this.clock.set_format_string(_("%l:%M:%S"));
+            this._labelFormat = "%l:%M";
         }
         if (show_seconds) {
-           this._labelFormat += ":%S";
+            this._labelFormat += ":%S";
         }
         if (in_vertical_panel) {
-           this._labelFormat = this._labelFormat.replace(/:/g,"%n");
+            this._labelFormat = this._labelFormat.replace(/:/g, "%n");
         }
         this._labelFormat = _(this._labelFormat);
 
--- a/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
@@ -138,22 +138,23 @@ class CinnamonCalendarApplet extends Applet.TextApplet {
            this._labelFormat += ":%S";
         }
         if (in_vertical_panel) {
-           this._labelFormat = this._labelFormat.replace(/:/g,"%n")
+           this._labelFormat = this._labelFormat.replace(/:/g,"%n");
         }
-        this._labelFormat = _(this._labelFormat)
+        this._labelFormat = _(this._labelFormat);
 
         if (this.use_custom_format) {
             if (!this.clock.get_clock_for_format(this.custom_format)) {
                 global.logError("Calendar applet: bad time format string - check your string.");
                 this.clock.set_format_string("~CLOCK FORMAT ERROR~ %l:%M %p");
+            } else {
+                this._labelFormat = this.custom_format;
             }
-        } else {
-            this._labelFormat = this.custom_format;
         }
     }
 
     _updateClockAndDate() {
         let timeFormatted = this.clock.get_clock();
+        timeFormatted = timeFormatted.capitalize();
         this._time.set_text(timeFormatted);
 
         let label_string = this.clock.get_clock_for_format(this._labelFormat);
--- a/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
@@ -145,7 +145,7 @@ class CinnamonCalendarApplet extends Applet.TextApplet {
                 this._labelFormat += _(":%S");
             }
             if (in_vertical_panel) {
-                this._labelFormat = this._labelFormat.replace(/:/g,"%n")
+                this._labelFormat = this._labelFormat.replace(/:/g, "%n")
             }
         }
     }
--- a/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
@@ -45,6 +45,11 @@ class CinnamonCalendarApplet extends Applet.TextApplet {
             this._initContextMenu();
             this.menu.setCustomStyleClass('calendar-background');
 
+            // Time
+            this._time = new St.Label();
+            this._time.style_class = 'datemenu-date-label datemenu-time-label';
+            this.menu.addActor(this._time);
+
             // Date
             this._date = new St.Label();
             this._date.style_class = 'datemenu-date-label';
@@ -81,6 +86,8 @@ class CinnamonCalendarApplet extends Applet.TextApplet {
             }));
 
             this.clock_notify_id = 0;
+
+            this.clock.set_format_string("%X");
 
             // https://bugzilla.gnome.org/show_bug.cgi?id=655129
             this._upClient = new UPowerGlib.Client();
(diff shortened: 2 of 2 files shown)


clefebvre:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=5343822&bid=17495588)

Krafting:
I say yes to this! doesn't seems like much, but it would be so useful to have the second here at all time, without seeing it! (I used it all the time when I was using windows 10)

sidorsett:
Hi, Krafting.

Thanks for the support!

Cheers.

mtwebster:
Hi, this won't work,

- The time/date shown on the panel *must* follow the preferences shown in the global date/time settings (this pr ignores that):
image
- If the user wants to change this, they turn on "Use custom data format" in the applet's settings (this pr also ignores this).

You also can't add translated partial strings together:
```
                this._labelFormat += _(":%S");
```
The translated (`_()`) formatting strings that are currently in the applet are translated based on that *exact* string, and you can't assume order (or punctuation) remains the same in different languages and locales.

I would consider adding the time to the calendar view, but the format should to be the exact same as the panel format (with seconds turned on, if you prefer, while still respecting the 24h option).

sidorsett:
Hello, Michael.

Thanks for the feedback. I may disagree with some points, but in the end I have learned something new (about translations) from it!

> The time/date shown on the panel must follow the preferences shown in the global date/time settings (this pr ignores that)

No, the PR does _not_ ignore that.

- For the date.

The parameter is handled in /usr/share/cinnamon/cinnamon-settings/modules/cs_calendar.py
```
settings.add_row(GSettingsSwitch(_("Display the date"), "org.cinnamon.desktop.interface", "clock-show-date"))
```

If to search for "clock-show-date" in the cinnamon repo, you will not find it anywhere else, including the calendar@cinnamon.org/applet.js. Moreover, I couldn't find it in the "blame" for the calendar applet, meaning that it has _never_ been used here.

I could add the date to the panel label based on this setting, but in another PR (because it is a different issue).



- For the time.

The global settings for time in the panel label work just fine (except the translation issue discussed further). I have tested it. Here is the code, which takes care of this:
```
            this.desktop_settings.connect("changed::clock-use-24h", Lang.bind(this, function(key) {
                this._onSettingsChanged();
            }));
            this.desktop_settings.connect("changed::clock-show-seconds", Lang.bind(this, function(key) {
                this._onSettingsChanged();
            }));
```
```
            let use_24h = this.desktop_settings.get_boolean("clock-use-24h");
            let show_seconds = this.desktop_settings.get_boolean("clock-show-seconds");
            if (use_24h) {
                this._labelFormat = _("%H:%M");
            } else {
                this._labelFormat = _("%l:%M");
            }
            if (show_seconds) {
                this._labelFormat += _(":%S");
            }
```
```
        let label_string = this.clock.get_clock_for_format(this._labelFormat);

        this.set_applet_label(label_string);
```



> If the user wants to change this, they turn on "Use custom data format" in the applet's settings (this pr also ignores this).

This is also not true. I have tested it, and it works fine for the panel label too. Here is the code sequence:

```
            this.settings.bind("use-custom-format", "use_custom_format", this._onSettingsChanged);
            this.settings.bind("custom-format", "custom_format", this._onSettingsChanged);
```
```
        if (this.use_custom_format) {
            if (!this.clock.get_clock_for_format(this.custom_format)) {
                global.logError("Calendar applet: bad time format string - check your string.");
                this._labelFormat = _("~CLOCK FORMAT ERROR~");
            } else {
                this._labelFormat = this.custom_format;
            }
        } else {
```
```
        let label_string = this.clock.get_clock_for_format(this._labelFormat);

        this.set_applet_label(label_string);
```


> You also can't add translated partial strings together.

This is indeed my fault. I knew that the translation takes place by the exact string, but I didn't know that `_()` states for the translation.
I will fix it with the new commit.

> The format should to be the exact same as the panel format (with seconds turned on, if you prefer, while still respecting the 24h option).

I agree to the 24h option. I will fix it with the new commit too.
Though, the string format in the applet window cannot and doesn't need to be _exact_ same as in the panel (even if appended with seconds). Because user may configure to display time and date in the panel. In this case, it doesn't make sense to display the exact same string in the applet again, because there is already the dedicate date label present in the applet window.
You cannot resolve this by removing the _existing_ date label in the applet window either, because the user may want to display only time in the panel, but the full date in the applet window is always necessary.
Moreover, separate strings for date and time in the applet window allow different styling for the labels (like `font-size:200%` for the time label in my commit).
Overall, it doesn't make sense to get the panel label append it with seconds and put it into the applet window, because this way you will force user to either overload the panel with the date or to miss the date everywhere. Hence, I believe that my proposal is better (except that it shall take the 24h setting into the account).


To sum up the flows:
- The translation of the panel label must be fixed.
- The time label in the applet window must additionally take 24h settings into account.

Agree?

Cheers!

sidorsett:
Done

clefebvre:


```diff

Issues
======
- Added 7
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=5343822&bid=17495588)

clefebvre:


```diff

Issues
======
- Added 7
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=5343822&bid=17495588)

sidorsett:
work in progress

sidorsett:
Hello, Michael.

Please, check if this is something you were looking for.

Cheers!

mtwebster:
Ok.  I should have been more clear last time.

I'm ok with adding the time to the calendar.  We can find the previously translated format for time to display there, and make it customizable from the applet's settings.

But first, 

Take a look here:
https://github.com/linuxmint/cinnamon-desktop/blob/master/libcinnamon-desktop/gnome-wall-clock.c#L63
and here:
https://github.com/linuxmint/cinnamon-desktop/blob/master/libcinnamon-desktop/gnome-wall-clock.c#L288-L340

These formats have been translated, and are the ones used based on the options set in Date and Time settings.  If there is no custom format set in the applet, these are what should be used, and it's accomplished by `this.clock.set_format_string(null)`.  You use `this.clock.get_time()` to get the date/time string based on those translations.  This is not being done here.  We're only getting whatever the contents of `this._labelFormat` is at the time.  It ignores the date preference altogether ("Display the date" setting).  And this is also unnecessarily adding new translatable format strings.

Also, I'm not even certain [this line](https://github.com/linuxmint/cinnamon/pull/9244/files#diff-a9c28db946e4a2b541a9422809387fa0R143) even works (though we shouldn't need it anyhow).

Honestly, I would leave what's currently there alone, and just worry about the time in the menu.  There's been a lot of tweaking of this code over the years to get it just right (it's very very easy to screw up formatting for different language and locales, not just from what we do, but also due to libraries we use).

You can add another switch and entry field in the applet settings.  If the switch is off, I would use `this.clock.get_clock_for_format()` along with the format strings from the gnome-wall-clock.c file.  And follow the same settings as the applet label does *excluding* the date option.

To be more clear, check the 24h and seconds settings, and pass these strings to `get_clock_for_format()`:
```

/* Default date/time format when show-seconds, use-24h are set */
#define NO_DATE_24H_SECONDS   (_("%R:%S"))

/* Default date/time format when just show-seconds is set */
#define NO_DATE_12H_SECONDS   (_("%l:%M:%S %p"))

/* Default date/time format when just use-24h is set */
#define NO_DATE_24H           (_("%R"))

/* Default date/time format with no options are set */
#define NO_DATE_12H           (_("%l:%M %p"))
```
(the extra enclosing parentheses around the format string can be left out - they're C-language specific only).

If the custom format setting is enabled, you just pass that format to `get_clock_for_format()`.

This is a bit lengthy, I apologize - I should have done this the first time.  I see value in having the time in the menu, I just don't think it's necessary to change so much to get there.

sidorsett:
Hello, Michael.

I appreciate your elaborated feedback very much! I need to dig into it. I will come with another PR, if I have a better proposal.

Cheers!

sidorsett:
Hello, @mtwebster 

I understand that you propose not to touch `this.clock.set_format_string()`, but just to generate one another label for time with `get_clock_for_format()`. I am afraid, it won't work though, because `this.clock.set_format_string(null)` is set with respect to the time settings, **including the `clock-show-seconds` option**. As a result, if `clock-show-seconds` is set to false, time label acquired with `get_clock_for_format()` will not update every second, but only once in a minute.

In other words, I used `set_format_string()` on purpose. I have to do that! But of course I can feed it with proper formats: `_("%R:%S")` or `_("%l:%M:%S %p")`, depending on `clock-use-24h` setting.

Aside of that I don't see enough benefit (and consistency) behind adding another switch and entry field for the custom time format in applet settings:
- There is an option for the custom **applet label** format, because people may be willing to have different information in their panel: some may desire for time and date, others--only for time; some may desire for time with seconds, others--without. This is the reason, why there is option for the custom format.
- There is no option for the custom **date** format in the menu, because the preferred date format has to be globally defined in the locale settings.
- Similarly to the date label, there is no need for the custom **time** format in the menu, because the preferred time format is defined in the locale settings.

I look forward to hearing your feedback on that.

Cheers!

mtwebster:
> In other words, I used set_format_string() on purpose. I have to do that! But of course I can feed it with proper formats: _("%R:%S") or _("%l:%M:%S %p"), depending on clock-use-24h setting.

This is what I meant, use those formats with get_clock_for_format().

I had forgotten that we only update on notifies from the wall clock (which is sad because I wrote it).  These timers used to be in the applet itself.  You'd have to set up your own 1 second update timer here if the global setting wasn't for seconds.  And it needs to start and stop running when the menu opens/closes.  There's a menu `open-state-changed` signal you can listen to.

Krafting:
Hi, is this PR dead, this is a feature I'd really like to have on Cinnamon!
Obviously I understand the mint team are entirely focused on Mint 20 but is this PR still worked on ?

sidorsett:
Hi, Krafting.

No, I decided to give up the feature. The reason is that I think that the proposed solution with `open-state-changed` won't work. Or better to say it would work crookedly. Though, I didn't try to implement it, my conclusion is based purely on a thought experiment.

I think, the result will be as follows:
- Say, you open menu at 00:00:58.9. The applet will show you 00:00:58.
- If it implemented the own counter, it would update exactly in 1 second. I.e. it will change to 00:00:59 at 00:00:59.9.
- Next, however, the gnome watch will trigger the update on the minute change. Hence, the applet will change to 00:01:00 at 00:01:00.
- Next trigger by the own timer won't cause any time update: at 00:01:00.9, the applet will still show 00:01:00.
- And the next update on the own timer will take place one second later: at 00:01:01.9, the applet will show 00:01:01.

I.e. we will have update intervals of:
- 1 second;
- 0.1 second;
- 1.9 seconds.

P.S. I am not from the MINT team. I am just a random guest, who tried to implement the desired feature oneself.

Cheers!

Krafting:
I understand the problem here, and this is kind of a shame, because this would be awesome.
There isn't any function that you could call to get the exact time, then find the number of ms left until the time change to the next second and then just run a 1s timer ?

sidorsett:
Hi, Krafting.

I don't offend, if you take over the PR.

Cheers!

Issue #3144, opened 2014-05-15
[Solved][Bug] "mini-shake" of LinuxMint Menu when scrolling between items


highwindmx:
I'm in LinuxMint17 RC, then
Let's see this video first: http://youtu.be/_1CLVP03usY
Due to the description of the items in the menu are not fully translated(English/Chinese)
the height of these characters is not consistent, the menu shows "mini-shake" when scrolling between these items.

May it be corrected by modifying the theme css file?

PR #9246, opened 2020-04-12
menu@cinnamon.org: alternative fix for jumpiness bug.

fredcw:
The current fix ([here](https://github.com/linuxmint/cinnamon/commit/08742aa68838f95c1de419c146f096d4c1a387d3)) to prevent "jumpiness" [#3144](https://github.com/linuxmint/cinnamon/issues/3144) when different scripts are used works by always setting the selectedAppBox height to 30 as get_theme_node().get_length('height') always seems to return 0 and get_theme_node().get_height() returns -1 meaning that selectedAppBox height is always set to 30. Also, fixing the selectedAppBox height upon applet initialization means that when the cinnamon theme, text scaling factor or user interface scaling are changed, the fixed height of selectedAppBox doesn't change until the applet is restarted. There's also the problem that setting the height of selectedAppBox in cinnamon theme files (.menu-selected-app-box) takes no account of text scaling factor so this should not be a preferred method anyway.

This alternative fix fixes these problems by setting the height to get_preferred_height() every time the menu is opened. Although strangely, when the menu is first opened after a change (eg. text scaling factor,) nothing happens until after the menu is opened for a second time. I don't know why this is. The line this.selectedAppBox.set_height(-1); is necessary to unset the previously set height. This means that the height setting in the theme css file is effectively ignored which I guess is a good thing because this setting was most likely only used as a fix for this bug anyway and would otherwise be a problem.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1231,6 +1231,9 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
 
             this.lastSelectedCategory = null;
 
+            this.selectedAppBox.set_height(-1); //unset previously set height
+            this.selectedAppBox.set_height(this.selectedAppBox.get_preferred_height(-1)[1]);
+
             let n = Math.min(this._applicationsButtons.length,
                              INITIAL_BUTTON_LOAD);
             for (let i = 0; i < n; i++) {
@@ -2466,11 +2469,6 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
         this.mainBox._delegate = null;
 
         this.selectedAppBox = new St.BoxLayout({ style_class: 'menu-selected-app-box', vertical: true });
-
-        if (this.selectedAppBox.peek_theme_node() == null ||
-            this.selectedAppBox.get_theme_node().get_length('height') == 0)
-            this.selectedAppBox.set_height(30 * global.ui_scale);
-
         this.selectedAppTitle = new St.Label({ style_class: 'menu-selected-app-title', text: "" });
         this.selectedAppBox.add_actor(this.selectedAppTitle);
         this.selectedAppDescription = new St.Label({ style_class: 'menu-selected-app-description', text: "" });


smurphos:
> There's also the problem that setting the height of selectedAppBox in cinnamon theme files (.menu-selected-app-box) takes no account of text scaling factor so this should not be a preferred method anyway.

If the theme uses em (as any well written theme should) to set a height that should take into account text scaling factor.

fredcw:
>If the theme uses em (as any well written theme should) to set a height that should take into account text scaling factor.

Thanks, I didn't know this, and you're right, it adjusts to scaling as well. So I think the best solution would be for theme files to include a relative height setting. In any case the applet as it stands just sets the height to 30px regardless and I don't know of a way to detect if the theme file sets a height  or how to add a relative one if they don't from within the applet. One benefit of my solution is that a suitable value is always set. So yeah, I think ideally it should be left to the theme file to set a height and that my solution is not the way to go. What does anyone else think?

smurphos:
@fredcw  - I've been messing a little with this this morning as it has always irritated me that I couldn't get the box to scale properly with changes in text scaling. It needs fixing for 4.6 in any case as it will be even easier for users to adjust the theme font size and potentially end up with an app description to big for the box.

Anyway on some testing your PR actually works really well with the height set by the theme in em. 
If it's set in px not so good when the end user sets a large font size or text-scaling factor, but that was a problem with the hardcoded size in any case. 

It also works OK if the theme doesn't set a height at all except that if the user does change the font size or the text scaling the menu doesn't adjust until it's reloaded.

As the existing code doesn't respect the height set by the theme at all, if I were you I'd go ahead and reopen this PR. But I'd also consider making PR's for the default Cinnamon theme and the Mint-themes at make their height declarations in em.

fredcw:
@smurphos I've just realised that get_preferred_height() does in fact respect the height set in the .css file. Therefore this fix only makes a difference when no height value is set in the .css file.

fredcw:
@smurphos I did another PR but I have no idea if I'll be passed or not.

fredcw:
@smurphos 
>It also works OK if the theme doesn't set a height at all except that if the user does change the font size or the text scaling the menu doesn't adjust until it's reloaded.

It works for me, you have to open the menu twice.

PR #9333, opened 2020-05-27
menu@cinnamon.org: a fix for jumpiness bug

fredcw:
The current fix ([here](https://github.com/linuxmint/cinnamon/commit/08742aa68838f95c1de419c146f096d4c1a387d3)) to prevent "jumpiness" [#3144](https://github.com/linuxmint/cinnamon/issues/3144) when different scripts are used works by always setting the selectedAppBox height to 30px as get_theme_node().get_length('height') always returns 0 and get_theme_node().get_height() returns -1 (due to the style not yet having been initialised) meaning that selectedAppBox height is always set to 30px regardless of any height set in the .css file. 

This fix sets the height to an appropriate height if no height value exists in the .css file. If a height is set in the .css, get_preferred_height() returns this value.

~~Ideally~~, this fix would be used whenever the style-changed signal is received in case the style is changed after applet initialisation. But this signal is not currently connected in the app so using it each time the menu is opened seems like a good alternative.

The line this.selectedAppBox.set_height(-1); is necessary to unset the previously set_height().

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1233,6 +1233,12 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
 
             this.lastSelectedCategory = null;
 
+            /* This is a workaround to prevent selectedAppBox from changing height when no height value is set
+             * in the .css style and thus causing the menu above to jump up and down. This has no effect when a
+             * height value is set in the .css style as get_preferred_height() returns this value in this case*/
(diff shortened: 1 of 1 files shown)


fredcw:
An alternative fix would be to respond to the style/theme_changed signal (if that's what its called,) which is also emitted when the app is initialised, with something like:

    if (this.selectedAppBox.get_theme_node().get_height() <= 0) { //if no height is set in the .css file
        this.selectedAppBox.style = "height: 3.5em;";
    }

I don't think that even this would work as get_theme_node().get_height() seems to always return 0 the first time it's called and setting the height to an arbitary value (3.5em) may not be best for all themes.

PR #9349, opened 2020-06-05
Remove build dependency on dbus-glib

eli-schwartz:
It's not actually used, but it pulls in a dependency on dbus itself, and the dbus-glib header is used to grab some constants from a dbus provided header. There's no reason to proxy this through dbus-glib.

Also clean up some PKG_CHECK_MODULES I noticed at the same time.

@Fantu relevant to your discussion about using it in cinnamon-settings-daemon, it needs to be cleaned up in a few other places before cinnamon-settings-daemon dropping it will mean much. Also https://github.com/linuxmint/cinnamon-control-center/pull/235

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -75,7 +75,7 @@ STARTUP_NOTIFICATION_MIN_VERSION=0.11
 
 # Collect more than 20 libraries for a prize!
 PKG_CHECK_MODULES(CINNAMON, gio-2.0 >= $GIO_MIN_VERSION
-                               gio-unix-2.0 dbus-glib-1 libxml-2.0
+                               gio-unix-2.0 dbus-1 libxml-2.0
                                gtk+-3.0 >= $GTK_MIN_VERSION
                                libmuffin >= $MUFFIN_MIN_VERSION
                                cjs-1.0 >= $GJS_MIN_VERSION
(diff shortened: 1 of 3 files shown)


eli-schwartz:
I have never in my life cared less about failing CI. This ridiculous service provider won't let me follow the details link to view build logs, until I log in and give them write access to all of *my* personal repos.

Fantu:
thanks, was already done also in debian packages but in slightly different way ( https://salsa.debian.org/cinnamon-team/cinnamon/-/commit/86302fb287d6176c8f479dedc2485e89f7666428 ) but if you already checked that dbus-shared.h it is sufficient instead of including other "extra things" with dbus.h I suppose yours is better

eli-schwartz:
dbus.h is a meta include that provides the entire protocol, but dbus-shared.h is enough of a public header to be the recommended way for alternative dbus implementations to provide these macros.

Given the external dbus-glib.h can get away with using dbus-shared.h, I see no reason not to use it here too -- and fewer headers is faster to lex and compile objects. We just need a few things...

Interesting that no one tried to upstream the debian patch o_0

Fantu:
therefore it is confirmed that your patch is better, I had not thoroughly checked the code and with circleci not working I had no immediate confirmation to be able to say it immediately

about merging debian patches I did/do, even if I don't have much time and for a period I had also stopped after having unfortunately waste a few tens of hours in numerous PRs unreviewed/unmerged for months which were then checked in a massive merge every half year just before the new version, therefore requiring rebases and further tests and if I had no time in those days and I did it later there was a risk that he would stay there for other months
I fear that this unfortunately discourages also others to contribute :(

I hope that in future there will be more review/merging throughout the period without waiting just before a new version is released, so will have less risk of continuous rebase and more time to better test the changes and further reducing bugs/unexpected cases, and there will be multiple contributions from multiple people

cinnamon is a very good DE, the others I tried even if more supported (and some with less bugs/unexpected cases, best performance and support on more newer technologies) have become in part too "minimal" (and optimal I suppose more for tablets/smartphones but not much  for desktop/workstation)

also since I don't like ubuntu much for some recent choices (for example using too much the snaps even though they are still "immature"), I suppose I'll retry Mint for future upgrade of my desktop and my workstation instead ubuntu with cinnamon (like what I did up to ubuntu 18.04)

thanks eli-schwartz for this and other commits that you did for help cinnamon and thanks to clem and any other developers that works on cinnamon

eli-schwartz:
The current CI mysteriously succeeds after I rebased it on master, so I'll take that as a good sign, I guess?

Fantu:
Is not related to this PR code, simply because some deps wasn't with updated CI and in any case CI check was failed, clem updated the latest (like cinnamon-session) only few day ago and rebase triggered a new CI check (now ok with all components with CI conf updated)

PR #9360, opened 2020-06-11
Fix workspace OSD

yparitcher:
the osd does not get destroyed and lingers after switching workspaces, when window effects is disabled.

Regression caused by 041ac072e77431cbe5ea1fb23611d8c4f6fde08c


yparitcher:
Please put out a point release after fixing this, it makes cinnamon unusable

PR #9418, opened 2020-07-01
Grouped Window List applet: Added option to hide the open window indicator on unfocused windows

AleFelix:
I added the option to hide the open window indicator. This can be useful when button labels are enabled and indicating which applications are open ends up being redundant.

Example:

- Normal behavior: https://gfycat.com/ifr/foolhardydigitalclumber?hd=1

- Indicator hidden: https://gfycat.com/ifr/unfithelpfulflies?hd=1

### DIFF
--- a/cinnamon.pot
+++ b/cinnamon.pot
@@ -6644,6 +6644,11 @@ msgstr ""
 msgid "Enable app button dragging"
 msgstr ""
 
+#. grouped-window-list@cinnamon.org->settings-schema.json->hide-indicator-on-unfocused-
+#. windows->description
+msgid "Hide the open window indicator on unfocused windows"
+msgstr ""
+
 #. grouped-window-list@cinnamon.org->settings-schema.json->launcher-animation-
 #. effect->description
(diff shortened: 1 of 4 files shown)


smurphos:
Mmm

`let windowClass = this.state.settings.hideIndicatorUnfocused ? 'window-list-item-box' : 'grouped-window-list-item-box';`

This does weird things with some third party themes that use the same styling for the window-list-item-box as they do for an active grouped-window-list-item-box. There are quite a few like this.

AleFelix:
Ohh sorry I didn't know that, I tested it with a couple of themes but it wasn't a very exhaustive sample. I personally didn't have problems with those two themes (however, I didn't try every color variation of Mint-X haha), but If you think that reusing that class could be problematic feel free to reject the PR.

(I actually tried another approach first, but it was not pretty. Adding the top/bottom/left/right class in the onEnter function and removing it when the focus was lost. It worked, but the css on my theme caused the focused items to have a sightly different position than the other buttons).

PR #9430, opened 2020-07-04
show launcher file from menu

beauxq:
Right click on an application in the menu and choose to show the launcher desktop file in the default file manager.

This is something I want and I think other people would like it too.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -339,8 +339,7 @@ class ApplicationContextMenuItem extends PopupMenu.PopupBaseMenuItem {
                         // so each one would have to be handled individually
                         //     else if (it's nautilus) { do this; }
                         //     else if (it's konqueror) { do that; }
-                        let last_slash = this._appButton.app.get_app_info().get_filename().lastIndexOf("/");
-                        let path = this._appButton.app.get_app_info().get_filename().substring(0, last_slash + 1)
+                        let path = GLib.path_get_dirname(this._appButton.app.get_app_info().get_filename());
                         fileapp.launch_uris([path], null);
                     }
                 }
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -331,7 +331,7 @@ class ApplicationContextMenuItem extends PopupMenu.PopupBaseMenuItem {
                 let fileapp = Gio.app_info_get_default_for_type('inode/directory', true);
                 if (fileapp) {
                     if (fileapp.get_id() == 'nemo.desktop') {
-                        Util.spawnCommandLine('nemo --no-desktop "' + this._appButton.app.get_app_info().get_filename() + '"');
+                        Util.spawnCommandLine('nemo "' + this._appButton.app.get_app_info().get_filename() + '"');
                     }
                     else {
                         // some file managers are capable of doing this and some are not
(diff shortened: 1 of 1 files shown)


beauxq:
show-launcher-file-from-menu

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -327,6 +327,27 @@ class ApplicationContextMenuItem extends PopupMenu.PopupBaseMenuItem {
...
+                        Util.spawnCommandLine('nemo --no-desktop "' + this._appButton.app.get_app_info().get_filename() + '"');
+                    }
+                    else {
+                        // some file managers are capable of doing this and some are not
+                        // and they could each do it differently
+                        // so each one would have to be handled individually
+                        //     else if (it's nautilus) { do this; }
+                        //     else if (it's konqueror) { do that; }
+                        let last_slash = this._appButton.app.get_app_info().get_filename().lastIndexOf("/");
+                        let path = this._appButton.app.get_app_info().get_filename().substring(0, last_slash + 1)
```
Odyseus, line 343:
```suggestion
                        let path = GLib.path_get_dirname(this._appButton.app.get_app_info().get_filename());
```

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -327,6 +327,26 @@ class ApplicationContextMenuItem extends PopupMenu.PopupBaseMenuItem {
 
     activate (event) {
         switch (this._action) {
+            case "show_launcher_file":
+                let fileapp = Gio.app_info_get_default_for_type('inode/directory', true);
+                if (fileapp) {
+                    if (fileapp.get_id() == 'nemo.desktop') {
+                        Util.spawnCommandLine('nemo --no-desktop "' + this._appButton.app.get_app_info().get_filename() + '"');
```
Odyseus, line 334: I forgot to mention that the `--no-desktop` CLI option might not be needed. Nemo (the process) no longer handles the desktop (the `nemo-desktop` process does it now). The `--no-desktop` CLI option isn't used by the `nemo` process and it doesn't trigger a parse error because the option was added as a *"dummy"* option apparently for retro-compatibility.

beauxq: Thank you for the info.

Is there any chance that a newer version of cinnamon could be running with an older version of nemo that needs the option?

Odyseus: > Is there any chance that a newer version of cinnamon could be running with an older version of nemo that needs the option?

I hardly doubt it. The `--no-desktop` CLI option was removed around Nemo 3.4.x (about 6 versions/3 years ago). I don't think that anyone using Cinnamon 4.6.x with a Nemo version older than 3.4.x will have a usable system nor if they could even install those versions together at all.

mtwebster:
I'm really not sure we need this here - as of 4.6 this submenu is getting a bit full:
Screenshot from 2020-09-03 11-08-00

And I don't think it would be useful except for a very small group of users. Maybe in the menu editor instead?

smurphos:
I think a feature along these lines would be good to have, but rather than opening the file manager it should access an editable launcher properties window similarly to right click > properties of a desktop launcher, or the edit UI of the panel-launcher applet. 

This would improve feature parity with mintmenu on MATE and the whisker menu on XFCE which both can do this via the menu right click context.

beauxq:
What is important to me is just getting the information of the location of the file.

Not all of these launcher files are in the same place. It's annoying to have to go looking in a few different places to find the file.

A properties window won't help unless that properties window tells me the location of the file.

I wouldn't mind it in the menu editor instead of this submenu - even in the properties window that can be accessed from the menu editor.

collinss:
I would think this functionality would be better in the menu editor, rather than the menu itself. It's pretty niche, and the average user isn't going to use it.

mtwebster:
I just finished some rework on the menu editor layout, and added a button for this.

https://github.com/linuxmint/cinnamon/commit/8465cc599601e8150aec6313368e7dd216d1905a

Issue #237, opened 2012-01-25
found an error in js/makefile.am


zenrox:
Running automake-1.11...
js/Makefile.am:22: whitespace following trailing backslash
js/Makefile.am:23: whitespace following trailing backslash
js/Makefile.am:28: whitespace following trailing backslash
js/Makefile.am:30: whitespace following trailing backslash
js/Makefile.am:54: whitespace following trailing backslash
js/Makefile.am:59: whitespace following trailing backslash

still compiles fine 
was useing sh autogen.sh
then make and make install

PR #9435, opened 2020-07-07
add suspend-then-hibernate (issue cinnamon-settings-daemon:#237)

miniskipper:
This consists of 3 PRs (involved repositories: cinnamon, cinnamon-session, cinnamon-settings-daemon)

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py
@@ -142,7 +142,7 @@ class Module:
 
         section = power_page.add_section(_("Power Options"))
 
-        lid_options, button_power_options, critical_options, can_suspend, can_hybrid_sleep = get_available_options(self.up_client)
+        lid_options, button_power_options, critical_options, can_suspend, can_hybrid_sleep, can_hibernate = get_available_options(self.up_client)
 
         size_group = Gtk.SizeGroup(mode=Gtk.SizeGroupMode.HORIZONTAL)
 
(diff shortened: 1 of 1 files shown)


clefebvre:


```diff

Complexity increasing per file
==============================
- files/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=5815536&bid=19038894)

miniskipper:
The CI failed, 404 Error at build.

mtwebster:
Hi, this looks good for the most part, with a couple of issues:

Suspending from the menu (shutdown->suspend for instance) doesn't respect the setting - this calls cinnamon-session over dbus. It has its own set of calls for interacting with systemd that would need similar treatment to what you did in c-s-d - see [here](https://github.com/linuxmint/cinnamon-session/blob/3e64d5d505281083dc7c26381b7bc885587fe7c2/cinnamon-session/csm-manager.c#L1280).

The delay time is not displayed or configurable - from what I can figure out, it's two hours by systemd documentation or three hours according to my /etc/systemd/sleep.conf.  It seems to ignore a sleep.conf in ~/.config/systemd - if we could modify it without root it could be made configurable in the UI.  Maybe this isn't that important.

If you can finish implementing this in cinnamon-session I'll merge it, and maybe something can be done for the delay later on.

Thanks!

miniskipper:
Well, that slipped through my fingers (because I never use the power-options from the menu ^_^  )

miniskipper:
Changes for cinnamon-session done.

Still searching a way for the user to change the delay setting.
https://www.freedesktop.org/software/systemd/man/systemd-sleep.conf.html does not mention user-config dirs but there is this: systemd/systemd#16052

but tbh: 2 hours seems to be a sensible default, probably good in most cases.

PR #9486, opened 2020-07-25
Menu applet: Match text within a word

### LINKED ISSUE #9444: [regression] Menu search no longer matches within words
**Issue**

After upgrading to Debian Bullseye, the `menu@cinnamon.org` applet no longer matches text within a word, you now have to type the beginning of a word. This makes it harder to find nicely typable matches for software you want to launch frequently.

**Steps to reproduce**

1. Search for, for example, `idu` (for me that lies very nicely under my hands' natural position)
2. Observe that there are no search results

**Expected behaviour**

It should match the screenshotting tool because its description contains the word "individual". This is only entry that contains that word, so another app can't sort above it and be launched by accident.

**Other information**

The regression was introduced in this commit: https://github.com/linuxmint/cinnamon/commit/d40ba4f900da90f09ecda78f3b139f7bef0fd4cf

ItzSwirlz:
@lucb1e patches this by having the menu check the specific index of the pattern.

Regression from commit: d40ba4f

Closes #9444

You can thank Atom for (cough) fixing whitespaces.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1242,7 +1242,7 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
              * height value is set in the .css style as get_preferred_height() returns this value in this case*/
             this.selectedAppBox.set_height(-1); //unset previously set height
             this.selectedAppBox.set_height(this.selectedAppBox.get_preferred_height(-1)[1]);
-            
+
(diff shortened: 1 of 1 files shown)


mtwebster:
We just merged some changes that allow re-ordering of items in search results, so that (hopefully) the best matches will naturally be at the top. Unfortunately the section of code in this PR is eliminated, and hopefully no longer necessary.

See https://github.com/linuxmint/cinnamon/commit/a6738bd306180b4e2e668c90c1d7d5d6c0578906

The new method may need tweaking as well, but at least we'll be at a clean starting point.
Thanks

PR #9501, opened 2020-07-31
Bundle libcroco

leigh123linux:
gnome-shell bundled it.

https://github.com/GNOME/gnome-shell/commit/47758d16ff0acd9edbd8ff6eb3922e59fb14f8bb

Now fedora gnome developers want to retire it.

https://pagure.io/releng/issue/9641

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -122,7 +122,7 @@ AC_CHECK_FUNCS(XFixesCreatePointerBarrier)
 CFLAGS=$saved_CFLAGS
 LIBS=$saved_LIBS
 
-PKG_CHECK_MODULES(ST, muffin-cogl-path-0 muffin-clutter-0 gtk+-3.0 libcroco-0.6 >= 0.6.2 cinnamon-desktop >= 2.4.0 x11)
+PKG_CHECK_MODULES(ST, muffin-cogl-path-0 muffin-clutter-0 gtk+-3.0 cinnamon-desktop >= 2.4.0 x11)
 PKG_CHECK_MODULES(GDMUSER, dbus-glib-1 gtk+-3.0)
 PKG_CHECK_MODULES(TRAY, muffin-clutter-0 gtk+-3.0)
 PKG_CHECK_MODULES(DESKTOP_SCHEMAS, cinnamon-desktop >= 2.4.0)
--- a/debian/control
+++ b/debian/control
@@ -17,7 +17,6 @@ Build-Depends:
  libcinnamon-desktop-dev (>= 3.8),
  libcinnamon-menu-3-dev,
  libcjs-dev (>= 3.8),
- libcroco3-dev (>= 0.6.2),
  libdbus-glib-1-dev,
  libgirepository1.0-dev (>= 1.29.15),
  libgl1-mesa-dev,
--- a/src/Makefile-st.am
+++ b/src/Makefile-st.am
@@ -96,11 +96,65 @@ BUILT_SOURCES += st.h
 CLEANFILES += stamp-st.h
 
 st_source_private_h =				\
+	st/croco/cr-additional-sel.h			\
+	st/croco/cr-attr-sel.h			\
+	st/croco/cr-cascade.h			\
+	st/croco/cr-declaration.h			\
+	st/croco/cr-doc-handler.h			\
+	st/croco/cr-enc-handler.h			\
+	st/croco/cr-fonts.h			\
+	st/croco/cr-input.h			\
+	st/croco/cr-num.h			\
+	st/croco/cr-om-parser.h			\
+	st/croco/cr-parser.h			\
+	st/croco/cr-parsing-location.h			\
+	st/croco/cr-prop-list.h			\
+	st/croco/cr-pseudo.h			\
+	st/croco/cr-rgb.h			\
+	st/croco/cr-selector.h			\
+	st/croco/cr-simple-sel.h			\
+	st/croco/cr-statement.h			\
+	st/croco/cr-string.h			\
+	st/croco/cr-stylesheet.h			\
+	st/croco/cr-term.h			\
+	st/croco/cr-tknzr.h			\
+	st/croco/cr-token.h			\
+	st/croco/cr-utils.h			\
+	st/croco/libcroco-config.h			\
+	st/croco/libcroco.h			\
 	st/st-private.h				\
 	st/st-table-private.h			\
 	st/st-theme-private.h			\
 	st/st-theme-node-private.h		\
 	st/st-theme-node-transition.h
+
+# please, keep this sorted alphabetically
+croco_source_c =					\
+	st/croco/cr-additional-sel.c			\
+	st/croco/cr-attr-sel.c			\
+	st/croco/cr-cascade.c			\
+	st/croco/cr-declaration.c			\
+	st/croco/cr-doc-handler.c			\
+	st/croco/cr-enc-handler.c			\
+	st/croco/cr-fonts.c			\
+	st/croco/cr-input.c			\
+	st/croco/cr-num.c			\
+	st/croco/cr-om-parser.c			\
+	st/croco/cr-parser.c			\
+	st/croco/cr-parsing-location.c			\
+	st/croco/cr-prop-list.c			\
+	st/croco/cr-pseudo.c			\
+	st/croco/cr-rgb.c			\
+	st/croco/cr-selector.c			\
+	st/croco/cr-simple-sel.c			\
+	st/croco/cr-statement.c			\
+	st/croco/cr-string.c			\
+	st/croco/cr-stylesheet.c			\
+	st/croco/cr-term.c			\
+	st/croco/cr-tknzr.c			\
+	st/croco/cr-token.c			\
+	st/croco/cr-utils.c				\
+	$(NULL)
 
 # please, keep this sorted alphabetically
 st_source_c =					\
@@ -148,6 +202,7 @@ noinst_LTLIBRARIES += libst-1.0.la
 
 libst_1_0_la_LIBADD = -lm $(ST_LIBS)
 libst_1_0_la_SOURCES =				\
+	$(croco_source_c)				\
 	$(st_source_c)					\
 	$(st_non_gir_sources)			\
 	$(st_source_private_h)			\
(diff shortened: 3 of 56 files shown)


clefebvre:


```diff

Issues
======
- Added 38
           

Clones added
============
- src/st/croco/cr-rgb.c  1
- src/st/croco/cr-additional-sel.c  2
- src/st/croco/cr-om-parser.c  14
- src/st/croco/cr-utils.c  6
- src/st/croco/cr-selector.c  1
- src/st/croco/cr-tknzr.c  27
- src/st/croco/cr-declaration.c  3
- src/st/croco/cr-parser.c  42
- src/st/croco/cr-statement.c  5
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=5945270&bid=19490196)

leigh123linux:
Added follow up commits

https://github.com/GNOME/gnome-shell/commits/master/src/st/croco

clefebvre:


```diff

Issues
======
- Added 36
           

Clones added
============
- src/st/croco/cr-parser.c  42
- src/st/croco/cr-selector.c  1
- src/st/croco/cr-rgb.c  1
- src/st/croco/cr-statement.c  5
- src/st/croco/cr-declaration.c  3
- src/st/croco/cr-additional-sel.c  2
- src/st/croco/cr-tknzr.c  27
- src/st/croco/cr-utils.c  6
- src/st/croco/cr-om-parser.c  14
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=5945270&bid=19490196)

federicomenaquintero:
By the way, I cherry-picked the fix to libcroco's latest CVE and put it here: https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1404 - maybe you'll find it useful :)

clefebvre:


```diff

Issues
======
- Added 36
           

Clones added
============
- src/st/croco/cr-tknzr.c  27
- src/st/croco/cr-additional-sel.c  2
- src/st/croco/cr-rgb.c  1
- src/st/croco/cr-utils.c  6
- src/st/croco/cr-statement.c  5
- src/st/croco/cr-declaration.c  3
- src/st/croco/cr-parser.c  39
- src/st/croco/cr-om-parser.c  14
- src/st/croco/cr-selector.c  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=5945270&bid=19490196)

leigh123linux:
> By the way, I cherry-picked the fix to libcroco's latest CVE and put it here: https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1404 - maybe you'll find it useful :)

Thank you.

clefebvre:


```diff

Issues
======
- Added 36
           

Clones added
============
- src/st/croco/cr-statement.c  5
- src/st/croco/cr-selector.c  1
- src/st/croco/cr-utils.c  6
- src/st/croco/cr-additional-sel.c  2
- src/st/croco/cr-om-parser.c  14
- src/st/croco/cr-tknzr.c  27
- src/st/croco/cr-declaration.c  3
- src/st/croco/cr-rgb.c  1
- src/st/croco/cr-parser.c  39
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=5945270&bid=19490196)

clefebvre:


```diff

Issues
======
- Added 36
           

Clones added
============
- src/st/croco/cr-rgb.c  1
- src/st/croco/cr-selector.c  1
- src/st/croco/cr-declaration.c  3
- src/st/croco/cr-om-parser.c  14
- src/st/croco/cr-utils.c  6
- src/st/croco/cr-statement.c  5
- src/st/croco/cr-tknzr.c  27
- src/st/croco/cr-additional-sel.c  2
- src/st/croco/cr-parser.c  39
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=5945270&bid=19490196)

eli-schwartz:
I do not comprehend how bundling this library in X applications is intended to solve the problem of it being unmaintained and containing CVEs. Now we still have the exact same code even in gnome-shell, but with the additional burden that the same patches need to be applied to gnome-shell, gettext, and cinnamon. How does this save work? How does this make libcroco-bundled-code-copy-number-3 more secure than libcroco-system-copy?

"Hiding" the library by slipping the files into gnome-shell's source tree didn't make gnome-shell more secure, and it won't make cinnamon more secure either. Creating an alternative for gnome-shell **and migrating cinnamon to it** would make cinnamon more secure.

In the meantime, it would be lovely if one copy of libcroco could be patched once and fix bugs in three programs. I acknowledge this would depend on gnome-shell actually caring about sharing the effort of providing maintenance patches for their vendored library, but maybe we could at least share those patches between *the other two* programs?

leigh123linux:
@eli-schwartz libcroco has been removed from F33 

https://src.fedoraproject.org/rpms/libcroco/c/9252608f0ca98b5061eec8c52c91509072eef697?branch=master

eli-schwartz:
That's factually incorrect, f33 ships `bundled(libcroco) = 0.6.12` in gettext.spec in addition to undeclared copies in cinnamon.spec and gnome-shell.spec, for a grand total of at least 3 copies of libcroco in f33. Just because there isn't an rpm named "libcroco" doesn't mean it isn't there, and the same applies to Linux Mint, Arch, Debian, etc. except with less explicit tagging of bundled libs.

I don't see how a gnome maintainer doing what's best for gnome-shell using a fedora commit bit in strictly unambiguous violation of https://fedoraproject.org/wiki/Bundled_Libraries#Requirement_if_you_bundle, is a good reason to make the cross-distro ecosystem worse by not consolidating CVE fixes for deprecated libraries in one package for that deprecated library.

On the other hand, I'm certainly not surprised.

federicomenaquintero:
> I do not comprehend how bundling this library in X applications is intended to solve the problem of it being unmaintained and containing CVEs. 

The problem with having an unmaintained library as a standalone package is that people actively look for vulnerabilities in it, and create CVEs for it, even if the library is not used in critical internet-facing components.  Before being bundled in gnome-shell, libcroco-standalone was used just by two packages: librsvg and gnome-shell itself.

For librsvg, we've replaced its use of libcroco by a Rust-based CSS parser.  When that was finished, I bundled libcroco into gnome-shell and [started replacing it by a Rust library](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/1934) ([merge request in progress](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1355)).  This is not a drop-in replacement; it's new code with a lot of CSS machinery borrowed from librsvg.  In the end I intend to provide a crate that can serve both projects.

As someone working on a distro (I work for Suse), bundling libcroco into gnome-shell means that I don't have to spend a few days in emergency mode every few months when someone files a CVE for libcroco.  Sure, the bundled copy still has bugs, but if it is not directly usable as a standalone library, it is no longer an automatic target for exploits.  Gnome-shell relies on trusted extensions and their corresponding CSS; it is not intended to load any possibly malicious CSS chunk from the internet.

In gitlab.gnome.org, we just moved libcroco to the "Archived" section - that's for unmaintained, obsolete, do-not-use modules.  Having it in the main set of modules means that people perceive it as code for general consumption, which it definitely isn't.

(I have no idea what gettext does with libcroco.  Inkscape also embeds a copy, by the way, to process CSS in SVG files.  Per codesearch.debian.net, those are the only two remaining packages that bundle libcroco.)

What would Cinnamon like to do about this change in gnome-shell?  I hope that my Rust stuff will be merged soon; I'm happy to assist Cinnamon in doing the same if you want to go that route.  I think you've modified st-theme-node.c on your own?  I'm open to ideas on things to change there, so they can be folded into the Rust code.

eli-schwartz:
>  Sure, the bundled copy still has bugs, but if it is not directly usable as a standalone library, it is no longer an automatic target for exploits. Gnome-shell relies on trusted extensions and their corresponding CSS; it is not intended to load any possibly malicious CSS chunk from the internet.

The problem is this smacks of "we are hiding it inside gnome-shell so CVE authors don't notice it and bother us". But I don't understand why you couldn't instead do....

"libcroco is deprecated and no longer actively maintained. It is known to be buggy, exploitable, and riddled with undiscovered CVEs and should not be used to parse untrusted input. Do not use it in new projects, try to port old projects over to something newer and more robust (we suggest XXXX), and do not expect issues without accompanying patches to be a high priority."

It's functionally the exact same thing as archiving it, except the remaining programs which use libcroco since the days of yore and probably don't parse untrusted input can share one implementation.
(Is it not possible to move it to the archive but still provide life support patches and use it from there in gnome-shell?)

I reiterate, I don't get the point of hiding it away via bundled code copies. It's not improving the code, only hiding it from "those annoying people who look for CVEs we don't want to hear about".

...

Getting cinnamon ported to use gnome-shell's replacement sounds interesting, though I am curious -- is this going to be a replacement library, or something that adds a direct build-dep on rust to build gnome-shell/cinnamon itself?

debarshiray:
I work on Fedora and some of its downstreams.

> The problem is this smacks of "we are hiding it inside gnome-shell
> so CVE authors don't notice it and bother us". But I don't understand
> why you couldn't instead do....

Yes, it *hides* libcroco inside GNOME Shell just like any of the other dozens of private interfaces in any codebase that can't be accessed by random other processes outside that codebase.

This means that libcroco-inside-gnome-shell is only used the way GNOME Shell uses it, which doesn't involve parsing untrusted CSS off the Internet, and since it's a private interface nobody else can access it either.

I don't think *hiding* is necessarily a bad thing in this case.

> "libcroco is deprecated and no longer actively maintained. It is known
> to be buggy, exploitable, and riddled with undiscovered CVEs and should
> not be used to parse untrusted input. Do not use it in new projects, try to
> port old projects over to something newer and more robust (we suggest
> XXXX), and do not expect issues without accompanying patches to be a
> high priority."
>
> It's functionally the exact same thing as archiving it, except the remaining
> programs which use libcroco since the days of yore and probably don't parse
> untrusted input can share one implementation.

Saying that isn't as effective as actually preventing random processes from actually using the librari. Removing the standalone library from a OS distribution is a more effective way of communicating that the OS vendor isn't responsible for it, beyond a few selective cases where the library is bundled as part of some package.

Third-party packages can surely bundle it, but then the responsibility for it squarely lies on the third-party.

Also, having a bunch of unfixed CVEs open against an OS distribution is not without its downsides. Just dealing with the negative publicity generated by CVE authors and having to explain the numbers is a drain in itself.

PR #9514, opened 2020-08-11
[sound@cinnamon.org] Update applet.js - Further fixes #6243

### LINKED ISSUE #6243: Sound applet should not show Launch player when Control Players off
When you turn off "Control Players" in the Sound applet config, the Sound applet should no longer show "Launch player > "  but it does.

(Or, if you disagree with that, then I am suggesting a new option to hide "Launch player >" in the Sound applet.)

Ammako:
When the Control Players setting is off, the Launch player menu item disappears, as it should. The problem is that the Launch player menu item will reappear after cinnamon is reloaded, either manually or from a restart/at next boot. This fixes it so that the Launch player menu item correctly stays hidden.


dustinhylton360:
Merge

Issue #9327, opened 2020-05-21
Volume adjustments from hotkeys does not display volume percentage


tdeeb:
```
 * Cinnamon version 4.4.8
 * Distribution - Linux Mint 19.3 Tricia
 * Nvidia GeForce RTX 2060
  - Driver: nvidia-driver-435.21-0ubuntu0.18.04.2
 * 64 bit
 ```

**Issue**
Increasing or lowering the sound volume with Volume up or Volume down hotkeys does not display the exact volume percentage. As shown in the screenshot below, there is a bar representing the volume, but the exact percentage is not shown and thus ambiguous.

![VolumeDisplay](https://user-images.githubusercontent.com/1388037/82615560-401baf80-9bba-11ea-8147-8c3cbf1f14ab.png)

**Steps to reproduce**
Press the Volume up or Volume down hotkey to increase/lower the sound volume and observe that the volume percentage is not shown.

**Expected behaviour**
The value of the current volume is shown in percent. For example, if the volume is at 25% and it's raised, it should show that the volume is now at 30%.

**Other information**
N/A

**Configuration info, if applicable**
N/A

smurphos:
Hi, if you want to modify your install yourself to add this feature - https://forums.linuxmint.com/viewtopic.php?p=1762715#p1762715

tdeeb:
That worked like a charm; thank you!

acmpo6ou:
This is a possible solution to https://github.com/linuxmint/cinnamon/issues/12535

PR #9532, opened 2020-08-19
Add Percentage Label to OSD for Volume Control

MysticRyuujin:
Code taken from https://forums.linuxmint.com/viewtopic.php?t=312162

Related to issue #9327

I'm guessing there's a reason this change hasn't been proposed on GitHub yet, but I did a quick search and I didn't see anyone attempt it so :sweat_smile:

It's really @smurphos  code

### DIFF
--- a/js/ui/osdWindow.js
+++ b/js/ui/osdWindow.js
@@ -110,6 +110,10 @@ OsdWindow.prototype = {
 
         this._level = new LevelBar();
         this.actor.add(this._level.actor);
+        
+        this._label = new St.Label();
+        this._label.style = 'font-size: 1.2em; text-align: center;'
+        this.actor.add(this._label);
 
         this._hideTimeoutId = 0;
         this._reset();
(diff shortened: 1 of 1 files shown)


clefebvre:
We talked about this before...

.. my TV does that, it shows me a number.. so when it's not loud enough, say I'm at 20....

... I increase the volume until it's perfect (i.e. loud enough but not too much)....

... and then because the number is not round, I decrease it a bit or increase it a bit to fall on something round... 

so if I'm at 24 or 26 I put it at 25... if I'm at 29, I put it at 30...

would you say that number is helping me set my volume? or it's getting in the way?

clefebvre:
I'll argue the opposite also. Say I want to use reduced volume to not annoy people in other rooms, it's not easy for me to assess what volume that might be without going to the other room, but if I did that in the past, I can easily remember a number and know that setting to that number means it won't annoy them. So in this case the value is very helpful.

So with that in mind, I'm OK with adding this.

clefebvre:
From a technical point of view, you can't show "N/A" like that. If the volume is undefined just show nothing (i.e. "").

MysticRyuujin:
It's slightly beyond my skillset but I could see there being a toggle setting in one of the system settings areas for audio/volume control to make it optional for the user.

I like having it there, I suppose others might OCD about it

Furvy:
Is there any chance of this getting better implemented? I opened a bug report on this which does not seem to be acknowledged.

https://github.com/linuxmint/cinnamon/issues/9758

I understand that I should probably try and code this better myself, but I was hoping experts would take a look at it rather than a complete novice like me with no commits attempting such a feat.

PR #9550, opened 2020-08-30
Use all monitor(s) geometry to determine hot corner positions (linuxmint#7726)

### LINKED ISSUE #7726: On dual screen move hot corners to the extended screen edges
```
~:: cinnamon --version
Cinnamon 3.8.6
~:: uname -a
Linux carlos 4.14.53-1-MANJARO #1 SMP PREEMPT Tue Jul 3 17:59:17 UTC 2018 x86_64 GNU/Linux
~:: lspci | grep VGA
00:02.0 VGA compatible controller: Intel Corporation 2nd Generation Core Processor Family Integrated Graphics Controller (rev 09)
~:: pacman -Qs video intel
local/xf86-video-intel 1:2.99.917+829+gd7dfab62-1 (xorg-drivers)
    X.org Intel i810/i830/i915/945G/G965+ video drivers
 ```

I have an external monitor on top of my laptop screen. The hot corners in my desktop are in the left and right top. But it's very hard to put the mouse pointer over them since the pointer passes through to the top screen. Now, since both screens form an extended screen which top edges are in the top monitor, I think the hot corners should be there. The idea is that a hot corner is easily reachable by just "throwing" the pointer over it, being between two screens doesn't help reaching the corner. Also, if I configured my top  monitor as the primary display (which I don't want to) the same would apply for the bottom corners.

codewithmichael:
Fixes #7726

**Description:**

Avoids hard-to-trigger "corner" points between monitors on dual/multi-monitor setups by moving hot corner positions to the outer edge.

**Change Summary:**

- Updated the `HotCornerManager.updatePosition()` to (optionally) accept multiple monitors for use in determining hot corner positions.
- Updated `LayoutManager._updateBoxes()` to pass all connected monitors instead of just `primaryMonitor`.

### DIFF
--- a/js/ui/hotCorner.js
+++ b/js/ui/hotCorner.js
@@ -55,11 +55,14 @@ HotCornerManager.prototype = {
         return true;
     },
 
-    updatePosition: function(monitor) {
-        let left   = monitor.x;
-        let right  = monitor.x + monitor.width - 2;
-        let top    = monitor.y;
-        let bottom = monitor.y + monitor.height - 2;
+    updatePosition: function(...monitors) {
+        let left=0, right=0, top=0, bottom=0;
+        for (var monitor of monitors) {
+            left   = Math.min(left,   monitor.x);
+            right  = Math.max(right,  monitor.x + monitor.width - 2);
+            top    = Math.min(top,    monitor.y);
+            bottom = Math.max(bottom, monitor.y + monitor.height - 2);
+        }
 
         // Top Left: 0
         this.corners[0].actor.set_position(left, top);
--- a/js/ui/layout.js
+++ b/js/ui/layout.js
@@ -145,7 +145,7 @@ LayoutManager.prototype = {
 
     _updateBoxes: function() {
         if (this.hotCornerManager)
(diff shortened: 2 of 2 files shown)


mtwebster:
Hi, there a few things here:

- This would be a regression for people that might being using hot corners on shared edges of a multi-monitor setup.
- There's an option for only showing a workspace overview (expo) that only displays the primary monitor - using a hot corner to activate this could be confusing with this option activated.

 It's probably a fairly common case also, since the issue goes away with adjacent panels on adjacent monitors - barriers are added by default (panel settings->allow pointer to pass thru edges of panels).

These barriers are probably the best solution for the original problem - if an 'inner' corner is active, place a barrier there to keep the pointer from overshooting.

Maybe in a later patch multi-monitor support could be added (adding corners to more than the primary monitor).

The pointer barriers are not terribly complicated - it's implemented in cinnamon's [panel.js](https://github.com/linuxmint/cinnamon/blob/master/js/ui/panel.js#L2340) file if you're interested in taking a look - if you'd like update this pr (or open a new one later).

Thanks

PR #9559, opened 2020-09-04
[FEAUTURE] [BETTER THEMING] add applet uuid formatted as style class name

bomdia:
Added uuid formatted ex from menu@cinnamon.org to 
menu-cinnamon-org-applet, this allow for a better css property apply, so 
theme more better tuned is possible

### DIFF
--- a/js/ui/applet.js
+++ b/js/ui/applet.js
@@ -200,7 +200,9 @@ var Applet = class Applet {
             setTimeout(() => this._getPanelInfo(), 0);
         }
     }
-
+    _setStyleClass(className){
+        this.actor.style_class = className;
+    }
     _getPanelInfo(instance_id) {
         if (!instance_id) instance_id = this.instance_id;
         let appletDefinition = AppletManager.getAppletDefinition({applet_id: instance_id});
--- a/js/ui/appletManager.js
+++ b/js/ui/appletManager.js
@@ -585,7 +585,7 @@ function createApplet(extension, appletDefinition, panel = null) {
         let metadata = extension.meta;
         let appletClass = '';
         if(typeof metadata === 'object' && metadata !== null && typeof metadata.uuid === 'string'){
-            appletClass = metadata.uuid.toLowerCase().replace(/([^a-z0-9]+)/gi, '-')+'-applet';
+            appletClass = ' ' + metadata.uuid.toLowerCase().replace(/([^a-z0-9]+)/gi, '-') + '-applet';
         }
         applet._setStyleClass('applet-box'+appletClass);
 
--- a/js/ui/appletManager.js
+++ b/js/ui/appletManager.js
@@ -580,6 +580,15 @@ function createApplet(extension, appletDefinition, panel = null) {
         // so we don't need to pass it to the constructor anymore, but would
         // require a compatibility clean-up effort.
         applet = module.main(extension.meta, orientation, panel.height, applet_id);
+
+        //set applet style class
+        let metadata = extension.meta;
+        let appletClass = '';
+        if(typeof metadata === 'object' && metadata !== null && typeof metadata.uuid === 'string'){
+            appletClass = metadata.uuid.toLowerCase().replace(/([^a-z0-9]+)/gi, '-')+'-applet';
+        }
+        applet._setStyleClass('applet-box'+appletClass);
+
     } catch (e) {
         Extension.logError(`Failed to evaluate 'main' function on applet: ${uuid}/${applet_id}`, uuid, e);
         return null;


brownsr:
I couldn't see where you had pulled the code from.  Regardless, we'd want to see that this does not cause themeing regressions in applets, so could you please include some examples of where you have successfully regression tested this against existing applets and themes.  Also a specific example of how someone might apply this change to a theme's CSS would be useful,  so documentation could be updated. A screenshot or two of the intended result would also be useful.

Regards

Simon

bomdia:
"I couldn't see where you had pulled the code from." what do you mean @brownsr ?

i'm running this right now and i don't see any regression,
howeaver i think that can't be any regression as i only added a className to the existing one to better identify any applet in css.

you could use this class for appling specific css code to an applet in your theme ex
you have an hover border image applied for all applet-box but in a particular applet like show desktop you want to use another image , some **scss** of this:

```
@each $pos in top, bottom { 
    .panel-#{$pos} .applet-box {
        text-align: center;
        &:hover {
            border-image: url("img/panel/panel-#{$pos}__inactive__open.svg") 4;
        }

         &:checked {
            color: $text-active;
        }
    }
}
.applet-box.show-desktop-cinnamon-org-applet, .applet-box.sw-mohammad-sn-applet {
    &:hover {
        border-image: url("img/panel/mini-semi-transparent-hover.svg") 2;
    }
    .applet-icon {
        width: 0;        
    }
    border-image: url("img/panel/mini-semi-transparent-bg.svg") 2;
    width: 1em;
}
```
mod

as you see the first is the common one for all and the second is the specific one for show-desktop

using a specific class name help theme developer to select the right applet that they want to apply a css rule

and because we talking about this think i'll ask for a better support of css rule from gtk object (ex StIcon) like the display one or the opacity

js/ui/appletManager.js
```diff
@@ -580,6 +580,15 @@ function createApplet(extension, appletDefinition, panel = null) {
...
         // require a compatibility clean-up effort.
         applet = module.main(extension.meta, orientation, panel.height, applet_id);
+
+        //set applet style class
+        let metadata = extension.meta;
+        let appletClass = '';
+        if(typeof metadata === 'object' && metadata !== null && typeof metadata.uuid === 'string'){
+            appletClass = ' ' + metadata.uuid.toLowerCase().replace(/([^a-z0-9]+)/gi, '-') + '-applet';
+        }
+        applet._setStyleClass('applet-box'+appletClass);
```
Cobinja, line 590: What exactly is the purpose of prepending 'applet-box' **and** appending 'applet'? I'd say one of them is sufficient.

bomdia: applet-box is the default class name and because i use a set method its overwrited so this, is in the _init of Applet class 
when creating the actor is setted to applet-box and is a common class in a lot of theme so i don't think is possible remove it, for the appending applet is done only if we have an uuid and is appendend to the parsed uuid (note the space  before it)

bomdia: so the custom class name ends up in something like menu-cinnamon-org-applet is more cosmetic but i thinks is more intuitive

Cobinja (commented):
Ah, sry, I didn't see the space. Guess I need new glasses 🙂

bomdia:
there is any news on this?

brownsr:
This definitely causes regressions for me.  I've not tried to debug the code, but it breaks the themeing in the vertical panels that I use.  To aid your fault-finding, that's using spices theme graphite-zero, if that's at all relevant.  At first glance it seems like all the themeing that is in applet-box.vertical gets missed out with this PR in place.

Regards

Simon

js/ui/applet.js
```diff
@@ -200,7 +200,9 @@ var Applet = class Applet {
             setTimeout(() => this._getPanelInfo(), 0);
         }
     }
-
+    _setStyleClass(className){
```
NikoKrause, line 203: I haven't tested it, but I think it should be as follows:
```
_addStyleClass(className){
    this.actor.add_style_class_name(className);
}
```

js/ui/appletManager.js
NikoKrause, line 590: Hence this needs to be changed to:
```
applet._addStyleClass('applet-box'+appletClass);
```

NikoKrause:
As @collinss mentioned the class shouldn't be overwritten, but you should add an extra style class on top. It checks for the added style class first, if it doesn't find the style class in the theme, it falls back to the original style class. For example I'm doing this in CinnVIIStarkMenu. I submitted a review to this PR, but I haven't tested it yet. Maybe @brownsr can test and confirm, if it fixes the issues?

brownsr:
With the suggested changes it is definitely better.  The applet box themeing appears to work as originally.  But systray themeing is still broken.  I'd suggest reverting, and let the author bring it back once he has tested all the various cases out.

Regards

Simon

bomdia:
@brownsr i was reviewing this thing and i need to know what was broken with systray and how it had to be for correct behavior, so i can fix this thing and commit again with patched thing

brownsr:
I found problems with vertical panels, the existing CSS appeared not to be
applied.  I happen to use theme graphite-zero, but you may find the issues
on others as well.  From memory the font size adjustments in things like
the weather applet were not applied, and hovering over things in the
systray caused excess bordering, an 'item within item' appearance that the
CSS was crafted to avoid.  I seem to remember (may not be right here, I
didn't take notes) that hover could be a bit 'bouncy' in places with the
icon moving slightly around on hover.  I suspect that the .vertical bits in
the CSS were getting over-ridden.

hope that helps you debug

Regards

Simon

bomdia:
@brownsr found it, the problem happen because of poor code, i rewroted this better with te add as @NikoKrause suggested but without re adding applet-box class as  we are not setting it but adding and i don't get Any systray weird item in item and theme seems the same ad usual if they don't use new descriptor to theme thing, i'll submit the changes because i've moved the logic in a new function for better readibility

PR #9569, opened 2020-09-10
menu@cinnamon.org: Allow items to be re-ordered to better reflect the search outcome.

mtwebster:
- Buttons are sorted before they're sorted into their respective
  lists, so their original order within the app box can always be
  restored.
- When a search is made, an index is assigned to resulting items,
  corresponding to the position of the pattern within the search
  string. The four app search strings (name, keywords, description,
  id) are weighted, so name matches come before keywords, and so on.
- Places will always display after apps, and recents after places,
  maintaining their original respective positions.

Sorting is done simply by comparing the search indices for all
matching buttons:

Examples:
- Given two apps with with the names "File" and "The File", and
  search string "file", "File" has an index of 0 (matched at start of
  string), "The File" has an index of 4 (matched at position 4),
  so "File" will appear before "The File".
- Given two apps with matches of "File" in the name for one, and
  "File" in the keywords, the first will have an index of 0 (matched
  at the start of the name) and the second will have an index of
  1000 plus the position of "File" within the keywords string.

Performance appears to be little affected when searching, and there
shouldn't be much imact at all on normal non-search operation of
the applet.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -47,6 +47,15 @@ const RefreshFlags = Object.freeze({
 });
 const REFRESH_ALL_MASK = 0b11111;
 
+const NO_MATCH = 99999;
+const APP_MATCH_ADDERS = [
+    0, // name
+    1000, // keywords
+    2000, // desc
+    3000 // id
+];
+const RECENT_PLACES_ADDER = 4000;
+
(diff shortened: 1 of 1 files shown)


clefebvre:


```diff

Issues
======
- Added 1
           

Clones added
============
- files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=6159009&bid=20120617)

Issue #8779, opened 2019-08-08
Grouped Window List does not clear icons for apps that were closed while they were minimized


mattvalley:
```
 * Cinnamon version (cinnamon --version) = Cinnamon 4.2.3
 * I am not using the daily builds PPA.
 * Distribution = Linux Mint 19.2
 * 32 or 64 bit = 64 bit
 * Attach /home/<username>/.xsession-errors, or /var/log/syslog
 * Graphics hardware *and* driver used

Device-1: Intel HD Graphics 530 vendor: ASRock driver: i915 v: kernel bus ID: 00:02.0 
chip ID: 8086:1912 
Device-2: NVIDIA GM206 [GeForce GTX 950] vendor: Micro-Star MSI driver: nvidia 
v: 390.116 bus ID: 04:00.0 chip ID: 10de:1402 
Display: x11 server: X.Org 1.19.6 driver: modesetting,nvidia 
unloaded: fbdev,nouveau,vesa resolution: 1920x1080~60Hz, 1920x1080~60Hz, 1920x1080~60Hz 
OpenGL: renderer: GeForce GTX 950/PCIe/SSE2 v: 4.6.0 NVIDIA 390.116 direct render: Yes 
```

**Issue**
Sometimes when I have opened multiple instances of an app and then when I try to close an instance the panel does not remove the icon for that recently closed instance and instead keeps the icon but the icon is now in a broken state.  That icon still shows the window title when hovering over it but does not show the thumbnail because it does not refer to an existing process anymore.

Once the icon gets into that state every time I left-click on the icon I get the following error in my `~/.xsession-errors` log:
```
(cinnamon:2233): Cjs-WARNING **: 15:14:49.816: JS ERROR: TypeError: window.get_workspace(...) is null
activateWindow@/usr/share/cinnamon/js/ui/main.js:1479:76
anonymous/onAppButtonRelease/handleMinimizeToggle@/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js:737:17
onAppButtonRelease@/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js:754:17
anonymous/AppGroup/<@/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js:157:103
```  
[...]

PR #9571, opened 2020-09-11
Removed logic that was keeping icons around after window closed

icarter09:
This PR addresses #8779. 

This PR will need tested since the logic that was keeping the icons around is closely related to logic that is used when moving windows around from one workspace to another. I did some testing and things looked to still be working correctly. I'll remove the commented code once further testing has been established.


gadass (commented):
I think both issues described by @mattvalley in #8779 can be addressed by changing

` if (!metaWindow.showing_on_its_workspace()) return;`
by
` if (!metaWindow.showing_on_its_workspace() && metaWindow.minimize()) return;`

PR #9580, opened 2020-09-15
Deprecated Clutter keysyms migration

contribucious:
# What

```c
/*
 * Compatibility version of clutter-keysyms.h.
 *
 * Since Clutter 1.4, the key symbol defines have been changed to have
 * a KEY_ prefix. This is a compatibility header that is included when
 * deprecated symbols are enabled. Consider porting to the new names
 * instead.
 */
```
**Source:** [deprecated/clutter-keysyms.h (muffin)](https://github.com/linuxmint/muffin/blob/9cbb7ea75d3855ee11a71c11e75000edc7c1ede7/clutter/clutter/deprecated/clutter-keysyms.h#L20-L27)

:fast_forward: This PR does this throughout the content of the Cinnamon repository, in an automated manner. :thumbsup:

# How

```c
#define CLUTTER_VoidSymbol 0xffffff CLUTTER_DEPRECATED_MACRO_FOR("Deprecated key symbol. Use CLUTTER_KEY_VoidSymbol instead.")
```
:arrow_right_hook: Above, an excerpt from [this long list](https://github.com/linuxmint/muffin/blob/9cbb7ea75d3855ee11a71c11e75000edc7c1ede7/clutter/clutter/deprecated/clutter-keysyms.h#L34-L2132).

`awk '{ print $2 }'` = `CLUTTER_VoidSymbol`
`awk '{ print $8 }'` = `CLUTTER_KEY_VoidSymbol`

Just need to replace `CLUTTER_` by `Clutter.` and we can then look for the first to replace it with the second.

＞ In each file, perform this action but for all entries in this list — using two different security mechanisms to ensure that only exact matches are replaced. Also, the case is preserved, to differentiate `Clutter.a` from `Clutter.A` for example.

# Remarks

:bulb: I've checked in an automated manner **all** other entries in this long list that **don't have** `CLUTTER_DEPRECATED_MACRO_FOR` as well but none are present in the case of Cinnamon <strong>(</strong>e.g. `Clutter.HomePage` and `Clutter.AudioPlay` — which also have their non-deprecated equivalent `Clutter.KEY_HomePage` and `Clutter.KEY_AudioPlay`, simply that not specified in the file but easily determinable according to the same pattern<strong>)</strong>.

:bulb: As a bonus, in a second full round (once the migration is done), I transform the few ` == Clutter.KEY_` found into ` === Clutter.KEY_`.
&nbsp;
&nbsp;

___
**Similar to:** https://github.com/GNOME/gnome-shell/commit/d3d165243c8457cc65e8864cee493f27fca15d59 ([source](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/808/diffs?commit_id=d3d165243c8457cc65e8864cee493f27fca15d59))

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1011,7 +1011,7 @@ class Menu extends Applet.AppletPopupMenu {
     }
 
     _onKeyPressEvent(actor, event) {
-        if (event.get_key_symbol() == Clutter.KEY_Escape) {
+        if (event.get_key_symbol() === Clutter.KEY_Escape) {
             this.close(false);
             return true;
         }
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -209,7 +209,7 @@ class SimpleMenuItem {
         if (this.activate &&
             (symbol === Clutter.KEY_space ||
              symbol === Clutter.KEY_Return ||
-             symbol === Clutter.KP_Enter)) {
+             symbol === Clutter.KEY_KP_Enter)) {
             this.activate();
             return Clutter.EVENT_STOP;
         }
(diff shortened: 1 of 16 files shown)


contribucious:
@mtwebster Hello Michael. When you have time, can you review this PR? This, in order to avoid a predictable conflict regarding my next PR (to come). Thank you in advance to you! :thumbsup:

contribucious:
Thanks @mtwebster! :thumbsup:

Issue #9591, opened 2020-09-19
menu@cinnamon.org: search in Hebrew doesn't work


ronyala:
```
* Cinnamon version (cinnamon --version) Cinnamon 4.6.7
   - Please specify if you are using the daily builds PPA - yes
     (https://launchpad.net/~linuxmint-daily-build-team/+archive/ubuntu/daily-builds).
 * Distribution - Mint 20
 ```

**Issue**

When searching for apps/files in cinnamon menu using Hebrew, there are no results.

**Steps to reproduce**

Search for app names using Hebrew.

**Expected behaviour**

Display a list of apps and file name results matching the search query.

**Other information**

This problem exists since commit a6738bd306180b4e2e668c90c1d7d5d6c0578906, currently working on a PR to fix this issue.

ronyala:
Closing as this is fixed as of commit 18f24baf6b80ac7e4c9cf2b0c1bd928280e1d125

PR #9592, opened 2020-09-19
Fix search in Hebrew

ronyala:
Fixes issue #9591;
Replaced the `\b` expression in the regexes with `(?:^|\s|;|,|.)` to add support for non-Latin languages.
As specified [here](https://javascript.info/regexp-boundary), the `\b` expression does not support non-Latin alphabets.
`(?:^|\s|;|,|.)` is meant to check for search query at the start of the string, after every whitespace (space, newline, etc), and after `;`, `,` and `.`.


ronyala:
Closing as this seems to be fixed as of commit 18f24ba (by removing regex prefix entirely).

PR #9594, opened 2020-09-19
Rename applets and desklets Add/Remove text to Enable/Disable

mven:
To provide consistency with the extensions panel and to align the description with the interaction on what is actually being performed, the applets and desktlets "Add" and "Remove" text has been changed to "Enable" and "Disable".

| Applets Before (Add)   |      Applets After (Enable)     |
|:----------:|:-------------:|
| applets-add | applets-enable |

| Applets Before (Remove)   |      Applets After (Disable)     |
|:----------:|:-------------:|
| applets-remove | applets-disable |

| Desklets Before (Add)   |      Desklets After (Enable)     |
|:----------:|:-------------:|
| desklets-add | desklets-enable |

| Desklets Before (Remove)   |      Desklets After (Disable)     |
|:----------:|:-------------:|
| desklets-remove | desklets-disable |

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_applets.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_applets.py
@@ -51,8 +51,8 @@ class ManageAppletsPage(ManageSpicesPage):
     directories = [("%s/.local/share/cinnamon/applets") % GLib.get_home_dir(), "/usr/share/cinnamon/applets"]
     collection_type = "applet"
     installed_page_title = _("Installed applets")
-    instance_button_text = _("Add")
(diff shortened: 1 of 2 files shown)


collinss:
Enable/Disable is actually less correct for applets and desklets than Add/Remove. The reason they're different is because some applets and desklets are multi-instance (so if you already have one instance of the applet, you're not really 'enabling' it by adding a second one to a panel), while extensions are strictly single-instance (in this case Enable/Disable actually _does_ make sense, but Add/Remove _doesn't_).

mven:
Ugh. I forgot about that distinction... Well - this was a fun learning exercise on how to develop for Cinnamon - I'll chalk it up as such. Thanks 👍

Issue #9153, opened 2020-02-04
Grouped windows list: windows disappear...


kolayne:
```
 * Cinnamon 4.4.8 not using daily ppa
 * Mint 19.3
 * NVidia Geforce 940MX (nvidia-driver-435)
 * 64 bit
 ```

**Issue**
If you switch to a workspace, holding a window on top, and move back, still holding this window, it disappears from the GWL on all the desktops.
Sometimes the window disappears after the first desktop switch.

**Steps to reproduce**
0. Create at least two desktops, switch to a left one
1\. Open any window (for example, Nemo)
2. Place the cursor on the window's title
3. Holding the mouse button, press right arrow key
4. Holding the mouse button, press left arrow key

**Expected behaviour**
After both of the desktop switches, the window appears as opened in the active desktop's GWL

**Other information**
I guess, this may be related to #9080 

<details>
<summary>GWL configuration info</summary>

```JSON
{
    "layout": {
        "type": "layout",
        "pages": [
            "generalPage",
            "panelPage",
            "thumbnailsPage",
            "contextMenuPage"
        ],
        "generalPage": {
            "type": "page",
            "title": "General",
            "sections": [
                "generalSection",
                "hotKeysSection"
            ]
        },
        "panelPage": {
            "type": "page",
            "title": "Panel",
            "sections": [
                "appButtonsSection"
            ]
        },
        "thumbnailsPage": {
            "type": "page",
            "title": "Thumbnails",
            "sections": [
                "thumbnailsSection",
                "hoverPeekSection"
            ]
        },
        "contextMenuPage": {
            "type": "page",
            "title": "Context Menu",
            "sections": [
                "contextMenuSection"
            ]
        },
        "generalSection": {
            "type": "section",
            "title": "Behavior",
            "keys": [
                "group-apps",
                "scroll-behavior",
                "left-click-action",
                "middle-click-action",
                "show-all-workspaces"
            ]
        },
        "appButtonsSection": {
            "type": "section",
            "title": "Application Buttons",
            "keys": [
                "title-display",
                "launcher-animation-effect",
                "number-display",
                "pinned-apps",
                "enable-app-button-dragging"
            ]
        },
        "hotKeysSection": {
            "type": "section",
            "title": "Hot Keys",
            "keys": [
                "cycleMenusHotkey",
                "show-apps-order-hotkey",
                "show-apps-order-timeout"
            ]
        },
        "thumbnailsSection": {
            "type": "section",
[...]

PR #9595, opened 2020-09-19
Fix for GWL windows disappearing

icarter09:
This PR addresses #9153.

Fixed logic that checks when window should be removed.

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appList.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appList.js
@@ -340,11 +340,10 @@ class AppList {
     windowRemoved(metaWorkspace, metaWindow) {
         if (!this.state) return;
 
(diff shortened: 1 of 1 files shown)


TechnologyMan101:
I just tested this PR. Not only does it not seem to fix the issue, but it also breaks the panel. Now it shows apps that I don't have open, and also breaks highlighting. All apps that I open will stay there in the panel until I restart Cinnamon. Now, since I was testing an app. I have Discord on the panel for all 4 workspaces.

Screenshot from 2020-09-19 13-40-39
This screenshot shows what it is doing. Except for Firefox, I have nothing open. The pinned apps are Firefox, Terminal, and Nemo. They also persist the ones that were open that have already been closed.

icarter09:
@TechnologyMan101 thanks for the heads up. Odd since it does fix the one issue for me. I'll delete this PR and go back to the drawing board.

TechnologyMan101:
I actually overwrote the file with your file, and then rebooted to find this result. Of course I made a backup of the original file.

Issue #9567, opened 2020-08-28
Panel sometimes shows only app(s) related to a workspace that I'm not using at the moment


curtisgibby:
```
 * cinnamon-control-center version 4.6.2
 * Distribution - Mint 20.0
 * Graphics hardware - Intel Corporation HD Graphics 630
 * 64 bit
 ```

**Issue**

Panel sometimes shows only the app(s) related to a workspace that I'm not using at the moment. That is, I'll be in Workspace 1, but the Panel will show the app(s) that I've got open in Workspace 2. This happens several times a day since I upgraded to Mint 20 from Mint 17, where I never saw this problem.

(Updated with an example screenshot below)

**Steps to reproduce**

Unknown.

**Expected behavior**

Panel shows the app(s) related to a workspace that I'm using at the moment.

**Other information**

To get out of this behavior, I have to switch to the workspace containing the apps that I'm currently seeing in the Panel (Workspace 2), then back to the original desktop (Workspace 1).

icarter09:
@curtisgibby any luck on figuring on the steps to reproduce this?

curtisgibby:
Nope, it's sporadic. The next time I see this behavior, I'll definitely try to figure out what I was doing just beforehand.

curtisgibby:
One other interesting note that may help diagnose the problem: every time the "wrong" workspace's apps show up, they're always from the _same_ wrong workspace.

- Workspace 1 sometimes shows Workspace 2's panels
- Workspace 3 sometimes shows Workspace 2's panels
- Workspace 1 _never_ shows Workspace 3's panels
- Workspace 3 _never_ shows Workspace 1's panels

TechnologyMan101:
It happened to me only once and I was deliberately lagging out the computer by switching workspaces rapidly for a while.

curtisgibby:
@ronyala I think your conjecture is correct -- the problem does appear to be caused by rapidly jumping back and forth between workspaces. I can never get it to occur if I slow down my shortcut keypresses to trigger "Switch to Left Workspace" / "Switch to Right Workspace", but I can consistently make it happen if I push `Ctrl+Alt+Left` or `Ctrl+Alt+Right` more than once in rapid succession. Good detective work! :detective:

curtisgibby:
Glad to see it fixed! Any idea when it might land?

PR #9606, opened 2020-10-01
Fix issue 9567

ronyala:
The technical explanation is in the commit description
This will hopefully fix issue #9567


mtwebster:
Did you try setting throttle's callFirst argument to false? I think this causes the callback to be reset ifThis should cause the callback delay to reset on repeat calls within the timeout period - I'd rather do something like that than eliminating the delay altogether (which affect responsiveness).

ronyala:
Did try now. While setting callFirst to false does solve the issue, it creates a 100ms delay in the applet when switching workspaces that is noticeable when window animations are off.
Perhaps it's better to shorten the delay on the throttle to 30-50ms (much less noticeable) and set callFirst to false instead of removing it?

mtwebster:
Yes I think -  that seems to work well.

PR #9632, opened 2020-10-12
xlet-settings.py: Take into account icon.svg prior to icon.png

claudiux:
Until now, when viewing an applet's settings, the `icon.png` icon was displayed in the panel, which could contrast with the applet icon itself and confuse the user.

From now on, if an `icon.svg` icon exists then this one is displayed first.


mtwebster:
This looks fine, but can you copy the info here into the commit message? (This way the reasoning is kept with the commit).

Thanks

claudiux:
@mtwebster Done.

Issue #9320, opened 2020-05-18
Tile size of a window in taskbar differs


AveryEmerous:
```
 * Cinnamon version - Cinnamon 4.4.8
 * Distribution - Manjaro 20.0.1
 * Graphics hardware *and* driver used - GTX 970, proprietary (440xx)
 * 32 or 64 bit - 64 bit
 * Attach /home/<username>/.xsession-errors, or /var/log/syslog
[xsession-errors.txt ](https://github.com/linuxmint/cinnamon/files/4647414/xsession-errors.txt)
 ```

**Issue**
Taskbar window tile size doesn't follow the title name but does after a restart of cinnamon desktop with ALT+F2 and r.

**Steps to reproduce**
- Have open windows with titles on.
- Restart Cinnamon.
- Voila.

**Expected behaviour**
Either to keep the sizes of the tiles as big as they are allowed or follow the size of the title.

icarter09:
@AveryEmerous can you provide a screen capture of what you are seeing? Thanks.

AveryEmerous:
Pre-restart:
![image](https://user-images.githubusercontent.com/58141819/82470794-754ad380-9ace-11ea-8f3e-0d45988e69a4.png)

Post-restart:
![image](https://user-images.githubusercontent.com/58141819/82470826-7f6cd200-9ace-11ea-8a40-45e708213ea0.png)

evg-zhabotinsky:
I can't reproduce this on Cinnamon 5.6.8 (LM 21.1), restarting Cinnamon in any manner doesn't seem to affect window button width any more.

PR #9638, opened 2020-10-14
[WIP] Added proper scaling of grouped window list's window buttons. (Fixing issue #9320)

dvivanov00:
Issue described in #9320 has been fixed. Window buttons will now always be as long as the text of the window name. Previously, such behaviour would only occur when the DE is restarted with Alt+F2 and "r" as the input command and only for pre-existing window buttons.

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -449,7 +449,11 @@ class AppGroup {
             return;
         }
 
-        let width = MAX_BUTTON_WIDTH * global.ui_scale;
+        let iconNaturalSize = this.iconBox.get_preferred_width(forHeight);
+        let labelNaturalSize = this.label.get_preferred_width(forHeight);
+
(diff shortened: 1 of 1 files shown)


clefebvre:


```diff

Issues
======
- Added 2
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=6339745&bid=20593568)

mtwebster:
We had some discussion on this, and I think having the buttons all the same (pre-restart with current cinnamon) is the best way to go.

This is more consistent with other applets (and other desktops), as well as it's what users are expecting.  Most folks don't restart cinnamon much, if at all, so any change from the current initial behavior could be considered a regression.

Can you reverse the behavior here to keep the button sizes all the same?
Thanks

evg-zhabotinsky:
I can't reproduce the bug that this patch tried to fix (see #9320 )
Also, the behavior implemented by this patch now can be achieved without it, mostly "as is" by checking "use the entrie space available", and fully with the simple fix in #11608 by also setting "button width" to zero.
Therefore, I think this PR can be closed.

PR #9643, opened 2020-10-16
Some post-meson port build fixes

collinss:
- debian build: use debhelper >= 11 for meson support
- build: re-add missing defines to suppress some extra noise in the logs.


Fantu:
thanks for debian changes, in debian/rules can be removed also "DHFLAGS=--parallel" that is default in newer debhelper

if bionic is not supported anymore (and only focal and buster) can do debhelper-compat (= 12), instead of debhelper (>= 11.1~) and remove debian/compat

mtwebster:
Added the other changes here: https://github.com/linuxmint/cinnamon/commit/f9ae00edd00f9ec66ba8cd31931a951f09ae167c

PR #9664, opened 2020-10-20
Simplify spice version checks

clefebvre:
Change how some of the metadata fields are interpreted.

- Don't read "multiversion" anymore, always assume it is TRUE.
- No longer interpret "cinnamon-version" as an array which should contain our version but as an array which should contain any version lower than ours.
- No longer let the user load incompatible versions.
- Always assume forward compatibility.

Basically this turns all spices into multiversion spices and cinnamon-version into a field which specifies the lowest required version of Cinnamon.

Important notes:

- Cinnamon 4.6 and older continue to read these fields the old way, so in spices cinnamon-version should remain an array and specify versions between 3.0 and 4.6, and multiversion should remain specified also for compatibility with 3.0-4.6.


clefebvre:


```diff

Issues
======
+ Solved 1
- Added 3
           

Complexity decreasing per file
==============================
+ files/usr/share/cinnamon/cinnamon-settings/bin/ExtensionCore.py  -4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=6371477&bid=20675375)

PR #9668, opened 2020-10-21
[WIP] window-list applet: use the hotkeys to quick switch between windows

veryangryman:
Add hot keys to switch between windows by its number.

By default these shortcuts is used:
 - Super+N - for windows with numbers 1..10
 - Super+Shift+N - for windows with numbers 11..20

Hotkeys is fully customizable. This feature can be disabled/enabled.
Code is simple and stable - tested around two months with various
versions of cinnamon.

### DIFF
--- a/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
@@ -109,7 +109,7 @@ class WindowPreview extends Tooltips.TooltipBase {
         if (this._windowActor) {
             return this._windowActor;
         } else {
-            log("metaWindow has no actor!");
+            global.log("[window-list] metaWindow has no actor!");
             return null;
         }
     }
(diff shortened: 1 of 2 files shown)


clefebvre:


```diff

Complexity increasing per file
==============================
- files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cinnamon/Cinnamon/pullRequest?prid=6376359&bid=20686837)

mtwebster:
Hi, I thought I had provided feedback on this weeks ago, I must have forgotten to submit it - I apologize.

I think this would be a nice feature to have, but I think the way you configure it is way too complicated. I'd prefer to have it more along the lines of how the grouped window list does it.

- The 'Show the number in a window button title' should be under the 'Switch to a window by the number' setting, and be visible only if that setting is enable (the former should depend on the latter).
- While you'd still need to set these up individually in the applet, we don't need to be able to configure every single possible shortcut. I'd prefer it like gwl with the modifiers fixed, and you're just enabling the feature.
- I'm not a huge fan of the [n] representation - it takes up a large amount of space. Again, gwl is nice here in that the numbers are superscripted and don't affect the text.

Thanks

PR #9697, opened 2020-11-11
grouped-window-list: Add FLASH_MAX_COUNT constant

tdgroot:
This way the amount of flashes is not hardcoded in the UI code anymore.

If this were to be merged, should I do the same for the `window-list` applet?


mtwebster:
You can do the window list if you'd like.  I generally prefer using constants for readability.

PR #9699, opened 2020-11-12
grouped-window-list applet: Add settings for attention flash

tdgroot:
Add settings for attention flash, making it configurable:

image
(theme: Adapta-Nokto)


mtwebster:
Hi,

While we appreciate the effort here, this is overkill. With something like this, we've found that in most cases, it's best to settle on 
 a single ideal behavior and use it.

We're definitely not opposed to having another look at these types of decisions as we go, but they should be about improving the existing behavior, not making it configurable.

Thank you

tdgroot:
@mtwebster Okay then.. 

Then I would like to know if the current flash behavior is debatable, because to me, it is completely useless in default Cinnamon. I would always miss the fact that applications are asking for attention (especially chat applications). Then again, it could be just about perfect to someone else, so I really think it comes down to personal taste, hence the addition of the configurations of the applet.

Issue #9310, opened 2020-05-15
Feature Request: Add mic controls to keyboard shortcuts


BeatLink:
Hey, it would be really helpful if you could add the following mic controls as possible commands to bind keyboard shortcuts to:

Mic Mute
Mic Unmute
Mic Toggle
Mic Volume Up
Mic Volume Down

Thanks in advance!

esa1975:
I would settle for a mic mute/unmute toggle like GNOME and Plasma have. Incidentally, if anyone knows of a decent workaround for the time being please share. I've done some things with scripts but none of it works particularly well.

vsudakov:
this feature is **must have** and absolutely necessary!

DemonTPx:
In my pull request above, I've added the "Mic mute" option, so it can be bound to a keyboard shortcut. I found out that cinnamon already supports this, but only the option is missing in the keyboard shortcut configuration panel.

To add a keyboard shortcut manually you can use something like this (I added Scroll_Lock as my shortcut button, which I never use anyway):

```
gsettings set org.cinnamon.desktop.keybindings.media-keys mic-mute "['XF86AudioMicMute', 'Scroll_Lock']"
```

To see what shortcuts are configured for mic-mute:

```
gsettings get org.cinnamon.desktop.keybindings.media-keys mic-mute
```

Note that the shortcut will only work if you have an input connected which can be muted.

vsudakov:
@DemonTPx thank you - your tips work for me. Though instead of `Scroll_Lock` I use `<Super>m` because `Scroll_Lock` doesn't work for me (I'm not sure what is the cause of that)

```
gsettings set org.cinnamon.desktop.keybindings.media-keys mic-mute "['XF86AudioMicMute', '<Super>m']"
```

rootwork:
I agree, it'd be nice to see this in the Keyboard > Shortcuts > Sound and Media panel.

Thanks, @DemonTPx, for creating a PR and for discovering we can set this already on the command line.

I have an eject button on my keyboard that is perfect for this (sadly I can't map it alone, but `<Primary>XF86Eject` worked).

PR #9704, opened 2020-11-14
Add mic mute keyboard shortcut option

DemonTPx:
This allows a keyboard shortcut to be configured for "Mic mute". Adding this and configuring a key already seems to make toggle muting the default input/mic to work for me.

Partly fixes https://github.com/linuxmint/cinnamon/issues/9310


DemonTPx:
If people can't wait for this pull request to be merged and want to configure a keyboard shortcut manually, try this:

To add a keyboard shortcut manually you can use something like this (I added Scroll_Lock as my shortcut button, which I never use anyway):
```
gsettings set org.cinnamon.desktop.keybindings.media-keys mic-mute "['XF86AudioMicMute', 'Scroll_Lock']"
```

To see what shortcuts are configured for mic-mute:
```
gsettings get org.cinnamon.desktop.keybindings.media-keys mic-mute
```

Note that the shortcut will only work if you have an input connected which can be muted.

Issue #9727, opened 2020-11-28
Cinnamon fails to build if disable_networkmanager=true


darkshram:
```
 * Cinnamon version 4.8.0
   - from github 4.8.0 release source
 ```

**Issue**
If cinnamon is build using the meson option 'disable_networkmanager=true', build fails because libnm is defined in src/meson.build and meson reports it as unknown variable. The workaround was to remove 'libnm,' (line 87) from src/meson.build if build is done without NetworkManager support.

**Steps to reproduce**
Try to build cinnamon without NetworkManager support (-D disable_networkmanager=true)

**Expected behaviour**
Will fail to build because libnm is an unknown variable.

`src/meson.build:73:0: ERROR: Unknown variable "libnm".`

**Other information**
The workaround was to remove 'libnm,' (line 87) from src/meson.build if cinnamon is built without Neworkmanager support (-D disable_networkmanager=true).

PR #9729, opened 2020-11-28
meson: Fix build when network manager is disabled

JosephMcc:
It's optional but we need to account for that in the libcinnamon dependencies
as well.

Fixes: https://github.com/linuxmint/cinnamon/issues/9727


eli-schwartz:
We don't need to account for it at all.

PR #9733, opened 2020-11-28
Lint the python code a bit

ItzSwirlz:
These are just simple normal general Python linting. I didn't deal with cinnamon-settings and skipped
some stuff as I didn't want to create a monster of problems.


ItzSwirlz:
Closing as this should probably be something left for 5.0 as it's a major version and in case anything bad happens with the code, it's also better to lint file by file one at a time.

Issue #8549, opened 2019-05-08
A bug (?) in the zoom accessibility tool


kolayne:
```
 * Cinnamon 4.0.10
 * Distribution - Mint 19.1
 * 64 bit
 ```

**Issue**
When using the zoom accessibility tool, mouse scrolling is caught by an opened window, on which the cursor is.

**Steps to reproduce**
Switch on the zoom tool in the accessibility system settings
Start using it (Alt + scrolling), holding the cursor on any window

**Expected behaviour**
Screen image becomes zoomed, nothing else happens

**Actual behaviour**
Screen image becomes zoomed, the window content under cursor scrolls

sshimko:
Chiming in to say this is an issue for me as well. It is an annoyance, not a show-stopped. Trying to zoom in to read a webpage results in the page scrolling. Once you are done zooming, you have to fix the page position to pick up where you left off.

Cinnamon 4.2.3
Fedora 30
64 bit

I glanced through the zoom and magnifier code including dbus/gio/js and nothing was immediately apparent, but if someone can point me in the right direction I can take a crack at a fix and submit a PR.

As an aside - I tried all the major window managers and all of their accessibility support was either broken (gnome-shell) or not featureful enough to make it an acceptable solution (Kwin). Cinnamon’s great. Thank you!

kolayne:
Cinnamon 4.4.8. Not fixed yet :(

kolayne:
Can reproduce in Cinnamon 4.6.6

nkreer:
I've noticed this too, but only when using the chromium browser (including any Electron apps etc.), so I'm guessing this is an issue with the way that engine handles input. It has become a mayor annoyance for me as I started using lots of different electron apps recently, like vscode. Perhaps we should open an issue with the Chromium team and hope they'll fix it and the change propagates quickly to Electron and other dependants (hopefully within years, not decades) or implement a solution where the system would catch the zoom shortcut and cancel the event before other applications even see it.

kolayne:
> so I'm guessing this is an issue with the way that engine handles input

I believe this kind of input should not get to applications at all. I mean, if I'm holding the `Alt` key and scrolling, cinnamon should capture this input, process it but not forward to the application my mouse is hovering above (of course, this should only happen if zoom is enabled).
[...]

PR #9747, opened 2020-12-04
Include shift key as magnifier shortcut

nkreer:
This pull request adds the Shift key as an option for mouse wheel magnification hotkeys to cinnamon-settings. In the current situation, this works a lot better than the other options because blink/chromium-based UIs (including cinnamon itself) would react to scrolling even though a hotkey was pressed and the magnifier was active. This is not the case with the Shift key.

Shift is the hotkey for horizontal mouse wheel scrolling, though, so it will interfere with that. But seeing that horizontal scrolling is less common in UIs, it is a suitable workaround for most users and applications. In cinnamon UIs, zooming while holding shift is not possible - it will just scroll vertically. From a usability perspective, this is more likely to be understood.

This issue was first described in #8549 and I've already proposed this workaround there.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_accessibility.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_accessibility.py
@@ -84,7 +84,7 @@ class Module:
             spin = GSettingsSpinButton(_("Magnification"), "org.cinnamon.desktop.a11y.magnifier", "mag-factor", None, 1.0, 15.0, step=0.5)
             settings.add_reveal_row(spin, "org.cinnamon.desktop.a11y.applications", "screen-magnifier-enabled")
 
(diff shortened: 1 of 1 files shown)


ItzSwirlz:
I assume you are aware about the differences between **Shift** and how there can be confusions with Shift_R and Shift_L?

nkreer:
The addition works with both Shift_R and Shift_L.

ItzSwirlz:
I see. Actually; even better. Make another commit, instead of just 'Shift' have an option for 'Shift_L' and 'Shift_R'

nkreer:
That's not possible with the way this works, I'm afraid. I tried setting <L_Shift> and <Shift_L> respectively, but that is not interpreted correctly. I'm not aware how this works under the hood - maybe the maintainers have a suggestion before just merging.

By the way, R_Shift too toggles horizontal mouse wheel scrolling, so there is really no difference as to which key is used.

mtwebster:
Shift (and any other modifier key) is just shift when used as a modifier. They're only left/right if used as standalone shortcuts.

PR #9753, opened 2020-12-07
Lowercase schema name org.cinnamon.screensaver

ItzSwirlz:


mtwebster:
1) Those aren't DBUS names/paths not schema names
2) You can't change these without modifications to things that attempt to communicate with this program. We don't necessarily know what those things are (cinnamon third-party applets, etc..)

3) Please stop submitting patches that you don't test.

Issue #9754, opened 2020-12-08
Cinnamon settings: cs_info: Failure when detecting CPU name leads to crash


Cyanoxygen:
Hi, 
I just encountered an issue trying to open the System Info page of Cinnamon Settings, this page can not be opened.

### Environment
ARM64 Raspberry Pi 4, Cinnamon 4.8.0

### Troubleshooting
I run `cinnamon-settings info` in the terminal and get this output:

```
Using pam module (python3-pampy)
/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py:427: DeprecationWarning: Gtk.Window.set_wmclass is deprecated
  self.window.set_wmclass(wm_class, wm_class)
Loading Info module
Traceback (most recent call last):
  File "/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py", line 731, in <module>
    window = MainWindow()
  File "/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py", line 432, in __init__
    self.go_to_sidepage(cat, path, user_action=False)
  File "/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py", line 189, in go_to_sidepage
    sidePage.build()
  File "/usr/share/cinnamon/cinnamon-settings/bin/SettingsWidgets.py", line 212, in build
    self.module.on_module_selected()
  File "/usr/share/cinnamon/cinnamon-settings/modules/cs_info.py", line 154, in on_module_selected
    infos = createSystemInfos()
  File "/usr/share/cinnamon/cinnamon-settings/modules/cs_info.py", line 95, in createSystemInfos
    processorName = procInfos['cpu_name'].replace("(R)", "\u00A9").replace("(TM)", "\u2122")
KeyError: 'cpu_name'
```
I then looked into the code detecting the CPU name:

https://github.com/linuxmint/cinnamon/blob/fa2293b25657a26808513431809434128febef43/files/usr/share/cinnamon/cinnamon-settings/modules/cs_info.py#L70

I do think this code can not work on some platforms (for my instance it is arm64). Then I referred the [code from neofetch](https://github.com/dylanaraps/neofetch/blob/6dd85d67fc0d4ede9248f2df31b2cd554cca6c2f/neofetch#L2163), and wrote a simple fix for it: https://github.com/Cyanoxygen/cinnamon/commit/b60b93165f76ef370bb248e37525863facb71e74.

Here's the result:
[...]

PR #9755, opened 2020-12-08
cs_info: fix crash caused by missing key under getProcInfo

Cyanoxygen:
It will try to detect CPU model and leave it being "Unknown" if all attempts failed.
Also see #9754.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_info.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_info.py
@@ -66,18 +66,25 @@ def getDiskSize():
 
 
 def getProcInfos():
+    # For some platforms, 'model name' will no longer take effect.
+    # We can try our best to detect it, but if all attempts failed just leave it to be "Unknown".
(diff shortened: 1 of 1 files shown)


clefebvre:
Cool but there are two issues..

- Unknown isn't translated, it should be _("Unknown")
- _("Unknown") cannot be translated as an adjective in Latin languages if it refers to multiple nouns which have different genre and are either singular or plural...

clefebvre:
image

It would be better to use None instead of Unknown within the hashmap, and then exploit this value to show less information when displaying it... 

For instance we could simply show the CPU name without the core number if we don't have it. Or we could show nothing at all, or something like _("Unknown CPU") if we can't find the name.

clefebvre:
There's a crash still when memTotal isn't detected.

Cyanoxygen:
Thank you very much! I'm going to fix these problems then.

Cyanoxygen:
> It would be better to use None instead of Unknown within the hashmap, and then exploit this value to show less information when displaying it...

For `getProcInfo()` we can only fill out the `cpu_name` and `mem_total` , the main parser will ignore if other key does not present.
And this is it! Just get the `Unknown CPU` and `Unknown Capacity` translated:
image

clefebvre:
Thanks :+1:

PR #9761, opened 2020-12-11
[WIP] magnifier: Configurability and general improvements to user experience

KreerC:
This PR adds configuration options for several magnifier features Cinnamon has retained from GNOME, but haven't been usable in the DE since. It also aims to improve the overall user experience of magnifier.

Changes to cinnamon-settings:

* Add additional settings for magnifier
  * Enable/Disable cross-hairs
  * Cross-hair opacity
  * Cross-hair thickness
  * Cross-hair color

Changes to magnifier.js:

 * No styling when lens is in fullscreen mode

I'll be adding more magnifier-specific things over the next couple of weeks. I'd be happy to receive some sort of mentorship to get more familiar with Cinnamons architecture.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_accessibility.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_accessibility.py
@@ -134,8 +134,30 @@ class Module:
             else:
                 self.zoom_stack.set_visible_child_name("screen")
 
-            crosshair = GSettingsSwitch(_("Enable crosshair"), "org.cinnamon.desktop.a11y.magnifier", "show-cross-hairs")
-            settings.add_reveal_row(crosshair, "org.cinnamon.desktop.a11y.applications", "screen-magnifier-enabled")
+            widget = GSettingsSwitch(_("Enable crosshair"), "org.cinnamon.desktop.a11y.magnifier", "show-cross-hairs")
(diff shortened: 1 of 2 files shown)


mtwebster:
Hi, thanks for pr - this seems to cause a regression if you enable lens mode - If you start disabled, then enable it, it ignores further setting changes after that. If you restart cinnamon with it already enabled, zoom is very broken from the start.

You can take a look at ~/.xsession-errors to see the errors it throws (this is the same logging that you find in the looking glass (windows-key + L).

To be honest, we've been aware of the crosshair stuff but didn't think they added much usefulness, and we're wary of adding too many options - we try to find a sane configuration that will be fine for the most users.

I'd prefer to just have 'enable crosshairs' and allow the user to select the color - the thickness and opacity are overkill - we should find a nice default configuration and just use it (personally I think thinner is better - 3 or 4 max). Those other options will still exist if there is some niche user that *must* have it some other way, but they're not visible to the other 99.9% of users.

We don't really do any sort of mentor-ship, but you can ask questions whenever you want (you can just ping us in github) and we'll try to guide you a bit. There's some broad overview and some coding guidelines here: https://linuxmint-developer-guide.readthedocs.io/en/latest/cinnamon.html.

Really the best way to learn quickly is to do what you're doing - look for things to fix/improve and go from there.

ItzSwirlz:
@nkreer you interested in working on this, fixing it and cleaning it up?

PR #9784, opened 2020-12-22
Power applet: Use spaces after commas in paramaters.

ItzSwirlz:
Linting with ESLint


ItzSwirlz:
deleting because i'm stupid

PR #9788, opened 2020-12-23
Simplify path to .xsession-errors

dandv:


LinuxOnTheDesktop:
If the idea is to help the bug reporter, please note that (1) .xsession-errors is a hidden file, (2) it lacks an extension and so cannot - without renaming - be attached to a github issue.

Issue #9758, opened 2020-12-09
Graphical glitch - Media keys OSD volume % font clipped off at bottom


Furvy:
```
 * Cinnamon version: 4.8.2
 * Distribution: Arch
 * Graphics hardware *and* driver used: AMD Radeon RX 5700 XT (NAVI10, DRM 3.39.0, 5.9.13-arch1-1, LLVM 11.0.0); Mesa 20.3.0-3
 * Display resolution: 1920x1200 at 96 DPI. User interface scale is at Automatic (1x).
 * 32 or 64 bit: 64 bit
 ```

**Issue**
- Media keys OSD volume numerical percentage is clipped off ~2px at the bottom by some sort of overlayed padding or element around the border that shows on top of the value.

- I am not clear if this glitch was due to something I modified as I've been mucking around with editing the menu applet in /usr/share/cinnamon/applets/menu@cinnamon.org as well. I just know that this was not an issue until a recent release of Cinnamon, and/or its accompanying components.

**Steps to reproduce**
- Media keys OSD size set to "Small" or "Medium". It appears like "Large" does not clip over significantly, or might ~1px. Is not affected by whichever Desktop Theme is selected. For purposes of testing, I reset everything to Adwaita Icons, Adwaita Controls and Cinnamon Desktop under Themes.

**Expected behaviour**
- The font indicating volume percentage numerically should not be clipped off in any OSD size setting.

**Other information**
- Editing cinnamon.css in whatever theme under section "* OSD Popups" does not appear to affect the font size in a direct manner. Adding "font-size: 12pt" under ".osd-window {" should do it from my best guess and experiments, but the actual size of the font changes in some sort of buried code or CSS not directly defined by this. Eliminating all CSS under the "* OSD Popups" section also does not correct the underlying issue. It appears to be Theme-unrelated or defined in some other style file.

- In dconf editor under "org.cinnamon.desktop.keybindings.media-keys" there is a property called "decrease-text-size" but I have no idea what to write in there as a custom value. "true" does not work, nor "12", nor "12pt" nor "font-size: 12pt".

- No idea if this is a font issue, I have Cantarell as the Default and Desktop fonts.

- Attached image shows from left to right "Small", "Medium", "Large"
![MediaOSD](https://user-images.githubusercontent.com/29053825/101563174-a234da00-397d-11eb-9910-00716d6a78e8.jpg)

Furvy:
After looking through the recent changes, appears directly related to this:

[...]

PR #9811, opened 2020-12-31
Attempt to fix clipping issue with osdWindow.js

Furvy:
https://github.com/linuxmint/cinnamon/issues/9758

I have literally zero idea what I am doing or how to code JavaScript, but my issue has been open more than 20 days with no response, so I am desperate at trying to get this resolved. I figure this is the only way to get someone to look at it?

This fixes the issue I described and also makes the Volume % font less blurry and more "crisp". There is some sort of issue with the OSD_SIZE and this._sizeMultiplier not playing nice with each other according to the previous values.

### DIFF
--- a/js/ui/osdWindow.js
+++ b/js/ui/osdWindow.js
@@ -12,7 +12,7 @@ const LEVEL_ANIMATION_TIME = 0.1;
 const FADE_TIME = 0.1;
 const HIDE_TIMEOUT = 1500;
 
-const OSD_SIZE = 110;
+const OSD_SIZE = 100;
 
 function convertGdkIndex(monitorIndex) {
     let screen = Gdk.Screen.get_default();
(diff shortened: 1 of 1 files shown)


Furvy:
Attached is the improvements at all OSD size settings using my fix
Untitled

clefebvre:
Thanks @Furvy we'll have a look post-release.

ありがとう！You called me Senpai?! :)

clefebvre:
blurriness is usually fixed by rounding the coordinates.

Furvy:
Thank you so much for looking at this!

I think I figured out why this would be an issue - My display resolution 1920x**1200**. I changed it to 1920x1080 and applied the changes here and there was negligible difference from the original - Slight sharpness change perhaps on 'Medium' OSD Size.

Seems to only benefit my odd display resolution? Perhaps other weird resolutions too? I very much appreciate your time taking a look.

Issue #9200, opened 2020-03-09
Application icon not showing in panel after moving to another workspace


ariasmn:
```
 * Cinnamon version - 4.4.8
 * Distribution - Mint 19.3
 * Graphics hardware *and* driver used - Could reproduce the problem in several machines. UHD 620, HD520, SVGA3D in VirtualBox.
 * 32 or 64 bit - 64 bit
 * Attach /home/<username>/.xsession-errors, or /var/log/syslog - Seems like there are no related errors in the logs. The tail content of each:

.xsession-errors
Cjs-Message: 14:40:51.015: JS LOG: [LookingGlass/info] Adding systray: mintupdate.py (24x24px)
Cjs-Message: 14:40:51.022: JS LOG: [LookingGlass/info] Resized mintupdate.py with normalized size (24x24px)
Cjs-Message: 14:41:04.875: JS LOG: [LookingGlass/info] Adding systray: mintreport (24x24px)
Cjs-Message: 14:41:04.877: JS LOG: [LookingGlass/info] Resized mintreport with normalized size (24x24px)

(tray.py:2343): Gdk-CRITICAL **: 14:41:04.884: gdk_window_thaw_toplevel_updates: assertion 'window->update_and_descendants_freeze_count > 0' failed
Cjs-Message: 14:42:15.915: JS LOG: [LookingGlass/info] Adding systray: mintupdate.py (24x24px)
Cjs-Message: 14:42:15.920: JS LOG: [LookingGlass/info] Resized mintupdate.py with normalized size (24x24px)
Cjs-Message: 14:42:20.020: JS LOG: [LookingGlass/info] Adding systray: mintupdate.py (24x24px)
Cjs-Message: 14:42:20.022: JS LOG: [LookingGlass/info] Resized mintupdate.py with normalized size (24x24px)

/var/log/syslog
Mar  9 14:40:41 osboxes pulseaudio[1298]: [pulseaudio] bluez5-util.c: GetManagedObjects() failed: org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
[...]

Issue #9080, opened 2020-01-01
On-all-workspaces windows are not displayed in grouped windows list


kolayne:
```
 * Cinnamon 4.4.6
   - Please specify if you are using the daily builds PPA: no
 * Mint 19.3
 * NVidia Geforce 940MX (nvidia-driver-435)
 * 64 bit
 ```

**Issue**
If I mark a window to be visible on all workspaces and then switch to another workspace, it is visible there as expected, but is not displayed in a grouped windows list.

**Steps to reproduce**
1) Open any application
2) Mark it as visible on all workspaces (Always on Visible Workspace)
3) Switch to another workspace

**Expected behaviour**
The window is shown in grouped windows lists for each of the workspaces

**Comment by JosephMcc** [comment]
I can reproduce this

**Comment by claudiux** [comment]
I can't reproduce this.
```
Cinnamon 4.4.6
Mint 19.3 (upgraded from 19.2)
Graphic card: Intel Corporation Xeon E3-1200 v3/4th Gen Core Processor Integrated Graphics Controller driver=snd_hda_intel latency=0
Kernel 5.3.0-24-generic
```

**Comment by JosephMcc** [comment]
Make sure show windows from all workspaces is disabled in the applets settings.

**Comment by claudiux** [comment]
@JosephMcc This option is disabled.
Tested with many applications (terminal, firefox, libreoffice...)

Procedure:

1. Open Terminal in workspace 1 -> Switch to workspace 2 -> Terminal icon doesn't appear in GWL (OK).
2. Switch to WS1 -> Select 'Visible on all WS' -> Switch to WS2 -> Terminal icon appears in GWL (OK).
3. Switch to WS1 -> Open a second Terminal -> Switch to WS2 -> Only the icon of the first Terminal appears (OK).
4. Switch to WS1 -> Select the second Terminal then select 'Visible on all WS' -> Switch to WS2 -> both 'Terminal' are accessible (OK).

**Comment by JosephMcc** [comment]
Well I can't reproduce it now either. But have been messing around. Will have to try again when I do a fresh boot. Weirdly some of these GWL problems seem to just "go away" after you mess with something. It's actually really frustrating.

**Comment by mtwebster** [comment]
This doesn't happen unless grouping is enabled it seems.

This is why it's important people supply the requested info, the issue template requests applet configuration be supplied.

**Comment by claudiux** [comment]
My configuration of GWL:
```
{
    "layout": {
        "type": "layout",
        "pages": [
[...]

PR #9819, opened 2021-01-01
I'm trying to fix two issues related to the GWL applet.

gadass:
1)  Does not clear icons for apps that were closed while they were minimized, related to #8779. 

2)  windows disappear while dragging when the arrow or numeric keys are used to switch workspaces, related to #9153.

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appList.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appList.js
@@ -333,6 +333,8 @@ class AppList {
             // Abort the remove if the window is just changing workspaces, window
             // should always remain indexed on all workspaces while its mapped.
             // if (!metaWindow.showing_on_its_workspace()) return;
+            if ((!metaWindow.showing_on_its_workspace() && metaWindow.minimize())
+            || metaWindow.has_focus() && metaWorkspace.index() !== global.screen.get_active_workspace_index()) return;
             this.state.removingWindowFromWorkspaces = true;
             this.state.trigger('removeWindowFromAllWorkspaces', metaWindow);
             return;
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appList.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appList.js
@@ -50,6 +50,8 @@ class AppList {
 
         // Connect all the signals
         this.signals.connect(global.screen, 'window-workspace-changed', (...args) => this.windowWorkspaceChanged(...args));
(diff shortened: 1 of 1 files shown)


files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appList.js
```diff
@@ -333,6 +335,8 @@ class AppList {
             // Abort the remove if the window is just changing workspaces, window
             // should always remain indexed on all workspaces while its mapped.
             // if (!metaWindow.showing_on_its_workspace()) return;
+            if ((!metaWindow.showing_on_its_workspace() && metaWindow.minimize())
+            || metaWindow.has_focus() && metaWorkspace.index() !== global.screen.get_active_workspace_index()) return;
             this.state.removingWindowFromWorkspaces = true;
             this.state.trigger('removeWindowFromAllWorkspaces', metaWindow);
             return;
```
kolayne, line 342: Not sure, just a suggestion: this one probably looks prettier, because it only uses one return in the block and (as to me) it's easier to understand this piece as "if we should abort then (if condition then do something); stop" than "if we should abort then (if condition then stop); do something; stop"
```suggestion
            if ((metaWindow.showing_on_its_workspace() || !metaWindow.minimize())
            && (!metaWindow.has_focus() || metaWorkspace.index() === global.screen.get_active_workspace_index()))
            {
                this.state.removingWindowFromWorkspaces = true;
                this.state.trigger('removeWindowFromAllWorkspaces', metaWindow);
            }
            return;
```

gadass: @kolayne thanks. I'm was just trying to solve it with minimal changes :) but I see your point and seems like a nice suggestion. Works fine here.

files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appList.js
```diff
@@ -50,6 +50,8 @@ class AppList {
 
         // Connect all the signals
         this.signals.connect(global.screen, 'window-workspace-changed', (...args) => this.windowWorkspaceChanged(...args));
+        // ugly hack: forces applet to refresh the list of all removed app instances in all workspaces
+        this.signals.connect(global.screen, 'window-removed', (...args) => this.windowRemoved(...args));
```
kolayne, line 54: Found a trailing whitespace here

gadass:
Those changes in appList.js have been useful to solve the described issues only when the "Show windows from all workspaces" setting is enabled. However, I found several tickets (#9200, #9080 and #9153) that reported similar issues when the "Show windows from all workspaces" setting is disabled. I will close this PR and look to a more general solution.

PR #9821, opened 2021-01-02
[WIP] various: Fix hidpi change detection.

mtwebster:
This was broken with fractional scaling - we have traditionally
equated the changing of Xft.dpi to mean the scale changed, so we
could update our ui (cinnamon runs in 1x always, but we scale our
elements to match the global scale).

When fractional scaling is disabled (switched off thru the gui),
these old methods are fine - changes to the global scale affect
the dpi as well.

With it enabled, however, the following differences manifest:
- The global scale becomes only the base scale factor from which
  the fractional scale will be achieved.
- Only changes to the fractional value will affect the Xft.dpi
  value; changing the global scale does not.

Unfortunately cinnamon (as well as some apps in certain situations)
need to know what the global scale is, and when it changes, so the
dpi is no longer a reliable value to monitor.

This uses the primary GdkMonitor to listen for changes to its
scale-factor property and notify the appropriate parties when it
changes.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1175,6 +1175,8 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
         this._updateKeybinding();
 
         Main.themeManager.connect("theme-set", Lang.bind(this, this._updateIconAndLabel));
+        global.connect('scale-changed', () => this.queueRefresh(REFRESH_ALL_MASK));
+
         this._updateIconAndLabel();
 
         this._searchInactiveIcon = new St.Icon({ style_class: 'menu-search-entry-icon',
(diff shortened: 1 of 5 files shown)


benmccann:
Thanks @mtwebster. This will be a big improvement! Just FYI, it looks like this one needs a rebase

benmccann:
This says it's a work-in-progress, but it's not clear to me, is there something specific that needs to be done still? Is there anything you'd want help with on this PR?

mtwebster:
I'll get back on this soon and finish it up.

starlightknight:
That'd be much appreciated. I'd be happy to help test if needed. 🙂 

I struggle with this issue when switching terminals on my KVM - it re-occurs every time I switch machines. So I'll be really happy when it's fixed.

benmccann:
Thanks @mtwebster! Just gently checking in - are you still planning to get back to this one? If not, let us know what would be needed to complete the PR. Thanks for making Cinnamon :heart:

mtwebster:
obsolete due to changes in 5.4

Issue #9823, opened 2021-01-03
[Bug] Battery notification icon is inaccurate


mystic-githuber:
Cinnamon version 4.8.5-1
  EndeavourOS Linux x86_64 (Arch)
  Kernel: 5.10.3-arch1-1
  Intel i5-10210U (8) @ 4.200GHz * 64 bit 
   Intel UHD Graphics
 
**Issue**
Battery notification icon displays inaccurate reading. As you see from the image below the icon is showing that there is 3/4 battery left when there is only 30% left. 

Using only the icon without the percentage indicator is problematic. 

![Battery Notifications](https://user-images.githubusercontent.com/66426048/103476810-ba4b1e00-4db0-11eb-8e53-424ba71769d0.png)

ItzSwirlz:
Here's what it really comes down to-we need a system-icon for certain points of the battery percentage.

The Adwaita Icon set has in /usr/share/icons/Adwaita/scalable/status-and that's PERFECT as it has battery level 0, 10, 20, 30, 40 , 50, all of them through 10's :D (and it's not in the legacy folder which has battery stuff so this is probably kept)

I'll see if I can put on my maintainer hat and open a PR

ItzSwirlz:
So far, after leaving my PC unplugged for about an hour on Mint 20.1 Beta, it seems somewhat accurate but not exactly perfect. We will see what happens soon, but I'll begin working on a PR I guess.

![Screenshot from 2021-01-03 17-50-49](https://user-images.githubusercontent.com/24401303/103490802-4b8bb600-4dec-11eb-8c3d-2d3ef4079ad5.png)

ItzSwirlz:
Aha! I found the problem.

@mtwebster 

At line 503 if statement:

```
  } else if (percentage > 5) {
                this._applet_icon.set_style_class_name('system-status-icon warning');
```

This applies to **any** icon-the above won't work, what you need to do is

```
                this.set_applet_icon_symbolic_name('battery-level-20');
```

Now that wouldn't be the correct symbolic icon but you should use set_applet_icon_symbolic_na,e. I guess it also makes sense to keep the style class name, so I will just reword it into a switch case and change it. Will have to test a bit first.

JosephMcc:
The code you are talking about does exactly what it's meant to. You should not be setting the icon that is being used there.

ItzSwirlz:
> The code you are talking about does exactly what it's meant to. You should not be setting the icon that is being used there.

[...]

PR #9825, opened 2021-01-04
Power applet: Use specific icons depending on percentage (#9823)

ItzSwirlz:
There is a few other ideas I have but I'm holding them off until after Mint 20.1/Cinnamon 4.10 or 5.0, whatever it's versioned.


ItzSwirlz:
Closing due to uncertainty with some stuff

PR #9858, opened 2021-01-14
Grouped-Window-List: Fix changes not saving after dragging a launcher

ronyala:
Fixed a problem where if you finish a launcher darg move outside of the applet area, the changes are not saved.

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -162,7 +162,6 @@ class AppGroup {
         this.signals.connect(this.actor, 'button-press-event', (...args) => this.onAppButtonPress(...args));
         this.signals.connect(this._draggable, 'drag-begin', (...args) => this.onDragBegin(...args));
         this.signals.connect(this._draggable, 'drag-cancelled', (...args) => this.onDragCancelled(...args));
-        this.signals.connect(this._draggable, 'drag-end', (...args) => this.onDragEnd(...args));
 
         this.calcWindowNumber();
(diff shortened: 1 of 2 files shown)


mtwebster:
Hi, can you tell me how to reproduce this? I seem unable to, and there's a lot of code changing here so I'm a bit nervous not being sure about it.

Could it have been affected by the other PR I merged yesterday?

Thanks

ronyala:
Hi,

The original problem I found was that if you drop a launcher/app outside of the applet area, the `acceptDrop` function wasn't called and the change wasn't saved.

You can reproduce this by dragging and dropping an app of GWL outside of GWL's area.

This wasn't a problem previously since the app position didn't change in real-time (the DND place holder wasn't destroyed after drag, but it was invisible anyway). Now that app positions change during the dragging itself, it made it _seem_ as if the new position was saved, when it actually wasn't.


I made a workaround by using the `onDragCancelled` function in `appGroup.js`, as it is called only when an app is dropped outside of the applet's area.

So now there are essentially three different 'events' that can happen when an app drag is finished:
1) accepting a new app via `acceptDrop` in `applet.js`
2) saving an app move via `acceptDrop` in `applet.js`
3) saving an app move via `onDragCancelled` in `appGroup.js`

A lot of the changes are moving stuff around and creating some more functions. Seemed simple to fix but ended up being more complex.

Oh, and the `misc cleanup` part of the commit is because I moved the dragging related variables to `.dragging`, and I made a _slight_ change in menu behavior when dragging (when they're closed).

`Could it have been affected by the other PR I merged yesterday?` I don't think so if you mean my other PR; It turned out to be a completely different issue.

mtwebster:
Ok. I missed the most obvious point 'dragging a **launcher**`

PR #9904, opened 2021-02-01
[WIP] xapp applet: add multi-row support

dcolascione:
This PR adds a multi-row layout feature to the xapp notification applet, allowing for more efficient use of space in panels. The PR additionally makes a few internal layout details of the xapp applet (e.g. padding, inter-row spacing) theme-able via CSS.


mtwebster:
Hi, this works fairly well - there are a few things I want to comment on:

- Things like row spacing don't appear to have any effect, depending on how much panel height there is to work on - it seems like it will sacrifice spacing/margin to maintain icon size. Of course this is also affected by how our icon sizes are handled (and handled differently whether they're symbolic or not.
- I'm not sure about laying out by column, though I suppose this is arbitrary. I don't think this warrants an additional setting, just a bit of consideration to figure out what's best.
- Maybe additional style classes can be used to apply to single and multi-row? I'm not crazy about how much gap is between each icon in multi-row.
- There are a lot of missing semicolons in your code (to be fair I noticed some existing code is missing it also) - you must use python a lot :)

Thanks!

mtwebster:

PR #9923, opened 2021-02-12
Add a memory limit mechanism to Cinnamon

clefebvre:
Check the memory used by Cinnamon at regular intervals. Restart Cinnamon when the memory reaches an unacceptable limit.

Rationale: Some memory leaks are hard to detect or identify (whether that's because they only happen on certain hardware or in certain niche use cases or whether that's because they're very small and have an imperceptible effect when monitored during a short period of time). Sometimes we don't even know if issues reporting leaks we can't reproduce are still relevant. Yet people still sometimes do report abnormal RAM usage, particularly as the result of using 3rd party spices or after leaving the computer idle for a long period of time. This PR provides a workaround by monitoring the RAM usage and restarting Cinnamon if it reaches a certain threshold.

image

Using the General section of the System Settings, the memory limit can be enabled/disabled. It's value can be configured and so can the frequency of the memory checks.

Cinnamon restarts are logged in ~/.cinnamon/restarts.log.

### DIFF
--- a/data/org.cinnamon.gschema.xml
+++ b/data/org.cinnamon.gschema.xml
@@ -589,6 +589,7 @@
     <child name="keyboard" schema="org.cinnamon.keyboard"/>
     <child name="desklets" schema="org.cinnamon.desklets" />
     <child name="sounds" schema="org.cinnamon.sounds" />
+    <child name="launcher" schema="org.cinnamon.launcher" />
 
     <key name="enable-vfade" type="b">
       <default>true</default>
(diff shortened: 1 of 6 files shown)


ItzSwirlz:
This will sound sort of dumb, but maybe enforce a minimum RAM requirement: We don't want someone accidentally setting their RAM to like 800 MB and Cinnamon acts horribly and an end-user would have no idea how to use it.

clefebvre:
The UI stops you from setting it lower than 1GB.

PR #9935, opened 2021-02-16
windows-quick-list applet: avoid fixed popup size

ikoveshnikov:
Height of the the windows-quick-list applet popup is hardcoded to 400px, this forces ScrollView to show the scroll bar too early. After removing the hardcoded option, the popup will be dynamically resized to show all the opened windows until it will reach screen height, only then scroll bar will be shown.

### DIFF
--- a/files/usr/share/cinnamon/applets/windows-quick-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/windows-quick-list@cinnamon.org/applet.js
@@ -11,7 +11,7 @@ const Clutter = imports.gi.Clutter;
 class WindowMenuItem extends PopupMenu.PopupBaseMenuItem {
     constructor(icon, label, params) {
         super(params);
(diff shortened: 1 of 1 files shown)


ikoveshnikov:
Here is demonstration of changes.

On current master the applet looks sparse and scroll bar appears after 6th window was opened:
Screenshot from 2021-02-17 11-45-33

After the changes all the unnecessary spacing disappeared an the applet auto resized to show all the 8 windows without scroll bar.
Screenshot from 2021-02-17 11-47-55

MinePro120:
Works like a charm!

mtwebster:
Thanks, can you do the same with the favorites applet?

ikoveshnikov:
Thank you for merging the fix! 

>  can you do the same with the favorites applet?

@mtwebster Sure! #9943 - I made a fix and checked it with a really long list of favourites.

PR #9939, opened 2021-02-18
cinnamon-xlet-makepot: ignore keywords if they're used as settings keys 

collinss:
This prevents the script from failing if one of our translated keywords in settings-schema.json is used as the key name for a settings definition. It works by skipping the check for keywords on the root node (1st run of the function extract_settings_strings()). Also display a warning and return if we attempt to save a value that is not a string so that the user of the script knows there's something wrong with the input file.


mtwebster:
Can you rebase please? Thanks

PR #9943, opened 2021-02-19
favorites applet: remove extra padding and avoid setting constant height

ikoveshnikov:
Same change as recent to windows-quick-list applet: the widget uses scroll view, so
it can grow to fill all available space by vertical direction, and then a scroll bar is shown.
Extra padding between elements was removed to keep the list more clean, it uses the 
same padding as menu applet now.


mtwebster:
Cool, thank you again.

PR #9947, opened 2021-02-20
Pimp the sound@cinnamon.org applet tooltip

sphh:
Currently the tooltip of the sound@cinnamon.org applet only shows the sound volume. Often I just want to check quickly, which track is playing. At the moment I have to click onto the icon to reveal this information. I don't want to have this information shown permanently in the panel, because it takes a lot of space.

This adds the following additional information to the tooltip:
- The player and its state.
- The artist.
- The title.

It is possible to configure what is actually added.

### DIFF
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -894,6 +894,11 @@ class CinnamonSoundApplet extends Applet.TextIconApplet {
         
         this.settings.bind("keyOpen", "keyOpen", this._setKeybinding);
 
+        this.settings.bind("tooltipShowVolume", "tooltipShowVolume", this.on_settings_changed);
+        this.settings.bind("tooltipShowPlayer", "tooltipShowPlayer", this.on_settings_changed);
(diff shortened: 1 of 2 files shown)


sphh:
I used the two "icons" 🎩 and 🎶 as a header for the artist and the song title. Please feel free to adjust them to your liking!!

mtwebster:
Hi this seems to work fine, but a few issues:

- Please change 'pimp' to something else in this PR title as well as the commit message. I don't personally have any issue with the word but github is public and for all ages.
- I don't think we need icons to be honest - the artist and track should be obvious enough I think.
- I was messing around with multiple players, and it seems to get confused sometimes and revert back to simply showing the volume.

Thanks

sphh:
Thank you for your comments.

* With "pimp" I had "to pimp" in mind, which I believe is not dingy at all. But your are right, the noun should not be used. The next expression which came to my mind, was "to beef up", which we also cannot use. I'll go for "add more information to the tooltip". (Now I just have to figure out, how to change the name of a branch and the commit messages …)
* I have no problems whatsoever to remove the icons.
* Do you have any idea, where that might come from? I just tap into the function, which changes the player in the drop-down dialog, which opens, when clicking onto the icon. Does the player gets confused there as well?

PR #9954, opened 2021-02-24
Add additional information to the sound@cinnamon.org applet tooltip

sphh:
Currently the tooltip of the sound@cinnamon.org applet only shows the sound volume. Often I just want to check quickly, which track is playing. At the moment I have to click onto the icon to reveal this information. I don't want to have this information shown permanently in the panel, because it takes a lot of space.

This adds the following additional information to the tooltip:

    The player and its state.
    The artist.
    The title.

It is possible to configure what is actually added.

@mtwebster

### DIFF
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -1253,17 +1253,10 @@ class CinnamonSoundApplet extends Applet.TextIconApplet {
                 tooltips.push(this.player._name + " - " + _(this.player._playerStatus));
             }
             if (this.tooltipShowArtist) {
-<<<<<<< HEAD
                 tooltips.push(this.player._artist);
             }
             if (this.tooltipShowTitle) {
                 tooltips.push(this.player._title);
-=======
-                tooltips.push(this.player._artist);
-            }
-            if (this.tooltipShowTitle) {
-                tooltips.push(this.player._title);
->>>>>>> ff72891b23bc922b79c9357908e26d89d894f480
             }
         }
         if (tooltips.length == 0) {
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -896,8 +896,7 @@ class CinnamonSoundApplet extends Applet.TextIconApplet {
 
         this.settings.bind("tooltipShowVolume", "tooltipShowVolume", this.on_settings_changed);
         this.settings.bind("tooltipShowPlayer", "tooltipShowPlayer", this.on_settings_changed);
-        this.settings.bind("tooltipShowArtist", "tooltipShowArtist", this.on_settings_changed);
-        this.settings.bind("tooltipShowTitle", "tooltipShowTitle", this.on_settings_changed);
+        this.settings.bind("tooltipShowArtistTitle", "tooltipShowArtistTitle", this.on_settings_changed);
 
         this.menuManager = new PopupMenu.PopupMenuManager(this);
         this.menu = new Applet.AppletPopupMenu(this, orientation);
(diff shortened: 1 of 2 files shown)


mtwebster:
Hi, looks good as just text. However, I'd like to reduce the options, there's no need for so many, and I don't want to remove the volume.

There should just be a single additional option to show or not show the track info. I apologize for not noticing this the first time around, I was focused on the other aspects of this pr.

It's reasonably easy to just update this existing pr, so you don't need to go thru the work of creating a new one - this way we can also squash all of these commits into one at the same time.

Hopefully this could be useful in the future as well..

1) Make the changes on your `add-information-to-sound-tooltip` branch.
2) `git commit -a`, then save the commit file (you can just type some garbage for the message, you'll see why in a moment)
3) `git log` should show you now have 5 commits that are your own, on top of existing cinnamon ones.
4) `git reset --soft HEAD~3 && git commit --amend` - this should bring up the editor with the message from your 'Add additional information..' commit. Edit it if you'd like, then save.
5) `git log` should show you just have a single commit now - containing the changes from all commits.
6) `git pull --rebase origin master` - this will temporarily remove your single commit, apply all new commits from the master branch since you originally cloned it to your local drive, then re-apply your commit.
7) `git log` should show your commit, and the next one below it should have a tag 'origin/master' (and maybe more).

Up until this point, *nothing* has been changed on github. If you messed up any of the above steps and are unsure, you can simply remove the cloned cinnamon folder, clone linuxmint/cinnamon again, and re-add your fork remote (or vice-versa).

8) Now, update the PR on github: `git push --force <name-of-your-remote> add-information-to-sound-tooltip`.

`<name-of-your-remote>` will be whatever name you assigned to your fork - do `git remote -v` if you ever want to see the list.

There's actually a more useful command (interactive rebase) that I'd normally use for step 4, but your emoji symbols seem to giving git some trouble so we can't use it.

Let me know if you have trouble.

sphh:
I can certainly remove the option which toggles the volume info. What is the reason for not making its display optional? Personally I do not need this information at all and I also think, it is somewhat redundant.

What about the option to show the player? Should this be incorporated into the new "single additional option" or stay as a separate option?

The reason, why I made separate options for the track title _and_ the artist is as follows: I regularly listen radio stations. If I am not mistaken, all internet radio stations send both the artist and the title in the title tag. The artist tag is always empty, hence I do not want to have an empty line displayed. A possible workaround is, to not display it, if it is empty. This could be done for the title and the artist. What do you think?

Is the procedure you outlined, so that this PR appears as a single (big) commit and not many small ones?

sphh:
I can certainly remove the option which toggles the volume info. What is the reason for not making its display optional? Personally I do not need this information at all and I also think, it is somewhat redundant.

What about the option to show the player? Should this be incorporated into the new "single additional option" or stay as a separate option?

The reason, why I made separate options for the track title _and_ the artist is as follows: I regularly listen radio stations. If I am not mistaken, all internet radio stations send both the artist and the title in the title tag. The artist tag is always empty, hence I do not want to have an empty line displayed. A possible workaround is, to not display it, if it is empty. This could be done for the title and the artist. What do you think?

Is the procedure you outlined, so that this PR appears as a single (big) commit and not many small ones?

sphh:
I can certainly remove the option which toggles the volume info. What is the reason for not making its display optional? Personally I do not need this information at all and I also think, it is somewhat redundant.

What about the option to show the player? Should this be incorporated into the new "single additional option" or stay as a separate option?

The reason, why I made separate options for the track title _and_ the artist is as follows: I regularly listen radio stations. If I am not mistaken, all internet radio stations send both the artist and the title in the title tag. The artist tag is always empty, hence I do not want to have an empty line displayed. A possible workaround is, to not display it, if it is empty. This could be done for the title and the artist. What do you think?

Is the procedure you outlined, so that this PR appears as a single (big) commit and not many small ones?

sphh:
@mtwebster: Have you had a think about my proposition regarding the artist and title? I would really love to cross this off my list. Thanks.

mtwebster:
Hi, I apologize for not getting back to this sooner - 

Thanks for explaining your reasoning a bit more.

I think I'm ok with your suggestion - reduce this to 3 options (if I understand you correctly) -
a) volume
b) player status
c) track/artist, but only showing non-empty info.

What is the reason behind showing the player info btw?

Thanks

sphh:
Yes, you understood me correctly. But option b) is actually "player + status".

Why I included the player and the status? Because the drop-down display also has this included. Simple as that ;-) But also you could have more than one player running and then it might be good to know, to which player the information belongs.

Do you think, I should remove it?

Thinking about it, it would be best to show all active players in the tooltip. But don't worry, that's too much for now!

sphh:
@mtwebster: Done with the latest commit https://github.com/linuxmint/cinnamon/pull/9954/commits/af5a7e965404f5742f454c32b2ad4b2de9138877

Collapsing artist/title into one option and not displaying if empty, works and looks great. (Finally I got rid of this "Unknown artist" field …) Thanks for the idea!

mtwebster:
Looks good now, only three small things - 

- If all tooltip options are disabled, let's just not show any tooltip (I think you can just set the tooltip to an empty string).
- Make the defaults to match the existing behavior (only volume enabled).
- Change 'artist/title' to 'artist and title'.

Thanks!

sphh:
@mtwebster: Done.

sphh:
Thanks, @mtwebster!

mtwebster:
Np, thanks for dealing with my endless demands :)

PR #9956, opened 2021-02-27
Fix cinnamon-xlet-makepot so it always gives the same output on every system

Gr3q:
The issue exists because [os.walk's list elements are in arbitrary order](https://stackoverflow.com/a/18282401). Because of that the resulting .pot file will be in different order on different systems, making reviewing PRs for them unnecessarily hard (and the script's output should be reproducible for it's own merit too).

Sorting os.walk's output in every place in the script fixes this issue.

Relevant issue: https://github.com/linuxmint/cinnamon-spices-applets/issues/3645

### DIFF
--- a/files/usr/bin/cinnamon-xlet-makepot
+++ b/files/usr/bin/cinnamon-xlet-makepot
@@ -45,7 +45,7 @@ def scan_json(dir, pot_path):
 
     for root, dirs, files in os.walk(dir):
         rel_root = os.path.relpath(root)
-        for file in files:
+        for file in sorted(files):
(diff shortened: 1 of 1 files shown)


NikoKrause (commented):
You missed one `for f in files:` line. It in the `remove_empty_folders` and probably not needed, but better safe than sorry.

NikoKrause:
Looks good to me. Thanks for the pull request!

NikoKrause (approved):

mtwebster:
Nevermind..

Issue #8955, opened 2019-11-13
Grouped Window List shows 'ghost' applications


LinuxOnTheDesktop:
```
 * Cinnamon version (cinnamon --version) 4.2.4
   - Please specify if you are using the daily builds PPA (https://launchpad.net/~linuxmint-daily-build-team/+archive/ubuntu/daily-builds). I am not.
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...) Mint 19.2
 * Graphics hardware *and* driver used: Intel UHD Graphics 620 3.0 Mesa 19.0.8
 * 32 or 64 bit: 64
 * Attach /home/<username>/.xsession-errors, or /var/log/syslog: I attach both.
[xsesh.txt](https://github.com/linuxmint/cinnamon/files/3843484/xsesh.txt)
[syslog.txt](https://github.com/linuxmint/cinnamon/files/3843500/syslog.txt)

 ```

**Issue**

Sometimes Grouped Window List (GWL) presents an application as running after that application has closed - or at least seems to have closed. Sometimes the problem  takes this form: (1a) a pinned application (e.g. Thunderbird) continues to show the 'running' indicator (the horizontal bar) even when it is not running. At other times, (1b) an icon for a non-pinned application (e.g. the Update Manager) persists after closing. Sometimes (2) clicking the panel icon produces something with the appearance of a small titlebar. Whether or not 2 obtains, clicking anything - the icon, the titlebar-thing if present - does nothing.

**Steps to reproduce**

The problem is intermittent and I do not know the exact cause. It tends to occur with one application or another at least once in every session tha I use my computer. The problem seems to be new, or at least much more common that it used to be (some two months or so ago).

**Expected behaviour**

Clearly, all of this - icons showing for applications that are neither running nor pinned, pinned applications showing as running when they are not, applications showing as running not responding - just should not happen.

**Other information**

Here are two other outstanding and I believe gross GWL problems: (1) #8578; (2) #8796.

jaszhix2:
Just a reminder, the previous maintainer of GWL (me - @jaszhix) has been banned since June, and have not heard from Clem or anyone part of the organization since then. I did not abandon GWL, it was work I was proud of, and petty, spiteful, politics ruined it for everyone else.

LinuxOnTheDesktop:
[...]

PR #9957, opened 2021-02-27
Cinnamon-settings spices: allow installing 3rd party spice from zip file

collinss:


lestcape:
Thanks @collinss, because even though I really wanted to do this myself, something inside me was holding me back from doing it. I hope this help to install easily non official spices and with that more third party developers will want to create useful spices for the Cinnamon users from their own software development gits.

Maybe you want to also include a warning message to the users  before install an spice, explaining that install it locally may not be so safe as install it from the spices website or something similar. Today maybe that have not sense (because not one is using this new method), but could be something useful in the future to avoid unnecessary inconvenience... Personally, I still cannot see a real difference in security between the extensions of the website of spices and any other, but there are those who believe deeply that there is this difference. If it actually exists that difference, I think it would be useful to make the user notice it or this option could end up disappearing, like how disappeared the freedom that third party developers used to have in Cinnamon some time ago.

ItzSwirlz:
Thing is you are going to have to trust the end user with this-which shouldn't be a big problem. Users shouldn't be downloading from unsafe sources, so it can be a security issue depending on where the spices are hosted.

lestcape:
@ItzSwirlz i disagree with you because, you can not ensure you have security with a simple inspection. So, all source are insecure in one level or another and to me insecure means insecure, the level is a human factor, so should be inconsequential.

Anyway, the focus on my comment was avoid unhappy users claiming things that cannot be resolved from inside the mint repositories.

clefebvre:
> Users shouldn't be downloading from unsafe sources, so it can be a security issue depending on where the spices are hosted.

The computer belongs to the user so that is entirely the user's choice. We should not go against it. Also, I think it's widely known and understood that downloading software outside of the official sources is a risk taken and accepted by the user. So I don't think we should really worry about this.

What we need to worry about is taking responsibility for the sources which are predefined by default. Whether it's a spice or a software application, if it's found by the user on a 3rd party website or repository, it's really none of our business whether it has bugs or even whether is has security issues. When that software comes from preinstalled sources though (cinnamon-spices or our APT repos) then our responsibility is clearly engaged because we give the guarantee of safety to the user and commit for him/her to delegate scrutiny to us. When that spice/software is installed by default, the bar is even higher, it needs to work PERFECTLY.

Now, with all that said, and before I get to the bad news (I'm closing the PR), I'd like to tell @lestcape and other devs that this initiative (as the origin of this PR) was mine. I started this project. I did it for you, and I did it to fulfill a promise I made you the last time we talked about this. I agreed at the time that we should make it easy to install 3rd party spices, and since then I also found out there were bugs and shortcomings associated with just moving files to ~, so these should be fixed as well.

The problem here is integration and UI mostly. We've been working on this for weeks now, about 4 of our devs joined in on the design and we can't still can't manage to get to a point where this looks OK. Some of the devs don't understand why we're spending so much time on this and we're ready to give up on the UI side and switch to something we've already worked on which is the CLI that does this.

`cinnamon-install-spice xlet uuid`

This CLI doesn't just install the spice, it unpacks the translations properly and is compatible with the update system (this is not relevant to 3rd party spices since they have no update source but it's very relevant for spices manually installed as ZIP from the spice website).

We'll close this PR and abandon the UI changes. On the spice website we'll replace the instructions to move files to the home folder advertise the use of this command instead. I recommend 3rd party sources do the same.

Last but not least I am not opposed to a re-implementation of a UI change to accommodate this but it's no longer in our roadmap. If  done again it will need to solve all the problems we bumped into.

I list some of the big issues below:

- exceptions/errors need to be handled and made visible
- a dedicated screen/page should be used to show information about the spice and/or warning messages prior to confirmation
- themes settings are different than other settings, there is no page there to remove manually installed spices, a redesign is needed.

clefebvre:
I'm going to add zip auto-decompression and and spice type auto-detection to the CLI.

lestcape:
I really don't get you.... So, the conclusion is there are not support for third party extension here right?

The mechanism to install extensions from spices is download it to a temporal folder and then make a local installation. So, is clear to me that what need a local installation is just only part of the procedure to install extensions from spice, but some how you said is differently? Is simple, is not differently and have not other new different issue to the already needed procedure.

That is why i only can concluded, you simple don't want to support any third party here and i see this is your personal decision as i don't see any other here than you writing strange reasons without any basis for me. It is in the end your decision what so you want to make, but do not try to convince us that there is something other than your personal decision here.

Don't get it wrong, I also make personal decisions, we all do that. In this case, my personal decision is also not to waste time developing code for an environment that, far from supporting others to contribute, blocks them. Good job!!

clefebvre:
Wow, indeed, you really don't get me.

lestcape:
> Wow, indeed, you really don't get me.

No, really not. It is clear to me that the code that currently exists is far from perfect, but it is not clear to me that this should be specially important for install third party, because it's in general a problem for any installation method. That is why I do not understand that you only cancel the installation of third parties and not the installation in general.

For example merging Gtk stuff in [Spice_Harvester](https://github.com/linuxmint/cinnamon/blob/master/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py#L127) was a wrong decision it should be better a service instead and provide a DBus Server to inform the client (a clutter or Gtk client) about the installation process. For example, you can create or delegate to an applet the responsibility to control exceptions in the server and also, it could work as a replacement to the cinnamon-settings page or better than that, because the settings is not always open when you are installing extension (example CLI) and the applet can be loaded at all times on the desktop connected as a DBus client waiting for an installation to start.

clefebvre:
These issues are specific to 3rd party. The CLI (which is dedicated to 3rd party) fixes all of them.

clefebvre:
> Let me clarify, because although it seems obvious it may not be. Putting a button cannot take weeks and also not needed 4 developer. Also fixing everything that may be wrong and are BIG issues has nothing to do with putting a button.

Anyone can add a button, it takes 5 minutes. Integrating it properly and getting everything sorted takes way longer. If you think you can do better than @collinss then by all means make a PR. If you introduce more issues than you solve then your PR will get closed.

As for me, I'm going to prioritize other things. As I said before, I'm doing this for you. So if you're not happy with it, I'm wasting my time.

lestcape:
> These issues are specific to 3rd party. The CLI (which is dedicated to 3rd party) fixes all of them.

> exceptions/errors need to be handled and made visible

The exception and errors can only occurs when the zip file have a wrong structure, a missing files or other things and can occurs in both scenarios, from spice and local. Anyway, when the creation of the zip file correspond to 3rd party we have a 3rd party bug and you already said: **"it's really none of our business whether it has bugs or even whether is has security issues"**. So, is or is not your business?

> a dedicated screen/page should be used to show information about the spice and/or warning messages prior to confirmation

True, I commented the same: ["Maybe you want to also include a warning message to the users before install an spice"](https://github.com/linuxmint/cinnamon/pull/9957#issuecomment-790605616), but is not need a page to show an info, a message dialog should be enough and is more common to show a message in a message dialog than in a form/page.

> themes settings are different than other settings, there is no page there to remove manually installed spices, a redesign is needed.

Sorry but this is not an specific to 3rd party issue at all. Also i was never consider the themes as something to target, but is nice that you also considered it. Anyway this is a pointless for 3rd party here.

> If you think you can do better than @collinss then by all means make a PR. If you introduce more issues than you solve then your PR will get closed.

This is difficult because apparently i don' t recognize what is a bug at same as you. So, i can not "fixed" what is not a problem to me. What is not broken not need to be fixed.

> I'm doing this for you. So if you're not happy with it, I'm wasting my time.

I sincerely apologize for bringing you to this point. I also apologize for what I am about to say, but, NO, please don't do anything for me!!! I've never done anything for you either, everything I did was for the users and fully depersonalized. Cinnamon should be also depersonalized. If you think that is correct do it, if it has more or lees priority for you do it with your priority and your rules, but don' t do that for me at all. 

Anyway, what you are doing is not at all what i think should be useful for 3rd party, instead what @collinss did and you considered wrong was what i considered useful. People are complex and difficult to understand right? Please note in my comment what was what I originally said and paid specially attention when i said **from a GUI source**:

["What is not so understandable is that in parallel there is no way to install extensions **from a GUI source** other than the official one at a user's decision (for example, a git repository of an external developer)."](https://github.com/linuxmint/cinnamon/issues/8955#issuecomment-559897996)

So, you are working in something that was not what i request. A CLI and nothing is the same to me, because i don' t want something to be used by a 3rd party, i want something to be used by NOVICE USERS to install extensions that was made by a 3rd party.

Yes, what i want is something easy to preserved, also if it occurred to you again to restructure everything in Cinnamon. Something that should not be a big deal to be maintained. That is to prevent than your ideas of changes will affect the work of others 3rd party again and with that the users of this extensions they created. But seem to be you see BIG issues in all places related with 3rd party and you really not want to resolve the situation.

lestcape:
So, please notice this as a summary of everything I said: 

The 3rd party bugs **"it's really none of our business"**, but we cannot merge this pull request because it create issues **"specific to 3rd party"**.

It' s seriously, you need to revisited your own position if you want that someone can understand you. I personally don' t get you at all. I can not help in the current situation  for that reason.

clefebvre:
It's very simple to understand. I explained it already in my first message. We're committed to supporting anything we provide.

We don't provide 3rd party spices, so they don't need to work, they're not "ours". The button itself is ours though, it needs to work fully. If the spice fails to install for any reason it needs to show why. If we let people install manually, we need to also let them remove as well. In @collinss's PR the position of the button is not OK, the design is not OK, the lack of being able to remove 3rd party themes is not OK, and the lack of error handling is not OK. You can decide that it's OK to add BETA or ALPHA features in your projects, that's entirely up to you, but it isn't your place to expect that of us. If we're not happy and consider something half baked, then we don't add it. That's our prerogative.

In regards to who the feature is for, I don't actually think or know if there's a need for this feature at the moment. We're certainly not doing it for Cinnamon users, we're doing it for potential 3rd party developers. Cinnamon users already have access to hundreds of spices through the ecosystem and there is no demand for 3rd party support. Even the bugs I talked about (translations and lack of updates) when using ZIP files are not well known to the public. Telling people to use the CLI instead of moving ZIP content to ~ will fix that, but even that isn't a big problem since users have embedded access to spices already and thus do not require the use of ZIP files in the first place. There is no real demand from the community on these aspects.

In regards to who is doing what and who wants what, you're seeing what you want to see and you've no idea what's actually happening. We worked on this as a team and didn't reach a satisfactory outcome, if you think @collinss really wants this and I really don't, you're completely mistaken, this is just a WIP on something we've been trying to design together for weeks and after which we agreed to move to CLI instead. It was not @collinss who wanted this feature at the start, it was me. This isn't a Stephen idea squashed by me, this is an idea started by me, which many in the team worked on: Stephen, myself and Niko primarily but also Simon, Isaac and recently Darryn. You're just seeing a WIP PR which isn't ready and doesn't solve the entire picture. We all agreed it wasn't working, we had no real demand for it, it required a UI redesign and the CLI was a better approach to this. I can understand you're disappointed since you wanted it to be graphical, but don't go imagining things.

What you can do is continue this PR we're abandoning and put a lot of work in getting it complete (with a redesign of the theme section and full error handling). It's a lot of work but if you manage to pull it off (pun intended), then we'll merge it.

What you can do also is understand that the CLI is pretty good at doing this and that the demand for this is extremely niche. There aren't many 3rd party spices out there. It is obviously a chicken and egg situation but the reality is still that the demand for this is extremely small.

Note that the CLI already existed since Cinnamon 3.8 and that since 3.8 it already fixes the bugs (lack of updates and lack of translations) triggered by manually moving spices into ~. What I was suggesting was to make it even simpler. Right now it takes 4 commands to install something:

```
wget https://cinnamon-spices.linuxmint.com/files/applets/weather@mockturtl.zip?time=1618745817
unzip weather@mockturtl.zip
cinnamon-spice-installer applet weather@mockturl
rm weather@mockturtl.zip
```

The idea is to simplify this into one command:

`cinnamon-spice-installer https://cinnamon-spices.linuxmint.com/files/applets/weather@mockturtl.zip?time=1618745817`

But as I said before, we're not doing this for us. If you don't like it, we are wasting our time. A full UI redesign is too costly for such a small demand, and even a little CLI improvement is not worth it if nobody cares about it. I'm not even sure who else is interested in this, so yes, if you aren't, then we've no reason to work on it at all.

lestcape:
Well, what you are saying in your last comment now makes sense and is more clear to me. Thanks for clarify it because to me was not the same of the previous comments.

Please, no, do not work on anything more personally made for me, that far from walking away makes me sad, as my intention is not resolved a personal problem. Mi intention was show up what is to me a problem and I think i already showed it. Non of you need to agree with me in that, so there is not sense in a discution about it. Maybe I'm the only one that think that way and the users are not losting anything and they are not selecting a different environment with more options for that reason. Thanks to every body involved for the effort anyway.

lestcape:
Just to link to implications: https://www.reddit.com/r/linuxmint/comments/ju4gy7/global_menu_in_linux_mint/gvfsnqy/?utm_source=reddit&utm_medium=web2x&context=3

Secret-chest:
> Some of the devs don't understand why we're spending so much time on this and we're ready to give up on the UI side and switch to something we've already worked on which is the CLI that does this.\n\ncinnamon-install-spice xlet uuid\n\nThis CLI doesn't just install the spice, it unpacks the translations properly and is compatible with the update system (this is not relevant to 3rd party spices since they have no update source but it's very relevant for spices manually installed as ZIP from the spice website).

Oh, nice! I found an emoji picker and wanted to install it

PR #9962, opened 2021-03-02
Added suspend to critical battery actions

tdeeb:
This PR adds suspend as a critical battery action. Useful especially after configuring the critical battery percentage in dconf.


mtwebster:
Hi - this is probably not a good idea. Most users don't know how to change these unexposed percentages in dconf, and I wouldn't want them to think suspending at 3% battery is a smart thing to do.

For users like yourself that change the critical threshold, there's no reason you can't manually set the 'critical-battery-action' setting in dconf to 'suspend' as well - it simply won't display in cinnamon settings (the combo box will be blank).

PR #9966, opened 2021-03-03
cinnamon-settings-users: Don't hold when user creates an existant user.

### LINKED ISSUE #9494: New user dialog (cinnamon-settings-users) hangs if the username already exists
```
 * 4.6.*
 * Mint 20
 * 64 bit
```

**Issue**
Attempt to create a new user.  If you use an existing name and try to add it, the dialog hangs, you must close it with X and open a new one.

```


  File "/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py", line 884, in on_user_addition
    new_user = self.accountService.create_user(username, fullname, account_type)
gi.repository.GLib.Error: act_user_manager_error: GDBus.Error:org.freedesktop.Accounts.Error.UserExists: A user with name 'mtwebster' already exists (1)
Error in sys.excepthook:
  File "/usr/lib/python3/dist-packages/apport_python_hook.py", line 144, in apport_excepthook
    with os.fdopen(os.open(pr_filename,
FileNotFoundError: [Errno 2] No such file or directory: '/var/crash/_usr_share_cinnamon_cinnamon-settings-users_cinnamon-settings-users.py.0.crash'

Original exception was:
  File "/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py", line 884, in on_user_addition
    new_user = self.accountService.create_user(username, fullname, account_type)
gi.repository.GLib.Error: act_user_manager_error: GDBus.Error:org.freedesktop.Accounts.Error.UserExists: A user with name 'mtwebster' already exists (1)
```

A couple of options:
- Use self.accountService.get_user() to see if the name exists already
- Try to create_user() and catch the UserExists error.

In both cases you need to allow the user to try again, which means instead of using dialog.run() and blocking for a response code, we just show() the dialog, connect to the 'response' signal, and check for uniqueness before closing the dialog.  Some sort of error message would be appropriate also (we can handle more than just this error if necessary, we need to see what other ways it can be broken).

ItzSwirlz:
This catches the AccountsService error for USER_EXISTS. Everything has 
to be moved to try/except because without it, it can lead to some 
unstability.

Fixes #9494

Extra Notes: For some reason using in or == doesn't work but __eq__ does. Making a new dialog can cause some unstability, and the *whole* try/except has to be moved because we don't want an undefined reference with the new_user variable. Believe me I got frustrated trying this.

I know this isn't the *best* but it's something for systems that want to make a Stable Release Update for this fix. Let me know if you want me to change anything

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py
+++ b/files/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py
@@ -392,6 +392,15 @@ class NewUserDialog(Gtk.Dialog):
         except Exception as detail:
             print(detail)
 
+    def user_exists(self, user_name):
+        users = AccountsService.UserManager.get_default().list_users()
+
+        for user in users:
+            if user.get_user_name() == user_name:
+                return True
+
+        return False
+
     def _on_info_changed(self, widget):
         fullname = self.realname_entry.get_text()
         username = self.username_entry.get_text()
@@ -399,6 +408,10 @@ class NewUserDialog(Gtk.Dialog):
         if re.search('[^a-z0-9_.-]', username):
             self.username_entry.set_icon_from_icon_name(Gtk.EntryIconPosition.SECONDARY, "dialog-warning-symbolic")
             self.username_entry.set_icon_tooltip_text(Gtk.EntryIconPosition.SECONDARY, _("Invalid username"))
+            valid = False
+        elif self.user_exists(username):
+            self.username_entry.set_icon_from_icon_name(Gtk.EntryIconPosition.SECONDARY, "dialog-warning-symbolic")
+            self.username_entry.set_icon_tooltip_text(Gtk.EntryIconPosition.SECONDARY, _("A user with the name '" + username + "' already exists."))
             valid = False
         else:
             self.username_entry.set_icon_from_icon_name(Gtk.EntryIconPosition.SECONDARY, None)


mtwebster:
How did you test this?

- You removed the executable bits from this file, so it no longer even runs either from a terminal or the menu (it runs `pkexec cinnamon-settings-users.py` so it must be executable).
- I'm not sure why you have `while True`?  This ends up running the try/except a second time if the username is actually unique (because that ends with `continue`) and you get a duplicate username error the second time.
- this is probably better off being checked as the user is actually typing the username (in `on_info_changed`) - this will prevent them from even attempting to create a new user until the name is unique.

Btw, for GLib.Errors btw, have a look here:
https://lazka.github.io/pgi-docs/#GLib-2.0/classes/Error.html#GLib.Error

You should do `if err.code == AccountsService.UserManagerError.USER_EXISTS:` instead there (I would avoid using .eq).

ItzSwirlz:
Had no idea about removing the executable bit. Because it was in root I tested with python3 cinnamon-settings-users.py

Iirc isn't while true the only way to use a try and except?

Yes, an on_info_changed would be better. Marked as WIP

ItzSwirlz:
I did your suggestion- here is the only downside: It causes some spam in the logs.

Screenshot from 2021-03-09 14-30-55

The NoneType thing I'm unclear of.. when running it in python3 console it normally doesn't AttributeError about it.

Then the log spam: It basically does that if the user doesn't exist and can be created. The only good thing I can see out if it is that when the user is creating the new user, this is a way of the log telling us the user is changing the information.

Thoughts?

ItzSwirlz:
AccountsService with Python is quite annoying. Not a lot of documentation; anyways list_users would return the object, and you can't use an in with it by saying elif username in list_users. The best you can do also since this is a different class is manually define a AccountsService.UserManager.get_default().

mtwebster:
https://lazka.github.io/pgi-docs/#AccountsService-1.0/classes/User.html#AccountsService.User

You should bookmark that site, it has all C libraries we commonly use and their python syntax.

This can probably be done in one line but I prefer readable code:
```
diff --git a/files/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py b/files/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py
index face6b925..82368e3ca 100755
--- a/files/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py
+++ b/files/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py
@@ -392,6 +392,15 @@ class NewUserDialog(Gtk.Dialog):
         except Exception as detail:
             print(detail)
 
+    def user_exists(self, user_name):
+        users = AccountsService.UserManager.get_default().list_users()
+
+        for user in users:
+            if user.get_user_name() == user_name:
+                return True
+
+        return False
+
     def _on_info_changed(self, widget):
         fullname = self.realname_entry.get_text()
         username = self.username_entry.get_text()
@@ -400,15 +409,15 @@ class NewUserDialog(Gtk.Dialog):
             self.username_entry.set_icon_from_icon_name(Gtk.EntryIconPosition.SECONDARY, "dialog-warning-symbolic")
             self.username_entry.set_icon_tooltip_text(Gtk.EntryIconPosition.SECONDARY, _("Invalid username"))
             valid = False
+        elif self.user_exists(username):
+            self.username_entry.set_icon_from_icon_name(Gtk.EntryIconPosition.SECONDARY, "dialog-warning-symbolic")
+            self.username_entry.set_icon_tooltip_text(Gtk.EntryIconPosition.SECONDARY, _("A user with the name '" + username + "' already exists."))
+            valid = False
         else:
             self.username_entry.set_icon_from_icon_name(Gtk.EntryIconPosition.SECONDARY, None)
         if username == "" or fullname == "":
             valid = False
         # If the user exists, don't let the user create it.
-        if username == AccountsService.UserManager.get_default().get_user(username).get_user_name():
-            self.username_entry.set_icon_from_icon_name(Gtk.EntryIconPosition.SECONDARY, "dialog-warning-symbolic")
-            self.username_entry.set_icon_tooltip_text(Gtk.EntryIconPosition.SECONDARY, _("A user with the name '" + username + "' already exists."))
-            valid = False
 
         self.set_response_sensitive(Gtk.ResponseType.OK, valid)
 
```

PR #9968, opened 2021-03-04
network applet: Indicate on the panel icon when there is no internet access

### LINKED ISSUE #9927: [Network Manager (network@cinnamon.org)] can not tell by looking at icon if connected to the internet
Mint 19.3
 Cinnamon 4.4.8

**Issue**
There is a gulf of evaluation in the network manager applet. users can not be certain if they are connected to the internet by just looking at the applet icon. The user only finds out their network status after, for example, a webpage fails to load. If the user is informed of the network status through an icon change of the network manager applet, they can take corrective measures sooner.

**Steps to reproduce**
connected to network, network has no internet access but icon remains the same.

**Expected behaviour**

for example in windows this is the icon when connected to the internet

and this is the icon when the user is not connected to the internet

mtwebster:
Perform a periodic connectivity check to indicate
on the panel when there is no internet (global) access.

Due to a bug in libnm, this uses a dbus to communicate directly
with org.freedesktop.NetworkManager for libnm versions before
1.28.0. For newer versions, the NM.Client api is used.

Since the check can take some time - potentially more than the
periodic update interval, we don't start a new check before
receiving a reply from the previous one.

fixes #9927

### DIFF
--- a/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
@@ -1,5 +1,6 @@
 const Applet = imports.ui.applet;
 const GLib = imports.gi.GLib;
+const Gio = imports.gi.Gio;
 const GObject = imports.gi.GObject;
 const Lang = imports.lang;
 const NM = imports.gi.NM;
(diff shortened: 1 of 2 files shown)


ItzSwirlz:
Screenshot from 2021-03-04 13-29-29
well, I guess this is close enough?

By the way @mtwebster when taking a screenshot, no matter where the white flash won't show on the panel, it will sort of go under it.

PR #9975, opened 2021-03-09
messageTray.js: prevent absurdly long notifications

kilbouri:
First off, I'm new here so I appologize if this is improper format or a request to the wrong branch. Anyways, this PR...

Truncates messages in the setMarkup function to prevent super-long notifications being sent to the user. If the length of the text is over MAX_NOTIF_LEN (default, 225) characters long, then it will be truncated with substring and have "..." appended to show the message has been shrunk.

### DIFF
--- a/js/ui/messageTray.js
+++ b/js/ui/messageTray.js
@@ -32,6 +32,8 @@ var LONGER_HIDE_TIMEOUT = 0.6;
 const NOTIFICATION_IMAGE_SIZE = 125;
 const NOTIFICATION_IMAGE_OPACITY = 230; // 0 - 255
 
+const MAX_NOTIF_LEN = 225;
+
 var State = {
     HIDDEN:  0,
     SHOWING: 1,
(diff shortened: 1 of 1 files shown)


brownsr:
The notifications allow a long description.  I think we'd want to be sure that a facility we deliberately allow is not being affected.  Under the notification settings (system settings, notifications) there is the facility to send a test notification.  Could you screen shot a before and after (with and without your PR) to demonstrate that there is nothing here that will truncate the test notification description ?

Regards

Simon

kilbouri:
The entire point of this PR was to prevent extreme length notifications. If that is intended behaviour and by design, then my PR is not a good fit for it. However, in my personal opinion (and several others, having talked to other Cinnamon users), a complete lack of restriction on length can be annoying when apps decide to send thousand-character notifications (ie, Discord, which allows up to 2000 characters).

As requested, here is a before-and-after with my changes.
**Before**:
image

**After**:
image

As you can see, the test notification is shortened, but the overall functionality is not harmed. Further, as mentioned in the initial comment, if this default of 225 characters is too short, then its simple enough to increase it, just by changing the added MAX_NOTIF_LEN variable in messageTray.js.

icarter09:
This should be a setting that can be toggled off/on, instead if of a hard coded solution.

NikoKrause:
What would be even better than making it configurable, would be to have a "Read more" button as done on many websites:

pic

kilbouri:
> This should be a setting that can be toggled off/on, instead if of a hard coded solution.

I'll definitely look into it, thanks for the feedback!

PR #10018, opened 2021-03-29
cinnamon-settings: Add standalone module for software-properties-gtk

ItzSwirlz:
-------
(No extended commit description)

It's in Ubuntu (come on, let's not bias against distros here) and can actually technically probably be found in others.

If you want I can push over a little bit to make the code more nicer.

I was thinking of having a specific Ubuntu module for Livepatch but that's something for another day and I need to think about it with other Ubuntu devs first (in things like MATE)


mtwebster:
Software-properties-gtk is a symlink to mintsources in Mint - under no circumstances do we want that installable there - plus this would give us two entries. You can patch this easily enough with quilt or something.

Cinnamon is *first and foremost* intended for Linux Mint. We absolutely support other distros and will go out of our way to make sure Cinnamon works everywhere, but unless it's some mechanism that's inextricably part of cinnamon that needs to have different behavior depending on where it's running (such as systemd/consolekit stuff), this type of detail ("distro glue") is up to maintainers to handle. The same with any other distro-specific things.

PR #10028, opened 2021-04-04
Implement optional "switch categories by clicking" option in menu applet 

Lurux:
This PR implements an optional "switch categories by clicking" option for the default menu applet in Cinnamon.

The option is now available as a toggle in the applet's settings, which have been rejigged a little bit to make the most important options more prominent.

This PR also adds an icon for the "all applications" category.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -908,7 +908,7 @@ class FavoriteButton extends SimpleMenuItem {
 
 class CategoryButton extends SimpleMenuItem {
     constructor(applet, categoryId, label, icon) {
-        super(applet, { name: label || _("All Applications"),
+        super(applet, { name: label,
                         type: 'category',
                         styleClass: 'menu-category-button',
                         categoryId: categoryId });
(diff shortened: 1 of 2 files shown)


Lurux:
Uhm also, I don't know how localization works here. Do I need to do anything on my end for translations to be added ?

Lurux:
**Update:** added icon for "all applications" category, squashed the PR into two commits to clean things up.

It seems that most PRs here get answered in less time than this, do I need to do anything special for this to be reviewed ?

icarter09:
There's nothing special that needs to be done. The PR will be reviewed in due time.

Lurux:
I understand, thank you :slightly_smiling_face:

mtwebster:
Hi, sorry for the delay - the category switching works well. But can you remove the All Applications icon? I don't think it really works here - I'd rather just leave it without an icon.

Thanks

Lurux:
@mtwebster - I mean, I sure can do that; I just always found a bit awkward that the "all applications" category doesn't have a dedicated icon. Tbh I think the menu feels much better with it :sweat_smile:

Oh, I realize I also enabled menu animations by default to bring it in line with the rest of the applets. Should I revert that too ?

clefebvre:
Yes please. Stick to 1 thing per PR to make it easier to merge/review/accept.

Lurux:
Ok, I understand. I'll go ahead and change that, and merge it into a single commit to keep things clean.

Should I open a separate PR to add the icon and animation by default, or is this a straight "no" for both of these ?

mtwebster:
Leave the animation off as well (mainly to keep from changing existing behavior on people - it's disabled to improve opening speed).

You can open a pr for the icon, though I'm personally against it (to me this is because it stands out compared to actual categories) but there can be more discussion I suppose.

Lurux:
I did the changes as requested !

I can spin up a live usb to test the latest changes, though everything should be good :slightly_smiling_face:

Lurux:
Yahoo ! :tada:

I have a question for next PR: should I stay on the same branch to implement the proposal PR to add an "all applications" icon, or should I create a new one ?

Anyways, expect to see me again around here ! :slightly_smiling_face:

PR #10029, opened 2021-04-04
Add setting to disable Q key in Alt-Tab switcher

### LINKED ISSUE #9787: Disable Alt+Tab Q feature
```
 * Cinnamon 4.6.7
 * Distribution - proprietary
 * 64 bit
 ```

**Issue**

Windows are closed by accident when I use the Alt+Tab window switcher, specifically when pressing Q while it's open. This problem is worsened by the fact that Q is so close to Tab on the keyboard, and also because I use Alt+Q itself as a shortcut in Chrome to toggle between tabs (so my brain decides very late whether it wants to hit Tab or Q, and thus the finger often lands in between).

Users have asked about this issue before:
* https://forums.linuxmint.com/viewtopic.php?t=284426
* https://www.reddit.com/r/linuxquestions/comments/fsssw9/disable_alt_tab_q_keybind_for_cinnamon/

This matters because closing windows can mean data loss or other kinds of pain. E.g. Chrome windows can have many many tabs open and there's no easy way to make Chrome ask before closing a window.

**Steps to reproduce**

* Open at least two windows.
* Press and hold Alt.
* Press and release Tab. The window switcher will open and select the "other" window.
* Press Q. The selected window is closed.

While this behavior is "as implemented", it's not intended by the user when they press `Alt+Tab+Q` or `Alt+Tab Q` very quickly. In particular this happens when hitting the Tab and Q buttons with one finger around the same time by accident.

**Expected behaviour**

Either nothing happens when Q is pressed (not sure how many users even use this as a feature to close windows on purpose), or the behavior should be configurable. I can think of two ways to configure this:

1. Add "Q" or "Alt+Q" as the default (but removable) shortcut in the "Keyboard" settings.
2. In the "Windows" settings on the tab "Alt-Tab" add a boolean to disable the Q=close behavior.

### PR DESCRIPTION
**Pull request by Philipp91**
Fixes #9787

The new setting allows disabling the feature in the Alt-Tab switcher which closes the currently selected window.

Such a setting is only added for this key/action (and not the other actions `_keyPressEvent()`) because it's the only destructive action.

### DIFF
--- a/data/org.cinnamon.gschema.xml
+++ b/data/org.cinnamon.gschema.xml
@@ -513,6 +513,11 @@
       <default>false</default>
       <summary>Show all windows from all workspaces</summary>
     </key>
+    
+    <key type="b" name="alttab-close-with-q">
+      <default>true</default>
+      <summary>Close the selected window with Q</summary>
+    </key>
 
     <key name="bring-windows-to-current-workspace" type="b">
       <default>false</default>
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_windows.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_windows.py
@@ -149,3 +149,6 @@ class Module:
 
             widget = GSettingsSwitch(_("Show windows from all workspaces"), "org.cinnamon", "alttab-switcher-show-all-workspaces")
             settings.add_row(widget)
+            
+            widget = GSettingsSwitch(_("Close the selected window with Q"), "org.cinnamon", "alttab-close-with-q")
+            settings.add_row(widget)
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by Philipp91** [comment]
How can I test this? I tried bringing up a Linux Mint instance on DistroTest, but executing `./test` there fails with `cp: cannot stat 'debian/tmp/user/share/cinnamon/js': No such file or directory`

**Comment by mtwebster** [comment]
I think the test script assumes you build cinnamon as a debian package:

```
apt build-dep cinnamon # you'll need to enable 'source repositories' in the software sources program
dpkg-buildpackage
```
Then I think the script will work.

We had a quick chat about this shortcut, and we'd be ok with just deleting it altogether - just delete the whole Key_Q/Key_q case there. I'm pretty sure none of us knew it existed until you brought this up - someone snuck it in years ago.

In this case, if you're just editing appSwitcher.js, you could simply copy that single file to /usr/share/cinnamon/js/ui and restart cinnamon to test it.

Philipp91:
Okay. Yes, just removing it definitely works. That's how I patched it in my machine and I'm happy with it already -- I'm mostly sending it as a PR to prevent back-sliding when Cinnamon is updated.

So this is superseded by #10031.

PR #10031, opened 2021-04-05
Remove window-close action on Alt-Tab Q shortcut

### LINKED ISSUE #9787: Disable Alt+Tab Q feature

Philipp91:
Fixes #9787

Closing windows like that (especially with the Q key, which is dangerously close to the Tab key that the user often intends to press) is a destructive and not undoable action, which makes for a bad user experience when it happens. As per https://github.com/linuxmint/cinnamon/pull/10029#issuecomment-813109829 it seems like the feature might have low enough usage that removing it is better than maintaining it and having a separate configuration option.

### DIFF
--- a/js/ui/appSwitcher/appSwitcher.js
+++ b/js/ui/appSwitcher/appSwitcher.js
@@ -9,7 +9,6 @@ const Mainloop = imports.mainloop;
 const Main = imports.ui.main;
 const Cinnamon = imports.gi.Cinnamon;
 
-const CHECK_DESTROYED_TIMEOUT = 100;
 const DISABLE_HOVER_TIMEOUT = 500; // milliseconds
 
 function sortWindowsByUserTime(win1, win2) {
@@ -107,7 +106,6 @@ AppSwitcher.prototype = {
         this._haveModal = false;
         this._destroyed = false;
         this._motionTimeoutId = 0;
-        this._checkDestroyedTimeoutId = 0;
         this._currentIndex = this._windows.indexOf(global.display.focus_window);
         if (this._currentIndex < 0) {
             this._currentIndex = 0;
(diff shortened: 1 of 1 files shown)


Philipp91:
I think I should also clean up `CHECK_DESTROYED_TIMEOUT` and `_checkDestroyed`, right?

PR #10039, opened 2021-04-09
Attempt to fix parallel build

t6:
Uncovered while building with samurai [0] but can probably also
happen with ninja eventually.

In file included from ../src/main.c:25:
In file included from ../src/cinnamon-global-private.h:28:
In file included from src/st/st.h:7:
../src/st/st-box-layout-child.h:24:10: fatal error: 'st-enum-types.h' file not found
 #include "st-enum-types.h"
         ^~~~~~~~~~~~~~~~~

[0] https://github.com/michaelforney/samurai


mtwebster:
Thanks, I thought we'd caught all of these.

PR #10043, opened 2021-04-13
[WIP] Stop using GTimeVal/g_get_current_time

ItzSwirlz:
Sources: (expect more)
https://github.com/GNOME/gnome-shell/commit/9b7f228f8ef224c36fd8809942ec78ce071fdbd7

------------------------------
Pause here. I could try getting it through with the cinnamon recorder but I don't really use it and I'm not sure really who does..

It also automatically exports to mp4 for some reason.. if we want to we could use it but we'd have to keep it up with patches.

They switched to a separate D-Bus seperate process which is friendlier, which is written in JavaScript which may be more preferred by Cinnamon maintainers. (I don't like JS much but yeah).

I'd prefer just in case anything big happens to be aligned with GNOME in a way (in case of a security issue or something), so input on rewriting the recorder system or keeping the C recorder we have?


mtwebster:
Why are you removing cinnamon_util_format_date - how is it related to the rest of the pr?? That's needed by a workaround we have in environment.js.

ItzSwirlz:
> Why are you removing cinnamon_util_format_date - how is it related to the rest of the pr?? That's needed by a workaround we have in environment.js.

Didn't know that.

src/cinnamon-recorder.c
```diff
@@ -509,20 +509,12 @@ recorder_draw_buffer_meter (CinnamonRecorder *recorder)
...
 static GstClockTime
 get_wall_time (void)
 {
-  GTimeVal tv;
-
-  g_get_current_time (&tv);
+  GstClock clock;
 
-  return tv.tv_sec * 1000000000LL + tv.tv_usec * 1000LL;
+  return gst_clock_get_time (&clock);
```
ItzSwirlz, line 517: It should just fit with the function type so using this function *should* work

src/cinnamon-recorder.c
```diff
@@ -1263,10 +1255,7 @@ recorder_open_outfile (CinnamonRecorder *recorder)
                   {
                     /* Appends date as YYYYMMDD */
                     GDate date;
-                    GTimeVal now;
-                    g_get_current_time (&now);
                     g_date_clear (&date, 1);
-                    g_date_set_time_val (&date, &now);
```
ItzSwirlz, line 1269: I'm not sure what the point of this is. This is just for YYYY-MM-DD and i don't know what time has to do with anything, it doesn't seem to be used in the final output

ItzSwirlz (commented):
a few things to look at:

src/cinnamon-recorder.c
mtwebster: used here: https://github.com/linuxmint/Cinnamon/blob/master/js/ui/main.js#L208-L211

Using the screen recorder I get a file named `cinnamon-20210531-1.webm`

src/cinnamon-recorder.c
mtwebster: should be ok

ItzSwirlz:
_i can't do crap i just rewrote the function using GstDateTime and other stuff the compiler recommended_

and of course the g_get_real_time cast still makes me want to cry

ItzSwirlz:
Finally done. Apparently it doesn't care if the function is a GstClockTime and it returns a gint64?

src/hotplug-sniffer/hotplug-sniffer.c
```diff
@@ -290,16 +288,15 @@ print_debug (const gchar *format, ...)
...
-  strftime (timebuf, sizeof timebuf, "%H:%M:%S", &broken_down);
+  now = g_date_time_new_now_local ();
+  timestamp = g_date_time_format (now, "%H:%M:%S");
 
   va_start (ap, format);
   s = g_strdup_vprintf (format, ap);
   va_end (ap);
 
-  g_print ("cinnamon-hotplug-sniffer[%d]: %s.%03d: %s\n", pid, timebuf, (gint) (now.tv_usec / 1000), s);
+  g_print ("gnome-shell-hotplug-sniffer[%d]: %s.%03d: %s\n",
```
mtwebster, line 298: cinnamon

ItzSwirlz: Fixed

Issue #10001, opened 2021-03-25
Network Manager applet - wireless network list settings


boughtonp:
When I'm connected to a wireless network, I don't care about other wireless networks I'm not connected to.

The network manager applet currently has no way of preventing this information being output, so I cloned it and added some settings.

First I changed the NUM_VISIBLE_NETWORKS constant (which controls how many networks to show before putting the rest in "More") to a setting (spinbutton) - this allowed a min value of 1 resulting in non-active networks going into the "More" menu instead, which was ok but still not quite what I wanted.

So I also added a boolean setting so that - when a wireless connection is active - no other wireless networks display (not even a more menu), but if the connection isn't active then it displays as normal.

Default settings are same as current behaviour - i.e. anyone who doesn't configure the applet will not see any difference.

Tested on Cinnamon 4.8.6 (Debian Testing) and it works for me, but I expect there's a better way of doing the settings than I used.

Assuming you agree with the idea, should I go ahead and raise a pull request against master, or use a different branch?

icarter09:
@boughtonp create a PR against the master branch. Thanks.

PR #10076, opened 2021-05-07
[WIP] network applet: Add settings for wireless network list.

boughtonp:
As per #10001 this adds a couple of settings in relation to the wireless network list.

The first setting replaces the existing constant of 5 before showing the "More" menu.
The second setting allows hiding the list when connected to a network.

### DIFF
--- a/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
@@ -1700,11 +1700,11 @@ CinnamonNetworkApplet.prototype = {
 
             this.settings = new Settings.AppletSettings(this, metadata.uuid, this.instance_id);
             this.settings.bind("keyOpen", "keyOpen", this._setKeybinding);
-            this.settings.bind("num-visible-networks", "num_visible_networks");
-            this.settings.bind("hide-other-networks-when-connected", "hide_other_networks_when_connected");
-            APPLET_SETTINGS.num_visible_networks = this.num_visible_networks;
-            APPLET_SETTINGS.hide_other_networks_when_connected = this.hide_other_networks_when_connected;
+            this.settings.bind("num-visible-networks", this._setNumVisibleNetworks);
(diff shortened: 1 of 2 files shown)


files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
```diff
@@ -1692,6 +1700,10 @@ CinnamonNetworkApplet.prototype = {
 
             this.settings = new Settings.AppletSettings(this, metadata.uuid, this.instance_id);
             this.settings.bind("keyOpen", "keyOpen", this._setKeybinding);
+            this.settings.bind("num-visible-networks", "num_visible_networks");
+            this.settings.bind("hide-other-networks-when-connected", "hide_other_networks_when_connected");
+            APPLET_SETTINGS.num_visible_networks = this.num_visible_networks;
+            APPLET_SETTINGS.hide_other_networks_when_connected = this.hide_other_networks_when_connected;
```
mtwebster, line 1706: These settings won't ever get updated - you'd need a cinnamon restart to make any changes apply.  I believe you can provide a callback to the .bind calls, then update APPLET_SETTING's values then.

boughtonp:
Ok, I've updated to use settings via callback.

It still seems to need an applet reload for changed settings to take effect - should that be immediate?

mtwebster:
I'm sorry for this being overlooked - I use filters when going thru pr's and I've neglected to check.

I tried this again, and it's still not updating  - 
```
           this.settings.bind("num-visible-networks", this._setNumVisibleNetworks);
           this.settings.bind("hide-other-networks-when-connected", this._setHideOtherNetworksWhenConnected);
```

These still need the  `"num_visible_networks"` and `"hide_other_networks_when_connected"` arguments, but with the callbacks *after* them - see https://github.com/linuxmint/Cinnamon/blob/master/js/ui/settings.js#L344-L355

And to update the actual applet, you would need some method to reconstruct the applet menu, or at least to modify those details that are relevant to the settings, and call this function in those callbacks.

PR #10086, opened 2021-05-14
notifications applet: Use a colorful icon for normal notifications

marmistrz:
With the current design it's still quite easy to miss notifications. Using a yellow dot for normal notifications makes them stand out a little more.

### DIFF
--- a/files/usr/share/cinnamon/applets/notifications@cinnamon.org/icons/normal-notif-symbolic.svg
+++ b/files/usr/share/cinnamon/applets/notifications@cinnamon.org/icons/normal-notif-symbolic.svg
@@ -5,10 +5,34 @@
    xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
    xmlns:svg="http://www.w3.org/2000/svg"
    xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
    height="16"
    width="16"
    version="1.1"
-   id="svg4">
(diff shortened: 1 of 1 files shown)


mtwebster:
Hi this doesn't work - when we add color to symbolic icons like this, we use a "class" property for the element we wish to color.
image
It looks like you tried modifying the alt-critical icon, which was set to 'warning', so the dot was showing up orange.

The only options here are error (red), warning (orange) and success (green) - I think using green might be ok for this

NikoKrause:
Hey,  we had the discussion of coloring the dot when I added those icons. The consensus was to go without colors. See https://github.com/linuxmint/cinnamon/pull/7947#issuecomment-431655194

marmistrz:
Could we make it configurable then? I stopped missing notifications since I've started using a colorful icon

marmistrz:
> Could we somehow make it configurable then?

ping @NikoKrause @clefebvre

AxXxB:
> Could we somehow make it configurable then?

I think that it's posiible to add an option into a notification applet settings. Could you please tell me more about the change you want to see?

marmistrz:
> > Could we somehow make it configurable then?
> 
> I think that it's posiible to add an option into a notification applet settings. Could you please tell me more about the change you want to see?

Yeah, I'm pretty sure it's technically possible to add an option into the applet settings but I don't want to spend time working on a feature that would be rejected by the project maintainers.

I thought about allowing the user to specify the icon name for normal notifications. In essence, we would use variable instead of a hardcoded string in https://github.com/linuxmint/cinnamon/blob/66579ed9fd7d395dfd08be7643aa38ea40420da4/files/usr/share/cinnamon/applets/notifications%40cinnamon.org/applet.js#L182 and initialize this variable based on settings, with `normal-notif` as the default.

PR #10090, opened 2021-05-14
Add icon for "all applications" in menu

Lurux:
The lack of icon for this category button causes that button to be less tall than the other ones, and messes up the category text alignment. It also feels awkward compared to the rest.

This PR adds a category icon, so looking at the Cinnamon menu no longer triggers my OCD. See: *happy perfectionist noises*

Currently the icon is set to be a display, but this could be anything else really.

Any thoughts on this ?


clefebvre:
I agree with the rationale but although it solves the described issue, it creates a new one. The choice of icon can only look good if icon themes have something appropriate.

As it is for instance in our own theme, this looks completely out of place:

image

Lurux:
@clefebvre, I agree that the icon choice is not easy - but, please bear in mind the current icons for "places" and "recent files" don't respect the category icon appearance, either.

I have also though about using the "disks" icon for this, but I thought it would feel weird.

The "other" category icon would have been a great choice, but it's already taken obviously. Maybe we should use a recolored version of it, or change the actual "other" category icon to have a question mark on it ?

Please let me know if you have any other ideas !
Do note I'm ready to work with whoever is in charge of the icon theme, though I'm not super good when it comes to graphics.

<sub>Bonjour de France !</sub>

clefebvre:
How do other menus do?

Marking as next for 5.2.

Lurux:
@clefebvre, do you mean other applets ?

If this is what you mean - Cinnamenu uses the icon called "computer", which in the Min-Y icon theme is the same as the "display" icon.
CinnVIIStarkMenu uses no icon for this category, like the current Cinnamon menu.

What does it mean for a PR to be marked as "next" ?

Lurux:
Anything else I need to do ?

leigh123linux:
Can this be made optional, I don't want a icon for an invalid group.

Lurux:
@leigh123linux, I'm not sure I understand what you mean by "invalid group" ?

Yes, that can be made optional. I will look into adding a setting. The simplest option in the meantime would probably be to delete the icon in applet.js, line 2651

clefebvre:
We can't refer to an icon name which we don't provide since icon themes don't ship with an icon for this abstract/generic category.

We can't provide an icon like it is done here because category icons are typically styled in a particular way in different icon themes and they're visible just below this one, so it will look out of place in any theme which hasn't yet provided an icon for it.

The best way to handle this is to use a symbolic icon in my opinion. "start-here-symbolic" provides the wanted icon and doesn't look the same as categories-other.

Lurux:
I don't understand - you proposed a solution, then closed this immediately. Does this mean this have been implemented elsewhere ?

haggen88:
How about the view-list-symbolic icon?

Secret-chest:
> How about the view-list-symbolic icon?

Nice idea! It would fit better than start-here-symbolic.

PR #10100, opened 2021-05-18
Use spread operator instead of concat

Connormiha:
I applied spread operator where appropriate.


mtwebster:
Hi - 

This seems to help readability a bit and clean things up. I want to warn against all-out trying to 'modernize' our js codebase, though. Most of us aren't javascript programmers - at least not in the typical sense of it.. we like to keep things as readable as possible, and use syntax reasonably similar to the other languages we use - c and python (we prefer to avoid things like promises, for example, and would love to be rid of the few that snuck into our codebase).

Anyhow, thanks!

mtwebster:
sorry.. accidentally closed instead of merging...

PR #10137, opened 2021-06-07
Add basic proxy auth for spices manager

petroid:
A quick and dirty fix to allow spices downloads behind corporate proxy with basic auth.
Credits to https://gist.github.com/beugley/13dd4cba88a19169bcb0


ItzSwirlz:
This seems fine but I am only concerned about the import because of the dependency for b64encode.

Is it built into python, like a standard module? Or is it through a package (besides pip)? Because even on Ubuntu Hirsute, literally Debian's Unstable, basically almost even with the latest releases the latest package is base58.
Screenshot from 2021-06-07 08-22-32

mtwebster:
er.. this is base64 as in character encoding, not a version number. It's part of the standard library.

files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
```diff
@@ -392,7 +393,14 @@ def _url_retrieve(self, url, outfd, reporthook, binary):
             proxy = proxygsettings.get_proxy_settings()
             if proxy and proxy.get('https'):
                 connection = HTTPSConnection(proxy.get('https'), timeout=15)
-                connection.set_tunnel(host)
+                if proxy and proxy.get('http'):
+                    proxy_url = urlparse("//" + proxy.get('http'))
+                    if proxy_url.username and proxy_url.password:
+                        proxy_auth = '%s:%s' % (proxy_url.username, proxy_url.password)
+                        headers = {"Proxy-Authorization": "Basic " + str(b64encode(proxy_auth.encode())).replace("b'", "").replace("'", "")}
```
SparkyBluefang, line 400: Wouldn't this be simpler?
```python
'Basic %s' % (b64encode(proxy_auth.encode()).decode())
```

petroid: Yes, definitely simpler. Python is not my native language ;)
Seems I've made a mess with commits here, sorry.

PR #10144, opened 2021-06-08
Notification Source simplification

itzexor:
This removes a bunch of code related to the notification source's former usage as an actual actor on the panel in the old message tray implementation. We only use the icon generation and source tracking features of this prototype these days, so we don't need any of this wasted CPU time making a new actor for the summary label and such.

Works for me on the notification sources I know of, but could probably use some additional testing before merge just in case. This code was all in play before, just not used...so there is still possibility for breakage if I missed any vestiges.

### DIFF
--- a/js/ui/messageTray.js
+++ b/js/ui/messageTray.js
@@ -620,77 +620,15 @@ Source.prototype = {
     _init: function(title) {
         this.title = title;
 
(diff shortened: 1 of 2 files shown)


clefebvre:
Moving to 5.2.

PR #10164, opened 2021-06-18
Keyboard applet: fix incorrect flag icon sizing when panel is vertical

AxXxB:
If the panel is vertical, the width of the flag icon always equals the width of the panel. These 3 commits fix this behaviour (and make some another small changes).

Thank you a lot for Cinnamon! =)

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
@@ -20,6 +20,7 @@ except Exception as detail:
 
 from http.client import HTTPSConnection
 from urllib.parse import urlparse
+from base64 import b64encode
 
 try:
     import json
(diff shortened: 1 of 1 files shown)


AxXxB:
Oops, I didn't know that my further commits will change PR, dumb-dumb me...

Last three commits doesn't relate to the name of PR, these are some sort of RFCs.

`grouped-window-list applet: Add new setting 'App button icon scaling factor'` - dirty way to add a setting that changes program icon size in Grouped Window List applet.

`notifications applet: Add new setting "Don't show notification count in tray"` - normal way =) to add a setting that always hides notification count (useful when panel is vertical).

Both settings need translations.

mtwebster:
Hi - I appreciate the enthusiasm :) but please keep separate sets of changes as separate PRs. Meaning, this pr should only be about the keyboard applet - I'd prefer to keep the let -> const changes separate also, to keep the real changes isolated (easier to pick out regressions this way).

Open additional prs for:
- the const->let changes   (is this change just for readability improvement? Not that I'm against it, I'm just curious, not being a js guru.
- gwl changes
- notification applet changes

This also allows me to potentially merge the flag fix immediately, if it resolves an issue, whereas the others are improvements, and probably not something I'm going to consider for a few weeks at least (we're in a feature freeze currently for mint 20.2).

Thanks!

AxXxB:
Ok, thanks a lot for feedback!

mtwebster:
sorry - I didn't see your own comment until I submitted mine, or I would have responded to that.

AxXxB:
`the const->let changes (is this change just for readability improvement? Not that I'm against it, I'm just curious, not being a js guru.`

Well, `const` is a constraint that will not let a programmer to try to change its value when there is no need for it. It's a good practice in JS now (and even in C as well as in other languages where it's available). Moreover, JS JIT possibly can make better native code with `const` variable because he know that its type will never change. So i prefer to use `const` anywhere when it's possible, but if it isn't done in Cinnamon, i can rewrite my changes in keyboard applet without `const` with no problem =)

mtwebster:
No, const is ok, what you said was what I more or less figured. We don't mind adopting newer js constructs, but keep in mind we use python and C heavily also, and keeping things readable for people who *don't* use js on a daily basis is important - things like promises come to mind (we have a few places they were added and regret it now because they're impossible to follow).

AxXxB:
Understood, thank you a lot!

PR #10165, opened 2021-06-18
grouped-window-list applet: Add new setting 'App button icon scaling factor'

AxXxB:
This is a dirty way to add a setting that changes program icon size in Grouped Window List applet. Although it's dirty, all works as expected, but it's a little unsupportable.

This setting is useful (at least for me), for example, when panel height is something about 50px, but 48/50px program icons looks (subjectively) too big and 32px - too small.

Needs translation for string "App button icon scaling factor".


clefebvre:
Sorry this is way too niche to be merged. I understand it provides a workaround for a particular case and I appreciate the effort to share it.

The reason we use a combo for icon sizes as opposed to just a scale to let people input any size they want is to avoid blurry icons.

AxXxB:
Understood and fully agree, sorry for taking your time. Thanks a lot!

AxXxB:
Understood and fully agree, sorry for taking your time. Thanks a lot!

PR #10168, opened 2021-06-18
keyboard applet: Remove unneccesary 'if' in offsets calculation

AxXxB:
This is a small change that will allow to fix a bug with wrong layout icon width when the panel is vertical.

Creating a separate PR for it to isolate this change for easing to pick out possible regressions.


AxXxB:
> a small change that will allow to fix a bug with wrong layout icon width when the panel is vertical
Commit with applet fix - https://github.com/AxXxB/cinnamon/commit/a92066440b6b9dca162ac4152cf719e0674735a1

I'm going to make a PR for it when this PR (and https://github.com/linuxmint/cinnamon/pull/10167) will be merged or rejected, because the code in the fix depends of them.

Issue #10172, opened 2021-06-24
depends on policykit-1-gnome, which is unmaintained


smcv:
The Debian maintainers of polkit noticed that the Debian package for Cinnamon 4.x depends on `policykit-1-gnome`, and the same seems to be true in Linux Mint's packaging of Cinnamon 5.x.

PolicyKit-gnome hasn't been used by GNOME since GNOME 3 (around 2010), and hasn't been used by GNOME Flashback since 2015, so it is no longer maintained by GNOME. The upstream project was [officially deprecated 6 years ago](https://gitlab.gnome.org/Archive/policykit-gnome/-/commit/6295fec06e3bf1f87ac97763be5b5158cde31c4d) and it has now been moved to a read-only archived state. This doesn't seem great for something that Cinnamon is using as a security-sensitive desktop component.

Please consider alternatives: either giving Cinnamon a built-in polkit agent similar to the one in GNOME Shell (which appears to have been removed from Cinnamon in 2013), or forking PolicyKit-gnome under the Cinnamon umbrella and taking responsibility for its maintenance and security, or using some other polkit agent.

Fantu:
@mtwebster @clefebvre why built-in polkit agent was removed?

eli-schwartz:
I suppose this must be a problem with xfce4-session too.

```
$ pacman -Sii polkit-gnome| grep "Required By"
Required By     : cinnamon  xfce4-session
```


My only thought on the topic is, it is depressing that gnome continually abandons reusable components in favor of reimplementing features in non-reusable ways inside of gnome-shell or other leaf applications.

smcv:
> I suppose this must be a problem with xfce4-session too

The xfce4-session in Debian doesn't have that dependency, so this could either be a bug in the Debian packaging (not having a dependency that it should) or in the Arch packaging (having a dependency that it doesn't need). I don't use XFCE, so I don't know who is correct here. (I don't use Cinnamon either, but it's reasonably obvious from the packaging and code that Cinnamon does rely on the polkit agent formerly used in GNOME.)

> gnome-shell or other leaf applications

GNOME Shell and its equivalents like Cinnamon aren't really leaf applications - they're the desktop shell that presents UI for desktop-level and system-level things.

eli-schwartz:
> GNOME Shell and its equivalents like Cinnamon aren't really leaf applications - they're the desktop shell that presents UI for desktop-level and system-level things.

I don't see how this technicality affects my statement.

[...]

PR #10177, opened 2021-06-25
[WIP] Bring back polkit authentication agent (and patch it...)

ItzSwirlz:
This was removed in 2.0.8 and left us to rely on policykit-1-gnome by 
other applications such as pkexec, which is unmaintained (last commit 
was in 2013).

Refer to #10172

-----------------------------------------

Ever since it was removed, GNOME patched a bunch of other stuff and crashes that I'll backport here. Then get the JS stuff working

- [X] Bring back the files
- [X] Backport patches (patch JS files)
- [ ] Work the JS UI
- [ ] Why does the build complain about redefining the POLKIT_I_KNOW_API_IS_SUBJECT_TO_CHANGE?
- [X] Final cleanups (debian packaging, pkexec, etc.)


Fantu:
@ItzSwirlz thanks for starting do this, even if probably was better wait other comments about the choice (in #10172).
If you want do a fast test you should remove policykit-1-gnome from debian/control of cinnamon and cinnamon-control-center and remove it from the test system.

ItzSwirlz:
And that leaves us with this ugly masterpiece. This is good-it means automatically pkexec uses our authentication agent. Now it's just the UI tuning. I'm pretty new to this stuff, if anyone wants to take a shot at it, go ahead but I'll do what I can to get it to what it was like previously.
Screenshot from 2021-06-25 17-21-58

ItzSwirlz:
Skipped patch https://github.com/GNOME/gnome-shell/commits/master?after=ee1db9ba0a111ccb340bcf3905891684c5731f38+69&branch=master&path%5B%5D=js&path%5B%5D=ui&path%5B%5D=components&path%5B%5D=polkitAgent.js because I couldn't find us having a this.placeSpinner. Not sure how we do on that so I left it.

Fantu:
@ItzSwirlz keep attention to js, don't show warning/error on build but on runtime, years ago when I used on web programming was also more difficult to debug that many other languages that is used.
Also check better when you porting patch from gnome-shell, especially when many years have passed from fork and many things have changed, all the code should always be checked well, even with patches that apply correctly with cherry-pick or minimal manual changes.
In these weeks I have very little time and I can't help you with this task.

mtwebster:
Honestly I'm not sure we're interested in this.

- what will our fallback mode use? I don't want something that's inaccessible in certain situations.
- one reason we got rid of this originally was we felt the less the cinnamon process had to do, the more stable it would be. This is still true.

If we have no choice but to take up gnome-polkit ourselves, so be it. We'll have to consider it.

ItzSwirlz:
> If we have no choice but to take up gnome-polkit ourselves, so be it

Use mate-polkit then, shall I close this?

Fantu:
@mtwebster @ItzSwirlz I think is better continue to discussion about the choice on https://github.com/linuxmint/cinnamon/issues/10172 so also other peoples subscribed can see these message

ItzSwirlz:
We will be going cross-DE

PR #10178, opened 2021-06-26
Change method of making "Clear list" label bold..

fredcw:
..in menu applet to avoid label sometimes being ellipsized.

Ellipsizing goes away after changing text size but comes back after restarting cinnamon. This seems to fix it.

Changes:
Screenshot1
to:
Screenshot2


NikoKrause (approved):
Looks good to me. Seems to fix the issue. Thanks!

PR #10182, opened 2021-06-27
Option to display non-critical notifications over full-screen apps

Ammako:
#9529 

~~Seems okay here... this is probably not quite ready to merge yet though, because this change requires adding `display-noncritical-in-fullscreen` to `org.cinnamon.desktop.notifications`, and I could not for the life of me figure out how to programmatically add or update this in the cinnamon repo.~~

~~For testing I manually edited `/usr/share/glib-2.0/schemas/org.cinnamon.desktop.notifications.gschema.xml` to add the new key, and recompiled using `glib-compile-schemas`. If you'd like to merge then please make sure to update things on your end so the new key can be properly added.~~

PR #10213, opened 2021-07-14
Option to display non-critical notifications over full-screen apps

Ammako:
#10182

New PR so I can split this into a separate branch instead of using master (not sure if it would have been possible to change the original PR, but this is easier.)

(Also included indentation fixes)

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_notifications.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_notifications.py
@@ -62,6 +62,9 @@ class Module:
         switch = GSettingsSwitch(_("Show notifications on the bottom side of the screen"), "org.cinnamon.desktop.notifications", "bottom-notifications")
         settings.add_reveal_row(switch, "org.cinnamon.desktop.notifications", "display-notifications")
 
+        switch = GSettingsSwitch(_("Show non-critical notifications in fullscreen"), "org.cinnamon.desktop.notifications", "display-noncritical-in-fullscreen")
+        settings.add_reveal_row(switch, "org.cinnamon.desktop.notifications", "display-notifications")
+
         button = Button(_("Display a test notification"), self.send_test)
         settings.add_reveal_row(button, "org.cinnamon.desktop.notifications", "display-notifications")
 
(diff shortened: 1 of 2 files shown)


js/ui/messageTray.js
```diff
@@ -824,15 +824,16 @@ MessageTray.prototype = {
...
-		// Settings
+        // Settings
         this.settings = new Gio.Settings({ schema_id: "org.cinnamon.desktop.notifications" })
-		function setting(self, source, camelCase, dashed) {
-			function updater() { self[camelCase] = source.get_boolean(dashed); }
-			source.connect('changed::'+dashed, updater);
-			updater();
-		}
-		setting(this, this.settings, "_notificationsEnabled", "display-notifications");
+        function setting(self, source, camelCase, dashed) {
```
clefebvre, line 829: Use simple code here. No need for a generic update mechanism and abstract introspection. Do it like it's done for bottom-notifications already.

files/usr/share/cinnamon/cinnamon-settings/modules/cs_notifications.py
```diff
@@ -62,6 +62,9 @@ def on_module_selected(self):
         switch = GSettingsSwitch(_("Show notifications on the bottom side of the screen"), "org.cinnamon.desktop.notifications", "bottom-notifications")
         settings.add_reveal_row(switch, "org.cinnamon.desktop.notifications", "display-notifications")
 
+        switch = GSettingsSwitch(_("Show non-critical notifications in fullscreen"), "org.cinnamon.desktop.notifications", "display-noncritical-in-fullscreen")
```
clefebvre, line 65: This is a bit confusing from a user's point of view. Why non-critical notifications... what about critical ones? Shall we have a simple setting for all notifications instead? "Show notifications in fullscreen", simply.

clefebvre (commented):
While testing, I find the feature buggy. Not the additional code here in this PR, but Cinnamon on master itself.... I play a movie, go fullscreen, issue a notification with notify-send from the other monitor, sometimes I see it, sometimes I don't... it doesn't seem to always care whether the notification is critical or not. It looks like it's more to do with focus and whether previous notifications appeared or not.

js/ui/messageTray.js
Ammako: I have nothing to do with this part of the code. I'm just fixing the indentation there from tabs to spaces.

Issue #10195, opened 2021-07-06
cinnamon-settings Segmentation fault


draganbjedov:
* Cinnamon 5.0.4
   - Using https://launchpad.net/~linuxmint-daily-build-team/+archive/ubuntu/daily-builds
 * Distribution - Ubuntu 20.04
 * NVidia
 * 64 bit

**Issue**
cinnamon-settings crash with Segmentation fault
```
$ cinnamon-settings
Segmentation fault (core dumped)
```
It crashes even if I try to get version
```
$ cinnamon-settings --version
Segmentation fault (core dumped)
```

**Steps to reproduce**
Run from terminal cinnamon-settings

**Expected behaviour**

Cinnamon settings started

**Other information**
It works when started with `sudo` but then some settings are empty.

mtwebster:
There's not much to do without more info - can you provide anything more? A trace or coredump?

draganbjedov:
@mtwebster  Here is the core dump:
[core.tar.gz](https://github.com/linuxmint/cinnamon/files/6775869/core.tar.gz)

sgilissen:
Just to add here that I have the same issue on 2 systems. 
One of them is on Fedora 32, the other on Fedora 33.

This is a core dump from the Fedora 32 machine
[cinnamondump.tar.gz](https://github.com/linuxmint/cinnamon/files/6806599/cinnamondump.tar.gz)

leigh123linux:
@sgilissen I have no f33 installation to process the coredump so it's useless.
Have you been messing around with python/pip?

sgilissen:
> @sgilissen I have no f33 installation to process the coredump so it's useless.
> Have you been messing around with python/pip?

Only in virtualenvs as I'm a Python dev. The core Python install is untouched, on both machines.

sgilissen:
> Can you run cinnamon-settings in a terminal and post the output.

Sure! 

`[13:33:14][sgilissen@chiron] ~:$ cinnamon-settings

(cinnamon-settings.py:56799): Gtk-WARNING **: 13:33:17.406: Theme parsing error: colors.css:71:44: Invalid number for color value

(cinnamon-settings.py:56799): Gtk-WARNING **: 13:33:17.406: Theme parsing error: colors.css:72:44: Invalid number for color value

(cinnamon-settings.py:56799): Gtk-WARNING **: 13:33:17.406: Theme parsing error: colors.css:74:53: Invalid number for color value

[...]

PR #10224, opened 2021-07-19
cinnamon-settings: Remove ~/.local and /usr/local from python's module search paths.

mtwebster:
Ref: #10195


yehudade:
I get removing `~/.local` but why did you remove `/usr/local`? especially given `usrmerge`
https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge/

PR #10260, opened 2021-08-08
Documentation corrections

Odyseus:
- Corrected erroneous example for the layout key on setting-schemas.json files.
- Changed the examples using the `cinnamon-json-makepot` command in favor of using the `cinnamon-xlet-makepot` command. Mainly to avoid confusion since the `--js` CLI parameter has now an inverted meaning as it did before.
- Corrected other typos.

### DIFF
--- a/docs/reference/cinnamon-tutorials/documenting-source.xml
+++ b/docs/reference/cinnamon-tutorials/documenting-source.xml
@@ -145,7 +145,7 @@
         The type of a property specified will be automatically converted to a link to the documentation of that type, if available.
       </listitem>
       <listitem>
-        In the description of itmes, <code>@word</code> will be automatically converted to code format, ie. enclosed within <code>&lt;code&gt;</code> tags.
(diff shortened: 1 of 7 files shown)


fredcw:
@Odyseus Can I ask a question? Where can these documentation files be viewed? where are they complied to?

Odyseus:
Hello, @fredcw.

Those files are the sources for the documentation installed with the `cinnamon-doc` package. Some of those sources are created manually and others are created from source files comments.

For an updated documentation, install the packages `devhelp` and `cinnamon-doc`. Then open **Devhelp** from your applications menu. It not only contains a more up-to-date documentation than the one on-line, but also has documentations about the **St** library and Cinnamon's JavaScript modules.

For documentation about GObject libraries (Clutter, GLib, Gio, etc.), exists [this on-line documentation](https://lazka.github.io/pgi-docs/) that can also be [downloaded for off-line use](https://github.com/lazka/pgi-docs) as a static HTML site, and also can be downloaded as [Devhelp books](https://github.com/pygobject/pgi-docs-devhelp).

PR #10263, opened 2021-08-11
GWL: Fix thumbnail preview issue when window fade-out setting is non-zero

ronyala:
In order to have fade-out animations, multiple preview overlays are needed at the same time.
The existing code did not handle this, instead not creating a new overlay and not showing a window preview if an existing overlay was fading out.

I did not find any existing issues that were reported about this.
I also did some related code cleanup.

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -732,7 +732,7 @@ class AppGroup {
                 } else {
                     this.hoverMenu.open();
                 }
-                if (this.state.overlayPreview) {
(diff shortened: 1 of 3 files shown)


clefebvre:
@ronyala can you describe how to reproduce the issue?

ronyala:
@clefebvre 
1. Set the `window fade-out time` in `grouped-window-list` to a value greater than 0 (Ex. 150ms).
2. Open multiple windows of the same program, it's most apparent with at least 4-5 windows
3. Hide the existing windows behind a different program so it's easier to see (Ex. multiple nemo windows hidden behind a firefox window)
4. Hover over the window thumbnails in gwl, at least half of them will not show the corresponding window (new overlays/previews are not created while there is an existing overlay)

clefebvre:
Perfect, I can reproduce both the issue and the fix.

Issue #10293, opened 2021-08-29
Window list: reordering windows not working correctly with vertical panel


benjamin-owen:
```
 * Cinnamon version (cinnamon --version)
   - Cinnamon 5.0.5
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...)
   - Mint 20.2
 * Graphics hardware *and* driver used
   - NVIDIA GTX 960
   - NVIDIA Proprietary Driver (460.91.03)
 * 32 or 64 bit
   - 64 bit
```

**Issue**
With a vertical panel (mine is 200 pixels wide), dragging windows in the window list to rearrange their order results in the window either jumping to the top or not moving correctly (position window is moved to is not accurate with position of mouse cursor during drag).  I believe this is due to the way Cinnamon checks if the panel is vertical.

**Steps to reproduce**
Make a vertical panel and add the window list applet.  Use dconf-editor to manually make the panel width larger than the 60px maximum (this issue may work with narrower panels).  Open 2 or 3 windows and attempt to rearrange them.  You should observe the drag functionality not working as expected.

**Expected behaviour**
When dragging windows, the dragged window should be moved with the position of the mouse cursor.  When the mouse is released, the window should be placed in the window list where the cursor was, regardless of how many windows are open.

**Other information**
On the release version of the Applet, the window jumps to the bottom of the list unless there are enough windows open.  With the current version on GitHub, the window jumps to the top, and even with more windows open, the window does not move as expected.

**Configuration info, if applicable**

```

{
    "section1": {
        "type": "section",
        "description": "Behavior"
    },
    "show-all-workspaces": {
        "type": "switch",
        "default": false,
        "description": "Show windows from all workspaces",
        "value": true
    },
    "enable-alerts": {
        "type": "switch",
        "default": true,
        "description": "Show an alert in the window list when a window from another workspace requires attention",
        "value": true
    },
    "enable-scrolling": {
        "type": "switch",
        "default": false,
        "description": "Enable mouse-wheel scrolling in the window list",
        "value": false
    },
    "reverse-scrolling": {
        "type": "switch",
        "default": false,
        "description": "Reverse the direction of mouse-wheel scrolling in the window list",
        "dependency": "enable-scrolling",
        "value": false
    },
    "left-click-minimize": {
        "type": "switch",
        "default": true,
        "description": "Left click to minimize window",
        "value": true
    },
    "middle-click-close": {
        "type": "switch",
        "default": true,
[...]

PR #10297, opened 2021-08-31
Fix window ordering with vertical panels

benjamin-owen:
This is my solution to issue #10293.  By checking the position of the panel (left or right) versus checking the width and height of the panel, I found more reliable checking.  Additionally, a variable is defined (depending on vertical or horizontal applet) to fix the position at which the window will be inserted while dragging.

### DIFF
--- a/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
@@ -1382,7 +1382,7 @@ class CinnamonWindowListApplet extends Applet.Applet {
             return DND.DragMotionResult.NO_DROP;
 
         let children = this.manager_container.get_children();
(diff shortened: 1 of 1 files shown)


mtwebster:
Hi, I can't reproduce the original issue - I created a vertical panel, set to 200px wide, added gwl. Dragging seems to work as expected (other than some flakiness when interacting between pinned and unpinned items).

Can you definitely reproduce this on a new panel and new instance of the applet? Maybe there's some configuration difference.

benjamin-owen:
I'm a certified idiot.  Both the pull request and issue should be referring to the normal "Window list" applet, NOT the grouped window list.  I just spun up a fresh LM20.2 VM and could reproduce the issue.  Please let me know if you need any additional information.

I will go ahead and edit the original issue so it is not referencing the GWL.

mtwebster:
Thanks, got it :)

benjamin-owen:
Resolves #10293

PR #10326, opened 2021-09-17
sound-applet: hide "Unknown Artist" on panel

archmunky:
When listening to streaming radio, metadata often includes Artist and Song in Title leaving the Artist field empty. This PR simply hides the " - Unknown Artist" text on the panel. This feature is switchable in settings and is set to false by default.


clefebvre:
I agree to hide it when the info isn't there, but there's really no need for this to be configurable.

archmunky:
Should I pull out the settings configuration code and push the update?

clefebvre:
@archmunky yes, if you have time.

clefebvre:
Thanks @archmunky

PR #10340, opened 2021-09-20
Add confirmation dialog when removing a panel

### LINKED ISSUE #10337: [Enhancement] Ask for confirmation when removing a panel
```
 * Cinnamon 5.0.5, not using daily builds ppa
 * Mint 20.2
 * Intel HD Graphics 620 | i915
 * 64 bit
 ```

**Issue**
I accidentally deleted my panel because clicking the remove button on the panel context menu removes the panel without confirmation.

**Steps to reproduce**
1. Right click on any panel
2. Click on "➖ Remove"

Screenshot from 2021-09-20 15-24-20

The panel is removed instantly.

**Expected behaviour**

A confirmation should be asked before removing the panel because otherwise this leads to unintentional loss of panel configurations.

fredcw:
fixes #10337

Although a panel can be recovered, this is not straightforward for a novice especially if it's your last remaining panel and it's easy to accidentally click on the wrong menu item and instantly delete your panel, I've done it myself. If you google "accidentally deleted cinnamon panel" you'll find lots of posts where people are asking how to recover their panel so this seems to be a not uncommon problem. I think just a confirmation dialog is the way to go.

### DIFF
--- a/js/ui/panel.js
+++ b/js/ui/panel.js
@@ -1593,7 +1593,11 @@ PanelContextMenu.prototype = {
 
         let menuItem = new PopupMenu.PopupIconMenuItem(_("Remove"), "list-remove", St.IconType.SYMBOLIC);  // submenu item remove panel
         menuItem.activate = Lang.bind(menu, function() {
-            Main.panelManager.removePanel(panelId);
+            let confirm = new ModalDialog.ConfirmDialog(_("Are you sure you want to remove this panel?\n\n"),
+                    function() {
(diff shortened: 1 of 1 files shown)


JosephMcc:
I'd say this is probably a good idea. Please get rid of the new lines though. I know this is being done in a couple other places (but not in all). They really all need to go away. It's up to themes to create that padding/spacing. Not for the strings to try and force them that way. It also makes the spacing inconsistent in different ones.

PR #10343, opened 2021-09-22
Window list: Add toggle for window labels on/off

benjamin-owen:
Currently, window labels are completely disabled for vertical panels with the window list applet. I've added a toggle to the preferences menu to "Show window labels next to icons," and this is enabled by default. For vertical panels, this allows users to show the window labels (same as the default behavior for a horizontal panel). Users can still disable this setting if they prefer the cleaner look of only icons/no labels for both horizontal or vertical panels.

(2nd PR since I messed up the first one)


benjamin-owen:
Here are some example images to better explain what's changed:

With toggle enabled (default):
labels_on

With toggle disabled:
labels_off

PR #10352, opened 2021-09-27
Adds first system monitor app to panel right click menu if it exists

RavetcoFX:
Parity feature with Mate and a few other desktop environments. 

Screenshot from 2021-09-26 19-19-12


clefebvre:
I disagree with this. Why not terminal, why not firefox... we made it easy to launch apps from pinning, from favs, from launchers, there's no need to arbitrarily add this here.

PR #10358, opened 2021-09-28
Add Edit menu item option when right clicking an app

RavetcoFX:
Spawns cinnamon-desktop-editor

Less annoying than right clicking the menu -> configure... -> Menu -> open menu editor -> finding your app
Also puts it at parity with Mate and XFCE

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -375,6 +375,9 @@ class ApplicationContextMenuItem extends PopupMenu.PopupBaseMenuItem {
                     return false;
                 });
                 break;
+            case "edit_item":
+                Util.spawnCommandLine("cinnamon-desktop-editor -mlauncher -o " + this._appButton.app.get_app_info().get_filename());
+                break;
             case "add_to_desktop":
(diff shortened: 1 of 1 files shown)


fredcw:
@RavetcoFX Just out of curiosity, for what reason do you need to edit the menu items?

RavetcoFX:
@fredcw I just do often, changing the name or icons. It's driven me crazy it took so many steps in Cinnamon

fredcw:
@RavetcoFX What's wrong with the names and icons that you need to change them? I'm curious because I've never wanted to do this.

fredcw:
The reason I asked is because I was wondering if apps needing their names or icons changed was a problem best fixed elsewhere. I just installed PrusaSlicer on manjaro and it's just called "PrusaSlicer" in the menu so it would seem that the problem is with the naming of the app by the packager. Users shouldn't be expected to fix badly named apps or for example, fix missing icons etc.

benjamin-owen:
I agree users shouldn't be _expected_ to fix these apps, but I don't see a reason the ability to do so should be less accessible.  Maybe some users _want_ the version number next to their applications or want to put a prefix in front of it for alphabetical reasons.  There are probably a number of reasons as to why a user would want easier access to these settings.

clefebvre:
I'm closing this PR because the feature is very niche.

PR #10370, opened 2021-10-05
Fix menu-editor with python-3.10

leigh123linux:
Fix https://bugzilla.redhat.com/show_bug.cgi?id=2010737

```
  File "/usr/share/cinnamon/cinnamon-menu-editor/cme/util.py", line 22, in <module>
    from collections import Sequence
ImportError: cannot import name 'Sequence' from 'collections' (/usr/lib64/python3.10/collections/__init__.py)
```


mmuehlfeldRH:
I applied the changes from this PR to my Fedora 35 and can confirm that it fixes the problem.

Thanks for fixing the issue so quickly.

PR #10389, opened 2021-10-13
Simplify window animations

clefebvre:


Secret-chest:
@clefebvre PLEASE. UNDO.
Users don't have to touch that. You can just leave it alone if it looks too complicated.

PR #10428, opened 2021-10-29
[WIP] Add copy to clipboard in system settings

nurgasemetey:
Added copy to clipboard button in system settings.

Screenshot:

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -70,6 +70,7 @@ Depends:
  adwaita-icon-theme,
  gsettings-desktop-schemas (>= 0.1.7),
  iso-flag-png,
+ inxi,
  libcanberra-pulse,
  libcinnamon-desktop4,
  libcinnamon-menu-3-0,
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_info.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_info.py
@@ -195,6 +195,43 @@ class Module:
                 widget.pack_start(button, True, True, 0)
                 settings.add_row(widget)
 
+            if os.path.exists("/usr/bin/inxi"):
+                
+                widget = SettingsWidget()
+
+                button = Gtk.Button(label=_("Copy to clipboard"))
+                button.connect("clicked", self.on_copy_clipboard_button_clicked)
+                widget.pack_start(button, True, True, 0)
+                settings.add_row(widget)
+    
+    def on_copy_clipboard_button_clicked(self, button):
+        try:
+            # taken from https://github.com/linuxmint/xapp/blob/master/scripts/upload-system-info
+            subproc = Gio.Subprocess.new(['inxi', '-Fxxrzc0'], Gio.SubprocessFlags.STDOUT_PIPE |  Gio.SubprocessFlags.STDERR_PIPE)
+            subproc.wait_check_async(None, self.on_copy_clipboard_complete)
+        except Exception as e:
+            print ("An error occurred while copying the system information to clipboard")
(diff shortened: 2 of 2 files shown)


Secret-chest (approved):
Nice

nurgasemetey:
> I think is good add also inxi to cinnamon recommends in debian/control as commits 2-3-4° seems fixes/improvements to the first I think is better amend the first and keep only it

Sorry for late reply. Thanks for your review.

 I think that I am spoiling this repo by adding `inxi`. For example, `Upload system information` is done as follows:

https://github.com/linuxmint/cinnamon/blob/1e29d6c1be588439625ddaef73bbaf9f7cc7dddd/files/usr/share/cinnamon/cinnamon-settings/modules/cs_info.py#L186-L196

`upload-system-info` is located here - https://github.com/linuxmint/xapp/blob/master/scripts/upload-system-info. Also `inxi` is already declared there - https://github.com/linuxmint/xapp/blob/66f407bc2d9d47903cdb4d372b33c6d7f0befb88/debian/control#L75-L87

I fear that adding `inxi` to debian control can spoil code design because it is already declared in `xapps`. What do you think? I can write script there and connect like `upload-system-info` is connected.

nurgasemetey:
> If you just need inxi's output, you can simplify these lines using [subprocess.check_output](https://docs.python.org/3/library/subprocess.html#subprocess.check_output) `inxi_out = subprocess.check_output(['inxi', '-Fxxrzc0'], encoding='utf-8')`
> 
> But inxi takes quite some time to complete gathering all infos. Therefore it would be better if you run the process asynchronously, so that the UI doesn't freeze in the meantime. You could use `Gio.Subprocess` [[doc]](https://lazka.github.io/pgi-docs/#Gio-2.0/classes/Subprocess.html#Gio.Subprocess) for this

Sorry for late reply. You are certainly right. I have some concerns about code design mentioned there - https://github.com/linuxmint/cinnamon/pull/10428#issuecomment-961020917. That's why I want to wait before implementing.

clefebvre:
Hi @nurgasemetey,

- add_copy_clipboard() is unecessary, you're only calling it once, it's small.. there's no need to create a function here.
- you should wrap that code in an if() statement which tests the existence of /usr/bin/inxi. I know xapps depends on inxi, but this is more future-proof, if we ever remove that dependency in xapps, this will continue to check. Also, that dependency in xapps is specific to our packaging, it's not necessarily in place in other distros.
- if the call is sync, like it is now, you'll freeze the UI. If it's not, then the user won't know when the clipboard is ready. Add an infobar when it's ready, or show a notification... 

I'll mark this was WIP in the meantime. Remove the [WIP] prefix when it's complete.

nurgasemetey:
@clefebvre 
Thanks for feedback. I will try to stick to your advices.

nurgasemetey:
@clefebvre 
I have done following things:
- added to debian control
- removed separate function
- added conversion from stream to string

Could you review?

dreng:
@nurgasemetey 
> Could you review?

I think you should remove the [WIP] prefix from the title like clefebvre requested before. Otherwise nothing will be reviewed.

DirkHaar:
This is really handy; only: as it is asked for in github, could you add 
- 32 or 64 bit
- active graphics driver and
- Cinnamon version?
It would reduce a lot of questions on reported issues, wouldn't it?

mtwebster:
That info is already included in the output.

DirkHaar:
Fine, It just isn't shown in the dialogue windows. I should have looked into the clipboard content.

PR #10441, opened 2021-11-02
Window button layout setting: improve rtl handling

ronyala:
Relpace `right` and `left` lables when on an rtl layout.
I would like some feedback as to whether using a label multiple times can affect translation handling (e.g. create two entries to translate).

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_windows.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_windows.py
@@ -35,8 +35,12 @@ class Module:
             settings = page.add_section(_("Buttons"))
 
             button_options = []
-            button_options.append([":minimize,maximize,close", _("Right")])
-            button_options.append(["close,maximize,minimize:", _("Left")])
+            if Gtk.Widget.get_default_direction() == Gtk.TextDirection.RTL:
+                button_options.append([":minimize,maximize,close", _("Left")])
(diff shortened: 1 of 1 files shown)


clefebvre:
Translations are fine with multiple occurrences as long as all occurrences mean the same thing.

Say you do:

a = _("taken")
b = _("left")

and elsewhere in the code you've got..

c = _("left") 
d = _("right")
e = _("right")

there would be a collision on "left" (used by b and c), but also a collision on "right" (used by d and e).

The collision on d and e would not matter, since they both have the same meaning, but it would matter on b (which wants to convey the idea of something being left, i.e. not taken) and c (which wants to convey the idea of not being placed on the right).

In this example we would typically rephrase the translation ID for a and b to something like "item taken" and "item left".

clefebvre:
@ronyala can you show me a screenshot of the issue on the master branch? i.e. I'd like to see this setting window in your own language. Please add a translation to let me know whether the combo shows "right" or "left".

ronyala:
Thanks for explaining how the translation strings work, they have the same meaning in this case so it's fine.

I've added a picture of the 'Windows' settings window:
צילום מסך מ־2021-11-08 17-16-51
The setting value is 'right', while the top window buttons are on the left.

clefebvre:
OK, thanks for this, that makes sense... so it looks like GTK itself reverses how it handles the setting string, which is handy I guess.

PR #10453, opened 2021-11-09
calendar applet: Add support for evolution-data-server

mtwebster:
There is a cinnamon-calendar-server process launched by the applet.

A full month+ (the entire 42 square grid) at a time is requested. The applet tells the helper what date range it wants events for, then waits for a signal of new events. Everything is asynchronous.

Some todo:
- [ ] no events label needs an icon?
- [ ] make everything prettier

Maybe:
- allow event deletion?
- show todo/tasks? Maybe a different 'view' that the calendar can default to opening, and a button/switcher to go between views. Allow tasks to be completed.


archmunky:
Thank you for this work! I've been working on a similar applet but it just calls one or more ics files to populate the event list. Currently stuck in RRULE handling.

PR #10474, opened 2021-11-19
[WIP] Add file path to photframe desklet

obaqueiro:
Add thee ability to show a partial filepath to the photoframe desklet.


mtwebster:
Hi, I notice a few issues with this:

- The directory depth is too much customization, just provide the full path and ellipsize the leading portion of text, so you'd get:

with a small picture frame:
**...fooberry/jam/bar.jpg**
with a larger frame:
**.../joe/pictures/2011/factory/fooberry/jam/bar.jpg***

You can use `this._label.clutter_text.set_ellipsize()` to control where the ellipses occur.

- The label needs to be on the proper background of the desklet - right now it's against the desktop wallpaper and its visibility entirely depends on the background colors:
image

mtwebster:

PR #10486, opened 2021-11-27
Grouped Widow List - Improvements in handleScroll

alefpereira:
  feel any change.
- Change "Cycle windows" wheel scroll behaviour and left click action options


alefpereira:
The lmde4 ci log indicates an error during installing related to unmet cinnamon dependencies, I may be wrong but I don't think they are related to that change.

PR #10488, opened 2021-11-28
photoframe should be able to skip hidden files

xushen-ma:
Photoframe needs to be able to skip hidden files to avoid showing thumbnail images and intentionally hidden files.

Also allow users to configure left/middle button click actions.

### DIFF
--- a/files/usr/share/cinnamon/desklets/photoframe@cinnamon.org/desklet.js
+++ b/files/usr/share/cinnamon/desklets/photoframe@cinnamon.org/desklet.js
@@ -19,11 +19,14 @@ class CinnamonPhotoFrameDesklet extends Desklet.Desklet {
         this.settings = new Settings.DeskletSettings(this, this.metadata.uuid, this.instance_id);
         this.settings.bind('directory', 'dir', this.on_setting_changed);
         this.settings.bind('shuffle', 'shuffle', this.on_setting_changed);
+        this.settings.bind('skip-hidden', 'skip_hidden', this.on_setting_changed);
         this.settings.bind('delay', 'delay', this.on_setting_changed);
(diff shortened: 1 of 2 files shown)


files/usr/share/cinnamon/desklets/photoframe@cinnamon.org/desklet.js
```diff
@@ -90,6 +93,11 @@ class CinnamonPhotoFrameDesklet extends Desklet.Desklet {
 
         let info;
         while ((info = fileEnum.next_file(null)) != null) {
+            /* Skip hidden files / directories */
+            if (this.skip_hidden && info.get_name().indexOf(".") == 0) {
```
mtwebster, line 97: Instead of checking for `.` use info.get_is_hidden and add `standard::is-hidden` to the enumerate_children() call above.

Another method of hiding files is to create a file `.hidden` in the folder, and adding filenames that you want hidden. `info.get_is_hidden` will account for that as well.

mtwebster:
Hi - 

I think skipping hidden files is a good thing, but please check my review comment on a better way to do this.

Also, I don't agree with making the buttons, or hidden files, configurable.

We're definitely not against letting the user customize things on the desktop, but it all adds up in the long run to make maintenance more difficult, so when we can we prefer to just go with what we feel makes the most sense for the most users. I believe that's the case here - we're not adding anything except the ability to swap button actions.

Also, in the future can you use separate commits for unrelated changes? This makes reviewing easier, as well as debugging.

Thanks!

xushen-ma:
Hi Michael,

Thanks for getting back to me. No problems, let me seperate the commits and submit again.

Kind Regards

Xushen Ma
E-mail: ***@***.***

Issue #2883, opened 2014-02-11
Alt-Tab is window-centric


alexs77:
The Alt-Tab switcher is window-centric. It should be app-centric. 
For cycling between windows of an app, one can also use Alt+^ (key to the left of 1, or above tab). This works in Cinnamon and in Gnome and in Unity.
In Cinnamon, one also sees the full window list of all applications, when pressing Alt+Tab. That's cumbersome, if one has many windows (eg. terminals) open of one application.

jeffwildcoast:
Is there still no fix for this?

I come from a MacOS environment for about the past sixteen years or so. But since they became so extremely fragile and irreparable (not to mention butterfly keyboard design flaws) I've gone back to Linux (Ubuntu 22.04 + Cinnamon DE) on a Dell Precision which feels much like what a MacBook Pro used to. But with replaceable components!

I really miss MacOS. And they had UX nailed with grouped application windows with Alt-Tab. And Alt-Tilda to move between individual application windows.

Cannot understand how anyone would think it could ever be considered to be proper behaviour to cycle through all open windows with Alt-Tab.

![image](https://user-images.githubusercontent.com/105626702/223754622-76e0a2ae-6384-4f52-b406-4f504b4ded78.png)

egoarka:
@jeffwildcoast check this, https://github.com/linuxmint/cinnamon-spices-extensions/issues/462#issuecomment-1869130197

tjohnman:
I too come mainly from macOS (but also from KDE), and I find this behavior very jarring. After a morning coding, I end up with way too many icons to go through when alt-tabbing. It's simply unusable for me. Jeff's screenshot is exactly how my workspace ends up looking on Cinnamon every day.

I'm probably using Cinnamon wrong, because I don't understand how anyone could work like this.

If I had the time right now, I would try learning how to, and developing an applet, but as things are right now, I rather take the other workaround suggested here (ditching Cinnamon).

b3rs3rk1ng:
I've implemented a solution for this in PR #13498.

The PR adds a configurable setting `alttab-switcher-group-by-app` that allows users to choose between window-centric (current behavior) and application-centric (requested behavior).

When enabled:
- Alt+Tab groups windows by application
- Alt+` cycles directly between windows of the same app
- Shows application names instead of window titles

When disabled, Cinnamon behaves exactly as before.

The setting appears in System Settings → Windows → Alt-Tab.

[...]

PR #10490, opened 2021-11-30
Grouped Widow List - Improvements in handleScroll

alefpereira:
  feel any change.
- Change "Cycle windows" wheel scroll behaviour and left click action options


alefpereira:
Apart those fixes and improvements I also have other functionality I would like to merge: hotkeys to Cycle App Forward and Backward. This adds an option to be able to quickly switch to next or previous window/app with a hotkey, this is different from Alt-Tab as it switches to the most recent window. This is what this question is about https://unix.stackexchange.com/q/564285 and its also something I would like to be able to do and I think  other people will also find it useful.  

I already have it implemented and I'm using locally, but I would like to know if I can put this code on this PR (#10490) or if I have to wait until this PR is merged and then create a new PR for this new hotkey functionality?
 
https://github.com/alefpereira/cinnamon/tree/gwl-add-hotkey-to-cycle-apps-and-windows

1-DimensionalBox:
Hi @alefpereira,

Thank you for making this pull request for Cinnamon! I found it based upon a Linux Mint Forum topic on [Cycling through windows grouped by application](https://forums.linuxmint.com/viewtopic.php?t=320832).

Based upon further reading I found in #2883 that "Cinnamon's Alt-Tab is window-centric" and that this is **very much intentional**. Core developers do not seem eager to make this app-switching behavior part of Cinnamon.

Perhaps an alternative would be too create a Cinnamon Spices applet?

PR #10499, opened 2021-12-05
Add Child and Elder type of user groups

KAMI911:
Hi Folks!

I know this is not a final code to be merged. I would like to present an idea to discuss it to use more than two type of users. We can add extra settings (creating different Desktop user experience with larger font, and better contrast) or add firewall that can block user requests based on their user groups. For this I created a PoC that utilize two new groups:
* child
* elder

Currently the Cinnamon user/group application was modified to show the possibilities. Based on selection it creates the required groups on demand.
   
The motivation behind the extra groups to create pre-defined settings for special group of users.

For example we can add firewall and/or proxy that blocks unwanted websites, do dektop settings during user creation, etc.

This is a work-in-progress example, to make Cinnamon desktop more youngster and elder friendly.

Let's discuss the idea, if possible.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py
+++ b/files/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py
@@ -364,6 +364,8 @@ class NewUserDialog(Gtk.Dialog):
             self.account_type_combo = Gtk.ComboBoxText()
             self.account_type_combo.append_text(_("Standard"))
             self.account_type_combo.append_text(_("Administrator"))
+            self.account_type_combo.append_text(_("Child"))
+            self.account_type_combo.append_text(_("Elder"))
             self.account_type_combo.set_active(0)
 
             self.realname_entry = Gtk.Entry()
@@ -564,6 +566,8 @@ class Module:
             self.account_type_combo = Gtk.ComboBoxText()
             self.account_type_combo.append_text(_("Standard"))
             self.account_type_combo.append_text(_("Administrator"))
+            self.account_type_combo.append_text(_("Child"))
+            self.account_type_combo.append_text(_("Elder"))
             self.account_type_combo.connect("changed", self._on_accounttype_changed)
 
             self.realname_entry = EditableEntry()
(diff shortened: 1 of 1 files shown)


Secret-chest:
I don't think this would work. Elder accounts need to be configured separately, vision problems are diverse. We could add an assistant in Accesibility. Same for child accounts.

clefebvre:
I think that's the distribution's job, not the DE's. Also I'd rather see this in the welcome screen.

KAMI911:
@Secret-chest  and what about content filtering? I mainly created this not just for custom settings but for support easy filter presets that built into the OS toolset. Adding a current (and updatable) blocking rules to iptables/nftable with blocking a specific group like:
iptables -A OUTPUT -m set --match-set content-block-net src -m owner --gid-owner child -j LOG --log-prefix 'CONTENT_FILTER 
iptables -A OUTPUT -m set --match-set content-block-net src -m owner --gid-owner child -j DROP
may make easier the life for a family Linux distribution. We could load the blacklist data to iptables' ipset for the greatest performance.

What do you think about?

KAMI911:
@clefebvre, How do you like to integrate into Welcome screen?

Secret-chest:
just don't add fun themes. Also we could improve mintnanny to block domains only for specific users.

Secret-chest:
> Also I'd rather see this in the welcome screen.

I don't understand. You create user accounts in that application, not in the welcome screen. You could however link to it in the welcome screen with something like: "To create more user accounts, [click here]"

KAMI911:
Probably we can use this work as blacklist:
https://dsi.ut-capitole.fr/blacklists/index_en.php
It is updated frequently.
I think it is good idea to add features to MintNanny. Probably we could add functionality to MintNanny to handle iptables/nftables or use firewalld. It seems ufw not handling group based rules at all.
New features:
* Using kernel based blocking firewall (ipsets) instead of hosts file based
* Create multiple ipsets based on blacklist categories
* Add/remove ipsets members in one group based user preferences
* Add/remove/update ipsets from blacklist (automatically?, daily?, manually?)
* Assign ipsets to groups (child, elder)

What do you think about?

KAMI911:
https://dsi.ut-capitole.fr/blacklists/download/

KAMI911:
Additionally we could have  a time based internet usage settings too.

Secret-chest:
> Additionally we could have  a time based internet usage settings too. 

How would this work?

KAMI911:
Probably we can use iptables capabilities like:

iptables RULE -m time --timestart TIME --timestop TIME --days DAYS -j ACTION

KAMI911:
It would be nice to have configure blacklist / autoupdate from Internet like mintnanny. Do you have tools for UI prototype or should I use Glade? Do you have idea how the app should look like?

KAMI911:
I started to work on backend functions. Can anyone who experienced with UI help?

KAMI911:
We can query all the „A” type DNS records via:
dig -f type.txt +short
Based on downloaded list we can cache DNS queries in files and add as list of IPs in ipsets.

KAMI911:
Do you have idea for the UI?

clefebvre:
Hi guys, sorry for the delay.

mintnanny is really barebone, it doesnt' do much at all. I'd love to see a parental control solution integrated into Mint, and then we'd link to it from the welcome screen, but developing such a solution requires a huge amount of work. This is not a simple thing.

Just like timeshift and ufw, this really is something that could make the desktop better. We need to find a good upstream solution though and then take time to properly integrate it.

KAMI911:
@clefebvre, do you know open source upstream solution? I am happy to help somehow.

PR #10538, opened 2021-12-16
Windows Settings: Add menu to button layout

leigh123linux:
https://github.com/linuxmint/cinnamon-control-center/issues/273


hsbasu:
@leigh123linux Thanks a lot. After seeing your PR, I feel so idiotic of myself. Had I studied the code properly, I could have created a PR myself. It was this much easier. 👍🏾

hsbasu:
When would this PR get merged? Really looking forward to it.

JosephMcc:
I highly doubt that it will be. The plan is to move Cinnamon away from using Metacity theming, instead using the Gtk theming will handle the window decorations. Since really no themes support the menu button you'll end up with something that won't really work as expected.

hsbasu:
@JosephMcc Should I close the PR?

Secret-chest:
> The plan is to move Cinnamon away from using Metacity theming, instead using the Gtk theming will handle the window decorations.

Will the titlebars be huge? Please shrink them

PR #10548, opened 2021-12-22
[WIP] convert CinnamonNetworkApplet to es6 Class

jonath92:
The Network applet is AFAICS the last applet from the applets directly included in cinnamon which is not using the es6 class syntax from the Applet classes defined in applet.js

### DIFF
--- a/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
@@ -1668,15 +1668,10 @@ NMMessageTraySource.prototype = {
     }
 };
 
-function CinnamonNetworkApplet(metadata, orientation, panel_height, instance_id) {
-    this._init(metadata, orientation, panel_height, instance_id);
-}
+class CinnamonNetworkApplet extends Applet.IconApplet {
 
-CinnamonNetworkApplet.prototype = {
-    __proto__: Applet.IconApplet.prototype,
-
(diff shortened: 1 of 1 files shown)


AxXxB:
Dont know if this PR will be accepted, just a thought. Classes which are to export should be defined as 'var ClassName = class ClassName ... ' or there will be a warning: https://github.com/linuxmint/cinnamon/pull/10568

mtwebster:
If this is going to be done, I'd prefer the entire file be redone, not just the main applet class - there are lots of other classes that need to be done also.

lestcape:
> If this is going to be done, I'd prefer the entire file be redone, not just the main applet class - there are lots of other classes that need to be done also.

I think you should discuses first what is the preferable way to go. "[Porting to ES6 classes has the potential to break extensions (as it's not possible to use Lang.Class to inherit from ES6 classes)](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/350)". Probably you want to migrate the whole Cinnamon code at once then, to affect extensions only one time.

Additional to this, you should take special attention of how will be converted the GObject clases, because [gjs not supports dynamic GObject classes](https://bugzilla.gnome.org/show_bug.cgi?id=688973). So, the changes on de Cinnamon classes can force extensions to create a GObject extension class and then assigning a GType to that class. But that can involved then a failure when the extension load, because of a duplicate GType name.

mtwebster:

PR #10564, opened 2021-12-28
Fix #10563

### LINKED ISSUE #10563: Clicking on minimize Button throws an error after removing the "Grouped window list" applet
```
- Cinnamon version = 5.2.6  (using the daily builds PPA)
- Linux Mint 20.2 
 ```

**Issue**
When clicking on the minimize Button of an application window, it is thrown an error in xsession-error and the window gets unresponsive when there is not "grouped window list" (or similar) applet installed. 

**Steps to reproduce**
1. Remove "Grouped window list" (and all applets which provide similar functionality) 
2. Restart Cinnamon 
3. Click on the minus icon of any window to minimize the window

**Expected behaviour**
The window is removed from the window and can be reopened by cycling through the open windows (with the keyboard shortcut) or similar methods

**Actual beaviour**
The follwing error is shown in the xsession-errors and the window is unresponsive. 

```
(cinnamon:26756): Cjs-CRITICAL **: 12:28:21.736: JS ERROR: TypeError: this.scale is not a function
traditional@/usr/share/cinnamon/js/ui/windowEffects.js:226:18
_startWindowEffect@/usr/share/cinnamon/js/ui/windowManager.js:621:25
_minimizeWindow@/usr/share/cinnamon/js/ui/windowManager.js:680:14
WindowManager/<@/usr/share/cinnamon/js/ui/windowManager.js:505:66
```

jonath92:
Fixes #10563. `this.scale` doesn't exist...

### DIFF
--- a/js/ui/windowEffects.js
+++ b/js/ui/windowEffects.js
@@ -50,7 +50,7 @@ class Effect {
     }
 };
 
-var Map = class Map extends Effect {
+class Map extends Effect {
     constructor() {
         super(...arguments);
 
@@ -134,7 +134,7 @@ var Map = class Map extends Effect {
     }
 }
 
-var Close = class Close extends Effect {
+class Close extends Effect {
     constructor() {
         super(...arguments);
 
@@ -180,7 +180,7 @@ var Close = class Close extends Effect {
     }
 }
 
-var Minimize = class Minimize extends Close {
+class Minimize extends Close {
     constructor() {
         super(...arguments);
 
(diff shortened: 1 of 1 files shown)


mtwebster:
Hi, thanks for the contribution, but we went with a different fix - I wanted to still maintain an 'effect' rather than an instantaneous change, and the effects also have some cleanup to perform that wouldn't have occurred otherwise.

Also, in the future, only address the specific issue in a pull request - don't perform any sort of refactoring or other unrelated changes.

Thanks

Issue #9378, opened 2020-06-18
Critical notifications are ignoring timeout configuration


TurpIF:
```
 * Cinnamon version: 4.4.8
 * Please specify if you are using the daily builds PPA: no
 * Fedora 32
 * Intel Corporation UHD Graphics 620 (Whiskey Lake) and Mesa Intel(R) UHD Graphics 620 (WHL GT2)
 * 64 bit
 ```

**Issue**
Critical notifications are never hidden even if timeout is specified in configuration panel and the notification applet is provided.
Browser extensions such as Pushbullet send critical notifications. This implies that once a notif is shown, user should manually close it.

**Steps to reproduce**
* Setup notification applet
* Go in Menu > Preferences > Notifications and check the option "Hide notifications after expiration"
* Run `notify-send -u critical "test"`
* Wait for 10 seconds
* The notifications is still shown

**Expected behaviour**
The notification should be hidden and moved in the notification applet.

**Other information**
The variable [NOTIFICATION_CRITICAL_TIMEOUT_WITH_APPLET](https://github.com/linuxmint/cinnamon/blob/master/js/ui/messageTray.js#L25) seems to show that critical notification should be hidden when the notification applet is available.

But [this condition](https://github.com/linuxmint/cinnamon/blob/master/js/ui/messageTray.js#L963) prevent critical notifications to be hidden.

PR #10566, opened 2021-12-29
[WIP] Allow expiration of critical notification

TurpIF:
This fixed the issue #9378 which is still present on latest Cinnamon version.

Old behavior was keeping the critical notification displayed even is a notification applet is provided.
The new behavior is keeping the critical notification displayed for 10s, then if a notification applet is provided, then notification is moved into it and is hidden, else the notification stay displayed as before.


clefebvre:
hmm... we rely on critical notifications precisely to be able to show notifications which DO NOT vanish.

For instance, in mintupdate, when the notification is shown, we do not want it to disappear until it is dismissed by the user.

mtwebster:

PR #10567, opened 2021-12-30
Make evolution-data-server integration optional

SparkyBluefang:
I worked on this as part of packaging Cinnamon for Gentoo.

This adds a build flag to optionally disable building the calendar server. This helps reduce dependency weight for users who don't use Evolution.

I'm not 100% sure of the JS change (I haven't worked with CJS/GI objects before), but it seems to work and I'm not seeing any obvious log warnings/errors.

### DIFF
--- a/files/usr/share/cinnamon/applets/calendar@cinnamon.org/eventView.js
+++ b/files/usr/share/cinnamon/applets/calendar@cinnamon.org/eventView.js
@@ -301,7 +301,7 @@ class EventsManager {
     }
 
     start_events() {
-        if (this._calendar_server == null) {
+        if (this._calendar_server == null && Cinnamon.CalendarServerProxy) {
             Cinnamon.CalendarServerProxy.new_for_bus(
                 Gio.BusType.SESSION,
                 // Gio.DBusProxyFlags.NONE,
--- a/meson.build
+++ b/meson.build
@@ -179,6 +179,9 @@ install_subdir(
     strip_directory: true,
 )
 
-subdir('calendar-server')
+
+if get_option('build_calendar_server')
+    subdir('calendar-server')
+endif
 subdir('python3')
 subdir('install-scripts')
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -13,6 +13,11 @@ option('build_recorder',
     value: true,
     description: 'Build the cinnamon recorder into source'
 )
+option('build_calendar_server',
+    type: 'boolean',
+    value: true,
+    description: 'Build the cinnamon EDS calendar server'
+)
 option('disable_networkmanager',
     type: 'boolean',
     value: false,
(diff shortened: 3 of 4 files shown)


mtwebster:
Hi, this looks fine, and I'll merge it but I was wondering if it might be more convenient to make it a 'soft' dependency since we're working with python and js... cinnamon-calendar-server.py tries to import EDataServer, if it fails, it exits, and the event manager in Cinnamon never gets loaded.

With gentoo, are end users expected to be able to recompile if they want the feature?

Maybe I'll do that as well as this build option.

SparkyBluefang:
My intention behind making it a compile flag is that it helps keeps the logs clean of python import and dbus error messages.

In Gentoo, we try to manage optional runtime dependencies through package flags - same as compile time dependencies and make flags. This allows the package manager to properly manage reverse dependencies. So, from Gentoo's packaging perspective, changing either runtime dependencies or build flags would still result in a re-install (re-compile).

I'm definitely not opposed to this being managed in Cinnamon at runtime, since that would basically mean more robust error/state handling :) I just wasn't sure how the dbus proxy managed the service lifecycle. Ideally, we wouldn't want it to repeatedly try to start a "broken" service and spam the logs with errors.

mtwebster:
I'm also worried users might want to use the calendar and reporting a bug that it doesn't work - now we have the same cinnamon version behaving differently.

What I meant before was, is the user the one that defines these flags, or the maintainer alone? If someone wants the feature, is it easy for them to get?  I may be misunderstanding things.

SparkyBluefang:
Gentoo package flags have defaults, and you also have a package manager system "profile" that defines all of the platform defaults, but all are user configurable and can be overridden.

With this change, the cinnamon build options have it enabled by default (so that any builder unaware of this option outside of Gentoo would get the correct expected build), and the Gentoo package would also have it enabled by default. The eds integration would only be disabled if the user explicitly disables it for the package or has disabled any platform-wide eds integration in the past.

I'm basically attempting to treat this the same way as the recorder/gstreamer and networkmanager features/dependencies are handled.
https://packages.gentoo.org/packages/gnome-extra/cinnamon

mtwebster:
@SparkyBluefang can you give this a try: https://github.com/linuxmint/cinnamon/pull/10600

I realized today that the 'evolution-data-server' package isn't actually a dependency for any of the libraries that use its services, so it's perfectly reasonable for any user to be missing it, causing the calendar applet to complain. I had been assuming all along that cinnamon's dependency on  gir1.2-edataserver-1.2 would ensure e-d-s as well (this is usually a pretty reliable assumption).

So, I think it's best all around to have this as a runtime check. I don't mind merging this PR as well, but let me know if you think it'll still be necessary.

SparkyBluefang:
Sorry it took so long to get back to this. #10600 looks good to me. Since it verifies that EDS is running, before trying to launch the daemon, I don't think the build changes are needed any more. Thanks!

PR #10569, opened 2022-01-01
Fix a couple typos in latest code

stevegrubb:
In accessibility.js, there is a missing break in the switch statement.

In cr-utils.c, out_index + 4 is used twice, the second should have been
out_index + 5.


mtwebster:
`Gtk.Label.set_xalign` is correct - the only `set_x_align` is in Clutter

stevegrubb:
OK, I'll drop that one and resubmit. Thanks for the review.

stevegrubb:
The pull request has been updated to remove the change to cs_sound.py.

PR #10575, opened 2022-01-06
Option to change the button width in the window list

martj42:
I don't know if this is cool or not but I found the default option too narrow so here's the ability to choose the exact width of the buttons.

Peace. ✌️


mtwebster:
Hi, I'm ok with this but it needs to be rebased to the current master branch - there was a bit of work around that for a bug fix recently, unfortunately.

ref: 
https://github.com/linuxmint/cinnamon/commit/566cad571e2bf6c9bcedf4b3e3226dff39ba3f95
https://github.com/linuxmint/cinnamon/commit/b60b6a291cb123f18c32bc112077899b34c5b929

Thanks

martj42:
I think it should work now.

PR #10489, opened 2021-11-28
build: unbreak with meson >= 0.60.2

jbeich:
Regressed by [meson@5cc166a66](https://github.com/mesonbuild/meson/commit/5cc166a667ff). Similar to https://github.com/linuxmint/cjs/pull/98. From [error log](http://www.ipv6proxy.net/go.php?u=http://package22.nyi.freebsd.org/data/130amd64-default-foo/2021-11-28_09h10m20s/logs/cinnamon-4.8.6_1.log):
```python
$ meson setup _build
[...]
src/st/meson.build:204:0: ERROR: "install_dir" must be specified when installing a target
```

**EDIT**: `lmde4` bustage on CircleCI looks unrelated:
```
The following packages have unmet dependencies:
 cinnamon : Depends: gir1.2-ecal-2.0 but it is not installable
E: Unable to correct problems, you have held broken packages.
[...]
dpkg: dependency problems prevent configuration of cinnamon-common:
 cinnamon-common depends on gir1.2-timezonemap-1.0; however:
  Package gir1.2-timezonemap-1.0 is not installed.
[...]
dpkg: error processing package cinnamon (--install):
 dependency problems - leaving unconfigured
dpkg: dependency problems prevent configuration of cinnamon-dbg:
 cinnamon-dbg depends on cinnamon (= 5.2.1); however:
  Package cinnamon is not configured yet.

dpkg: error processing package cinnamon-dbg (--install):
 dependency problems - leaving unconfigured
[...]
Errors were encountered while processing:
 cinnamon-common
 cinnamon
 cinnamon-dbg
ERROR: 'sudo apt install --yes --allow-downgrades ../*.deb || sudo dpkg -i ../*.deb' exited with return code 256
```

PR #10596, opened 2022-01-11
Revert "build: disable gir install via list to pacify meson >= 0.60.2 (#10489)"

eli-schwartz:
This reverts commit 8fc2df08b40aa3e1958ed2fde853c50676d8cf48.

This commit was wrong, because it tried to work around a bug in a single version of meson by using something that isn't, wasn't, and won't be a valid value.

The fixed version of meson 0.60.x has been out for a while now, which once again accepts `false`, and 0.61.0 also accepts `false` but was known at the time of this workaround to not work in meson-git master (now meson 0.61.0).

Using `false` is acceptable and the failure to accept it has been qualified as a meson regression. Using `[false]` is just... trying to fuzz meson with random objects until you get something that slips its way through the argument checker and produces desired effects on the python implementation level.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -196,7 +196,7 @@ cinnamon_gir = gnome.generate_gir(
     includes: cinnamon_gir_includes,
     install: true,
     install_dir_typelib: pkglibdir,
-    install_dir_gir: [false],
+    install_dir_gir: false,
     extra_args: [
         '-DST_COMPILATION',
         '--quiet',
--- a/src/st/meson.build
+++ b/src/st/meson.build
@@ -213,7 +213,7 @@ st_gir = gnome.generate_gir(
     includes: st_gir_includes,
     install: true,
     install_dir_typelib: pkglibdir,
-    install_dir_gir: [false],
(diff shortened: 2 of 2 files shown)


Fantu:
@eli-schwartz with meson 0.61 there is another fail to build I saw on debian:
../docs/reference/cinnamon-js/meson.build:11:6: ERROR: gnome.gtkdoc keyword argument 'dependencies' was of type array[CustomTarget] but should have been array[Dependency | SharedLibrary | StaticLibrary]

I don't understand what is a good fix from a fast look, can you tell me please and/or do a patch also for it?

eli-schwartz:
I will do a patch for the dependencies kwarg tonight or tomorrow. Thanks for the heads up.

Fantu:
@eli-schwartz thanks, tested build on debian unstable with your patch and is ok: http://debomatic-amd64.debian.net/distribution#dpr-cinnamon-unstable/cinnamon/5.2.7-1~1
I'll use them in next debian build (to make it working)

PR #10603, opened 2022-01-14
[WIP] GWL - bug fixes: handleScroll, cycle apps, thumbnails scroll. Improvements: reverse scrolling, cycle apps hotkeys

alefpereira:
  feel any change.
- Change "Cycle windows" option text in wheel scroll behaviour and left click action settings
- Add new option to reverse mouse scrolling direction - also revert horizontal scrolling.
- Add new hotkeys settings to "Cycle Apps Forward" and "Cycle Apps Backwards". This greatly synergize with libinput-gestures and xdotool to create custom touchpad gestures improving user experience.
- Add new option to "Activate last focused if all minimized" on Cycle Apps

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -1000,7 +1000,10 @@ class AppGroup {
 
         let hasFocus = getFocusState(metaWindow);
         if (hasFocus && this.groupState.hasOwnProperty('lastFocused')) {
-            this.listState.set({lastFocusedApp: this.groupState.appId});
+            this.listState.set({
+                lastFocusedApp: this.groupState.appId,
+                lastFocused: metaWindow,
+            });
             this.groupState.set({lastFocused: metaWindow});
         }
         this.onFocusChange(hasFocus);
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -480,6 +480,28 @@ class AppGroup {
         this.label.hide();
     }
 
+    cycleWindows(step) {
+        let lastFocused, focusedIndex, z, count;
+        // Identifies last/current focused window from group (focused group or thumbnails)
+        lastFocused = this.groupState.lastFocused;
+        focusedIndex = findIndex(this.groupState.metaWindows, function(metaWindow) {
+            return metaWindow === lastFocused;
+        });
+        count = this.groupState.metaWindows.length - 1;
+
+        // Set next Window index (z) to focus
+        z = focusedIndex + step
+        // cycle on borders
+        if (z < 0) {
+            z = count;
+        } else if (z > count) {
+            z = 0;
+        }
+
+        let _window = this.groupState.metaWindows[z];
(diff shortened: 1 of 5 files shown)


AxXxB:
I have a thought that isReverseScrolling should also affect horizontal scrolling, but dont know if it is a behaviour that corresponds UX of other shells/OSes. What do you think?

AxXxB:
Also there is an option for scroll reverse in the mouse settings: https://github.com/linuxmint/cinnamon/blob/9f63f52c7b5c805b8cd63e18335b7f3edaf3141c/files/usr/share/cinnamon/cinnamon-settings/modules/cs_mouse.py#L37 - maybe it's better to use this one?

alefpereira:
> I have a feeling that isReverseScrolling should also affect horizontal scrolling, but dont know if it is a behaviour that corresponds UX of other shells/OSes. What do you think?

Yes, I didn't thought about that, it makes sense. I'll change reverse option to also affect horizontal scrolling.

Initially I included this feature inspired on `window-list@cinnamon.org` - at the moment I was just considering mouse. Later included better support for horizontal scrolling - thinking about touchpad, but  didn't thought whether the reverse option should affect or not horizontal scrolling. In the Mouse/Touchpad option, the reverse option also affects horizontal scrolling for touchpad, so it makes sense GWL to follow this behaviour. 

> Also there is an option for scroll reverse in the mouse settings:
> 
> https://github.com/linuxmint/cinnamon/blob/9f63f52c7b5c805b8cd63e18335b7f3edaf3141c/files/usr/share/cinnamon/cinnamon-settings/modules/cs_mouse.py#L37
> - maybe it's better to use this one?

imho this options should be separated. The Mouse/Touchad reverse options corresponds to page/frame/window scrolling, while the GWL is related to app/window switching in the bar, users may have different preferences for them. What do you think?

AxXxB:
> imho this options should be separated. The Mouse/Touchad reverse options corresponds to page/frame/window scrolling, while the GWL is related to app/window switching in the bar, users may have different preferences for them. What do you think?

Yes, it makes sense for sure.

alefpereira:
@AxXxB just pushed a new commit changing GWL reverse scroll option to also affect horizontal scrolling. Also rebased.

mtwebster:
Hi, I get this error when toggling 'Mouse wheel scroll action'
```
(cinnamon:367549): Cjs-CRITICAL **: 11:03:32.424: JS ERROR: TypeError: can't convert undefined to object
each@/usr/share/cinnamon/js/misc/util.js:500:27
averageProgress@/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js:566:13
onProgressChange@/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js:585:29
windowAdded/<@/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js:915:43
setTimeout/id<@/usr/share/cinnamon/js/misc/util.js:587:18

```
At the time I was writing an image to a usb stick, so the progress indicator was going, which may have something to do with it. Doesn't seem to actually break anything, and doesn't show up in normal use.

Would you mind squashing these down into just a couple or a single commit? Alternately I can do it when I merge. Thanks

kimimaru4000:
I'd love to see this merged! @alefpereira When you have time, can you address the above feedback?

mtwebster:
Closing due to age and lack of activity. It can be re-opened in the future if requested. I'd also like to suggest that if this is revisited, it's broken up into multiple PRs, as these are easier to evaluate and troubleshoot. Thanks

PR #10619, opened 2022-01-22
FIX #10618

### LINKED ISSUE #10618: Cinnamon gets unresponsive when clicking ESC while dragging an applet
```
Cinnamon version: 5.2.7 (using daily build)
Distribution - Min 20.3 
```

**Issue**
Cinnamon gets unresponsive when clicking ESC while dragging an applet and the following error is shown (sorry for uploading a picture instead of text but I don't know how to copy the text while cinnamon is unresponsive :D): 

**Steps to reproduce**
1. Open panel edit mode
2. Drag an applet - doesn't matter which one
3. Click ESC while dragging

**Expected behaviour**
Cinnamon is not unresponsive

jonath92:
Fixes https://github.com/linuxmint/cinnamon/issues/10618

### DIFF
--- a/js/ui/dnd.js
+++ b/js/ui/dnd.js
@@ -190,7 +190,7 @@ var _Draggable = new Lang.Class({
             } else if (this._dragActor != null && !this._animationInProgress) {
                 // Drag must have been cancelled with Esc.
                 // Check if escaped drag was from a desklet
-                if (this.target._delegate.acceptDrop){
+                if (this.target?._delegate.acceptDrop){
                     this.target._delegate.cancelDrag(this.actor._delegate, this._dragActor);
                 }
                 this._dragComplete();


ItzSwirlz:
My two cents-it's just a question mark and for the most part readers like me would skip over it.

Only way I view it is as someone learning flutter/dart where ? Means nullable (although null safety is a thing)

jonath92:
> As a rule we like to keep the js code readable to folks who aren't everyday js users - this is sort of an example, albeit a minor one. This stuff can make debugging more difficult later on as well.

> Above all else, NO Promises.

I guess I have no chance to convince you @mtwebster from the opposite but I would like to have at least tried. In my opinion your statement is fundamentally wrong. I am convinced that both promises and optional changing operator make javascript code significantly more readable for both everyday js developers and for not everyday js developers. Honestly which one is more readable? 

1: 
```js
let nestedProp = null 

if (grandparent && grandparent.parent) {
    nestedProp = grandparent.parent.nestedProp
}
```
2: 
```js
const nestedProp = grandparent?.parent?.nestedProp
```

Another advantage of using modern syntax is that you are learning interesting stuff when you read the code. In my opinion you shouldn't discourage people from using modern syntax but encourage people to do this!

mtwebster:
I'm ok with the second example - our codebase is filled with the first. Personally, I don't think it's really appropriate everywhere it currently exists - it's just sloppy programming. Most of the useful stuff related to Cinnamon are calls to C libraries - you should know precisely what to expect, in and out, and not need to resort to this. We end up with a lot of undefined behavior because of it.

I'm not against new syntax, I'm pro-readability - someone a few years ago added a lot of Promises to our code, when it was a new thing - some of this is an unreadable mess. I prefer using our existing Mainloop functions, at least compared to what we ended up with there, so it makes us a bit defensive. Maybe he just wasn't that good at it - I'm open to pursuasion

fredcw:
I'll give my opinion on js promises later.

PR #10631, opened 2022-01-26
GWL: fix rtl layout issues

### LINKED ISSUE #10280: Grouped Windows list applet not supporting RTL languages
When setting interface language to an RTL Language and the applet showing the application name; the applet(grouped windows list) writes the name (of the app) from right to left but it starts writing from the left side of the button.

ronyala:
* fix label allocation out of bounds in rtl layout (fix #10280)
* fix window count badge position in rtl layout
* allocation code refactor
* moved some appGroup styling into theme
* minor code cleanup

One issue I couldn't solve is getting the St.Label to position rtl text right-to-left. I tried setting `direction` to RTL is JS and using align or `direction` css properties... I could use some help with that.

Before:
Before

After:
After


AxXxB:
> One issue I couldn't solve is getting the St.Label to position rtl text right-to-left. I tried setting `direction` to RTL is JS and using align or `direction` css properties... I could use some help with that.

This is the one option i found:
```javascript
/* appGroup.js */

const Cinnamon = imports.gi.Cinnamon;
const Meta = imports.gi.Meta;
const Clutter = imports.gi.Clutter;
const St = imports.gi.St;
const Main = imports.ui.main;
const Tweener = imports.ui.tweener;
const DND = imports.ui.dnd;
const Tooltips = imports.ui.tooltips;
const PopupMenu = imports.ui.popupMenu;
const {SignalManager} = imports.misc.signalManager;
const {each, findIndex, unref} = imports.misc.util;
const {createStore} = imports.misc.state;

// https://github.com/kavirajk/isRTL
const rtlChars = [
	/* arabic ranges*/
	'\u0600-\u06FF',
	'\u0750-\u077F',
	'\uFB50-\uFDFF',
	'\uFE70-\uFEFF',
	/* hebrew range*/
	'\u05D0-\u05FF'
].join("");
const reRTL = new RegExp("[" + rtlChars + "]", "g");
const isRTL = function(text) {
	text = text.replace(/[0-9\s\\\/.,\-+="']/g, ''); // remove multilengual characters from count
    const firstChar = text[0] || '';
	return !!(firstChar.match(reRTL) || []).length;
};

...

    allocate(actor, box, flags) {
        const allocWidth = box.x2 - box.x1;
        const allocHeight = box.y2 - box.y1;
        const childBox = new Clutter.ActorBox();
        const direction = this.actor.get_text_direction();
        const textDirection = isRTL(this.label.get_text()) ? Clutter.TextDirection.RTL : Clutter.TextDirection.LTR;
        
        ...
        
        // Set label position
        if (this.drawLabel) {
            const labelNaturalHeight = this.label.get_preferred_size()[3];
            const labelYPadding = Math.floor(Math.max(0, allocHeight - labelNaturalHeight) / 2);

            childBox.y1 = box.y1 + labelYPadding;
            childBox.y2 = childBox.y1 + Math.min(labelNaturalHeight, allocHeight);
            if (direction === Clutter.TextDirection.LTR) {
                childBox.x1 = Math.min(this.iconBox.x + this.iconBox.width, box.x2);
                childBox.x2 = box.x2;
            } else {
                childBox.x1 = box.x1;
                childBox.x2 = this.iconBox.x;
            }

            if (textDirection == Clutter.TextDirection.LTR) {
                this.label.set_style('text-align: left;');
            } else {
                this.label.set_style('text-align: right;');
            }

            this.label.allocate(childBox, flags);
        }
        
        ...
        
    }
    
    ...
    
}

...
```
ScreenshotRTL

ronyala:
Hi, working off of your solution I wrote a new function for handling LTR text direction that also better takes into account special characters (who do not really have a direction):
```js
const _reLetterRtl = new RegExp("\\p{Script=Hebrew}|\\p{Script=Arabic}", "u");
const _reLetter = new RegExp("\\p{L}", "u");
const getTextDirection = function(text) {
    for (const l of text)
        if(l.match(_reLetterRtl))
            return Clutter.TextDirection.RTL;
        else if(l.match(_reLetter))
            return Clutter.TextDirection.LTR;
    return Clutter.TextDirection.None;
}
```
Perhaps this could be added to util.js as it seems useful in general, in any case where you dont want to align all texts to the same direction regardless of their script.
You're welcome to check the commit I made :)

mtwebster:
Hi, I don't understand the reason for getTextDirection() - widget.get_default_direction() or .get_direction() aren't correct here? 

You're using them other places, including at the top of the allocate function - why isn't that sufficient?

I'm not trying to be difficult, just to make sure I understand - as an English-only speaker, I'm probably underqualified here :)

ronyala:
I am using getTextDirection() to get the direction of the text itself, regardless of system or applet direction, so ltr text will be aligned to the left in an rtl system and vice versa :)
Both widget.get_default_direction() and .get_direction() are the system/widget directions and not the direction of the text.

AxXxB:
Have some layout issue with Mint-Y-Dark theme:
Screenshot_19

ronyala:
@AxXxB Actually during development I was aware that Mint-Y-Dark had some issues due to me changing the CSS, which is why I was using the cinnamon theme when testing. I should've pointed that out. I assumed that it's a matter of the rest of the themes somehow updating in the 'background', as I didn't find any documentation on how they are handled.
I can create PRs in Mint-Y and Mint-X manually if needed

mtwebster:
I wonder if we'd be better making a new style class name for the number - if we do this, all themes that don't have it defined should fall back to the default cinnamon theme (this is what the 'important' property does), so these changes won't depend on ours and everyone else's themes being updated.

mtwebster:
I wonder if we'd be better making a new style class name for the number - if we do this, all themes that don't have it defined should fall back to the default cinnamon theme (this is what the 'important' property does), so these changes won't depend on ours and everyone else's themes being updated.

mtwebster:
Actually it seems like it may be enough to remove the hardcoded width and height from the 'badge' constructor:
https://github.com/linuxmint/cinnamon/pull/10631/files#diff-ff87e15fbe5d4a462aeef966bf403d5543648b31289a55f4b835cdcfcb32f362R133-R134

If you do that, the badge scales to accomodate the text size:
image
(that's unmodified Mint-Y-Dark-Blue)

It doesn't look so good on a few user themes I tried, but it doesn't cut off the number at least.

ronyala:
I've added these changes to the PR.
I had some issues with my RTL layout, so while solving that I ended up taking the oppertunity to remove `y_anchor` from `numberLabel` and add updating of `x_anchor` in allocation, so the badge number is always centered (not a fixed offset like it was)

JosephMcc:
I'd much prefer that the badge number not be centered. It should stay offset like it is currently.

ronyala:
@JosephMcc I Reverted it

Issue #10258, opened 2021-08-07
The menu should support .desktop actions 


Lurux:
```
 * Cinnamon version: 5.0.4
 * Distribution:Mint 20.2
 ```

**Issue**

When right clicking the Firefox icon in the taskbar, custom .desktop actions associated with the application appear in the context menu (in this example, "new window" ans "new private window").

I think it would make sense if that was the case in the main Cinnamon menu, too.

**Steps to reproduce**

Right click the Firefox icon in the taskbar, and in the main menu, and compare the available entries
(This applies to other applications with custom .desktop actions, too)

**Configuration info, if applicable**

Tested on a clean Linux Mint Cinnamon installation

Issue #9470, opened 2020-07-18
Feature request: launch applications' desktop actios from the menu


n1c0saurio:
```
 * Cinnamon version 4.6.5
 * Distribution - Fedora 32
 * 64 bit
 ```

It would be useful to launch desktop actions directly in the Cinnamon Menu, via right-clicking on an application and/or made them searchable, like in Elementary's launcher.

**Expected behaviour**
Right click on, e.g., Firefox and being able to launch a new window or a private window, or search for "private window" and display the desktop action entry's from Firefox, Chromium or whatever other apps that match the search term.

ItzSwirlz:
GNOME's Menu is similar to that-but would this be something that goes in [https://github.com/linuxmint/cinnamon-menus](cinnamon-menus)?

n1c0saurio:
I see other issues related to the menu here, so I think there's no other place to post it. Should I close this and open in that repo?

ItzSwirlz:
Sorry for being a latie, you can keep it open but it might want to go to cinnamon-menus. Not 100% sure

PR #10654, opened 2022-02-04
Menu applet: Show application actions just like in window list

slowscript:
Show .desktop actions in right-click menu of menu applet the same way the window list does.

Screenshot:
menu-actions

The only thing I'm not sure about are the icons. Should I copy the ICON_NAMES array to the menu applet? Can we assume the window list applet is present or maybe should the array be moved somewhere globally accessible?

Issues #10258 and #9470


mtwebster:
Move it to `js/misc/util.js` and call it `DESKTOP_ACTION_ICON_NAMES`. It's probably better to have no icon if there's no match in that list. It looks like the action definition can specify an icon but it looks like it's a pain and it probably wouldn't look very good at that size.

slowscript:
I've moved the `ICON_NAMES` table to `util.js` and the key normalization logic to a new function `getDesktopActionIcon`. Actions in the menu now show the same icons as in the window list, or no icon if not found.
menu-actions2

Issue #10407, opened 2021-10-21
Question- Cinnamon Grouped Window List applet, is 6px "space" or padding between each window or icon normal?


bendover22:
```
 * Mint 20.1 desktop, Cinnamon 4.8.6
   - Please specify if you are using the daily builds PPA: NO
 Nvidia GeForce GT240, driver NVIDIA Driver Version: 340.108 64bit

This is a question - not a bug (AFAIK).
**Issue**
Is it "normal" in the Main Cinnamon Window List w/ App Grouping applet, that if you display full window buttons (all showing the app icon and part of the window title (in each "box"), that the separate boxes - for different apps, have a 6px space between each box / button?  See screenshot.

After further investigation, I answered my own question (see post #2).  The **6px space** is set in the **applet's appGroup.js file.**
**I'd LIKE TO CHANGE THIS FROM A QUESTION TO FEATURE REQUEST.**
In Cinnamon Grouped Window List applet, the fixed 6px "margin" on the panel between each open window icon (or full button w/ label) is quite a bit, as the icons also have padding or borders around them, which separates the icons pretty well w/o an extra 6px space. (Maybe 1 to 3px spacing).  Could a setting be added in the applet's config UI, to adjust the spacing ("margin") between label buttons or icons, instead of everyone having to use the same 6px spacing?

I've looked at several included Mint-X, Mint-Y themes in LM 20.1, and 3rd party theme, Cinnamox Heather.
I've played w/ many properties / values (in all mentioned themes) under selectors that sounded like they should control the 6px space, but no.  Not after restarting Cinnamon, or selecting different panel theme - then back to the current theme, or after reboot.  Many, many hours on this.

Oddly, on a 32 bit laptop running LM 19.3 XFCE, using ***exact* same Cinnamox Heather** theme vers. as the desktop w/ LM 20.1 Cinnamon, the XFCE panel's *individual* .grouped-window-list-item-box(es) don't have a 6px or any other space between every window list box.  Something probably is / is not in XFCE where it doesn't add that space.  

In XFCE4, the only "spacer" between panel "full length" open app windows , is if you allow them to have a 1px Left or Rt. border.
That's plenty of space to see where one (window box or just icons) ends & the next starts.

If settings in Cinnamon, the Grouped Window List app only shows a label on the focused box / button, & all other grouped (or single apps) show an icon, they *still* have the 6px spacing / padding, between each running app's **icons.**  

[...]

PR #10669, opened 2022-02-14
Grouped window list: Spacing is defined by CSS

AxXxB:
Adding ability to adjust a spacing between app groups via theme CSS. This is done as in window list applet but using following classes:

```
/* ===================================================================
 * Grouped window list (grouped-window-list@cinnamon.org)
 * ===================================================================*/
.grouped-window-list-box {
    spacing: 2px;
}
.grouped-window-list-box.vertical {
    spacing: 3px;
}
```

If there is no spacing defined in CSS, the old value of 6px is used.

Related to https://github.com/linuxmint/cinnamon/issues/10407

### DIFF
--- a/data/theme/cinnamon.css
+++ b/data/theme/cinnamon.css
@@ -1424,6 +1424,12 @@ StScrollBar StButton#vhandle:hover {
  * Grouped window list (grouped-window-list@cinnamon.org)
  * ===================================================================*/
 
+.grouped-window-list-box {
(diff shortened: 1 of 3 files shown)


mtwebster:
Hi can you squash this into a single commit? It's a small enough set of changes, and all for addressing the same issue.

Thanks

AxXxB:
@mtwebster Sure! Done.

Issue #10659, opened 2022-02-07
Power Manager: Duplicate battery entries make applet extremely wide


hillexed:
```
 * Cinnamon version (cinnamon --version) Cinnamon 5.0.7
 * Distribution - Linux Mint 20.2 Cinnamon
 * Graphics hardware *and* driver used
 * 32 or 64 bit
 ```
 * Attach ~/.xsession-errors, or /var/log/syslog: 
[xsession-errors.txt](https://github.com/linuxmint/cinnamon/files/8015052/xsession-errors.txt)

**Issue**
I have two batteries in my laptop. Sometimes the power applet shows two or more copies of the same charge levels, leading to the applet taking up more and more space on my taskbar. ![batteries_oh_no](https://user-images.githubusercontent.com/22244918/149026531-a8e18f35-1d1e-4abd-9e86-4ba930aaf442.png)

Plugging in a charger and unplugging the charger makes these duplicate entries sometimes, but not always, disappear. I recorded this video of what it looks like when I plug in and unplug a charger:

https://user-images.githubusercontent.com/22244918/152785220-5051dcbc-a7a1-485c-a854-356aca33fc3c.mp4

**Steps to reproduce**
Unsure. Plugging in and unplugging a charger seems to change the number of duplicates sometimes?

**Expected behaviour**
Only two batteries display in the taskbar

**Other information**
I recently replaced my external battery and this still happens.

**Configuration info, if applicable**
```{
    "labelinfo": {
        "type": "combobox",
        "description": "Display",
        "default": "nothing",
        "options": {
            "Show percentage": "percentage",
            "Show time remaining": "time",
            "Show percentage and time remaining": "percentage_time",
            "Hide label": "nothing"
        },
        "value": "percentage_time"
    },
    "showmulti": {
        "type": "switch",
        "default": false,
        "description": "Always show all batteries",
        "tooltip": "Check this to always show multiple batteries in the panel, regardless of whether one of them is marked primary.",
        "value": true
    },
    "__md5__": "db29025b0930e93063f745d0adcc675a"
}
```

I've attached the results of running upower --log here: 
[upower_log.txt](https://github.com/linuxmint/cinnamon/files/8015035/upower_log.txt)

 and the results of running `dbus-send --session --print-reply --dest=org.cinnamon.SettingsDaemon.Power /org/cinnamon/SettingsDaemon/Power org.cinnamon.SettingsDaemon.Power.GetDevices` here: 
[...]

Issue #10306, opened 2021-09-07
Power@cinnamon.org app displays double the number of batteries installed


Dongerrs:
```
 * Cinnamon version (cinnamon --version) 5.0.5
   - Please specify if you are using the daily builds PPA
     (https://launchpad.net/~linuxmint-daily-build-team/+archive/ubuntu/daily-builds).
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...) EndeavourOS (Arch based)
 * Graphics hardware *and* driver used - Lenovo T470 - Intel HD 620 - mesa
 * 32 or 64 bit 64Bit
 * Attach ~/.xsession-errors, or /var/log/syslog
 ```

**Issue**
Battery display/power@cinnamon.org/PowerManger will sometimes display double the number of batteries. Lenovo T470 has internal and removable battery. When config set to 'Always show all batteries' sometime on boot it will display 4 batteries. Turn off the setting for 'Always show all batteries'   correct the behaviour.

**Steps to reproduce**
Intermittent, but have two batteries in a machine and turn on 'Always show all batteries' config

**Expected behaviour**
Displays the correct number of batteries

**Other information**
Seems like could be a regression as earlier versions of cinnamon (4.8.6) running on debian 11 on same machine does not exhibit the behaviour.

**Configuration info, if applicable**
If this bug report is related to an Applet, Desklet or Extension, please paste (or use a pastebin service) the offending extension's settings, if it has any.  You can obtain this by opening its configuration, clicking the 'hamburger' button in the upper-right corner of the window, and selecting "Export to a file". **Please be sure to review the contents and remove any personal data it may contain.**

```

Please paste here
{
    "labelinfo": {
        "type": "combobox",
        "description": "Display",
        "default": "nothing",
        "options": {
            "Show percentage": "percentage",
            "Show time remaining": "time",
            "Show percentage and time remaining": "percentage_time",
            "Hide label": "nothing"
        },
        "value": "percentage_time"
    },
    "showmulti": {
        "type": "switch",
        "default": false,
        "description": "Always show all batteries",
        "tooltip": "Check this to always show multiple batteries in the panel, regardless of whether one of them is marked primary.",
        "value": true
    },
    "__md5__": "db29025b0930e93063f745d0adcc675a"
```

Dongerrs:
Just experienced it after a Suspend of the laptop

mtwebster:
[...]

PR #10670, opened 2022-02-14
power applet: fix device duplicates

AxXxB:
Fixes https://github.com/linuxmint/cinnamon/issues/10659, maybe also https://github.com/linuxmint/cinnamon/issues/10306


hillexed:
Wow, that was fast! Thank you! Maybe I'd add some comments explaining the difference between a device and a deviceItem, because I can see future devs getting confused?

mtwebster:
I think it's fine as is - looking at the entirety of `_devicesChanged` makes it clear enough.

PR #10734, opened 2022-04-04
Calendar Applet: single day event now shows end time

Pjsrcool:
In the current version of the Calendar Applet, in the Events view, the end times or end days is only shown for events that last more than one day. This small PR changes so that the end time is shown for events that are only one day. I personally find this feature to be useful for time management.

To the maintainers, I have some questions. Was it intentional to NOT show the end times for single day events? If so, should I add a toggle in the settings so that users can switch it on or off?

### DIFF
--- a/files/usr/share/cinnamon/applets/calendar@cinnamon.org/eventView.js
+++ b/files/usr/share/cinnamon/applets/calendar@cinnamon.org/eventView.js
@@ -972,6 +972,8 @@ class EventRow {
             } else {
                 // a timed event: "12:00 pm"
                 final_str += this.event.start.format(time_format);
+                final_str += ARROW_SEPARATOR;
+                final_str += this.event.end.format(time_format);
             }
 
             this.event_time.set_text(final_str);


mtwebster:
Hi, I wouldn't say it was intentional, just at the time I was going by my use pattern - I usually know how long something's going to be, but forget when it starts.

I'm fine with adding this - I don't think a setting is needed - we can add it later if it bothers anyone.

files/usr/share/cinnamon/applets/calendar@cinnamon.org/eventView.js
```diff
@@ -972,6 +972,8 @@ class EventRow {
             } else {
                 // a timed event: "12:00 pm"
                 final_str += this.event.start.format(time_format);
+                final_str += ARROW_SEPARATOR;
+                final_str += this.event.start.format(time_format);
```
mtwebster, line 976: Should be `.end` not `.start`?

Pjsrcool: My Bad. Just fixed it.

PR #10735, opened 2022-04-10
cinnamon-settings: noticeably accelerated startup time

okaestne:
This pull request boosts the startup time of cinnamon-settings noticeably by making the following changes:

1. ~Compile all python files in the `cinnamon-common` package during installation by using the python3 debhelper~ Edit: dropped as discussed below
2. Defer loading of heavier python packages (like `requests` or `imtools`, both ~50ms on my AMD Zen 3 dev machine). This has a noticeable effect, because all loaded python modules (`cs_*.py`) and their dependencies will be loaded at start time. Instead, import them before they are really used. Thus, the startup time will be shorter in any case.
3. The largest change attacks the before mentioned behavior that **all** modules are loaded in *any case*. This means that the "import (time) heavy" settings modules like the cinnamon spices x-lets are also load, if settings pages are started as standalone apps (i.e. using the menu shortcuts).

   The last commit therefore changes the way the the main `cinnamon-settings` is initialized. It now directly distinguishes between **starting into the "system settings"**, i.e. giving the overview over all settings modules, and the **standalone-startup of a given module** via the `argv`. Lazy-loading CCC and the real standalone modules (i.e. external binaries) are loaded in both cases as it made ~5ms difference while being more complex to handle.

As a result, I can see huge gains in startup performance. I tested both on my powerful development machine and on my Raspberry Pi 4. I didn't know how to reliably measure the time from start to visible GUI, so I just took the bash `time` internal and compared the CPU user and sys times. Here are some values (each one best of three measurements):

### Before

|             Command           |    Dev PC    |         RPi 4          | 
|----------------------------------|-----------------|-----------------------|
| `cinnamon-settings`                     | `user 0m0,350s`<br>`sys 0m0,058s` | `user 0m1,942s`<br>`sys 0,273ms` |
| `cinnamon-settings background` | `user 0m0,431s`<br>`sys 0m0,062s` | `user 0m2,203s`<br>`sys 0,352ms` |
| `cinnamon-settings user`             | `user 0m0,395s`<br>`sys 0m0,056s` | `user 0m1,959s`<br>`sys 0,325ms` |

### After

|             Command           |    Dev PC    |         RPi 4          | 
|----------------------------------|-----------------|-----------------------|
| `cinnamon-settings`                    | `user 0m0,206s`<br>`sys 0m0,027s` | `user 0m1,252s`<br>`sys 0m0,212s` |
| `cinnamon-settings background` | `user 0m0,258s`<br>`sys 0m0,035s` | `user 0m1,415s`<br>`sys 0m0,254s` |
| `cinnamon-settings user`            | `user 0m0,179s`<br>`sys 0m0,020s` | `user 0m1,053s`<br>`sys 0m0,203s` |

The commits contain also smaller other fixes and improvements to the existing code, hope that's okay for you.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/ExtensionCore.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/ExtensionCore.py
@@ -1,5 +1,6 @@
 #!/usr/bin/python3
 
+from functools import lru_cache
 import os
 import re
 import html
@@ -30,20 +31,22 @@ ROW_SIZE = 32
 
 UNSAFE_ITEMS = ['spawn_sync', 'spawn_command_line_sync', 'GTop', 'get_file_contents_utf8_sync']
 
-curr_ver = subprocess.check_output(['cinnamon', '--version']).decode("utf-8").splitlines()[0].split(' ')[1]
-curr_ver_elements = curr_ver.split(".")
-curr_ver_major = int(curr_ver_elements[0])
-curr_ver_minor = int(curr_ver_elements[1])
-
 LANGUAGE_CODE = "C"
 try:
     LANGUAGE_CODE = locale.getlocale()[0].split("_")[0]
 except:
     pass
 
+
+@lru_cache(maxsize=None)  # fetch only once
+def get_cinnamon_version():
+    version_str = subprocess.check_output(['cinnamon', '--version'], encoding="utf-8").split()[1]
+    return [int(part) for part in version_str.split(".")]
+
+
 def find_extension_subdir(directory):
-    largest = ['0']
-    curr_a = curr_ver.split('.')
+    largest = [0]
+    curr_a = get_cinnamon_version()
 
     for subdir in os.listdir(directory):
         if not os.path.isdir(os.path.join(directory, subdir)):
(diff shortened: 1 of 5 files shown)


Fantu:
@okaestne thanks for the improvements
about dh-python was always keeped in debian package and I readded here in muffin rebase branch when I did some packaging improvements (as I saw it was removed): https://github.com/linuxmint/cinnamon/commit/6c93a890b3701934615f1d0dc68b156d08c9c7c9
about "--no-ext-rename" (in dh_python3) is because you spotted an issue without it? and should not be the path (usr/share/cinnamon) absolute instead relative? the compile is not done in package build but in package install (is in postinst/postrm)

okaestne:
@Fantu thanks for your quick feedback. Should I drop the PyCompile commit in favor of your commit? The `--no-ext-rename` was just blindly taken from some other package tbh. It definitely works like this, also with the path.

Fantu:
there aren't library files in the path or other that can cause unexpected cases where "--no-ext-rename" is useful (FWIK) and there was not issue spotted or reported by users in debian package related to python compiled files
maybe it is better to keep the one I had put back which is more tested and same as the debian packages (so there are also fewer differences)

okaestne:
@Fantu Dropped it :)

PR #10747, opened 2022-04-16
sound applet ui overhaul

ssmid:
I made some changes to the sound applet. What do you think?

- all volume controls from context menu moved to main menu
  - Within 5 years of using Linux Mint I haven't found the application volume quick setting, just because I wasn't aware of the context menu. I'm pretty sure there are others.
- mute switches replaced by mute button for each slider
  - The slider's icons are now mute buttons, therefore the switches aren't necessary anymore.
- application volume sliders also show the application names
  - Often, applications only have the default icon and you have to hover over them for a tooltip to find the right one. Now, their names are shown above the slider.
- input device controls not hidden anymore when no one is recording
  - for example when using Audacity, you had to click the record button before the applet was showing the input controls
- menu rearranged
- a few refactorings

Other things that could be improved:
- merge submenus "launch player" and "choose active player" into one
- hovering mute button should have visual effect, as an indicator it is clickable (how?)
- remove bug "click to the left of icon triggers slider"

Affected: #10620

## old vs new:
5_comparison

### DIFF
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -81,11 +81,19 @@ class ControlButton {
 }
 
 class VolumeSlider extends PopupMenu.PopupSliderMenuItem {
-    constructor(applet, stream, tooltip, app_icon) {
+    /**
+     * A volume slider either for applications or sound input/output
+     * @param applet This sound applet
+     * @param stream A sound stream to connect the slider to. Nullable
+     * @param text An optional heading. Nullable
+     * @param tooltip The text which will show up when the users hovers over it. The slider's current value will be appended as percentage
+     * @param app_icon The icon of the application or "null" when used for input/output
+     */
+    constructor(applet, stream, text, tooltip, app_icon) {
         super(0);
         this.applet = applet;
 
-        if(tooltip)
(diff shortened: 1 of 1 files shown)


mtwebster:
Hi,

I'll be honest, I'm not crazy about this - it makes the popup way too busy. 

I'd be ok with merging the launch/choose player sections, and you could make an option to always show the input device chooser (maybe pair it with always showing the input mute switch.

Overall, these other controls are in the secondary menu because they're not likely to be needed by most users, and it lets us keep the primary menu uncomplicated. I can appreciate some people want more control, and this is what [spices](https://cinnamon-spices.linuxmint.com) are for.

ssmid:
Thanks for your honesty. This was just an idea of how the menu could look like, I'm open for suggestions. I can see your point with this approach being too busy, I'll make another attempt and if you think it doesn't fit the average user, I'll just make a spice out of it.

Regarding the **launch/choose player** section: As you can see in the first screenshot, those were already there. In my opinion though, at least the hard-coded "launch player" submenu could be removed. I don't think most users will click onto a speaker symbol in the bottom panel to start they're favourite music player. It's three clicks vs one or two.

About the **mute buttons**: Those don't take up any more space or increase complexity, and are more intuitive imho. Maybe, you want to keep them as well.

Have a great day.

PR #10794, opened 2022-05-17
Fix: Missing icon causes output sound device to not show in list

### LINKED ISSUE #10778: Pulseaudio virtual simultaneous output sink not visible in cinnamon settings
* Cinnamon 5.2.7
 * Arch linux

**Issue**
Pulseaudio's "Simultaneous output" option creates a virtual output sound device to send audio to all output devices at once. After updating my system a few days ago, this virtual device is no longer visible in cinnamon.

In the below screenshot, you can see that the virtual output is still visible in both discord and pavucontrol, but is not shown in cinnamon's sound panel. Prior to upgrading, the virtual device was visible in cinnamon's settings.

Screenshot from 2022-05-06 10-00-27

If I configure the virtual simultaneous output as the default sink via pavucontrol and then open the cinnamon sound settings, I observe:
- None of the visible output devices are highlighted (as expected since the virtual device is default but not visible)
- Adjusting the volume in the cinnamon settings, the change is reflected to the virtual device in pavucontrol
- Doing the sound test in the cinnamon settings, the sound being is played on the first sink in the list, instead of the default virtual device.
 
**Other information**
`.xsession-errors`: 
[xsession-errors.txt](https://github.com/linuxmint/cinnamon/files/8641930/xsession-errors.txt)

Here is the list of packages which were upgraded right before this started occurring. Cinnamon itself wasn't updated, so I'm not sure what packages I would try downgrading to test.

[upgrade-package-list.txt](https://github.com/linuxmint/cinnamon/files/8641740/upgrade-package-list.txt)

ion201:
Fixes #10778 

The main fix is in the first commit: https://github.com/linuxmint/cinnamon/commit/1fb49a95d8391a5a2a2746aabaaed45ff2febfe5. If the desktop theme doesn't have the `audio-card` icon, an exception is raised and the device would not get added to the list. I didn't know what to use instead (does IconTheme have an appropriate icon we could use that's guaranteed to exist??), so I left it empty. This will result in text only with no icon.

I also noticed that this file had some style issues too, so I also refactored the file to:
- Use `is None` instead of `== None`
- Renamed local variables which were masking the built-in functions `type` and `id`

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_sound.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_sound.py
@@ -708,10 +708,10 @@ class Module:
             if (iconName is not None and "bluetooth" in iconName):
                 icon = iconTheme.load_icon("bluetooth", 32, 0)
             elif iconTheme.has_icon("audio-card"):
+                # The audio-card icon was removed from adwaita, so may be absent in the current theme
                 icon = iconTheme.load_icon("audio-card", 32, 0)
             else:
-                # Default icon?
-                pass
+                icon = iconTheme.load_icon("sound", 32, 0)
 
         getattr(self, direction+"DeviceList").append([device.get_description() + "\n" +  device.get_origin(), "", False, deviceId, icon])
 
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_sound.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_sound.py
@@ -707,8 +707,11 @@ class Module:
         if icon is None:
             if (iconName is not None and "bluetooth" in iconName):
                 icon = iconTheme.load_icon("bluetooth", 32, 0)
-            else:
+            elif iconTheme.has_icon("audio-card"):
                 icon = iconTheme.load_icon("audio-card", 32, 0)
+            else:
+                # Default icon?
+                pass
 
         getattr(self, type+"DeviceList").append([device.get_description() + "\n" +  device.get_origin(), "", False, id, icon])
 
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_sound.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_sound.py
@@ -90,12 +90,12 @@ class Slider(SettingsWidget):
         self.set_spacing(5)
         self.set_margin_bottom(5)
 
-        if sizeGroup == None:
+        if sizeGroup is None:
             sizeGroup = Gtk.SizeGroup.new(Gtk.SizeGroupMode.HORIZONTAL)
 
-        if step == None:
+        if step is None:
             step = (maxValue - minValue) / 100
-        if page == None:
+        if page is None:
             page = (maxValue - minValue) / 10
         self.adjustment = Gtk.Adjustment.new(value, minValue, maxValue, step, page, 0)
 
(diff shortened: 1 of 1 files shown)


smurphos:
It's odd that the error is the linked issue showed up in any case - Mint-Y-Dark-Red inherits from Mint-Y-Dark,Adwaita,gnome, & hicolor and audio-card exists in Adwaita and gnome. Adwaita is a cinnamon dependency so should always be installed when Cinnamon is installed.

ion201:
Right, I was so focused on getting my system working, I forgot to root cause the failure...

Thanks for poking me, I double checked and found this change:
```
adwaita-icon-theme-41.0-1-any.pkg.tar.zst
------------------------------------------------------
> find /usr/share/icons/Adwaita/ -name '*audio-card*'
/usr/share/icons/Adwaita/32x32/legacy/audio-card.png
/usr/share/icons/Adwaita/32x32/devices/audio-card-symbolic.symbolic.png
/usr/share/icons/Adwaita/96x96/devices/audio-card-symbolic.symbolic.png
/usr/share/icons/Adwaita/24x24/legacy/audio-card.png
/usr/share/icons/Adwaita/24x24/devices/audio-card-symbolic.symbolic.png
/usr/share/icons/Adwaita/scalable/devices/audio-card-symbolic.svg
/usr/share/icons/Adwaita/22x22/legacy/audio-card.png
/usr/share/icons/Adwaita/48x48/legacy/audio-card.png
/usr/share/icons/Adwaita/48x48/devices/audio-card-symbolic.symbolic.png
/usr/share/icons/Adwaita/64x64/devices/audio-card-symbolic.symbolic.png
/usr/share/icons/Adwaita/16x16/legacy/audio-card.png
/usr/share/icons/Adwaita/16x16/devices/audio-card-symbolic.symbolic.png
```

```
adwaita-icon-theme-42.0+r1+gc144c3d75-1-any.pkg.tar.zst
------------------------------------------------------
> find /usr/share/icons/Adwaita/ -name '*audio-card*'
/usr/share/icons/Adwaita/32x32/devices/audio-card-symbolic.symbolic.png
/usr/share/icons/Adwaita/96x96/devices/audio-card-symbolic.symbolic.png
/usr/share/icons/Adwaita/24x24/devices/audio-card-symbolic.symbolic.png
/usr/share/icons/Adwaita/scalable/devices/audio-card-symbolic.svg
/usr/share/icons/Adwaita/48x48/devices/audio-card-symbolic.symbolic.png
/usr/share/icons/Adwaita/64x64/devices/audio-card-symbolic.symbolic.png
/usr/share/icons/Adwaita/16x16/devices/audio-card-symbolic.symbolic.png
```

Updating this PR appropriately now...

ion201:
After some more searching, I see that the root cause of this is actually https://gitlab.gnome.org/GNOME/gtk/-/issues/4754

The icons were already back in gtk4 a couple months ago, but the release with this fix for gtk3 was only staged 18 hours ago: https://gitlab.gnome.org/GNOME/gtk/-/tags/3.24.34

So presumably this issue will be fixed upstream soon, but I'm going to go ahead and leave this PR open as-is in case others might agree that the updated logic (ie, use no icon rather than raise an exception) is preferable anyway.

mtwebster:
Ah I hate that bug. They don't think it is (or at least they didn't when first reported).

Try 'yast_soundcard' as a fallback - it's part of the 'gnome' icon set which should always be there. Otherwise just use 'sound' so we have something there.

mtwebster:
Ah I hate that bug. They don't think it is (or at least they didn't when first reported).

Try 'yast_soundcard' as a fallback - it's part of the 'gnome' icon set which should always be there. Otherwise just use 'sound' so we have something there.

ion201:
Thanks for the suggestions. I don't see "yast_soundcard" here, so can use "sound".

PR #10805, opened 2022-05-26
Fix docs and re-enable

collinss:
We may or may not want `st-background-effect.h` and/or `cinnamon-recorder.h` to be included in the docs in the future, but for now, docs without them is still better than no docs at all.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -156,3 +156,15 @@ Description: Debugging symbols for the Cinnamon desktop
  and flexible.
  .
  This package contains the debugging symbols.
+
+Package: cinnamon-doc
+Section: doc
+Architecture: all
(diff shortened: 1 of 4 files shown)


Fantu:
in the case you don't already know the faster/easy thing is use "git rebase -i linuxmint/muffin-5.2-test" (after the "git fetch linuxmint") keeping only the commit "Fix docs and re-enable" line and deleting the others so do a rebase on top the updated muffin-5.2-test branch

collinss:
@Fantu Those extra commits weren't there when I submitted the pull request. I suspect it's because the muffin-5.2-test branch was force pushed twice yesterday.

PR #10928, opened 2022-07-13
fix unresponsive upper pixel in top window list

olebowle:
With cinnamon 5.4.2 (possibly some versions before as well) on Archlinux, the uppermost pixel of the window-list applet is unresponsive, if the bar orientation is set to the top. All other orientations (bottom, left, right) seem fine.

Not sure if this is the right way to do it, but adding `1` fixes the issue.


JosephMcc:
Have you tried changing themes? This isn't reproducible here using the Mint-Y theme.

olebowle:
Minty leaves me with two dead pixels on the top. New-Minty works without an issue. With my only pre-installed `cinnamon` (see screenshot) I have one dead pixel.
Bildschirmfoto vom 2022-07-14 16-10-45

JosephMcc:
Ok. This is just a bug in the default cinnamon theme I can look at fixing then. For other themes that you have an issue with, you will need to report bugs against them.

PR #10947, opened 2022-07-19
fixed cinnamon-looking-glass/page_log style

yutent:


JosephMcc:
What problem or issue are you fixing here? You don't offer any explanation for why you are doing this.

yutent:
> What problem or issue are you fixing here? You don't offer any explanation for why you are doing this.


It ruins the theme color design.

image


image

* ----- ---------------------------------------------------------------
**`There is no need to define the color here, it will inherit the theme design.`**

image

image

leeransetton (approved):

mtwebster:
I'm good with this, but you end up without any distinctions between types of logging output.

Instead of leaving them all the same, instead of no foreground color, you could use, for example:
```
foreground_rgba=self.get_style_context().lookup_color("error_color")[1]
```
for the error tag.

ref: https://github.com/linuxmint/mint-themes/blob/master/src/Mint-Y/gtk-3.0/gtk-dark.css#L3988-L3990

```
>>> x = Gtk.Box()
>>> x.get_style_context().lookup_color("error_color")
(True, color=Gdk.RGBA(red=0.988235, green=0.254902, blue=0.219608, alpha=1.000000))
```
For the trace color, leave it the normal color but maybe make it slightly bold (`Pango.Weight.SEMIBOLD`) or something else to make it stand out a bit.

mtwebster:
https://github.com/linuxmint/cinnamon/commit/36ba660a5f438ec85a6f4ccb6d90a24a1698317e

PR #10951, opened 2022-07-20
`main.js` refactor

Rudxain:
A little bit of refactoring. I avoided `for ... of` loops because I noticed only `forEach` (and the classic `for`) is used to iterate


Rudxain:
Oh ok, got it. So, should I remove the destructuring, and convert all `forEach`s into classic `for`s?

PR #10971, opened 2022-07-27
cs_default.py: Add entry for terminal exec-arg

### LINKED ISSUE #10938: Users should be able to set exec-arg key for the default terminal in "Preferred Applications"
```
 * Cinnamon version (cinnamon --version) Cinnamon 5.2.7
 * Distribution - Fedora 36
 * 64 bit
 ```

**Issue**
After setting the default terminal in "Preferred Applications" (`cinnamon-settings default`), "Open in Terminal" uses the chosen terminal as expected. However, applications that are set to open in the terminal no longer opens.

This is because "Preferred Applications" only sets the `exec` key in the `org.cinnamon.desktop.default-applications.terminal` GSettings schema. The `exec-arg` key of that schema is not set, and defaults to the flag that gnome-terminal uses, which is `--`. This means that switching the default terminal to any terminal that does not use `--` to specify a command would make "Run in Terminal" actions stop working in nemo, for instance. 

**Steps to reproduce**
1. Write a Nemo action that includes `Terminal=true`.
2. Install a terminal that uses anything other than `--` to specify a command. (For instance, [alacritty](https://alacritty.org/) uses `-e`.)
3. Set that terminal as your default terminal in "Preferred Applications".
4. Trigger the action.

**Expected behaviour**
The selected terminal opens, and runs the action.

**Other information**
One workaround for this issue is to run `gsettings set org.cinnamon.desktop.default-applications.terminal exec-arg <flag>` in the terminal.

### PR DESCRIPTION
**Pull request by veractor**
This adds a text box next to the terminal chooser button in "Preferred Applications", to specify the command-line argument to provide the terminal with a command to execute. The text box I admit takes up a lot of space, and that's because I have no idea how to allocate a specific proportion of a Gtk.Box to its child widgets. I think it still looks good enough though.

Fixes #10938.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_default.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_default.py
@@ -280,6 +280,23 @@ class DefaultTerminalButton(Gtk.AppChooserButton): #TODO: See if we can get this
         command_key = self.active_items[index_num]
         self.settings.set_string("exec", command_key)
 
+class TerminalExecArgEntry(Gtk.Entry):
+    def __init__(self):
+        super(TerminalExecArgEntry, self).__init__()
+
+        self.connect("changed", self.onChanged)
+
+        self.settings = Gio.Settings.new(TERMINAL_SCHEMA)
+        self.key_value = self.settings.get_string("exec-arg")
+
+        self.get_buffer().set_text(self.key_value, -1)
+
+        self.set_placeholder_text("exec-arg")
+        self.set_tooltip_text(_("Command-line option for your terminal to execute a passed-in command."))
+
+    def onChanged(self, entry):
+        self.settings.set_string("exec-arg", entry.get_buffer().get_text())
+
 class DefaultCalculatorButton(Gtk.AppChooserButton):
     def __init__(self):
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: files/usr/share/cinnamon/cinnamon-settings/modules/cs_default.py [hunk 1]
```diff
@@ -552,9 +569,15 @@ def on_module_selected(self):
             widget = SettingsWidget()
             button = DefaultTerminalButton()
             label = MnemonicLabel(_("Te_rminal"), button)
-            size_group.add_widget(button)
+            entry = TerminalExecArgEntry()
+
+            box = Gtk.Box()
```
mtwebster, line 574: Try making this a vertical box - I think it'll be cleaner keeping the picker button the same size as the rest on the page, with the entry below.

veractor: Done! Thanks for the suggestion.

mtwebster:
This is merged, but I had to do it manually:

https://github.com/linuxmint/cinnamon/commit/d6d8f3453c847e1e59803ea564ebedc5c1b0f096

Thanks

Issue #10881, opened 2022-06-26
locate-pointer: concentric circle animation barely or not visible within dark themes or backgrounds


DirkHaar:
```
 * Cinnamon version 5.2.7
   - no daily builds, standard packages
 * Distribution - (Mint 20.3 Una
 * Graphics hardware Intel 620
 * 64 bit

**Issue**
When using the locate-pointer function on dark themes, the animation of dark concentric circles is barely or sometimes impossible to be seen, making it absolutely useless.

**Steps to reproduce**
In "Mouse ad Touchpad" activate "Show position of pointer when the Control key is pressed". 
Try to see the animation on any dark background. It's not necessary to install a dark theme to see the current animation is wrong as it is only useful on a bright background.

**Expected behaviour**
The animation is needed to change its colours, no matter if it changes between black and white, as a colour gradient or rainbow or at best as an adjustable settings, according to mouse arrow theme or colour theme.

https://user-images.githubusercontent.com/12502925/175824907-c954392c-0b29-4940-94c7-75f4dc2d0a72.mov

Secret-chest:
> ```
>  * Cinnamon version 5.2.7
>    - no daily builds, standard packages
>  * Distribution - (Mint 20.3 Una
>  * Graphics hardware Intel 620
>  * 64 bit
> 
> **Issue**
> When using the locate-pointer function on dark themes, the animation of dark concentric circles is barely or sometimes impossible to be seen, making it absolutely useless.
> 
> **Steps to reproduce**
> In "Mouse ad Touchpad" activate "Show position of pointer when the Control key is pressed". 
> Try to see the animation on any dark background. It's not necessary to install a dark theme to see the current animation is wrong as it is only useful on a bright background.
> 
> **Expected behaviour**
> The animation is needed to change its colours, no matter if it changes between black and white, as a colour gradient or rainbow or at best as an adjustable settings, according to mouse arrow theme or colour theme.
> 
> https://user-images.githubusercontent.com/12502925/175824907-c954392c-0b29-4940-94c7-75f4dc2d0a72.mov
> 
> 
> 

Maybe it could be inverted.

Also I would like to see a hotkey setting - I would very much prefer it on right control so it doesn't interfere with shortcurs (or sprinting in Minecraft)

DirkHaar:
Inverting wouldn't solve the problem making the circles brighter and invisible on white/light grey background.
"Inverting the background" or kind of flickering of the pointer is on would be okay.

[...]

PR #10975, opened 2022-07-28
Add back support for the locate pointer

JosephMcc:
We lost this in the mutter rebase. This adds back the functionality and makes it
themable in the cinnamon theme.

Requires:
https://github.com/linuxmint/cinnamon-desktop/pull/213
https://github.com/linuxmint/muffin/pull/627

### DIFF
--- a/data/theme/cinnamon.css
+++ b/data/theme/cinnamon.css
@@ -76,6 +76,13 @@ StScrollBar StButton#vhandle:hover {
     font-weight: normal;
     text-align: center;
 }
+.ripple-pointer-location {
+  width: 50px;
+  height: 50px;
+  border-radius: 25px 25px 25px 25px;
+  background-color: rgba(160,160,160,0.3);
+  box-shadow: 0 0 2px 2px rgba(160,160,160,1.0);
+}
 /* ===================================================================
  * Shared button properties
  * ===================================================================*/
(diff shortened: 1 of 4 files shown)


DirkHaar:
Please consider #10881 - in 20.3 the concentric circles where not visible on dark backgrounds.
The different circles sizes should have alternating brightness or colour.

data/theme/cinnamon.css
```diff
@@ -76,6 +76,13 @@ StScrollBar StButton#vhandle:hover {
     font-weight: normal;
     text-align: center;
 }
+.ripple-pointer-location {
+  width: 50px;
+  height: 50px;
+  border-radius: 25px 25px 25px 25px;
```
rcalixte, line 82: @JosephMcc This PR failed due to this being a redundant longhand value.
Changing it to `border-radius: 25px;` should fix it.

rcalixte:
> Please consider #10881 - in 20.3 the concentric circles where not visible on dark backgrounds. The different circles sizes should have alternating brightness or colour.

Maybe an oscillation of black to white to grayscale? Would those three color palettes suffice? Maybe make it an option. I agree that sometimes the black indicator can be tough to see.

data/theme/cinnamon.css
rcalixte, line 82: @JosephMcc This PR failed due to this being a redundant longhand value.
Changing it to `border-radius: 25px;` should fix it.

JosephMcc: The failure means nothing in this case. The css is valid written either way.

rcalixte: The failure means it's introducing a new issue though? If the check isn't needed, maybe remove or disable it then? Otherwise, it's a quick fix?

PR #10989, opened 2022-08-01
cs_keyboard.py, main.js, and windowManager.js: Separate the keybindings and keyboard menu options for switching workspace up/down and toggling window/workspace selection

rcalixte:


rcalixte:
This is related to:
linuxmint/cinnamon-desktop#214
linuxmint/muffin#628

mtwebster:
Hi, I'm really sorry I haven't reviewed this sooner - adding the Next tag made it disappear from my PR filter and I forgot to remove it after our release.

There are a couple of issues here:
- We don't actually support up/down workspaces but you're adding their keybindings to cinnamon settings. This is only going to generate bug reports I think.
- You're going to force anyone that may have modified their show windows/workspaces shortcuts to wonder why they've been reset.

We resisted in the past, but I'd probably be ok with having an (optional) 2-d workspace grid, and  if that were done, we could then add shortcuts. There are also a couple of ways to do proper migration of old setting keys to new if we get to that point.

So for now I'm going to hold off on this. If I get a chance at some point I may have a look at implementing a workspace grid as well.

rcalixte:
@mtwebster Would it makes sense to pull in the functionality from [this workspace grid applet](https://github.com/linuxmint/cinnamon-spices-applets/tree/master/workspace-grid%40hernejj)? There are already intersections/overlaps between this applet and Expo plus some other workspace functionality. The applet was updated for 5.4 compatibility as well.

rcalixte:
@mtwebster To make clear what I'm thinking, assuming this is a Cinnamon 6.2(?)/Linux Mint 22.0 target release, based on the current applet settings:

Workspace grid (2D) and switcher_01

- These settings would move to Workspace Options under current settings (with proper dependencies for when to be enabled/disabled):
  - Number of columns
  - Number of rows

(It might make sense to have the number of workspaces tucked under an option for Dynamic/Fixed Workspaces.)

- The workspace grid applet and the workspace applet included in Cinnamon would be combined. As a result:
  - Style and Scroll Wheel Behavior would also be conditional options for the combined workspace applet.

- The key binding option would be integrated into system key binding options by the related PRs.


The total end result would be additional system settings and a single applet in place of the current two.

Please feel free to let me know what you think!

clefebvre:
Hi @rcalixte,

First of all sorry this is taking so long. I think we should brainstorm on this before rushing into an implementation.

In our roadmap we're planning to review Scale/Expo and workspace management as a whole to see how it can be better integrated and exposed a little bit more to the user. Right now it's not something most people use or are aware of.

I'd recommend to put this on pause for now. We're tagging for 5.8 right now (Mint 21.2 is coming this summer) and this roadmap item hasn't been looked at.

Feel free to use https://github.com/orgs/linuxmint/discussions/categories/pull-requests?discussions_q= to start a discussion about this, not just about configuration but about workspace management in general.

rcalixte:
@clefebvre No need for apologies! I'm grateful for the consideration. I'd mentioned in [this comment](https://github.com/linuxmint/cinnamon/pull/10989#issuecomment-1314586994) that this would at best be for a major release (I'm indifferent as to which future one) to try to minimize disruption for users, with as much communication beforehand as possible. (Many users upgrade but many also do clean installs of xx.0 releases as I'm sure you're aware.)

I have the PRs running locally across multiple machines but have yet to implement anything beyond that related to this. I've also effectively taken up maintaining [the applet for this functionality](https://github.com/linuxmint/cinnamon-spices-applets/tree/master/workspace-grid%40hernejj). I have no problem rebasing to keep the PRs current (since I'm already rebasing to build locally anyways). You can probably guess that this means quite a bit to my workflow. 😅 

I like the idea of a discussion thread for this. I'll get that draft done and try to publish it soon. Thanks for the feedback!

rcalixte:
Discussion: linuxmint/discussions#37

Secret-chest:
Cam I still change the keyboard shortcut?

rcalixte:
> Cam I still change the keyboard shortcut?

Of course. No restrictions would be introduced.

Secret-chest:
> Cam I still change the keyboard shortcut?
> 
> Of course. No restrictions would be introduced.
Well that's good, can't live without my Super+Tab and Shift+Super+Tab

mtwebster:
Had conflicts, I merged manually: https://github.com/linuxmint/cinnamon/commit/2de831f39392aef5e9d090da9af5b3eef7bbb436

PR #10993, opened 2022-08-02
settings-schema.json, applet.js and appGroup.js : Add an option to manage spacing on Grouped Window List applet

RBFraphael:
I've added an option to manage spacing between icons on Grouped Window List applet, so users can manage spacing as they want/need.


mtwebster:
Hi thanks for the contribution but this isn't something we want to add at this point. In fact it was one of the capabilities we removed when we first incorporated the GWL as our default window list - it was a very niche feature, and not something we felt was worth making configurable vs just choosing a good spacing and using it.

As of Cinnamon 5.4, this can be adjusted by themes.

PR #11090, opened 2022-08-20
cs-info: Add system icon

ItzSwirlz:
A vendor can disable this. It simply adds a Gtk Image under
the system info data which can be configured
by a gschema.

---
examples:
Screenshot from 2022-08-19 22-01-32
Screenshot from 2022-08-19 22-00-16
Screenshot from 2022-08-19 21-59-10
Screenshot from 2022-08-19 21-58-11

(Of course, the dimensions would have to be adjusted correctly by the vendor. I just downloaded these online, so like the LM logo would need configuring)

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_info.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_info.py
@@ -96,11 +96,11 @@ def getProcInfos():
 def getSystemIcon():
     schema = Gio.Settings(schema="org.cinnamon")
 
-    if schema.get_boolean("system-icon-enabled"):
-        iconPath = schema.get_string("system-icon-path")
-        return iconPath
-    else:
-        return
+    iconPath = schema.get_string("system-icon-path")
+    if iconPath is None:
+        return None
+
+    return iconPath
 
 
 def createSystemInfos():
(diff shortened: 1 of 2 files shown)


Secret-chest:
The LM logo should be colourful.

ItzSwirlz:
The LM logo is not a default preset, i just downloaded it online and set the gschema (which the distro maintainers will set)

files/usr/share/cinnamon/cinnamon-settings/modules/cs_info.py
```diff
@@ -93,6 +93,16 @@ def getProcInfos():
...
+def getSystemIcon():
+    schema = Gio.Settings(schema="org.cinnamon")
+
+    if schema.get_boolean("system-icon-enabled"):
+        iconPath = schema.get_string("system-icon-path")
+        return iconPath
+    else:
+        return
+
+
```
mtwebster, line 105: Let's just have the single 'system-icon-path' key - we don't need a separate 'enabled' setting.  If the path is empty, it's disabled.

Also, if your function returns something, it should always return something - even though in this case it amounts to the same thing, explicit is better, and a good practice. In this instance, it's best to return `None`.

ItzSwirlz:
gonna have to do some testing when I get home

ItzSwirlz:
sorry - it's all good

@leigh123linux and @Fantu - ping to ensure you are aware of the attention needed to the gschema next release

PR #11151, opened 2022-09-06
replace equality None check with identity None check

hduelme:
In python an identity (`is`) check should be used to check for `None`, instead of an equality check (`==`).

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_themes.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_themes.py
@@ -248,7 +248,7 @@ class Module:
 
                     dump = True
 
-                if theme_path == None:
+                if theme_path is None:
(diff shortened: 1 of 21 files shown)


files/usr/share/cinnamon/cinnamon-settings/modules/cs_themes.py
```diff
@@ -248,8 +248,8 @@ def refresh_chooser(self, payload):
 
                     dump = True
 
-                if theme_path == None:
-                    continue;
+                if theme_path is None:
+                    continue
```
mtwebster, line 252: Your other PR for semicolons broke this one, can you rebase it?

hduelme: @mtwebster rebased. Should be fixed

PR #11152, opened 2022-09-06
fix: variable name causing potential NameError

hduelme:
The typo causes, that `old_version` might not be initialized in case of an `KeyError`.
In that case `activity_logger.log("%s %s %s %s %s %s" % (log_timestamp, self.spice_type, action, uuid, old_version, new_version))` would fail with an `NameError`

### DIFF
--- a/python3/cinnamon/harvester.py
+++ b/python3/cinnamon/harvester.py
@@ -412,7 +412,7 @@ class Harvester():
 
     def write_to_log(self, uuid, action):
         new_version = "<none>"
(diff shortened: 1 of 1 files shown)


mtwebster:
Hi, can you start putting your PR description in the commit itself - it will be pasted automatically into the PR description, and it keeps us from having to do a 'squash' commit (so the PR number gets included for reference).

Thanks

hduelme:
Hi thanks for the feedback. From now on I will do so

PR #11154, opened 2022-09-07
remove trailing semicolons from cinnamon-calendar-server

hduelme:


mtwebster:
You try to switch between C and python all day long :)

Thanks

PR #11164, opened 2022-09-11
Add support for key bindings to Inhibit Applet

rcalixte:


rcalixte:
Just a note here:
I debated implementing notifications with the key bindings to notify of the action state. I'm leaning toward this being a better design and user experience but I am open to feedback.

rcalixte:
I've decided to switch this to a draft until I can implement proper OSD notifications with the key binding actions.

rcalixte:
OSD notifications have been implemented with the key binding actions so this is ready for review.

mtwebster:
Hi, looks mostly fine, just two things:

- Can you make the notifications disabled icon more consistent with the inhibit one - using the x (with the 'error' symbolic class) in the corner rather than it being part of the bell itself. This style is used in a few other places as well
- Please put the whitespace changes in a separate commit.

Thanks

rcalixte:
>     * Can you make the notifications disabled icon more consistent with the inhibit one - using the x (with the 'error' symbolic class) in the corner rather than it being part of the bell itself. This style is used in a few other places as well

>     * Please put the whitespace changes in a separate commit.

Both of these are done. Please feel free to review @mtwebster. Thank you!

PR #11175, opened 2022-09-19
fix: Actually use `XDG_DATA_HOME` to download themes and icons instead of home

Gr3q:
Mainly to not clutter up the user home folder when downloading themes, it was always creating `~/.themes`. `~/.icons` was created when a theme had extra icons.

Changes:
* Change themes install dir to `XDG_DATA_HOME/themes`
* when we initialize the `Spice_Harvester` we only ensure (create) the `install_folder` not the other directories
* Make sure the default icons dir is in `XDG_DATA_HOME`

Checked that themes still load from old location.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
@@ -159,8 +159,9 @@ class Spice_Harvester(GObject.Object):
             self.enabled_key = 'enabled-%ss' % self.collection_type
 
         if self.themes:
-            self.install_folder = '%s/.themes/' % (home)
-            self.spices_directories = (self.install_folder, )
+            self.install_folder = os.path.join(GLib.get_user_data_dir(), 'themes')
+            old_install_folder = '%s/.themes/' % (home)
+            self.spices_directories = (self.install_folder, old_install_folder)
         else:
             self.install_folder = '%s/.local/share/cinnamon/%ss/' % (home, self.collection_type)
             self.spices_directories = ('/usr/share/cinnamon/%ss/' % self.collection_type, self.install_folder)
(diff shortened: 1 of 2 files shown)


fredcw:
This seems to break theming with gtk2 apps, eg. Hexchat, GIMP (when set to use system theme in preferences), gwaterfall, hardinfo. It seems gtk2 can't find themes if not in ~/.themes

Gr3q:
I wonder `~/.gtkrc-2.0` can be configured to include another folder if Cinnamon still uses it. I guess I'll have to look into it, otherwise this will have to be reverted:(

Gr3q:
@mtwebster I've not found a way to include additional theme/icon folders for gtk2. 

To fix this one of 2 things I can do:
* Flip the folder ordering of this PR. Future downloaded themes will work. Manual intervention is still required, themes downloaded to `.local` will still not work.
* I add code to migrate anything from `.themes` if it's not a symlink to `.local/share/themes`, then delete `.themes` and create a symlink: `ln -s ~/.local/share/themes .themes`. Manual intervention is not required, but if users manually downloaded themes before into `.themes` they might be confused.

@fredcw for now this what I can recommend for you, just do `ln -s ~/.local/share/themes .themes` until it's fixed

fredcw:
I would go with the first option, the second option just throws good money after bad. Since I haven't seen anyone else complain and it's only a theming issue with very few apps, I think this is a minor enough of a problem that no intervention is necessary. The manual intervention required is just to uninstall and then reinstall the theme, assuming that uninstalling the theme will remove it from whichever folder it's in as it does now. Thus the problem will resolve itself over time.

mtwebster:
I'm with the first also - I'd considered the second when you guys first brought this up but it's too clever and that usually means trouble :)

PR #11186, opened 2022-09-21
Use `XDG_STATE_HOME` for logfiles instead of `~/.cinnamon`

Gr3q:
Although I've not seen any of these logfiles on my system.

glass.log entry was removed because it doesn't exists anymore.

### DIFF
--- a/files/usr/bin/cinnamon-launcher
+++ b/files/usr/bin/cinnamon-launcher
@@ -12,7 +12,7 @@ import psutil
 import threading
 import gi
 gi.require_version('Gtk', '3.0')  # noqa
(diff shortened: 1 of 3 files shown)


mtwebster:
Can you try/except with the new GLib function, and fall back to the current location? The spec change is recent enough that not all current versions of glib support it yet (particularly LMDE5/Debian Bullseye).

Thanks

Gr3q:
Done, should I add some `print`s when it happens or is it ok like this?

mtwebster:
Nah that's just an opportunity for a bug report :) 

Thanks

Issue #11213, opened 2022-10-09
window-list@cinnamon.org : autosave the windows position


sangorys:
### Is your feature request related to a problem? Please describe.
Each time cinnamon starts or crashes, the position of the windows is lost. We need to manually assign them again

### Describe the solution you'd like
When we do drag and drop for windows, save the window positions to disk too (not only to RAM)

### Describe alternatives you've considered
Develop the "pin" function which locks the window position : 

### Additional context
I guess there is a .js file for window-list@cinnamon.org.
If you teach me where it is, maybe I should be able to develop it and propose it to you as pull request. I have started to develop my first applet. I don't think this feature is difficult to code

sangorys:
OK. Actually, I spend the evening to implement a solution and I have found a solution. It works on my computer, the order is kept after restarting Cinamon or Cinamon crash.

I don't know where is the location of window-list@cinnamon.org, I would like to propose a git pull.

Somebody to help me ?

mtwebster:
If you mean their position in the window list (*not* the actual window positions), the applet should already be doing this. If it's not, it's a bug.

sangorys:
> 

It can be consider as a bug !

1. If you change the window order by drag and drop
2. Then you restart cinnamon (alt F2 then r)
3. Then the Window order is lost

This issue is also in all Cinnamon Spices applets for window-list.

I proposed a solution with this pull request : https://github.com/linuxmint/cinnamon/pull/11217

PR #11217, opened 2022-10-10
window-list : autosave the windows position / order to disk to keep it good if Cinnamon restarts or crashes

sangorys:
I succeed to implement a solution to this issue : https://github.com/linuxmint/cinnamon/issues/11213

:)

Again, I started developping applets and javascript and git 2 weeks ago. So I am a complete newbie. Don't hesitate to improve what I have done and to give me advices

I am 80% sure there is the same bug on grouped-window-list


mtwebster:
As I told you in the original issue, this is already implemented in the applet. If it's not working correctly, it should be fixed, not re-implemented.

Also, as a general rule, we try to avoid disk i/o in Cinnamon as much as possible, and *especially* synchronous i/o, so this isn't really a viable solution.

sangorys:
> As I told you in the original issue, this is already implemented in the applet. If it's not working correctly, it should be fixed, not re-implemented.
> 
> Also, as a general rule, we try to avoid disk i/o in Cinnamon as much as possible, and _especially_ synchronous i/o, so this isn't really a viable solution.

OK. You are the expert. So, I have 2 points : 

- I am going to submit an issue
- I am not a fan about disk i/o also but I don't know any other technical solutions. Do you have an idea about how to save the window order even after a crash without using the disk ?

alexmcwi:
I haven't looked at the code to figure out exactly what it is that @mtwebster had an issue with, but your code resolved a majorly aggravating issue I started having after replacing my old monitor (i.e. window-list was ignoring the order of the icons I set after power-cycling the monitor). So, whatever you did, I'm most grateful.

See https://forums.linuxmint.com/viewtopic.php?p=2332770#p2332770 for the issue I was having.

PR #11228, opened 2022-10-14
Add "Notifications disabled" icon clue to Inhibit Applet

BrunoNZ:
#### Problem:
There is no way to know if notifications are On or Off just by looking at Inhibit Applet icon, because there is no visual clue for it

#### Suggestion on how to solve it:
Adjust the applet to change the icon when notifications are disabled.

#### What i did:
* I added two new icons, using an "empty screen" variation
* I added `InhibitAppletIcon` class to manage icon name and statuses (inhibit and notifications), and also to serve as a helper for displaying the icon.
* I initialized the new class inside `CinnamonInhibitApplet`, which is accessible inside `InhibitSwitch` that also changes the icon
* I modified `set_applet_icon_symbolic_name` calls to use `InhibitAppletIcon.setAppletIcon`
* I added the call of `InhibitAppletIcon.toggleNotificationStatus` inside `notificationsSwitch` bind

#### Result:
Before:

[Peek 13-10-2022 23-44.webm](https://user-images.githubusercontent.com/1660320/195750475-cf93f4c7-c57c-497d-b983-55c7eabdb95b.webm)

After:

[Peek 13-10-2022 23-42.webm](https://user-images.githubusercontent.com/1660320/195750495-4bd350d0-381c-45b1-bdff-b061e1301462.webm)


mtwebster:
Hi could I get you to rebase this please?

This won't currently merge due some previous changes from https://github.com/linuxmint/cinnamon/pull/11164

Thanks

BrunoNZ:
For sure, @mtwebster .

BrunoNZ:
Done!

Ps: I've noticed that this new icon is not symmetric and also the line width is too thin compared to the others (cinnamon's defaults), but i don't believe my inkscape skills are good enough to be able to fix it.

rcalixte:
@BrunoNZ For what it's worth, I think your original implementation of the applet changes was fine. I'm not sure if the OSD needs to match the applet icons. If it would be easier to have them separate, I don't mind using new OSD icons and you can keep the applet changes as they were plus the rebase.

rcalixte (commented):
Just throwing in my two cents here... I think this was fine after the rebase and that this last commit can be dropped.

BrunoNZ:
@rcalixte , thank you for the feedback!
Actually, i think that the code is  cleaner and safer this, because it avoid two separate conditions and bugs concerning changing icon's filenames and forgetting to change it in these conditions. It was quite easy to do this changes, once the new class already has a method to return icon's filename.

However, i confess i didn't test it using OSD yet, because i don't know how to compile the whole project to enable it on this applet. So, if it somehow breaks your changes, i don't think it's necessary to undo this last commit. (or at least the code that returns icon's filenames, because the rest is necessary to fix some bugs of my code).

BrunoNZ:
@rcalixte , thank you for the feedback!
Actually, i think that the code is  cleaner and safer this, because it avoid two separate conditions and bugs concerning changing icon's filenames and forgetting to change it in these conditions. It was quite easy to do this changes, once the new class already has a method to return icon's filename.

However, i confess i didn't test it using OSD yet, because i don't know how to compile the whole project to enable it on this applet. So, if it somehow breaks your changes, i don't think it's necessary to undo this last commit. (or at least the code that returns icon's filenames, because the rest is necessary to fix some bugs of my code).

rcalixte:
@BrunoNZ If anything breaks with OSD, it'll be my fault first. 😅 

I was referencing your original PR with the applet icons modified as you attached in the .webm files earlier (and referenced below). I thought that implementation was cleaner. The icons I added for the OSD notification changes were strictly with OSD in mind which is possibly why they aren't playing well with the applet changes.

> After:
> Peek.13-10-2022.23-42.webm
I actually really liked this.

You can test locally by overwriting the files in `/usr/share/cinnamon/applets/inhibit@cinnamon.org` with your work and then restarting Cinnamon. You can also copy the `inhibit@cinnamon.org` folder from your branch to the applets download directory (the default is `~/.local/share/cinnamon/applets`) and then enabling it via Applets. Either option should work. Feel free to shout if you want an extra pair of eyes for testing!

BrunoNZ:
Ohhh, now i get it! haha I thought you were worried that these changes may brake something of your merged PR. My bad!!

Most of the changes were really necessary. I'm using my branch since i've opened the PR and i noticed that there was a bug when cinnamon was initialized with disabled notifications (silence notifications = true), because the default was enabled. So i had to force this initialization to sync it.

However, i totally agree that the code was cleaner before. I was thinking about refactoring this part of the initialization, to first initialize all the switches, and then initialize the icon passing all the values together, but there so many details around this that made me a bit afraid to modify it. I'll give it another go later today.

And about testing, I figured out how to test the applet itself, but i still don't know how to access the logs and i'm not sure if i have to compile something else to enable your hotkeys/OSD feature. About that, it would have helped me a lot to have a guide about how to run and debug applets. I've found [this wiki](https://projects.linuxmint.com/reference/git/cinnamon-tutorials/extension-system.html), but it only covers the source code.

BrunoNZ:
Done!!

I moved all notificationsSwitch logic to a new class (NotificationsSwitch), which greatly simplified the main class code. Now it is replicating the same style used for `InhibitSwitch`.

Everything is working fine, except line 173 (`this.toggled(this._switch.state);`).

Every time i execute `this.toggle_inhibit_power()`, in the end of `CinnamonInhibitApplet constructor`, the applet crashes. Commenting this line, it does not crash, but i think it doesn't work as expected. However, the same happens with the old code, so i assume there is something wrong with my system or there is a small bug on it.

rcalixte:
> And about testing, I figured out how to test the applet itself, but i still don't know how to access the logs and i'm not sure if i have to compile something else to enable your hotkeys/OSD feature.
The simplest way maybe is just to put a copy of the applets directory from your working area to the one under `~/.local/share/cinnamon/applets` and update with each commit.

For debugging, you can check the system logs under `/var/log/syslog` (something like `tail -F -n0 /var/log/syslog | grep cinnamon.*`) or you can also check the logs under Looking Glass (`cinnamon-looking-glass &`). Setting up a key binding for it that works for you will help if you want to go that path as well.

> Everything is working fine, except line 173 (`this.toggled(this._switch.state);`).
> 
> Every time i execute `this.toggle_inhibit_power()`, in the end of `CinnamonInhibitApplet constructor`, the applet crashes. Commenting this line, it does not crash, but i think it doesn't work as expected. However, the same happens with the old code, so i assume there is something wrong with my system or there is a small bug on it.
I'll take a look too!

BrunoNZ:
Thank you so much for the tips, @rcalixte !!! This CinnamonLookingGlass is awesome!!!

And i was wrong. It does work in master. I've broken something that is causing `sessionProxy` to be null  :thinking: 
I'll try to fix it.

BrunoNZ:
Finally!! With the help of the logs i've managed to figure out more things about the applet, and fix some bugs.
Now it's working!!!

I've noticed that the setAppletIcon is being called more times than it should, when `inhibitSwitch` is toggled, but i don't think it will make much difference.

Ohh, and btw, i've just noticed that there is a configuration panel for the applet. LOL!!!
I was trying to add the hotkey inside keyboard settings and i was not finding, obviously... 

[cinnamon-2022-11-07T222655-0300.webm](https://user-images.githubusercontent.com/1660320/200451919-3f194a84-92b2-4622-830e-fd0a4bf4a17a.webm)

rcalixte:
I've thought about this for a little bit... I'll admit it's not what I had in mind. Just my two cents, I still like the original applet changes better. This is still great though, but I think it might make the applet icons and the OSD notifications more ambiguous.

How would you feel about making it a pair of options to use one set of applet icons or the other (one option for the applet and OSD each)? I'm thinking this can be done using the settings import and ternary conditionals to keep things simple.

(I'm certainly not a design expert though. 😅)

mtwebster:
We definitely don't want options to choose icons here - that's overkill. But the bell icon is already used by the notification applet so it can't really be used here. At the moment I don't have any suggestions - there aren't a lot of pixels to work with there. I do want this functionality but it's probably going to have to wait for now.

BrunoNZ:
And what about the old icons? I've moved to these ones only because they were in the merged branch, but I think the old ones are pretty ok, maybe changing the "Z"s for couple of lines, mimicking the notifications bubbles. Thinking about what the applet does, inhibit notifications and energy/screen saver, makes sense to a "screen icon", because both of them translates to screen. A blank screen can be interpreted as no notifications/disturbance, and the "warning" tag means that energy saver is off, and user must be cautious about it.

@rcalixte ,I didn't quite understand your worries. Are you talking about the icons by themself, or the fact that the icon shown on OSD is ambiguous because both status are displayed at once. I'm not using the hotkeys/OSDs but i can see your point. 

I'll make some tests later today to how it goes.

rcalixte:
@BrunoNZ My concerns are about the applet tray icon being ambiguous in the latest version of the implementation versus the original implementation you proposed. (I can't remember despite watching the video several times which the solid bell correlates to and which one the hollow one does. This also feels like it would be confusing rather than intuitive. I actually found your original implementation rather intuitive.) Apologies if the new icons threw you off.

@mtwebster Would you be okay with the original approach that was offered here for the applet tray functionality? Also, I'm open to selecting a new pair of icons for the OSD to make the OSD notifications differentiate from the applet tray icons. That seems like it would wrap out this functionality?

@BrunoNZ @mtwebster In all, I think the result should be 4 icons for the applet tray icon and then 4 _different_ icons for the OSD notifications. Does that work for you? If so, I can work on the PR for the power management OSD notifications (as the notification OSD icons would already be different).

BrunoNZ:
@rcalixte , I'm sorry for the confusion! I really thought the new icons were the new standard, so i didn't bother to change them.

I've made some changes in the code, to make the OSD independent, and updated the icons to a variation of the old version. My fear is that it gets confused to a "battery icon", with the multiple stripes, but imho it different enough, once the screen frame is very familiar.

However, i think that using different icons in the OSD causes even more confusion, like shown it the video. Notice that when Notifications are off and I inhibit energy saver the icon used is the "full screen with the warning", which made me think that i accidentally reactivated notifications.

I'm trully in doubt about what the right answer to this problem is.

[cinnamon-2022-11-15T005748-0300.webm](https://user-images.githubusercontent.com/1660320/201824833-ab085b46-ca45-4005-825a-9eece0d00c0c.webm)

rcalixte:
@BrunoNZ No need for apologies! We'll get there! My two cents here, if you revert the OSD icons to what they were before and leave the tray applet icons, that would work?

BrunoNZ:
I did some experimenting with OSD exclusive icons, and came out with couple of options. I've first tried with "crossed out" icons, but i didn't quite like it, i don't know why.. ahhaha Then i went for a simpler solution, using a "faded out" approach, which pleased me a little more.

You was totally right about it, as it definitely looks more coherent now. Maybe the icons aren't in their best form yet, but it seems to be a good solution.

Captura de tela de 2022-11-15 13-51-41

Captura de tela de 2022-11-15 13-52-11

rcalixte:
@BrunoNZ Take a look at @mtwebster's comment [here](https://github.com/linuxmint/cinnamon/pull/11164#issuecomment-1281124091) pertaining to the 'error' symbolic class.

Would you be okay with replacing the faded icons with the consistent symbolic class? I'm wary of the two icons being too similar still but I'll leave the final say to the wisdom of the crowd.

PR #11244, opened 2022-10-31
build: Fix some hardcode path issues

bobby285271:
On NixOS all packages are installed into their own immutable prefix (i.e. we have `/nix/store/a8gwxrbqypfbiwxml1bmwgcf4hhp0hn3-cinnamon-common-5.4.12/bin/cinnamon` instead of `/usr/bin/cinnamon`), however some paths here are hardcoded and things break in NixOS, so far we make this desktop basically working by maintaining a set of `sed` commands ([listed here](https://github.com/NixOS/nixpkgs/blob/41004f4e0ea36afca9b6edd88c8d23f7e2c3ac86/pkgs/desktops/cinnamon/cinnamon-common/default.nix#L145-L175)) but it will be really nice if we can do this upstream so we can revisit less things on each bump. This PR contains **some** (but not all) of the fix.

Thanks for reviewing!

(Note that I moved some files from `files/` to `data/` in this PR, feel free to request changes / force push the PR if the file structure does not look good)


files/usr/bin/cinnamon-settings
```diff
@@ -120,9 +121,9 @@ if len(sys.argv) > 1:
             os.execvp("/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py", (" ",) + tuple(sys.argv[1:]))
     elif os.path.exists("/usr/share/cinnamon/cinnamon-settings/modules/cs_%s.py" % module):
         os.execvp("/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py", (" ",) + tuple(sys.argv[1:]))
-    elif os.path.exists("/usr/bin/cinnamon-control-center"):
-        os.execvp("/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py", (" ",) + tuple(sys.argv[1:]))
-    elif os.path.exists("/usr/bin/gnome-control-center"):
+    elif shutil.which("cinnamon-control-center"):
+        os.execvp("cinnamon-control-center", (" ",) + tuple(sys.argv[1:]))
```
mtwebster, line 125: This still needs to call cinnamon-settings.py - cinnamon-control-center is only used for debugging purposes mostly. 

This particular check doesn't seem all that useful in the first place - we have some C modules that are part of cinnamon-control-center, but we load them via cinnamon-settings.  We really should be looking at names of those modules themselves (in `...lib/.../cinnamon-control-center-1/panels`) rather than just the binary. But it's not a big deal in reality.

bobby285271: Done, thanks for the heads up!

data/services/meson.build
```diff
@@ -0,0 +1,13 @@
+service_files = [
+  'org.Cinnamon.Melange.service',
+  'org.Cinnamon.Slideshow.service',
+]
```
mtwebster, line 4: Should org.cinnamon.CalendarServer.service be included also?

bobby285271: I moved all dbus service files into the same folder (including the $\text{org.\textbf{\textcolor{red}{c}}innamon.CalendarServer.service}$ and the $\text{org.\textbf{\textcolor{red}{C}}innamon.HotplugSniffer.service}$) in 4c3b83e4532ef21f5b265ed30f8b8cac2f4d4603, PTAL.

mtwebster:
I was looking around at other hardcoded /usr/* paths (there are lots) - I suppose you could add the paths to config.js and maybe add some sort of config.py as well (I was thinking in the python3/cinnamon path so it's accessible by everything).

Thanks

Issue #11272, opened 2022-11-06
Typescript support (or at least declarations) inside Cinnamon


Gr3q:
### Is your feature request related to a problem? Please describe.

This includes allowing writing code in Typescript not just in Javascript. This includes:
1. Provide Typescript declarations under `imports.gi`, generating Cinnamon specific libs during build and provide for those we explicitly depend on. 
2. Provide Typescript declarations in the JS folder as `d.ts`
3. Allow writing code in typescript and generating `js` and `d.ts` files during build - this might need some extra effort so typescript knows about those libraries in `imports`.

### Describe the solution you'd like

1. I've got a [repository](https://github.com/Gr3q/GIR2TS) that does just that (output is stored [here](https://github.com/Gr3q/types-cjs)), it is pretty solid and used with some applets already. I'm happy to transfer over the repo to the Cinnamon team and still maintaining it myself.
2. This is a relatively easy task if the 1st step is done (technically 1. can be provided outside of Cinnamon itself for development) - downside is that these will be maintained alongside the JS files.  I'm more than happy to commit to write them and maintain them myself.
3. If imports are correctly resolve before (in TS) and after build (in JS) the codebase can be slowly converted .

### Additional context - why?

1. Personally I'd like to add more features to `js->ui->windowManager`, this is why I opened this request.
2. Could reduce regressions in the JS codebase because it will be typed.
4. Could provide better support for applets/extensions/desklets with correct types in editors with docstrings
5. Could reduce cognitive load to people working on the JS codebase and maybe help people open more PRs because the codebase functionality is easier to read, because everything in `imports` is not opaque as it is now or requires to read up documentation (if it exists - I'm looking at the Cinnamon specific libs in imports especially.)
6. Could enforce better code quality

### Possible problems
1. One more thing added to the toolchain Cinnamon is built with
2. (Slightly) higher level of entry of writing JS code and a bit more work to setup the dev environment 

Please respond if you think it's worth having a discussion about, especially that I can pledge to to most of the work around this.

ItzSwirlz:
This would be interesting though I'm not sure what usage this would cover other than being ideal. GJS's documentation mentions how it deals with C types

Gr3q:
`ts-for-gir` is for `gjs` or `node`, it generates declarations differently so it's not compatible with `cjs`, but maybe I'm wrong.

Also it doesn't generate docstrings.

NathanC:
[...]

PR #11276, opened 2022-11-06
[WIP] PoC Typescript support

Gr3q:
Currently with shitty structure for internal use (within Cinnamon repo) only.  c lib declarations are copy-pasted from [GIR2TS](https://github.com/Gr3q/GIR2TS) for now

TODO:

- [x] ts declaration generation from `.gir` files doesn't respect the `shadowed-by` property, needs to be fixed.
- [x] ts declaration generation from `.gir` files doesn't handle array types well in some cases, needs to be fixed.
- [ ] ts declaration generation from `.gir` files doesn't handle ~user_data and~ types for callbacks (generics), needs to be fixed
- [ ] Better internal structure on where to output `js` and `d.ts` files

https://github.com/linuxmint/cinnamon/issues/11272


ItzSwirlz:
Yeah.. all the dependencies are a slight problem. This would also mean that this TS would have to integrate with CJS otherwise everyone would have to install Node

But having to bootstrap the other Gir's isn't ideal - it's like what Gtk-Rs is doing and for one repo that's a bit crazy

ItzSwirlz:
GNOME's sound recorder just ported to TS. They are using Yarn and this tool: https://gitlab.gnome.org/ewlsh/gi.ts

It will probably be way cleaner if we go with the gi.ts route. There is still a types folder but IMO assuming that it is okay to have TS and it's bindings in the main cinnamon repository (meaning it is a requirement to be built and maintained by distributors) then I think this is feasible

Next move forward personally would be trying to work with gi.ts

Gr3q:
I'm in the process of trying out gi.ts but currently I have trouble building the declarations for Cinnamon with it.

For examples of spices, here they are:
https://github.com/linuxmint/cinnamon-spices-applets/tree/master/weather%40mockturtl
https://github.com/linuxmint/cinnamon-spices-applets/tree/master/radio%40driglu4it
https://github.com/linuxmint/cinnamon-spices-extensions/tree/master/gTile%40shuairan

They are either using the declarations in the applets repo (it was requested to be bundled for review) or the npm package. Both of them are basically the files included in this WIP

Gr3q:
I finally had time to check https://gitlab.gnome.org/ewlsh/gi.ts against the current weather applet codebase.

After a minor fix to `gi.ts`s codebase (Cinnamon's Meta doesn't have constructs it's checking for) it works fine. You can specifically provide where it should pull .gir files from and specify all top-level dependencies for Cinnamon, it will pull in the other libs theose depend on.

It doesn't inject them under imports, a declaration file is needed somewhere: 
```typescript
import * as Atk from "atk";
import * as St from "st";
import * as Gio from "gio";
import * as GLib from "glib";
import * as Soup from "soup";
import * as Clutter from "clutter";
import * as Gtk from "gtk";
import * as GObject from "gobject";

declare global {
    const _: (text: string) => string;
    const logError: (...args: any[]) => void;
 
    namespace imports {
        namespace gi {
            export {St}
            export {Clutter}
            export {Atk}
            export {Gio}
            export {GLib}
            export {Soup}
            export {Gtk}
            export {GObject}
        }
    }
}
```

The location of the file should depend on where we store/generate the declaration files or where injection actually happens.

It also can be broken up to more files - Typescript automatically merges `global` declarations - but in the end an applet or any codebase using them need to reference them in their `tsconfig.json` for it to work, so a standardized place would be useful.

mtwebster:
Closing this - I'm not interested in dealing with typescript any more than I'm already forced to with spices.

PR #11291, opened 2022-11-17
added type hints to util.py

hduelme:
I have added return type hints to both `util.py` files.

If you are interested I would also start adding type hints to variables, as it helps reading the code.

I also found, that in `cme/util.py` the method `getUserMenuXml` may fail with an `TypeError`. If the method `getSystemMenuPath` returns `None` then the concatenation in line 125 fails.


files/usr/share/cinnamon/cinnamon-menu-editor/cme/util.py
```diff
@@ -78,46 +78,46 @@ def getUniqueUndoFile(filepath):
             break
     return new_filepath
 
-def getItemPath(file_id):
+def getItemPath(file_id) -> str | None:
```
Gr3q, line 81: This syntax is only supported from 3.10+, you need to use `Optional` to support 3.6+

Gr3q (changes requested):

files/usr/share/cinnamon/cinnamon-menu-editor/cme/util.py
hduelme: @Gr3q , @mtwebster thanks for the feedback. 
I changed the `str | None` to `Optional[str]` in all places. Now it should support 3.6+

PR #11301, opened 2022-11-25
Add a setting to warp mouse pointer after Alt+Tab action completes

rcalixte:
* Add a second commit to fix whitespace

### DIFF
--- a/data/org.cinnamon.gschema.xml
+++ b/data/org.cinnamon.gschema.xml
@@ -429,6 +429,11 @@
       <summary>Show all windows from all workspaces</summary>
     </key>
 
+    <key type="b" name="alttab-switcher-warp-mouse-pointer">
+      <default>false</default>
+      <summary>Warp mouse pointer to the new focused window</summary>
+    </key>
+
     <key name="bring-windows-to-current-workspace" type="b">
       <default>false</default>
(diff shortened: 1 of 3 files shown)


mtwebster:
Works well (though this behavior would drive me nuts).

Not crazy about the setting description though - how about:
```
Warp mouse pointer to the new focused window
```

rcalixte:
> Works well (though this behavior would drive me nuts).

I tested in other modes and I agree. My use case/conditions are:
- Sloppy window focus
- Focus only raises window when clicked
- Alt+Tab focuses and raises window

With this model, I am able to work between multiple windows (across multiple screens and workspaces) without needing the mouse to operate.

Would it make sense to make this setting a dependency of the "Window focus mode" setting? (I ask this being wary of someone somewhere wanting it for an entirely different condition. 😅)




> Not crazy about the setting description though - how about:
> 
> ```
> Warp mouse pointer to the new focused window
> ```

Done.

mtwebster:
Nah it's ok here I think.

Btw, you updated the settings schema summary instead of the string in cs_windows.py (I maybe was too vague, sorry)

rcalixte:
> Btw, you updated the settings schema summary instead of the string in cs_windows.py (I maybe was too vague, sorry)

No need to apologize! Absolutely an oversight on my part. My apologies!

PR #11318, opened 2022-12-04
LM 21.1: Fix old thumbnail cache folder

Tibladar:
This was forgotten in PR #11176 and results in many useless requests where thumbnails are redownloaded because they can't be found in the new location but are saved to the old one.

Can be easily tested with `DEBUG=1 cinnamon-spice-updater --list-simple applet`:
Without this PR `Downloading thumbnail for ...` is always shown, even if the images got downloaded seconds ago

This reduces server load by a lot and should be picked up before stable release of Linux Mint 21.1


mtwebster:
Good catch! Thanks

PR #11335, opened 2022-12-20
Provide `.gir` files in cinnamon package

Gr3q:
Cinnamon should provide `.gir` files it generates at the correct place so they can be used by other tools.

Maybe it should be in `usr/share/cinnamon` like gnome-shell does it for it's shell, I'm not quite sure.


Gr3q:
It looks like to me that it's already there

Gr3q:
@mtwebster Added

Fantu:
After a RC bug in debian cinnamon package for fail to build after recent gobject-introspection changes, I'm trying to do changes for fix/improve package things relating to gir, also with the help of @smcv that give useful information
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1057391
for now, seems confirmed that gir files should be moved to private path and probably also new package needed
for other things I'm still doing some tests/checks and keep asking to be sure to do correct and good things, actual changes are in https://salsa.debian.org/cinnamon-team/cinnamon (I'm not sure will be definitive)
I hope to have same changes in both debian and upstream package in the future, so upstream opinions are welcome @clem @mtwebster

smcv:
> for now, seems confirmed that gir files should be moved to private path

What I said on the Debian bug is: because the library and the typelibs are in a private location (not on the default search path and only intended to be used within a narrow group of closely-related packages), the GIR XML should also be in a private location.

How GNOME Shell does it would be a good model, because GNOME Shell + Mutter is structurally similar to Cinnamon + Muffin.

> probably also new package needed

I would say this is probably not necessary, because the library is private.

Fantu:
@smcv thanks for reply
I was waiting for reply on my latest message in Debian bug but about cinnamon seems you replied here.
Recap:
https://salsa.debian.org/cinnamon-team/cinnamon/-/blob/master/debian/patches/move-gir-to-private-path.patch
and add in d/rules:
```
override_dh_girepository:
	dh_girepository -XCinnamon-0.1.gir -XSt-1.0.gir /usr/lib/$(DEB_HOST_MULTIARCH)/muffin /usr/share/cinnamon
```
will be ok and good also for upstream and next version with support of latest gobject-introspection changes and avoid "unwanted" -dev deps?

About cinnamon-screensaver to avoid -dev deps will be ok only this:
```
override_dh_girepository:
dh_girepository -XCScreensaver-1.0.gir
```
or new package should be added?

PR #11385, opened 2022-12-31
[WIP] Modularity to Panel's context-menu

danimesq:
This enables to load custom menu items at Panel's context menu.

Has used `imports` according to specification by [cinnamon-tutorials/importer](https://projects.linuxmint.com/reference/git/cinnamon-tutorials/importer.html). Based on [this Mint's forum question](https://forums.linuxmint.com/viewtopic.php?p=2163518).

This is a first step into making Cinnamon's menus and functions modular (also I would like to have modules at cinnamon-settings for its main menu).

----

New panel context-menu items could be installed by adding three lines in `js/ui/custom/panel/PanelContextMenu/addons.js`, as well as adding `js/ui/custom/panel/PanelContextMenu/{addon-folder}` containing the run.js with the functions that will be called by `js/ui/custom/panel/PanelContextMenu`'s addons.js then by Cinnamon's standard panel.js.

This measure, beyond being neat/cool for the sake of being cool, avoids risking users to break their whole Cinnamon installation just to add a well-needed shortcut to a menu.
And avoids them having to re-do their customizations at every update that overwrites the desired file.


danimesq:
Hey @mtwebster, I've added you permissions in the PRs repo so you can "presentially" fine tune some things if you think this proposal is viable.

lestcape:
@danimesq @danimesq forgetting the fact that what you do, seem to be not correctly, and taking in hand that the modularity is a synonyms of extension and the fact that Cinnamon already have extensions. I have one more general questions to you:

How should help the creation of some new types of extensions (modularity) for a thing that can be done directly from the current Cinnamon extension mechanism that is already include in Cinnamon?

What i said is that if you want to include a new option in the menu panel, you can create an normal Cinnamon extension and form there add this option to the place you want without need to modify any existing Cinnamon code. Is like what is doing the extension mentioned in the mint forum that you are taking as an example.

So, why you think as good the attempt to do what is already possible?

Also, the cinnamon-settings already have a mechanism of extensions (i.e. modules) that can be used/extended to achieved what you want in more/less the same way. The only problematic thing there, is that in this case and for an unknown reason to me the cinnamon developers decided don't give to the  cinnamon-settings extension the whole application context in an easy way. So, to do some things you need to parse the Gtk components starting from the Gtk.ApplicationWindow to find the Gtk.widget you want and then apply your patch. This mechanism should be something weak and and easy to break with a core update the cinnamon-setting, but possible to implement.

danimesq:
If I were Sergio Moro, I would hire you to help me re-interpret some loopholes in books and write a narrative (or a GPT3 commenter, who knows)

danimesq:
But passing through any possible passive-aggression, here's a comprehensive lecture.
You have some good points that could make me doubt myself and question why this PR is a need.

Let's cover:

> and taking in hand that the modularity is a synonyms of extension and the fact that Cinnamon already have extensions

Not necessarily/exactly.

Modularity implies diversity of options (some harder for newbies, some easier so here's the importance of equity of options/alternatives), not a particular API of extensions.

> How should help the creation of some new types of extensions (modularity) for a thing that can be done directly from the current Cinnamon extension mechanism that is already include in Cinnamon?

Sorry for having to parrot myself, but there's a point about that topic in the exact PR's description/OP/first comment: [this Mint's forum question](https://forums.linuxmint.com/viewtopic.php?p=2163518).

If the current extensions spice API were a maximum/sufficient alone, users wouldn't be coping together to make an easier way of implementing a context menu item, with a single file and few JS lines.

My vision is that extension spices are for advanced stuff, such as conditional transparent panels and desktop watermarks.

Simpler stuff that users build themselves, are covered in this PR's initial proposal for modulaty starting by the panel's context menu.

> What i said is that if you want to include a new option in the menu panel, you can create an normal Cinnamon extension and form there add this option to the place you want without need to modify any existing Cinnamon code

That's exactly the idea (disconsidering the exclusivity/limitation of a single general-purpose API that don't really fit the simpler, general-purpose), to prevent users like [these great guys](https://forums.linuxmint.com/viewtopic.php?p=2163518&sid=0a31db1222ea3f7cb9d2b352a26e708d#p2163518) from modifying files and potentially breaking their system sooner (in the next restart after implementation) or later (after updates overwritting general files that should instead call custom codes not covered by updates/mainstream enforcing). Sorry for having to parrot myself twice, but your counterpoint is also already covered in the PR's description.

> So, why you think as good the attempt to do what is already possible?

Most of the things we have on Linux, were already possible on Windows. The matter isn't about making the impossible, but making a better possible (it doesn't mean reinventing the wheel).

> Also, the cinnamon-settings already have a mechanism of extensions (i.e. modules) that can be used/extended to achieved what you want in more/less the same way. The only problematic thing there, is that in this case and for an unknown reason to me, the cinnamon developers decided don't give to the cinnamon-settings extension the whole application context in an easy way. So, to do some things you need to parse the Gtk components starting from the Gtk.ApplicationWindow to find the Gtk.Widget you want and then apply your patch. This mechanism should be something weak and easy to break with a core update the cinnamon-settings, but possible to implement.

About cinnamon-settings, it was just an off-topic comment in the PR's description; its not included in these commits and won't be.

lestcape:
I don't see any reason to have a passive-aggression to you, but who know in what way the the Open IA wrote the code this time.... :). 

In my belief system, the code is wrote to be understand and change only by a software developer... So, I cant not see how the complexity is reduced that way, creating another code and adding another new concept more, for something that can be easy resolved with an extension.

If you want to create something easy to modify by a user, do not provide a code to them; instead you can create a json file or an xml where  the user can describe de menu options and then you can build the menu from that file. The users can edit the file to easy add, remove or modify an option in a well know file format or also you then can wrote an external app to help the users to modify the file in a more easy way and visual.

You don't have to convince me, but it's clear that if you want to add something to a software, you have to expose at least why you think it's useful and if you will end to be Sergio Moro trying to expose the case, that should be good and not bad for the people that need to understand what you want. Good luck...

mtwebster:
This is a pretty niche feature, and like @lestcape has been explaining, there are mechanisms in place to accomplish this

The first responder in that forum thread even encouraged this - https://forums.linuxmint.com/viewtopic.php?p=2163120#p2163120

Another example - take an [extension I made](https://github.com/linuxmint/cinnamon-spices-extensions/blob/master/mnemonic-5-4-window-menu%40mtwebster/files/mnemonic-5-4-window-menu%40mtwebster/5.4/extension.js) - it completely replaces the Cinnamon's right-click window menu with a modified one.

Extensions require only three methods -- init (setup, or do nothing), enable and disable. You can define a settings ui using just a json file. You're free to take any existing code, either from core Cinnamon, or from [applets](https://github.com/linuxmint/cinnamon-spices-applets), [desklets](https://github.com/linuxmint/cinnamon-spices-desklets), and [extensions](https://github.com/linuxmint/cinnamon-spices-extensions) as a starting point.

Going back to the forum post - things should appear where they make sense contextually. The panel menu is for broad control of Cinnamon's panels and it offers a shortcut to system settings. There are at least half a dozen applets for resource monitoring, all of which offer a quick way to open the system monitor app. There are plenty more applets that are specifically geared towards custom actions.


Lastly, there is no place here or on the forums for treating others with anything but respect and kindness. It's possible to disagree while still holding to that.

PR #11418, opened 2023-01-12
Fix new lsblk for getDiskSize

gkxl:
[Fix new lsblk for getDiskSize](https://github.com/linuxmint/cinnamon/commit/3580783b8c4b0ad2c839ea6a5b402b68cd938a44)


clefebvre:
No, we want to show the total size here. This screen is there to show what's in the computer. Ideally we'd even list the HDD and their manufacturer info.

gkxl:
@clefebvre hi, 
I noticed that due to case insensitivity the lsblk "SIZE" parameter needs to be in capital letters.

image

Issue #9560, opened 2020-09-05
Grouped window list option "Enable Super+<number> shortcut to switch/open apps" should not override explicit keyboard shortcut settings


HairyDude:
* Cinnamon version 4.6.7
* 64 bit

**Issue**

I have used Super+N to switch to workspace N for many years, across multiple DEs. After upgrading to Mint 20, these shortcuts are usurped by the window list for switching between windows. This is surprising and unwanted behaviour that is not documented in the one place you look to set up keyboard shortcuts, namely the Keyboard app.

**Steps to reproduce**

1. Make sure the Workspace Switcher applet is on your panel.
1. Right click the Workspace Switcher applet and add five workspaces (for a total of six).
1. Open Keyboard and switch to the Shortcuts tab.
1. Go to Launchers -> Home folder and unbind every keystroke you find there.
1. Go to Workspaces -> Direct navigation and for each N from 1 to 6, add the binding Super+N to "Switch to workspace N".

**Expected behaviour**

Super+N switches to workspace N. For example, Super+4 switches to workspace 4.

**Actual behaviour**
With the windows I have open right now:

* Super+1 switches to workspace 1. (Although I have pinned a custom launcher for KeePass in the first position in the window list, it doesn't open KeePass.)
* Super+2 opens Nemo at my home directory and does not switch to workspace 2. (Nemo is pinned in 2nd position.)
* Super+N (N>=3) switches between the windows of whatever program happens to be in position N... except that if I'm on an empty desktop 4, they switch to desktop 5 and then do whatever they would have done there. Then Super+3 switches to desktop 6 where Hexchat is open.

**Other information**

Super is the Windows key.

This behaviour can be turned off in preferences for the grouped window list, after which the bindings work as expected. But it should not usurp explicit settings, and should not be turned on by default when migrating from earlier versions of Cinnamon if you have explicitly bound any of those keystrokes to anything.

HairyDude:
The setting seems to be reset on every login, but the Grouped Window List preferences show it as disabled, until you "enable" it. Then disabling the preference actually turns it off. It should remain off when I log out.

In addition, Super+Shift+1, which I have bound to "Move to workspace 1" (i.e. move the focussed window to workspace 1), fails and instead launches the first item in the grouped window list, regardless of the Grouped Window List setting. Super+Shift+N for other N (which have corresponding bindings) still work as expected.

Issue #11393, opened 2023-01-05
Grouped window list super+<number>


vikrend:
```
 * Cinnamon version 5.6.5
 * Distribution Arch
 * 64 bit
 ```

### Issue

After every cinnamon start/reload grouped-window-list's super+\<number\> shortcut still works with being turned off in applet settings. I have super+shift+\<number\> shortcuts in my system to move windows to another workspace and this bug blocks it. After manual on-off toggle in applet settings all works as supposed

### Steps to reproduce

* Turn off super+\<number\> shortcuts in applet settings
* Set some system hotkeys with super+\<number\> in it
* Fresh start/reload cinnamon
* Applet interferes with your shortcut

### Expected behaviour

The super+\<number\> applet hotkeys stay turned off

### Configuration info, if applicable

[applet settings pastebin](https://pastebin.com/vtvyn48s)

nonwibb:
I am experiencing the same issue. For me, I am able to set purely the shortcut `super+<number>`, but it's shortcuts with more things like `super+shift+<number>` that are blocked by this applet.

t3pfaffe:
I erroneously reported this same error in the cinnamon applets repository a short while ago ([cinnamon-spices-applets#4634](https://github.com/linuxmint/cinnamon-spices-applets/issues/4634)). This is also related to #10811 #9560. 

In my case, while it does seem to disable the `Super+<num>` hotkeys when the settings value `super-num-hotkeys` is set to false it does not also disable the `Super+Shift+<num>` key binds. The culprit appears to be this section of code: https://github.com/linuxmint/cinnamon/blob/eafc411664c28af8cb9bfc447b00b8e219b97f25/files/usr/share/cinnamon/applets/grouped-window-list%40cinnamon.org/applet.js#L460-L465

As you can see, it will still call the `bindNewAppKey` function even if the `SuperNumHotKeys` setting value is false. I will submit a pull request shortly, hopefully fixing the issue. I would also like to find out if this is actually the desired end behavior, since I am not sure how the author (@art-vasilyev) would have unintentionally excluded this call from the condition block when it was added in 9e57733fd880be86ebac61ee3fe2205acf21e09e. If this is the desired behavior we will have to add another toggle specifically for the `Super+Shift+<num>` key binds to avoid them being overridden.

[...]

Issue #10811, opened 2022-05-28
Shortcuts Super+[2,3,4] don't work as expected


tikhont:
* Cinnamon version - 5.2.7
 * Distribution - LMDE 5

**Issue**

I'd have assigned shortcuts Super+[1,2,3,4] to switch to corresponding workspaces. After upgrade LMDE 4 -> 5, only Super+1 works as expected. Super+[2,3,4] switch to (or minimize if it's active) corresponding windows - not workspaces. Although in Keyboard -> Shortcuts -> Direct Navigation, all are still set as expected. I've tried to reassign all four bindings - and it never said one was assigned to something else.

When I assign Super+2 to another command (run Terminal, for instance) - it works perfectly.

**Steps to reproduce**

Are described above.

**Expected behaviour**

Shortcuts to switch to windows [1, 2, 3, 4] are assigned in Keyboard preferences and can be reassigned as any other.

smurphos:
Sounds like interference with the Grouped Window List default shortcuts. Right click on any pinned app in the grouped window list > preferences > configure > uncheck 'Enable Super+<number> to launch/switch apps'

It would be a big improvement to Cinnamon to have some integration between the applet keybinding system and the 'general' one, as you noticed the keyboard app doesn't know about keybindings reserved by panel applets.

tikhont:
It helped, thank you.

Anyway, I consider it a bug. For example, keyboard preferences can warn about this interference and propose to turn off the corresponding function.

sctincman:
This has also been happening to me as well since the upgrade and is continuing on 5.4.4 (Arch Linux)

On log-in the applet behavior will override my keybindings (causing grouped window list to launch apps when using Super+[1,2,3,4]). The "Enable Super+ to launch/switch apps" option toggle will present as "off", but toggling it "on" then back to "off" will restore the expected behavior of my keybindings for that session alone.

Logging out and back in will result in the grouped window applet taking the keybindings over again, and will need to repeat above steps to get custom keybindings to take effect again.

Not sure why this is not being sticky. I did notice the following on log-in (repeats for most of my custom keybindings)
`Jul 20 22:54:04 mmo /usr/lib/gdm-x-session[18401]: Window manager warning: move-to-workspace-1 - Overwriting existing binding of keysym 31 with keysym 31 (keycode a).`

HairyDude:
[...]

PR #11464, opened 2023-01-31
Fix super+shift+number not disabling in grouped-window-list applet

t3pfaffe:
- Setting the value of the applet settings parameter 'super-num-hotkeys' to false should now avoid registering the super+shift+number app launcher keybinds whereas it previously it only avoided registering super+number keybinds
- These changes should fix the unexpected behavior reported in #11393.
- You can read a further explanation of the issue below originally from my comment on the original issue.
> I erroneously reported this same error in the cinnamon applets repository a short while ago ([cinnamon-spices-applets#4634](https://github.com/linuxmint/cinnamon-spices-applets/issues/4634)). This also seems related to #10811 #9560. 
> 
> In my case, while it does seem to disable the `Super+<num>` keybinds when the settings value `super-num-hotkeys` is set to false it does not also disable the `Super+Shift+<num>` keybinds. The culprit appears to be this section of code: https://github.com/linuxmint/cinnamon/blob/eafc411664c28af8cb9bfc447b00b8e219b97f25/files/usr/share/cinnamon/applets/grouped-window-list%40cinnamon.org/applet.js#L460-L465
> 
> As you can see, it will still call the `bindNewAppKey` function even if the `SuperNumHotKeys` setting value is false. I will submit a pull request shortly, hopefully fixing the issue. I would also like to find out if this is actually the desired end behavior, since I am not sure how the author (@art-vasilyev) would have unintentionally excluded this call from the condition block when it was added in 9e57733fd880be86ebac61ee3fe2205acf21e09e. If this is the desired behavior we will have to add another toggle specifically for the `Super+Shift+<num>` keybinds to avoid them being overridden.
> 
> I can only speculate as to why this behavior only started occurring a couple of months ago. I assume some update within the Cinnamon DE made it so that applets register their keybindings prior to keybindings set in the global Cinnamon keyboard layout section. This would explain why disabling and re-enabling the `super-num-hotkeys` setting temporarily fixes the problem since doing that unbinds and attempts to rebind all the hotkeys; reverting them to the values set in the global keyboard settings.
> 
> _Originally posted by @t3pfaffe in https://github.com/linuxmint/cinnamon/issues/11393#issuecomment-1411117279_

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/applet.js
@@ -459,11 +459,11 @@ class GroupedWindowListApplet extends Applet.Applet {
     bindAppKeys() {
         this.unbindAppKeys();
 
-        for (let i = 1; i < 10; i++) {
-            if (this.state.settings.SuperNumHotkeys) {
+        if (this.state.settings.SuperNumHotkeys) {
+            for (let i = 1; i < 10; i++) {
                 this.bindAppKey(i);
+                this.bindNewAppKey(i);
             }
-            this.bindNewAppKey(i);
         }
         Main.keybindingManager.addHotKey('launch-show-apps-order', this.state.settings.showAppsOrderHotkey, () =>
             this.showAppsOrder()


HairyDude:
While you're at it, please add a check that the relevant binding is not already used in keyboard shortcut settings, and avoid setting the binding if it is. This would fix #9560.

t3pfaffe:
Is there any further action needed to get this PR merged?

PR #11473, opened 2023-02-01
Use pseudo class 'hover' instead of class ...

### LINKED ISSUE #10935: Menu applet: Two categories highlighted at the same time.
```
 * Cinnamon version 5.4.2
 * Mint daily builds PPA and Manjaro
 * Bug not present in Mint 20.3
 ```

**Issue**
When option "Change categories on hover" is _off_

two categories can be highlighted at the same time - hover highlight is not removed.

simplescreenrecorder-2022-07-16_22 54 44

fredcw:
... menu-category-button-selected when hovering category buttons when "change category on hover" option is off.
Fixes #10935

Simplify some of the code in _onMenuKeyPress and _buttonEnterEvent.

Note: As menu-category-button:hover is not supported in any themes yet, the hover class will have no style applied to it unless PR #11451 is also merged.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -933,11 +933,12 @@ class CategoryButton extends SimpleMenuItem {
     }
 
     activate() {
-        if(this.applet.searchActive)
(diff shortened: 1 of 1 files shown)


mtwebster:
So we'll need to add support for :hover in our mint themes I presume?

Issue #11455, opened 2023-01-28
Show an Icon representation of the window in the workspace-switcher applet


anaximeno:
### Is your feature request related to a problem? Please describe.

When having multiple windows open in many workspaces, it is difficult to know from the workspace-switcher applet, which workspace is the one that contains the window I want to access.

### Describe the solution you'd like

A solution could be to have an option in the config to display the window's icon in the center of its respective frame representation in the applet.

### Describe alternatives you've considered

I currently navigate through all workspaces until I find the workspace that contains the window.

### Additional context

I tried to implement this myself, but until now I did not find any good way to retrieve the Icon of the window object that is used in the applet.

rcalixte:
@anaximeno Do you mean making a view like Expo but in the panel?

anaximeno:
> @anaximeno Do you mean making a view like Expo but in the panel?

Yeah, at least, showing the icon (when available) of the window that is being displayed in the applet.

**Here's what currently happens on Cinnamon**

![Captura de ecrã de 2023-01-31 13-03-11](https://user-images.githubusercontent.com/63073056/215783427-2d36ce88-7533-41cf-91dc-6fd06ab15c81.png)

**Here's an example from XFCE4**

![Captura de ecrã_2023-01-31_13-00-58](https://user-images.githubusercontent.com/63073056/215783700-97fa2c29-7499-4eec-9c78-25fb5a117e3e.png)
  
  
_* I have the same applications open in both examples._

PR #11530, opened 2023-02-24
workspace switcher: Add option to display windows icons

anaximeno:
Fixes https://github.com/linuxmint/cinnamon/issues/11455

The goal is to show the icon of the windows on the workspace-switcher applet.

To do:
- [x] Get the icon of the windows
- [x] Show the icon of the window in the center of its respective rectangle
- [x] Add settings to enable and disable this feature


anaximeno:
So far, I've managed to get the St.Icon (or St.Bin) object that represents the icons of the windows, but I didn't find how to draw those icons in the Cairo context.

lestcape:
@anaximeno While maybe your idea can be done, this probably is not the right way to go. You should avoid complex task of painting things from the cjs side and instead create it as an actor and let Muffin to draw it from the c side.

So, I think that instead of try to find a way to draw a texture/St.Icon you should probably try to find a way to set a Clutter actor (your texture) at top of the drawing area. Maybe this code should help you:

https://github.com/linuxmint/cinnamon/blob/master/js/ui/boxpointer.js#L47-L52

The other possibility that i not recommended is try to find the real icon of the app instead of a texture of it and then load it as a pixbuf, to then you can write it in the cairo context.

anaximeno:
Thanks for the suggestion @lestcape.

anaximeno:
If anyone wants to implement this let me know here and I will close my request in favor of the request. I have little free time to do all the necessary research before implementing this the right way, but I may complete it when I have more free time.

Another piece of code I'm using as a reference:
https://github.com/linuxmint/cinnamon/blob/ae4523dd7d88b764bcaa0753389be6f91d964e6d/files/usr/share/cinnamon/applets/grouped-window-list%40cinnamon.org/appGroup.js#L273-L294

anaximeno:
I believe it's **ready for review**.

Tested under **LM 21.3** with **X11** and with **Wayland** (worked without any issues)
Tested under **LMDE 6** (worked without any issues)

Thanks again @lestcape!

mtwebster:
Hi,

So this seems to work pretty well, but with one big issue - the icons are blurry.

I took a look at things, and you're loading the icons correctly, but you end up scaling the entire applet (or some other parent actor) which ruins the icons. You can see this by slowly increasing panel height (like mouse-wheel the slider) and observing the icons' sharpness changing as you go...

I didn't look at how the applet is graphically constructed, but possibly you need to add/improve size calculations so the layout is rendered without scaling (accounting for theme padding/margins maybe), or render the icons separately, on top of the scaled WindowGraphs.

mtwebster:
Hmm assuming these are svgs (most app icons will be I think), maybe it's related to it being rendered 'in between' pixels. Maybe try rounding their position to a whole number? I'm not really well-versed in that sort of thing, but I know we've had icon sharpness issues in the past, when an icon wasn't aligned properly on its canvas.

lestcape:
@anaximeno the icon have 12 px by default. It's difficult that it will not be blurry at that size. The smallest icon size Cinnamon uses have 24 pixels and normal app icon have at less 16px, while most of common apps icons are 32px at less. So, use 12px means that the icon was scaled for sure.

I recommended you select always standard icons sizes with more than 24px, and do not allow other size, because other size means a re-scaled the icons. You should see: /usr/share/icons/hicolor for the standard icons size. And also you can have an idea of who may apps have an icon in the select a size and how it luck like if you enter to the folder of the size you want. For example for the size 32px it should be: /usr/share/icons/hicolor/32x32/apps

I don´t think you can improve your implementation in that regards, but you can smart selected the preferred icon size to improve the result.

anaximeno:
I would prefer not limiting the feature to only sizes greater or equal to 24px because that would make it almost unusable on smaller screens and panels (here's an example with the panel height = 28):

- **24 px**
screenshot-auto_1707924409

- **8 px**
screenshot-auto_1707924437

And to be fair the blurriness is barely noticeable on smaller screens, though I suspect it might be more noticeable on wider screens, I will test with a wider monitor and see how bad it is.

lestcape:
Separate point: You should take the scalefactor on account in the icon size. Currently you only have on account the position, not the size. It probably should help to be possible display the icon with a better resolution in a screen with a big scalefactor.

anaximeno:
> Separate point: You should take the scalefactor on account in the icon size. Currently you only have on account the position, not the size. It probably should help to be possible display the icon with a better resolution in a screen with a big scalefactor.

I believe you are talking about the `global.ui_scale` of the monitor, for what I've observed, I believe Cinnamon (or Muffin) already takes care of applying scaling to the icons, so I believe it wouldn't be necessary to apply to scale to the size of the icon on the code. But I have to take it into account while calculating the position where the icon should be placed because the real size of the icon is bigger (because of the scaling) than the one I sent while creating the icon instance.

anaximeno:
I will recreate this PR with a cleaner commit log.

anaximeno:
Replaced by #12202

PR #11531, opened 2023-02-25
docs: add images or videos section in the bug report template

anaximeno:
I suggest adding a section for users to add Images or Videos of the bug that is happening (when applicable) which may help debug and reproduce the error or bug.


mtwebster:
I feel like there are already too many 'sections' as it is, and we're probably going to move to a form template instead of just a markdown file soon.

PR #11548, opened 2023-03-09
cs_themes: Add a simplified UI

clefebvre:
TODO:

- [x] Don't show styles which point to non-existing themes
- [x] Support a default mode/variant for a given style
- [x] Refresh UI properly
- [x] Harmonize "Themes" vs "Look & Feel" name
- [x] Remove custom legacy/dark sort in custom theme selectors
- [x] Go straight to custom page if the user uses a custom selection
- [x] Remove Custom in combo when it's not needed
- [x] HiDPI support for dots and appearance images

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/appearance-dark.svg
+++ b/files/usr/share/cinnamon/cinnamon-settings/appearance-dark.svg
@@ -1,0 +1,318 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   width="112"
+   height="80"
+   id="svg4729"
+   version="1.1"
+   sodipodi:docname="appearance-dark.svg"
+   inkscape:version="1.1.2 (0a00cf5339, 2022-02-04)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:dc="http://purl.org/dc/elements/1.1/">
+  <sodipodi:namedview
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1"
+     objecttolerance="10"
+     gridtolerance="10"
+     guidetolerance="10"
+     inkscape:pageopacity="0"
+     inkscape:pageshadow="2"
+     inkscape:window-width="2560"
+     inkscape:window-height="1352"
+     id="namedview55"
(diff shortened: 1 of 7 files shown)


Secret-chest:
Would you add an API so custom theme devs can use this interface as well?

Secret-chest:
Even though the UI is pretty nice, I think it would be better added in the welcome screen and keep the themes menu as-is.

ItzSwirlz:
Does the color selector only appear if it exists for the theme? Some themes do have multi-color but vary and can only be told depending on the suffix (e.g. Yaru-*)

Secret-chest:
> Does the color selector only appear if it exists for the theme? Some themes do have multi-color but vary and can only be told depending on the suffix (e.g. Yaru-*)

The themes have to be manually added to the dropdown. By default only mint themes and adwaita seem to be available, adwaita seems to only have blue colour.

mtwebster:
Advanced Settings is the old <= 5.4 view - nothing is being taken away. All themes will appear like before, no changes required. For themes to show up in the new page, the definition file would need to include them.

There's only one Adwaita because they only have a light and dark (blue) theme, no other variants.

Mint Welcome is not meant as a normal configuration tool, it's more for initial setup or for less experienced users perform certain major things (like the panel style) without getting lost.

This new theme page is not a novice, introductory tool, it's a less-cluttered way to change the desktop appearance, which is all most users really have a need for.

clefebvre:
> Does the color selector only appear if it exists for the theme? Some themes do have multi-color but vary and can only be told depending on the suffix (e.g. Yaru-*)

We don't look at the theme name to guess available color variants (we'll stop doing that as well in folder-color-switcher btw). We use variant definitions stored in /usr/share/cinnamon/styles.d/.

Secret-chest:
Would it be possible to be able to ship these kinds of definitions when installing a Cinnamon Spices theme?

Secret-chest:
> This new theme page is not a novice, introductory tool, it's a less-cluttered way to change the desktop appearance, which is all most users really have a need for.

I agree, and I think it's got potential since new definition files CAN be shipped with theme debs.

clefebvre:
> Would it be possible to be able to ship these kinds of definitions when installing a Cinnamon Spices theme?

No, it would need write permissions to /usr/share/cinnamon/styles.d/ which is a system directory.

Also, it's not something we'd want. A style is much more than a Spice theme, it's the combination of a GTK theme, an icon theme etc... it's a complete set.

When you're at the point of installing Cinnamon themes, you're already in the Custom tab. If you use a custom set like this the settings app will default to that tab for you whenever you open it again.

Secret-chest:
Looks good then. Serious themes are just going to get installers now (to add them here), aren't they?

Secret-chest:
> Looks good then. Serious themes are just going to get installers now (to add them here), aren't they?

Not sarcasm.

clefebvre:
> Looks good then. Serious themes are just going to get installers now (to add them here), aren't they?

Either that or distributions. The way it works is that they can override each others. Files in /usr/share/cinnamon/styles.d/ are read in alphabetical order. If two files define the same style name, the latter one overwrites the former.

This allows Cinnamon to ship with a predefined style for Adwaita for instance which distributions can override, or which GNOME/Adwaita can override.

PR #11561, opened 2023-03-15
feat(menu@cinnamon.org): Add Menu editor entry to applet context menu

axel358:
Discussion here https://forums.linuxmint.com/viewtopic.php?p=2227731#p2227731

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1174,6 +1174,10 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
         this.menu = new Menu(this, orientation);
         this.menuManager.addMenu(this.menu);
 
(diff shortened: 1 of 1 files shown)


files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -1174,6 +1174,10 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
         this.menu = new Menu(this, orientation);
         this.menuManager.addMenu(this.menu);
 
+        const edit_item = new PopupMenu.PopupIconMenuItem("Menu editor", "document-edit", St.IconType.SYMBOLIC);
```
mtwebster, line 1177: Change to "Edit menu" please, and it needs to be translated, so wrap it like this: `_("Edit menu")` (like line 1172 above)

axel358:
CI fails even on master on a clean fork

PR #11564, opened 2023-03-17
5.6 maintenance (suggestion): Create icon to the desktop (Right click menu on folder\file)

jujielbrus:
Hello!

I'm new to Linux and just installed Mint

If there any way to contact creators of release to ask for add "Create icon to the desktop" when you push right button on a folder or a file?

It takes me 15-20 minutes to find out that it only can be possible by Ctrl+M or "Edit" on Top menu and only then Create a link and then move it to a desktop

Would save time of many other's who are switching from other OS

Thank you!

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,23 @@
+cinnamon (5.6.8) vera; urgency=medium
+
+  [ Michael Webster ]
+  * grouped-window-list: Make appGroup's flashButton sane, fix invalid source id.
+  * cinnamon-screenshot.c: Disable unredirection when taking a screenshot.
+
+  [ hdulme ]
+  * remove unused import cairo
+
+  [ Bobby Rong ]
+  * cs_themes: load theme thumbnail from XDG_DATA_DIRS (#11497)
+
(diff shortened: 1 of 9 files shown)


Fantu:
@jujielbrus hi, I not understand your PR, I don't see your commit and seems you want merge merge 5.6-maintainance branch into master but it does not make sense because that branch contain only some fixes taken from master (so already present)
I suppose you should create an issue instead (and not pull request) with your question.
and anyway your question is related to the filemanager so is related to nemo instead

mtwebster:
Create a file "link-to-desktop.nemo_action" with these contents:
```
[Nemo Action]
Active=true
Name=Create link on the desktop
Exec=sh -c 'ln -s %F $(xdg-user-dir DESKTOP)'
Icon-Name=user-desktop-symbolic
Selection=single;
Extensions=any;
```

Copy it to `~/.local/share/nemo/actions` (~ is your home folder)

More info about nemo actions:
https://github.com/linuxmint/nemo/blob/master/files/usr/share/nemo/actions/sample.nemo_action

PR #11591, opened 2023-03-26
Remove simple-system-monitor@ariel as a known conflict

rcalixte:
The desklet has been cleaned up by linuxmint/cinnamon-spices-desklets#816 but this needs to be merged before linuxmint/cinnamon-spices-desklets#832.


mtwebster:
Added to 5.6-maintenance, but not sure how soon it might get a tag.
https://github.com/linuxmint/cinnamon/commit/deb3926b44ee4268d1051dc668be32762991ba43

PR #11628, opened 2023-04-19
Calendar Applet: Custom Tooltip Date Format

Dosenpfand:
Currently only the main date format is configurable if the custom format is activated. This PR makes the date that is shown in the tooltip configurable too.


mtwebster:
Hi, a couple of things:

- By default, the tooltip should show the same as it does now (DATE_FORMAT_FULL). Right now it's using the custom value whether it's enabled or not.
- The default custom value can be the same as the normal label's custom value.

Also, as tempting as it is, try to avoid random white-space fixes except for where you're modifying code - it can draw attention away from what's actually changing in the commit. No need to remove them in this case, however - just mentioning it.

Thanks

Dosenpfand:
Hi @mtwebster, thanks for the feedback! I tried to address both findings and would be glad if you could have another look.
Cheers!

PR #11643, opened 2023-04-30
Revert to `~/.themes` and `~/.icons` directory as default instead of XDG equivalents because GTK2 apps couldn't use them

Gr3q:
Fixes https://github.com/linuxmint/cinnamon/pull/11175#issuecomment-1457237711


Gr3q:
@mtwebster Discussion reminder https://github.com/linuxmint/cinnamon/pull/11175

PR #11656, opened 2023-05-07
System settings - Add the Screenshots module

claudiux:
Solves the problem discussed here: https://github.com/orgs/linuxmint/discussions/9#discussion-5167987


JosephMcc:
I see a couple issues here. One, there is no need to use a settings stack for only one page. Modules with only a single page don't use it. The bigger issue though is what if gnome-screenshot isn't installed? Cinnamon doesn't depend on it. It can be removed and a person can install whatever screenshot tool they want. A distro that ships Cinnamon could also use some other screenshot tool.

claudiux:
@JosephMcc 
I plan to use two pages.
I just fixed the 'bigger issue' but I also plan to take into account other screenshot-tools using D-Bus.

PR #11664, opened 2023-05-14
cinnamon-settings.py: Additional modules when Mint modules are absent

rcalixte:
* Removing trailing whitespace as well

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
@@ -69,25 +69,35 @@ CONTROL_CENTER_MODULES = [
 ]
 
 STANDALONE_MODULES = [
-    # Label                              Executable                             Icon                     Category          Keywords for filter
-    [_("Printers"),                      "system-config-printer",               "cs-printer",            "hardware",       _("printers, laser, inkjet")],
-    [_("Firewall"),                      "gufw",                                "cs-firewall",           "admin",          _("firewall, block, filter, programs")],
-    [_("Firewall"),                      "firewall-config",                     "cs-firewall",           "admin",          _("firewall, block, filter, programs")],
-    [_("Languages"),                     "mintlocale",                          "cs-language",           "prefs",          _("language, install, foreign")],
-    [_("Input Method"),                  "mintlocale-im",                       "cs-input-method",       "prefs",          _("language, install, foreign, input, method, chinese, korean, japanese, typing")],
-    [_("Login Window"),                  "pkexec lightdm-settings",             "cs-login",              "admin",          _("login, lightdm, mdm, gdm, manager, user, password, startup, switch")],
-    [_("Login Window"),                  "lightdm-gtk-greeter-settings-pkexec", "cs-login",              "admin",          _("login, lightdm, manager, settings, editor")],
-    [_("Driver Manager"),                "cinnamon-driver-manager",             "cs-drivers",            "admin",          _("video, driver, wifi, card, hardware, proprietary, nvidia, radeon, nouveau, fglrx")],
-    [_("Nvidia Settings"),               "nvidia-settings",                     "cs-drivers",            "admin",          _("video, driver, proprietary, nvidia, settings")],
-    [_("Software Sources"),              "pkexec mintsources",                  "cs-sources",            "admin",          _("ppa, repository, package, source, download")],
-    [_("Package Management"),            "dnfdragora",                          "cs-sources",            "admin",          _("update, install, repository, package, source, download")],
-    [_("Package Management"),            "yumex-dnf",                           "cs-sources",            "admin",          _("update, install, repository, package, source, download")],
-    [_("Users and Groups"),              "cinnamon-settings-users",             "cs-user-accounts",      "admin",          _("user, users, account, accounts, group, groups, password")],
-    [_("Bluetooth"),                     "blueberry",                           "cs-bluetooth",          "hardware",       _("bluetooth, dongle, transfer, mobile")],
-    [_("Bluetooth"),                     "blueman-manager",                     "cs-bluetooth",          "hardware",       _("bluetooth, dongle, transfer, mobile")],
-    [_("Manage Services and Units"),     "systemd-manager-pkexec",              "cs-sources",            "admin",          _("systemd, units, services, systemctl, init")],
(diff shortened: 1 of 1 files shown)


files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
```diff
@@ -467,12 +477,24 @@ def load_ccc_modules(self):
 
     def load_standalone_modules(self):
         """Loads all standalone settings modules."""
+        is_mint: bool = False
         for item in STANDALONE_MODULES:
             samodule = SettingsWidgets.SAModule(item[0], item[1], item[2], item[3], item[4], self.content_box)
             if samodule.process():
                 self.sidePages.append(SidePageData(samodule.sidePage, samodule.name, samodule.category))
+                if "mint" in item[1]:
+                    is_mint = True
```
anaximeno, line 486: Aren't there more effective ways of checking if the platform is mint? This code will be executed whenever its parent block is reached, and only the first pass in the condition would be enough for defining the platform, so that might even be executed when it is no more needed, also it will be dependent on the result of the `samodule.process()`, which may be good or bad depending on your goal there.

rcalixte: > Aren't there more effective ways of checking if the platform is mint? This code will be executed whenever its parent block is reached while only the first pass in the condition would be enough for defining the platform.

Agreed that it's not an elegant solution. I just double-checked the `platform` module and I actually think there's a function there that can be used instead.

> Also, in that place it will be dependent on the result of the `samodule.process()` in the parent if condition, which may be good or bad depending on your goal there.

This was intentional since the processes are either there or they are not. This would create a blind spot for users who remove the binaries from their path but that's something we can't try to solve for right now.

Thanks for the review!

rcalixte: @anaximeno Feel free to take a look at the change. It has a few blind spots but it should suffice for our needs.

clefebvre:
don't we test to see if the executable is present before showing the module?

if so we don't need to say "if Mint, else"

it's a bit weird to have Mint on one side and then all other distros on the other within the code.

rcalixte:
> don't we test to see if the executable is present before showing the module?
> 
> if so we don't need to say "if Mint, else"

The goal was to avoid collisions and duplicates in Mint where other installations of Cinnamon do not have the additional modules present. Rather than chasing down multiple upstreams to patch Cinnamon's settings modules, this would carve out a space for those modules and expand the list of modules on non-Mint installations of Cinnamon.

> it's a bit weird to have Mint on one side and then all other distros on the other within the code.

Most of the additional modules could be moved to the main definition but the main exception is `software-properties-gtk`. @clefebvre Are the plans in future versions of Mint to use a different namespace for this? Otherwise, this would conflict on Mint installations. If this is incorrect, we can consolidate the additional modules and simplify things. The original goal was to leave Mint undisturbed.

rcalixte:
> OK, an easy way to solve this would be to add a module blacklist in gsettings. Mint could set this to "['software-properties-gtk']", other distros could set this to suit themselves.

Wouldn't this be more difficult to maintain? Not only for other downstreams but for users as well? This would create two locations to configure one facet within Cinnamon.

clefebvre:
Yeah, you could only show software-properties-gtk if mintsources isn't present. That would work and be quite simple.

fredcw:
I curious about something. As each item seems to have an associated .desktop file, would it not be better to get the icon name for each item from it's .desktop file rather than have to specify it in this file. This would also ensure that the icon used in system settings would always be consistent with the icon used in the menu applet.

clefebvre:
> I curious about something. As each item seems to have an associated .desktop file, would it not be better to get the icon name for each item from it's .desktop file rather than have to specify it in this file. This would also ensure that the icon used in system settings would always be consistent with the icon used in the menu applet.

Yes. Good point.

rcalixte:
> Yeah, you could only show software-properties-gtk if mintsources isn't present. That would work and be quite simple.

Updated. I'm open to modifying the conditional if needed. We can check inside the `load_standalone_modules` function as well after determining whether the process was found. That was the original implementation when this PR was opened.



> I curious about something. As each item seems to have an associated .desktop file, would it not be better to get the icon name for each item from it's .desktop file rather than have to specify it in this file. This would also ensure that the icon used in system settings would always be consistent with the icon used in the menu applet.

I encountered the challenge with icons when I was testing since I had installed so many packages but had no way to differentiate them when they were all the same. If we're going to source .desktop files for icons, would it make sense to source the localized Name as well? (This is all assuming sourcing the .desktop files is trivial enough as well. I know it's possible, I primarily want to try to keep things simple.)

fredcw:
@rcalixte Oh, I wasn't suggesting you do the icon thing as part of this PR, it's a separate issue for a different PR. Maybe I'll take a look at it sometime. Then again, maybe it's not worth bothering with. "if it ain't broke, don't fix it" as it were.

lestcape:
I don't think it's a good idea to hardcode a list of modules at all. I thinks that all modules should be provided in a module file and there should be a function implemented by modules to evaluate if a module can be enabled or not. Thus, the responsibility of loading or not a module will be delegated to the module itself, which is the one who really knows its own requirements to be loaded.

Another improvement that I think should be nice is to have something to override the default modules. This is useful, for example, when there is more than one software to do the same thing and both are available on the system, but Mint prefers to use one, while a different Linux distribution prefers to use the other software. This can be done simply by using two different module folders, one which would be the primary "default" and the other would be where the distribution should put its "custom" module version. The cinnamon-settings should look first in the "custom" folder and then in the "default" folder and modules with the same name should only be loaded the first time they found. Using this organization, the modules of a distribution can even be added and maintained in a totally different and exclusive package of the distribution, without any need to modify the default code that comes with Cinnamon. Furthermore, the external distro can even excluded a module they don't like by providing a module witch the same name whose evaluation function is always return false.

The only downside to this is that it can be costly in terms of performance to have to load each module file to test its availability, but I believe the benefit would outweigh the prejudice and the real cost of doing this would be in terms of milliseconds. This if we are careful to separate the module file where the evaluation function is from the actual functionality of the module, so that no additional library needs to be loaded in a module that is not actually going to be loaded because it evaluation function return false.

rcalixte:
> Yeah, you could only show software-properties-gtk if mintsources isn't present. That would work and be quite simple.

This is now done in the latest rebase.

mtwebster:
Can you rebase this?

The only issue is the addition of this line:
https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py#L90
and the comma at the end of the previous line.

PR #11678, opened 2023-05-30
cinnamon-settings: use CSD

fredcw:
also change window title to sidepage title when changing to sidepage as "All Settings" button (and app icon) indicates still in system settings app rather than standalone.

fixes https://github.com/linuxmint/cinnamon-control-center/issues/312

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
@@ -165,10 +165,10 @@ class MainWindow(Gio.Application):
                 self.go_to_sidepage(sidePage, user_action=True)
 
     def _on_sidepage_hide_stack(self):
-        self.stack_switcher.set_opacity(0)
+        self.top_bar.hide()
 
     def _on_sidepage_show_stack(self):
-        self.stack_switcher.set_opacity(1)
+        self.top_bar.show()
 
     def go_to_sidepage(self, sidePage: SettingsWidgets.SidePage, user_action=True):
         sidePage.build()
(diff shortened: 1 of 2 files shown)


fredcw:
Some of the settings modules have a specific requested window height e.g Date & Time module (560px) and Online Accounts module (560px). Is there any specific reason for this when most modules don't specify a fixed height. Could these specified heights be removed as they would otherwise seem unnecessary?

mtwebster:
I don't think this is going to work for now, at least without a significant re-think of how we present everything:

Having a headerbar *and a toolbar* isn't a good look, as they're not really meant to go together - I think much of point of the headerbar is to merge toolbar and titlebar 'content' and save vertical space.

We also can't really allow the window to get any taller than it already is, to accommodate low resolution netbook users ( < 800px ). You can scroll content, but it's  nicer to have the same initial content presented, regardless of resolution, and we try to design the more complex pages, like display settings, roughly around the current size.

I'm not  really sure about losing the title either - I know a lot of gnome apps ditch it altogether but there might be resistance to this.

fredcw:
>Having a headerbar and a toolbar isn't a good look

I originally put the stack switcher in the main content area which looked better to me but I thought  it might be too radical a change for some so I left it in a toolbar.

>We also can't really allow the window to get any taller than it already is

Not sure what you mean. The window is about 10-15 pixels higher, but that can be easily fixed. Tbh, I didn't even notice. I assumed that setting the window size would include the title bar but apparently not.

>You can scroll content, but it's nicer to have the same initial content presented, regardless of resolution, and we try to design the more complex pages, like display settings, roughly around the current size.

I agree. I think that the pages that request a larger size should be changed to not have that request because it doesn't seem necessary when you have scrolling. I was about to submit another PR to do exactly that and also at present when no height is requested the window can still be larger if the content is larger. I was going to change that as well so basically all of the module pages are 600px or less by default.

>I'm not really sure about losing the title either

I agree. I don't think having a title and the stack switcher in the title bar would look too good.

PR #11680, opened 2023-06-01
system settings csd+window size changes

fredcw:
@mtwebster Since I already made the changes, I thought I post them here for testing in case it gives you some ideas on what you want to do.

I think that keeping the same window size when going to or from modules makes a big difference instead of frequent window size changes which looks jarring. Also, it was annoying when you resized the window only for it to change back to a default size as soon as you switched to a settings page or back.

Also, if you're not planning to make any major changes before 5.8 is released, would this not be a good temporary improvement for the next mint release, before you revamp things later?

Edit: btw. trying to have widgets fit a certain fixed window size doesn't work because of theming, font scaling etc.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py
@@ -43,8 +43,8 @@ class SidePageData(typing.NamedTuple):
 
 
 WIN_WIDTH = 800
-WIN_HEIGHT = 600
-WIN_H_PADDING = 20
+WIN_HEIGHT = 585
+WIN_V_PADDING = 10
 
 MIN_LABEL_WIDTH = 16
 MAX_LABEL_WIDTH = 25
@@ -226,20 +226,20 @@ class MainWindow(Gio.Application):
         m, tb_n = self.top_bar.get_preferred_size()
         
         # Resize vertically depending on the height requested by the module
-        use_height = WIN_HEIGHT
-        total_height = cb_n.height + tb_n.height + WIN_H_PADDING
+        use_height = 0
         if not sidePage.size:
-            # No height requested, resize vertically if the module is taller than the window
-            if total_height > WIN_HEIGHT:
-                use_height = total_height
+            # No height requested, don't change window size
+            pass
         elif sidePage.size > 0:
             # Height hardcoded by the module
-            use_height = sidePage.size + tb_n.height + WIN_H_PADDING
+            use_height = sidePage.size + tb_n.height + WIN_V_PADDING
         elif sidePage.size == -1:
             # Module requested the window to fit it (i.e. shrink the window if necessary)
-            use_height = total_height
+            use_height = cb_n.height + tb_n.height + WIN_V_PADDING
 
-        self.window.resize(WIN_WIDTH, use_height)
+        if use_height > 0:
+            self.restore_window_size = self.window.get_size()
+            self.window.resize(self.restore_window_size.width, use_height)
 
     def deselect(self, cat):
         for key in self.side_view:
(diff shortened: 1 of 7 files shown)


axel358:
Kay seems like even without the stack switcher on the headerbar this still looks good. I also agree that constantly changing the window sizes when navigating through modules is not a very good experience, so even if the csd part is not accepted for any reason at least the window size fixes should be. Just a suggestion, i feel like an arrow (<-) for navigating back to the main page feels more intuitive since text only buttons on the Mint-Y theme are flat n could be read as a label. Other than that, awesome work

JosephMcc:
Since the other one was already closed but you decided to open another one for the same thing, I'll comment here. I'm totally with mtwebster on this. Also with him on all his reasons. This isn't an improvement over what we already have. In this case just putting the stack switcher down in the window content does not look better. It just uses up more space. It's  also really weird in windows like the panel settings where you would have a stack switcher and then more buttons just underneath for changing panels.

In the end I don't see what the benefit is of switching this to CSD. It seems to only be done "Just because". I only see negatives but someone can correct me if there is an actual gain to doing this.

fredcw:
@JosephMcc The point is that it eliminates to toolbar altogether on the main page and on the module pages that don't have a stack switcher (which is the slight majority of them). Yes, on the pages with a stack switcher the usable height is slightly reduced by the CSD title bar being slightly taller than the non CSD title bar but I think this is more than compensated for by the elimination of the toolbar on other pages. I submitted the original PR as a suggestion for consideration and in an effort to be helpful, I neither expect nor even necessarily recommend that it be merged, it's a was suggestion, nothing more. I'm not even entirely convinced of the CSD title bar being an aesthetic improvement myself.

The reason for this second PR is mainly to demonstrate the changes I made to the way the windows are sized (i.e allowing the user's resizing of the window to not be reset when changing to/from module pages), and I'm happy to make a separate PR for that if it's wanted.

If you think that keeping this PR open is unhelpful then I'm happy to close it. Thanks.

mtwebster:
@fredcw I was focused entirely on the headerbar change, sorry.

I know the approximate 'content' size isn't perfect, especially with scaling, fonts, etc... but I guess we're trying for a best effort, in cases like a low resolution monitor, or a Virtualbox or a real hardware where resolution is a problem pre-installation/drivers. We can't keep the user from breaking that, but we can start out as usable.

I think it would be ok if the window behaved like a normal window, and didn't change sizes going from module to module. But I still think we should have some ideal size for the individual pages, for when launching `cinnamon-settings <module>`.  In that case something like hotcorners would look silly with a mostly empty window.

Or, for example, with this PR, the Panel settings page now ends up with General Panel Options partially cut off:
image
Which is probably fine if the user started in the main settings then navigated there, but when they open Panel settings directly from the panel, I think it's more important to have that section fully visible (if the resolution allows - if it doesn't, the window size will be smaller anyhow I think, and it can be scrolled to):
image

I made a quick attempt to shorten the display module a while back without any luck - it should easily fit inside the normal window size.

I think we use fixed heights in the past because we had trouble making the content fit the window, or vise-versa, and this was a lazy way to do it, when certain widgets (like scrolled windows) were used - though I may be wrong. I don't care if we stop doing this, but it would require fixing the content of those pages so it all 'just worked'.

So:
- `cinnamon-settings`: consistent window size
- `cinnamon-settings <module>`: size to content, or some nice part of content if there's way too much.

When we first made the changes so that opening modules directly would eliminate the 'back' button, we decided to only give the window a specific title when opening directly, and keeping "System Settings"  when not, regardless of what page you go to, so I wouldn't change that. I don't think there was a specific reason, it was just collective opinion.

JosephMcc:
For me, the resizing works well outside of a couple panels that do produce the jarring effect that someone mentioned. Two that come to mind are the Panel and Mouse/Touchpad modules. But I think we could come up with a better solutions then doing away with the resizing. In Panel maybe some settings could be split to a second page so it doesn't open so tall. The mouse settings would require a bit more thinking. Maybe we could break the settings down differently than just Mouse/Touchpad.

No matter what when opening things from the main Cinnamon settings icon view you are always going to get presented with different size windows because some of them open separate applications.

fredcw:
The more I look at this, the less comfortable I am with the changes I made. Maybe it's best to leave it as it is for the time being.

PR #11684, opened 2023-06-04
Select monitor to show notifcations

EvilGremlin:
Very simple - choose monitor index, defaults to main if unavailabe.
I don't really understand what i'm doing - i had to edit and regenerate `org.cinnamon.desktop.notifications.gschema.xml`, are xmls generated automatically on build? What else i have to add?


JosephMcc:
You need to add your new setting to the xml file, something like is being done here: https://github.com/linuxmint/cinnamon/commit/3b5c560e8607f41019a9c9feb8d64e387bceb43e

EvilGremlin:
Thanks!

mtwebster:
The schema (org.cinnamon.desktop.notifications) already exists in the cinnamon-desktop package.

You just need to add the key itself here:
https://github.com/linuxmint/cinnamon-desktop/blob/master/schemas/org.cinnamon.desktop.notifications.gschema.xml.in

And you'll need to open an additional PR there also.

Also, have you tested this with different arrangements of multiple monitors? (I'm not sure monitor 0 is always the primary anymore in Cinnamon).

EvilGremlin:
thanks!

> I'm not sure monitor 0 is always the primary anymore

Looks like it is, but it doesn't matter anyway. It'll show on primary if set index don't exist. https://github.com/linuxmint/cinnamon/blob/a07f259e18c3274b5b4bac71661ba09cf20ca742/js/ui/messageTray.js#L945

JosephMcc:
@EvilGremlin Sorry for misdirecting you on where to put the gschema. I forgot that they existed on cinnamon-desktop.

kacperpaczos:
@EvilGremlin, sometimes I need to display notifications on the screen where I have the mouse cursor. Can you implement this functionality? 

For example, when I have multiple screens and the sound is muted, if I focus on a task on the second screen, I will miss the notification.

In another scenario, when I have a full-screen application, such as YouTube, playing something on my main screen, I can't see the content. I have to minimize the app or otherwise manually check it.

EvilGremlin:
Interesting! I will look into it. It'l be easy if cinnamon have a function to track current screen already, and quite tedious if i'd have to calculate it from cursor position myself.

rcalixte:
Apologies for intruding here. This is something I'd been working on for a while and had mostly implemented. (I pulled the incomplete development to open the PRs). The PR at #11703 will add an option for "Primary," "Active," and "Fixed" options.

rcalixte:
@EvilGremlin If you're interested, the two things I pulled were:
* option for showing notifications over fullscreen windows (it kind of works but I need to polish it so I pulled it for now)
* option for "All Screens" in the settings

The latter of those two options is the one I had to carve out significantly from the PR before opening it. I thought it could be done via JavaScript but I think some more hooks need to be exposed via the C libraries first. Feel free to play with it because I hit a wall and was waiting on circling back until I had adequate time for it. The main challenge was getting notifications to display more than once and if I got to that point, I wanted to make sure they would sync on all screens (so that if one screen was dismissed, all would dismiss, whether manually or automatically dismissed). This is what held me up from opening the PR sooner since I wanted it to be feature complete but I would love to see what someone else can come up with.

EvilGremlin:
It's ok to do it in few small PR's... it's not like we have to rework whole notify system.
Notify over fullscreen - yeah, would be much appreciated, especially it it can be done over vaapi players.

PR #11690, opened 2023-06-07
Panel icon size up to 128px

EvilGremlin:
I found current max very limiting, especially in multi-monitor configs. There can be all kinds of combinations and needs (personally i use 40 and 64px now). Potentially even 128 might be not quite enough in some setups, as we're pushing above 4K now (yes, i know about UI  scaling, it's not always applicable)

### DIFF
--- a/js/ui/panel.js
+++ b/js/ui/panel.js
@@ -42,7 +42,7 @@ const APPLETS_DROP_ANIMATION_TIME = 0.2;
 const PANEL_PEEK_TIME = 1500;
 
 const EDIT_MODE_MIN_BOX_SIZE = 25;
-const VALID_ICON_SIZE_VALUES = [-1, 0, 16, 22, 24, 32, 48];
+const VALID_ICON_SIZE_VALUES = [-1, 0, 16, 22, 24, 32, 48, 64, 72, 80, 96, 128];
 
 /*** These are defaults for a new panel added */
 const DEFAULT_PANEL_VALUES = {"panels-autohide": "false",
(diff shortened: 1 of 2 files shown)


clefebvre:
Just one thing about the icon sizes... most themes only have 32, 48, 64, 96, 128... there are no 40, 56..etc. These are there to make sure the icons are crisp.

EvilGremlin:
Good point... irrelevant with SVG icons, but with will 56 be downscaled from 64 or upscaled from 48?

js/ui/panel.js
```diff
@@ -42,7 +42,7 @@ const APPLETS_DROP_ANIMATION_TIME = 0.2;
 const PANEL_PEEK_TIME = 1500;
 
 const EDIT_MODE_MIN_BOX_SIZE = 25;
-const VALID_ICON_SIZE_VALUES = [-1, 0, 16, 22, 24, 32, 48];
+const VALID_ICON_SIZE_VALUES = [-1, 0, 16, 22, 24, 32, 48, 64, 72, 80, 96, 128];
```
mtwebster, line 45: Are you missing additional sizes here? (40, 56)

mtwebster:
> will 56 be downscaled from 64 or upscaled from 48
For named/theme icons, the starting size/file is chosen by the GtkIconTheme:
https://github.com/GNOME/gtk/blob/8ff9b2f83ff491cbfcbf9b30c706bd917679e7cc/gtk/gtkicontheme.c#L2975

I tried this out, and one issue is that it immediately changes the look of my panels.

The default panel size is 40. Currently, that size's optimal color icon is 32px (`if (maxSize < 48) return 32;...`).

The new behavior would now return 40 (`else if (maxSize < 48) return 40;...`).  This ends up, at least in my case, with a bit of a broken layout, with larger icons offset in my panel:

image

It's important to us to not impact users' existing desktop

JosephMcc:
@clefebvre is right. This should not use sizes that are not typically contained in icon themes. As soon as you scale you will get blurry icons. You also cannot depend on themes using SVG as many do not.

EvilGremlin:
Yeah, it needs some testing and polish, to only downscale when there is higher res. 
But if there's no higher resolutuion - it's totally ok to upscale. Ideally, it should be done using some algorythms from emulators, so 32>256 won't turn into blurry splotch.

PR #11693, opened 2023-06-08
sound@cinnamon.org: Truncate applet panel text by grapheme instead...

### LINKED ISSUE #11691: sound@cinnamon.org: Bad ellipsis cutting mid-UTF8 char
### Distribution

Mint 21.1 Cinnamon

### Package version

5.6.8

### Graphics hardware in use

_No response_

### Frequency

Always

### Bug description

When a song title is too long, its name may be cut with no respect of UTF-8 characters.

### Steps to reproduce

Listen to a music with UTF-8 characters in the name so that it get ellipsised. I don't know DBus enough to provide you with a command to emulate a music title being played but it might even be simpler.

In my case, it's the artist name that is cut of in the taskbar.
bug_utf8

### Expected behavior

UTF-8 characters should be considered one single atomic thing and shouldn't be cut in the middle. Actually, even whole graphemes (i.e. emojis using ZWJ) should not be split in half.

### Additional information

_No response_

fredcw:
... of codepoint.

Applet panel text (song information) is truncated by js string length resulting in multiple codepoint characters, e.g. emoji, being shown incorrectly. Count string length and truncate string by unicode grapheme instead.

fixes #11691

### DIFF
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -1296,8 +1296,9 @@ class CinnamonSoundApplet extends Applet.TextIconApplet {
             else {
                 title_text = player._title + ' - ' + player._artist;
             }
-            if (this.truncatetext < title_text.length) {
-                title_text = title_text.substr(0, this.truncatetext) + "...";
+            const graphemes = Util.splitByGrapheme(title_text);
+            if (this.truncatetext < graphemes.length) {
+                title_text = graphemes.slice(0, this.truncatetext).join("") + "...";
             }
         }
         this.set_applet_label(title_text);
--- a/js/misc/util.js
+++ b/js/misc/util.js
@@ -866,3 +866,25 @@ function getDesktopActionIcon(action) {
     else return null;
 }
 
+/**
+ * splitByGrapheme:
+ * @str (string): The string to be converted
+ *
+ * Converts a string, possibly containing multiple codepoint unicode characters (e.g. emoji), into
+ * an array of unicode graphemes. For example: "👩‍👩‍👧‍👧😃".length === 13, but splitByGrapheme("👩‍👩‍👧‍👧😃")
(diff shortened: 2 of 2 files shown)


fredcw:
splitByGrapheme doesn't work, it misses many edge cases. :(

edgarogh:
Indeed, seeing by the huge file [here](https://github.com/orling/grapheme-splitter), this may require a library. I've never contributed to Cinnamon so I don't really know if they'd accept including a file as big as this in their codebase (and since there's no package manager, it would effectively require copying it). At least the license is compatible (MIT).

fredcw:
@edgarogh That's exactly what I was wondering. It's nearly 2000 lines of js code so it would quite a big addition to cinnamon's js code and it's memory footprint. Also the function is only updated to unicode version 11 so it would have to be updated to the current version which I think is 15. I don't know of another way to do this perfectly in js.

The alternative is that I could write a much much shorter version (perhaps only 2 dozen lines and based on the splitByGrapheme function in my PR above) that would cover nearly all emoji and all the combining character blocks. It would be far from perfect but it would be a lot better than nothing and cover latin and other scripts. It would miss all the Hangul combining characters and a few other non latin scripts were you would get a count of 2 characters instead of one for instance but would catch the emoji which could otherwise be counted as anything upto 8 characters or more. Personally I think this solution would be good enough as it would catch nearly all of the worst offenders. Maybe @mtwebster has an opinion?

fredcw:
We don't need to do this in js, we just need to set the label width and the Pango.EllipsizeMode of the label's clutter_text to PANGO_ELLIPSIZE_END and Pango will do it automatically. 🫣

fredcw:
@edgarogh It turned out to be much easier than I thought. I couldn't find a way to set a maximum label width in pixels while still having the text ellipsized but it turns out I didn't need to. Cinnamon has its own St library to handle applets. panels, etc which uses Clutter to draw stuff and Clutter uses Pango to render text. And because Pango knows how to render unicode properly, I could use clutter method get_chars() to get the correct number of graphemes from the label text.

New PR #11704

PR #11704, opened 2023-06-15
sound applet: truncate applet panel text by grapheme instead...

fredcw:
...of javascript string length.

Applet panel text (song information) is truncated by js string length resulting in multiple codepoint characters, e.g. emoji, being shown incorrectly. Truncate string by number of unicode graphemes instead.

Fixes: #11691

### DIFF
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -1296,11 +1296,15 @@ class CinnamonSoundApplet extends Applet.TextIconApplet {
             else {
                 title_text = player._title + ' - ' + player._artist;
             }
-            if (this.truncatetext < title_text.length) {
-                title_text = title_text.substr(0, this.truncatetext) + "...";
-            }
         }
         this.set_applet_label(title_text);
+        
+        //truncate text by number of unicode graphemes.
+        let truncated_text = this._applet_label.clutter_text.get_chars(0, this.truncatetext);
+        if (truncated_text.length != title_text.length) {
+            truncated_text = this._applet_label.clutter_text.get_chars(0, this.truncatetext - 3) + "...";
+            this.set_applet_label(truncated_text);
+        }
     }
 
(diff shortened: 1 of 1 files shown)


fredcw:
Sorry, I should have tested more thoroughly. This doesn't work either.

mtwebster:
What about the accepted answer here: https://stackoverflow.com/questions/62341685/javascript-unicode-aware-string-slice

It's crazy there's no baked in javascript function for this.

fredcw:
@mtwebster Actually there is a new addition to js to do this [Intl.Segmenter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter) but while it's been implemented in Chrome, Safari and Opera, it hasn't been implemented by firefox yet so it's not in cjs. In any case I've got the solution now (I even tested it properly this time xD). Pango of course knows how to do it properly, I just had to figure out how to get it to do it with an arbitrary string without having to render it first.

```
function splitByGlyph(str) {
    const glyphs = [];
    const buffer = new Gtk.TextBuffer();
    buffer.set_text(str, -1);
    const iter = buffer.get_start_iter();
    const iter2 = buffer.get_start_iter();
    
    while (!iter2.is_end()) {
        iter2.forward_cursor_position();
        glyphs.push(buffer.get_text(iter, iter2, false));
        iter.forward_cursor_position();
    }
    return glyphs;
}
```

it's based on code in this article https://ar.al/2021/10/31/how-to-count-unicode-glyphs-in-vala-using-gtk/

I'll create another PR later when I have time - and probably double check it again ;)

mtwebster:
Can it use ClutterTextBuffer instead?
https://gjs-docs.gnome.org/clutter7~7-textbuffer/

fredcw:
@mtwebster cutter.text and textbuffer methods seem to only use units of character and doesn't seem to have an equivalent of forward_cursor_position(). Pango.LayoutIter has a next_cluster() method but I don't how you'd use that. Is there any disadvantage to using Gtk over Clutter?

Issue #9449, opened 2020-07-12
Blank icon for "Disks" (gnome-disks) in "System Settings".


DorianB2017:
```
 * Cinnamon version (cinnamon --version) - 4.6.6
   - Please specify if you are using the daily builds PPA - No
     (https://launchpad.net/~linuxmint-daily-build-team/+archive/ubuntu/daily-builds).
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...) - Mint 20.0
 * Graphics hardware *and* driver used
 * 32 or 64 bit - 64 bit
 * Attach /home/<username>/.xsession-errors, or /var/log/syslog
 ```

**Issue**
After upgrading from Linux Mint v.19.3 to Linux Mint v.20.0, an empty/a blank icon for "Disks" (gnome-disks) appears in "System Settings" (for all icon themes - Adwaita, Gnome-* , Hicolor, Numix* , etc., except the Mint-X* and Mint-Y* themes).
Please fix this bug, too.

For exemple:
- the Gnome/Gnome-colors-common icon theme:
![Captură de ecran din 2020-07-12 21-56-39a](https://user-images.githubusercontent.com/30018486/87254428-46782a80-c48b-11ea-8c75-fe30768ec8c4.png)

- the Numix theme:
![Captură de ecran din 2020-07-12 22-03-24](https://user-images.githubusercontent.com/30018486/87254502-db7b2380-c48b-11ea-8041-656c05597ae3.png)

xenopeek:
I can reproduce that. Switched icon theme to Adwaita and Disks doesn't show its icon in System Settings but it does have an icon in the menu and on the panel when running.

Odyseus:
Hello, everybody.

The icon used there is called "gnome-disks" (it was changed several monthds ago by 40a33842f2d820c11657bcb3a610601abe77a8ec). The commit message reads:

> The "disks" name doesn't exist in all themes, whereas gnome-disks is
> supported in Mint-X, Mint-Y and in hicolor for any other theme that
> doesn't have their own.

The problem seems to be that the hicolor icon theme doesn't ship that icon anymore.

Ping @mtwebster.

PR #11705, opened 2023-06-15
nvidia-settings icon fix

aneios:
set the icons for nvidia-settings and gnome-disks to better identify them in the cinnamon settings view


mtwebster:
Hi, there are reasons they're this way:

gnome-disks: see https://github.com/linuxmint/cinnamon/issues/9449 - gnome names icons after their app ids now, and not every theme still has a 'gnome-disks' icon. Searching through my theme directories, most gnome-disks icons look exactly like org.gnome.DiskUtility anyhow.

nvidia: Again, not every theme has an nvidia-settings icon (Adwaita for one), but we guarantee cs-drivers will be there since we ship it with Cinnamon. The nvidia-settings package provides one, but it's not in a location that anything can use reliably (it's there for the app itself to load only). If they were to fix that, then we could change it to nvidia-settings.

aneios:
I see your point about gnome-disks.

However, about nvidia-settings, I just checked and there is a specific nvidia-settings icon in both Mint-X and Mint-Y icon themes. Given that those are the default themes shipped and recommended by Linux Mint, is that not sufficient?

Most of the widely used icon themes (Yaru, Papirus, Qogir etc) also have a separate specific icon for nvidia-settings.

PR #11706, opened 2023-06-15
color-scheme follow simplified themes

aneios:
added a settings update for x.apps.portal.color-scheme that follows the theme mode from the new (from 5.8) simplified themes view

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_themes.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_themes.py
@@ -172,6 +172,7 @@ class Module:
 
             self.settings = Gio.Settings.new("org.cinnamon.desktop.interface")
             self.cinnamon_settings = Gio.Settings.new("org.cinnamon.theme")
+            self.xsettings = Gio.Settings.new("org.x.apps.portal")
 
             self.scale = self.window.get_scale_factor()
 
@@ -259,8 +260,8 @@ class Module:
             settings = page.add_section(_("Miscellaneous options"))
 
             options = [("default", _("Let applications decide")),
-                       ("prefer-light", _("Prefer light mode")),
-                       ("prefer-dark", _("Prefer dark mode"))]
+                       ("prefer-dark", _("Prefer dark mode")),
+                       ("prefer-light", _("Prefer light mode"))]
             widget = GSettingsComboBox(_("Dark mode"), "org.x.apps.portal", "color-scheme", options)
             widget.set_tooltip_text(_("This setting only affects applications which support dark mode"))
             settings.add_row(widget)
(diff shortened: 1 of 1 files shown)


Lurux:
@clefebvre, it looks like this PR implements the solution we suggested in linuxmint/mint21.2-beta#35.

This would offer a more consistent and expected behavior, without preventing users from individually adjusting the "dark mode" option in the advanced settings if they want to.

I hope this can be merged before release if it meets your quality standards !

anaximeno:
I believe a better approach could be to have an option 'follow the theme' (which could be the default one) in the advanced settings and when it is chosen the profile gets changed automatically (like this) when dark, light, or mixed is chosen in the styles section. This solves the situation where the user wants 'prefer light mode' to be active, but chooses a dark style in the styles section, while still wanting the prefer light mode to be active, the way it is right now the mode would be modified to the 'prefer dark mode' in this situation. What do you think about this?

Lurux:
Well they'd just have to do such a setup via the advanced settings, same as if they wanted to mix a green theme with orange icons. Your scenario seems like enough of an edge case for this not to be a problem, not to mention you have to visit the advanced settings in both cases anyways.

Not that I don't think this is a good idea - it would make the behavior a bit clearer - but, remember there's not much time left until release, so if this works well enough I'd prefer to see this merged ASAP rather than missing the mark because we wanted to add a minor improvement.

anaximeno:
> Your scenario seems like enough of an edge case for this not to be a problem, not to mention you have to visit the advanced settings in both cases anyways.

With what I proposed the user only would have to go to the settings one time and choose the option he wants to use, the way it is now the option would change every time he chose a different style. I have the impression that it would be a more consistent behavior than having the setting change automatically independent of the preference of the user.

I agree with you that time is a factor here, but since we are talking about system software we should be careful and thoughtful with the changes we made.

Lurux:
Given the status quo is the setting not changing *at all*, I think an overwhelming amount of users will complain choosing a dark style doesn't set the dark mode (if this doesn't get merged) as compared to those who would complain you have to use the advanced settings to use light mode with a dark style - I mean seriously, who would want to do that anyways ?

I don't want to ping the team too much, but IMHO it would be idiotic not to merge this before release.

aneios:
> With what I proposed the user only would have to go to the settings one time and choose the option he wants to use, the way it is now the option would change every time he chose a different style. I have the impression that it would be a more consistent behavior than having the setting change automatically independent of the preference of the user.

I think the "advanced" theme settings (which up until 21.2, were just the normal theme settings) work perfectly fine and are simple enough for everyone to understand and use.

The main value proposition I see in the new "simplified" settings is to obtain a coherent unified theme in as few clicks as possible. The fact that the settings choose for you (being opinionated) is a feature, not a bug. If you want precise specific options, the more detailed ("advanced") theme settings aren't going anywhere and will continue to work exactly as before.

Also, for what it's worth, I've been running this modification on my Mint (hardware) as well as on Manjaro and Endeavour (VMs) for the past few weeks and have encountered zero unexpected behaviours.

Sooo... it works on my machine, lol :)

anaximeno:
I gave a bit more analysis of this situation and it would make sense since the change would be similar to how the other components are changed automatically when you choose light, dark, or mixed styles. Sometimes you gotta apply a bit of the KISS principle to the situation :)

aneios:
> Sometimes you gotta apply a bit of the KISS principle to the situation :)

I just want to thank you for making me find out that there is an actual acronym for this principle :rofl:

Lurux:
Nice ! 🎉

PR #11726, opened 2023-06-23
cs_gestures.py: Cleaning up Python

rcalixte:
* Remove unnecessary semicolons
* Clean up whitespace
* Use f-strings for string interpolation
* Use `enumerate` to generate a counter for an iterable
* Clean up conditional block to use map and return structure for improved readability
* Remove unused exception variable
* Clean up comparison to `None`
* Remove unnecessary `.keys()` call


mtwebster:
Those sneaky semicolons...

PR #11727, opened 2023-06-25
theme settings page: add a small l/r margin

fredcw:
Some l/r margin is needed when a larger font size than the default is used.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/themes.ui
+++ b/files/usr/share/cinnamon/cinnamon-settings/themes.ui
@@ -14,6 +14,8 @@
         <property name="halign">center</property>
         <property name="valign">start</property>
         <property name="margin-top">45</property>
+        <property name="margin-left">20</property>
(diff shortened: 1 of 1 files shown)


files/usr/share/cinnamon/cinnamon-settings/themes.ui
```diff
@@ -14,6 +14,8 @@
         <property name="halign">center</property>
         <property name="valign">start</property>
         <property name="margin-top">45</property>
+        <property name="margin-left">20</property>
+        <property name="margin-right">20</property>
```
mtwebster, line 18: Can you use margin-start, -end instead of left/right (these are deprecated properties).

fredcw:
Ah, this has since been fixed (and in the correct way) in https://github.com/linuxmint/cinnamon/commit/c954fd284eb9d4a6c68bd1e5ab714bc0c0295c64

Issue #11373, opened 2022-12-30
Very slow animations (especially compared to KDE and GNOME)


Freemason-12:
```
 * Cinnamon version (5.6.5)
   - No daily build PPA
 * Distribution - Mint 21
 * Intel HD Graphics 520
 * 64 bit
 ```

### Issue
Animations of overview effect and Expo are slow as hell (very low FPS)

### Steps to reproduce
trigger overview effect or Expo

### Expected behaviour
Everything smooth

### Other information

https://user-images.githubusercontent.com/97738675/210048433-d56bb814-052e-4bdb-b2be-a29addbf00ad.mp4

### Configuration info, if applicable
Animation issue has been a very big problem for for a lot of time. Issue seems to persist since Cinnamon 4.6.7 but before Cinnamon had an option in settings > general for animation type, and setting this specific setting to 'something else that default solved the problem (at least partially because animations were smooth but some flickering happened sometimes). Now in Cinnamon 5.4 and later this option disappeared and animations got worse

**Comment by Freemason-12** [comment]
**Another video on Mint 20.3 with Cinnamon 5.2.7**
the setting I was talking about earlier above in _VSync method_. When set to _presentation time_ (default) it was like this:

https://user-images.githubusercontent.com/97738675/210053276-80d4acbe-4ce9-45d1-adee-ec2fd4cd856d.mp4

And it became smoother when setting it to a different parameter. But not in all cases:

https://user-images.githubusercontent.com/97738675/210053653-01507ee6-1e70-4e90-879d-fa4d61de07e4.mp4

**Comment by xhpohanka** [comment]
I have similar experience:

```
System:
  Kernel: 6.1.9-arch1-1 arch: x86_64 bits: 64 compiler: gcc v: 12.2.1 Desktop: Cinnamon v: 5.6.7
    tk: GTK v: 3.24.36 dm: LightDM Distro: Arch Linux
Machine:
  Type: Laptop System: LENOVO product: 20RH001PMC v: ThinkPad P43s serial: <superuser required>
    Chassis: type: 10 serial: <superuser required>
  Mobo: LENOVO model: 20RH001PMC v: SDK0J40697 WIN serial: <superuser required> UEFI: LENOVO
    v: N2IETA0W (1.78 ) date: 10/14/2022
Battery:
  ID-1: BAT0 charge: 51.5 Wh (100.0%) condition: 51.5/50.5 Wh (101.9%) volts: 12.9 min: 11.6
[...]

PR #11734, opened 2023-06-29
cinnamon ui: improve overview animation

anaximeno:
The hidden windows were being shown before the end of the overview animation. Now the window will return to its original place before the real hidden window is shown in place.

May fix https://github.com/linuxmint/cinnamon/issues/11373

### DIFF
--- a/js/ui/overview.js
+++ b/js/ui/overview.js
@@ -407,7 +407,7 @@ Overview.prototype = {
             // Make other elements fade out.
             Tweener.addTween(this._group, {
                 opacity: 0,
-                transition: 'easeOutQuad',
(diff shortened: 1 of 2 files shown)


kacperpaczos:
Can you make a videos with before and after effects?

anaximeno:
> Can you make a videos with before and after effects?

Sure, thanks for suggesting it.

### Before

https://github.com/linuxmint/cinnamon/assets/63073056/dbbbce0d-bc2e-419b-8304-e0417eaaf911

### After

https://github.com/linuxmint/cinnamon/assets/63073056/141061df-cca6-482d-95af-a8b24c9c3c3d

Issue #8534, opened 2019-04-26
Pinned icons are not synchronized in others workspaces after changes


medmedin2014:
```
 * Cinnamon 4.0.10
 * Distribution - Mint 19.1
 * 64 bit
 ```

**Issue**
After adding new pinned windows from menu or modifying the order of pinned apps by moving them by mouse inside grouped list windows applet, the change is not reflected to others work spaces.

**Steps to reproduce**

**Expected behaviour**

**Other information**

Zorandon:
This bug is still present in Cinnamon 4.2.2. It is important to note though that a new order of icons (or added icons) is correctly represented on on all workspaces after a logout/login.

PR #11740, opened 2023-07-02
grouped window list: fix drag and move app icon on the app list

anaximeno:
#### This fixes the following bug:
1. Go to any workspace but the first one
2. Reload the grouped-window-list applet
3. Go to another workspace
4. Try to move the pinned apps in the grouped window list applet (It will probably not work)
5. Try to drag any app from the menu and drop in the grouped window list (It will also not work, probably)

#### .xsession-errors error log:

Screenshot from 2023-07-02 01-37-01

---
  
I've also fixed https://github.com/linuxmint/cinnamon/issues/8534

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/applet.js
@@ -548,7 +548,7 @@ class GroupedWindowListApplet extends Applet.Applet {
     }
 
     refreshCurrentAppList() {
-        let appList = this.appLists[this.state.currentWs];
+        let appList = this.getCurrentAppList();
         if (appList) appList.refreshList();
     }
 
@@ -643,7 +643,16 @@ class GroupedWindowListApplet extends Applet.Applet {
     }
 
     getCurrentAppList() {
-        if (typeof this.appLists[this.state.currentWs] !== 'undefined') {
+        let metaWorkspace = global.workspace_manager.get_workspace_by_index(this.state.currentWs);
+
+        let refWorkspace = findIndex(
+            this.appLists,
+            (item) => item.metaWorkspace && item.metaWorkspace === metaWorkspace
+        );
+
+        if (refWorkspace !== -1) {
+            return this.appLists[refWorkspace];
+        } else if (typeof this.appLists[this.state.currentWs] !== 'undefined') {
             return this.appLists[this.state.currentWs];
         } else if (typeof this.appLists[0] !== 'undefined') {
             return this.appLists[0];
@@ -694,6 +703,8 @@ class GroupedWindowListApplet extends Applet.Applet {
 
             this.state.set({scrollActive: true});
 
+            let appList = this.getCurrentAppList();
+
             let isAppScroll = this.state.settings.scrollBehavior === 2;
             let direction, source;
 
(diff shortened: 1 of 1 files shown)


fredcw:
@anaximeno I was thinking about replacing all the usages of the util.js functions, each, find, findIndex, filter, and map, with js natives in all of cinnamon's code including all xlets (appart from a few other places, it's mainly just grouped window list applet that uses them). Maybe they were originally used as polyfills but I don't see how they're useful now.

In any case, it's not worth changing your code at this point. Just wondering if you had any opinion on this?

anaximeno:
@fredcw, native implementations generally benefit more from optimizations made to the language and its interpreter, so it seems to be a positive change.

claudiux:
I don't see any improvement in Cinnamon 5.8.3 (Mint 21.2) where I've never seen this type of error message in `~/.xsession-errors`.

anaximeno:
@claudiux 

Following the steps above:

### Before applying the patch

https://github.com/linuxmint/cinnamon/assets/63073056/0b344093-4664-4421-b021-9c98dc0dba29

### After applying the patch

https://github.com/linuxmint/cinnamon/assets/63073056/a24b9d94-7689-49e6-a58a-40c25e0c1b2f

### Why this happens?

The gwl stores the workspaces where it is active inside a variable `this.appLists` which is a basic JS (CJS) array. The workspaces are stored lazily, therefore, if I started from the workspace  4, the first element in the list would be the workspace number 4. What was causing the problem is that it was being used the current workspace's index to get the variable corresponding to the current workspace in the variable `this.appLists`, but since the current workspace index doesn't always corresponds to the index its is stored in the `this.appLists` variable, that error were being triggered.

This problem can occur even if the applet started from the first workspace, which is the usual for most of the users, but then went to another workspaces in a non sequential order, we would have the same problem where the index of the current workspace (now, in this situation, besides the first workspace) could not correspond the correct representation stored in the `this.appLists` variable making the wrong instance of the gwl being used which would invalidate the drag and drop (and other direct operations) being made.

***Please let me know if I were not clear in what I tried to explain here, and I'll give a more in dept explanation when having more time***

claudiux:
@anaximeno These explanations are very clear, thank you. But with Cinnamon 5.8, this applet doesn't let you change the order of the icons, with or without applying your patch.

claudiux:
Please test it. I can't merge this PR as it is.

anaximeno:
> Please test it. I can't merge this PR as it is.

I've tested now the two scroll options I found in the gwl settings:

1. General/Mouse wheel scroll action
2. Thumbnails/Cycle windows on mouse wheel scroll

They all seem to be working as expected.

anaximeno:
> Which version of cinnamon do you have?

5.8.2

fredcw:
Looks good to me, fixes the bugs. Tested on cinnamon 5.8.3.

Also impressed that someone has managed to figure out how this applet works. I've tried a few times to figure it out and always given up. That whole state object (state.js) completely confuses me. One of the first things I did in Cinnamenu was get rid of it.
Also, if `this.applist` in `class Applist` is an array of workspaces, why isn't it called `this.workspaces`? That's why you end up with things like `count = appList.appList.length - 1;`

anaximeno:
> Also, if `this.applist` in class Applist is an array of workspaces, why isn't it called `this.workspaces`? That's why you end up with things like `count = appList.appList.length - 1`;

AppList isn't concretely an array of workspaces, it seems to be more of an encapsulation of the context of the list of apps in the gwl applet in its specific workspace. The meta workspace is stored inside the AppList class, and that's what I am using this to identify what is the correct AppList instance for the current workspace.

I believe they are using the state to synchronize relevant info from all instances of the `AppList` classes, but I agree it might be a bit confusing though.

Btw, this https://github.com/linuxmint/cinnamon/blob/master/files/usr/share/cinnamon/applets/grouped-window-list%40cinnamon.org/appList.js is the file where AppList is defined.

Issue #11106, opened 2022-08-23
[BUG] Grouped Window List not showing open programs


bphilpott:
```
 * Cinnamon 5.4.11
 * Distribution - Mint 21 
 * Graphics hardware *and* driver used - Device-1: AMD Lucienne driver: amdgpu v: kernel
 * 64 bit
 ```

### Issue

The Group Windows List applet is having an issue in which it does not display the open programs properly.

For example, if I have Firefox open, it should have a line under it and when I hover it should show my open Firefox instances. Or same with the Nemo file browser (or other applications). But it sometimes acts as if a particular application is open. If I hover over it, I see no previews like normal and clicking on it should bring it to the foreground, but instead opens a new instance. I can alt-tab over to the program but the GWL does not acknowledge that the program is open, even if it is in the forefront. I've tried to remove Window List and add it back in with fresh settings but it behaves the same.  

This happens at random.  It can work fine for a while and then will stop working for no reason that I have yet to observe. This system was upgraded from LM20.3 to LM21.  I have multiple screens and will frequently switch screens as I am often mobile with my laptop

### Steps to reproduce

Open programs that display in the Grouped Window List.  Open program may eventually stop being shown as open in GWL.

### Expected behaviour

Open program should be shown as open until it is terminated.

### Configuration info, if applicable

Here are the exported settings from the Applet Preferences
https://pastebin.com/NrbbWVdT

mtwebster:
Have you tried resetting the applet's settings to defaults (same menu where you exported them)?

Can you attach your `~/.xsession-errors` of when this is happening?

Does the other window list show previews ok? (You can't click them like GWL but they should be shown).

bphilpott:
I didn't reset the settings from the menu, but I removed the applet and added it back and it had reset my settings.  I'm assuming that is basically the same thing, but if not I can manually reset them.

Yes, I just added the other window list to my panel, next to the GWL, and it shows everything. In fact, when I added it, it made the GWL to show everything as expected all of a sudden.

This behavior started when I did the LM20.3->21 upgrade.  I don't remember if I noticed it that very day, but extremely shortly thereafter.

Here is my .xession-error file:
https://pastebin.com/bRDQa84A

Huecuva:
[...]

PR #11759, opened 2023-07-18
Fix thumbnails not appearing for windows created while menu is open

jminer:
With this change, I don't have #11106 happen anymore.

For me, the bug happened whenever a window is created while the hover menu is open.

Thumbnails objects were being created and added to the list of thumbnails in the `AppThumbnailHoverMenu`. However, then `refreshThumbnail` is called, which calls `getThumbnail`. In `getThumbnail`, `metaWindowActor` is still null and `metaWindow.get_compositor_private` returned null, causing it to remove the thumbnail again before it was ever visible.

With the `setTimeout(..., 0)`, `getThumbnail` is called on the next event loop turn, so that the actor will be returned by `get_compositor_private`.

If the menu isn't open when the window is created, then the window is added to the `queuedWindows` array and added later on when the actor is available.

I don't understand the code great (like when events are triggered and the actor lifecycle), so I'm not sure if there would be a better fix. This seems like a fairly safe fix, and it's currently quite broken since it misses all windows that are opened while the menu is open. I haven't noticed issues from it. Does this seem like a reasonable fix or is there possibly a better one?

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/menus.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/menus.js
@@ -821,7 +821,9 @@ class WindowThumbnail {
         }
 
         this.labelContainer.child.text = this.metaWindow.title || '';
-        this.getThumbnail(thumbnailWidth, thumbnailHeight);
+        setTimeout(() => {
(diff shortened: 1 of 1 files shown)


mtwebster:
> I don't understand the code great

Don't worry, a lot of us don't either. Looks ok to me,

PR #11779, opened 2023-07-24
Always show panel if a menu open on it

Gr3q:
instead of the current policy of "Don't hide the panel if a menu is open on it". With this, if you open an applet with a keybinding, the panel for it will also show.

I made these changes for another PR, so I thought I open it. Discuss it among yourselves if it's really needed. Maybe it can be turned into an optional toggle (feature).

### DIFF
--- a/js/ui/main.js
+++ b/js/ui/main.js
@@ -368,7 +368,7 @@ function start() {
 
     global.reparentActor(global.top_window_group, global.stage);
 
-    global.menuStackLength = 0;
+    global.menuStack = [];
 
     layoutManager = new Layout.LayoutManager();
 
--- a/js/ui/panel.js
+++ b/js/ui/panel.js
@@ -3548,6 +3548,26 @@ Panel.prototype = {
     },
 
     /**
+     * _panelHasOpenMenus:
+     * 
+     * Checks if panel has open menus in the global.menuStack
+     * @returns 
+     */
+    _panelHasOpenMenus: function() {
+        if (global.menuStack == null || global.menuStack.length == 0)
+            return false;
+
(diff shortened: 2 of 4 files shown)


fredcw:
@clefebvre I don't really see what this merge achieves except that the option to not show the panel when the menu is opened by shortcut key is no longer available (and thus now needs to be removed from the menu applet https://github.com/linuxmint/cinnamon/issues/12536)

Was that the intention?

The original issue that I think this was intended to fix (#10970) was fixed with #11788

btw, if the intention is to remove the option and always show the panel then could that not be more easily be achieved by simply removing the option from the menu applet.

JosephMcc:
I think when this PR was opened and again when it was merged, I think everyone just missed that the option was even there. There was no reason for this unless the option is going to be removed.

Gr3q:
This PR is was:

* Opened before  https://github.com/linuxmint/cinnamon/pull/11788 was opened and merged
* It's not only for the menu applet, it's for all applets
* Was made to fix https://github.com/linuxmint/cinnamon-spices-applets/issues/4964 

Before the fixes the opening behaviour was visually problematic, the panel height (in auto-hide or intelligent mode) was not always considered when showing an applet menu when it should've been, resulting in in the menu either hiding behind the panel or it's position changing after opening, neither being good.

With more work it might be possible to always show the applet menu in the correct y position but that won't change the fact that it could move around (which is never a good thing), so that would need to be prevented too.

In retrospect this fix might not be the best for everyone, because it seems some people want different show/hide behaviour for each applet.

Nevertheless, I don't really mind what happens to this PR, will it be reverted or not, at the time this was the easiest fix I could think of that fixes all of the problems above.

Issue #11547, opened 2023-03-09
default menu behaviour polish


beastr45:
```
 * Cinnamon version (cinnamon --version)
   - Please specify if you are using the daily builds PPA
     (https://launchpad.net/~linuxmint-daily-build-team/+archive/ubuntu/daily-builds).
 * Distribution - (Mint 21, Arch, Fedora 37, etc...)
 * Graphics hardware *and* driver used
 * 32 or 64 bit
 * Attach ~/.xsession-errors, or /var/log/syslog
 ```
cinnamon version 5.6.7
### Issue
with the default cinnamon menu included in linux mint it is possible to make the taskbar and the menu to overlap. 

### Steps to reproduce
In order to reproduce this bug set the taskbar to hide. Then press the windows icon key and the taskbar will overlap with the menu

### Expected behaviour

the expected behaviour would be for the menu to place itself above the taskbar or to have some sort of logic to position the menu under certain conditions.

### Other information
I have checked if this is fault of my desktop theme, it is not.

### Configuration info, if applicable
If this bug report is related to an Applet, Desklet or Extension, please paste (or use a pastebin service) the offending extension's settings, if it has any.  You can obtain this by opening its configuration, clicking the 'hamburger' button in the upper-right corner of the window, and selecting "Export to a file". **Please be sure to review the contents and remove any personal data it may contain.**

This is related to the default menu applet in Linux mint. I am not sure if this behaviour replicates with other menu's or if the same menu desk-let is in use in normal cinnamon. nevertheless someone could very easily fix this in around 5 minutes.

fredcw:
Duplicate of #10970

fredcw:
This issue was fixed by 11788 above and can be closed.

Issue #11429, opened 2023-01-17
Application menu: shrunk down vertically when scaling 125%


firefoxlover:
Freshly installed Linux Mint 21.

Scaled the screen to 125%, the application menu kinda faded into the downside of the screen.

Going to settings and disable "force the panel to be visible" fixed it.

fredcw:
Is this issue #10970 ?

Issue #11035, opened 2022-08-10
Bar autohide + Super key


lucasgabmoreno:
```
 * Cinnamon 5.4.9
 * Distribution: Linux Mint Debian Edition LMDE 5
 * Intel HD 5500(i915) + Nvidia 940M (nvidia-driver)
 * 64 bit
 ```

**Issue**

After updating to last Cinnamon version, when press Super (Winkey) in autohide bar, this overlaps Menu.

![2022-08-10_19-04](https://user-images.githubusercontent.com/83520011/184028506-2233432f-1cc3-4100-a1ab-140e7bfcf043.png)

**Steps to reproduce**
Super (Winkey)

**Expected behaviour**
No overlap Menu and bar

mtwebster:
Hi, can you try out the Mint cinnamon themes (any of them) to see if this happens there?  This may be a theme issue.

lucasgabmoreno:
Nope, same problem with Mint-Y-Dark (in different computers).

![Captura de pantalla de 2022-08-10 21-04-56](https://user-images.githubusercontent.com/83520011/184043507-d5da0ca7-39e0-453a-a531-0b4ad079569b.png)

rcalixte:
I've noticed this as well but haven't had a chance to dig into this part of the code yet. Essentially what is happening is when the panel is set to auto-hide and the Super key is pressed (or whatever Open Menu accelerators are configured) the Menu is displayed before the panel is unhidden which causes the panel to appear above the bottom of the menu dialog box. The same thing can happen if the Menu has "Open the menu when I move my mouse over it" enabled.

It seems like a race condition where the panel needs to be expanded first and only once fully expanded, then the Menu should be opened.

@lucasgabmoreno A workaround I've found when needed is to mouse over the panel to expand it before opening the Menu or disabling the option "Force the panel to be visible when opening the menu" in the settings.

rcalixte:
Duplicate of #10970

lucasgabmoreno:
Thanks @rcalixte! But I can't find "Force the panel to be visible when opening the menu" en settings.
![Captura de pantalla de 2022-08-13 10-31-36](https://user-images.githubusercontent.com/83520011/184496367-ad672262-99ed-40fc-a4aa-abe6cca5f5ce.png)

mtwebster:
Right-click the start menu button and select 'Configure' - it's in the applet's settings, not panels.

mtwebster:
[...]

PR #11788, opened 2023-07-24
Menu applet: Fix bug where menu is partly behind panel

### LINKED ISSUE #10970: Panel/Menu applet: Auto-hide panel partly occludes menu applet when opened with meta-key

fredcw:
Call peekPanel() before toggling menu open to ensure menu is positioned correctly.

Note: this PR makes no difference to current behaviour either way (doesn't fix the bug nor cause any other issues) unless PR #11770 is also merged. _updateKeybinding() is also the correct place to call peekPanel() anyway because it's only needed when menu is opened with key-press.

Fixes #10970, #11035, #11429, #11547


fredcw:
Replaces PR #11770

Issue #7745, opened 2018-07-17
System Settieng / Power management, shows one battery only


luismanson:
```
 * Cinnamon version 3.8.7
 * Distribution - Mint 19
 * i915 / Intel 620
 * 64 bit

```
**issue** 

Hello, i have noticed that in system settings under power management, the batteries tab shows only one battery, while my system has two, which are correctly shown under power@cinnamon 

**Steps to reproduce**

have two batteries

**Expected behaviour**

two batteries being listed

**Other information**

real-or-random:
I can confirm this. `upower` shows two batteries but the power management panel shows only the first one.

real-or-random:
I think the bug is here: https://github.com/linuxmint/Cinnamon/blob/7e96eb9a6a121be29020ed8ea2fb4e45341d6da5/files/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py#L392
As far as I can tell, the battery will only be displayed if it's full or empty. By the way I don't see a reason why an additional battery should be treated differently than the primary one when it comes to displaying.

Vahan86:
@luismanson, is this still an issue in Cinnamon 4.2?

luismanson:
I'm not currently using Cinnamon. I can try 4.2 next week if you need

real-or-random:
I can confirm that this is still an issue in  Cinnamon 4.4

icarter09:
@real-or-random you do have `Always show all batteries` in the `Power Manager` settings enabled correct?

real-or-random:
> @real-or-random you do have `Always show all batteries` in the `Power Manager` settings enabled correct?

This bug report is about the  "Power Management" *settings module*, which does not have such a setting.

The "Always show all batteries" setting is a setting in the "Power Manager" applet. (And enabling it does not affect the settings module of course.)

sublime238:
When I lock the screen , both batteries are visible in upper right corner. So here it works correct as in panel.

sublime238:
Sometimes both are visible, when unplug and plug again the power , or I take out the second battery and put it in again.
![image](https://user-images.githubusercontent.com/14123399/72217591-58a2fb80-3530-11ea-94c7-6c3f6ebe8d90.png)

airasim:
I'm running Cinnamon 4.8.6 and still have the same problem.

Like in issue #9024 `upower -d` shows a virtual combined device named "DisplayDevice" after the two physically batteries.

[...]

PR #11790, opened 2023-07-25
Show dual batteries properly in Power Management / Cinnamon Settings

libf-de:
I'm using Cinnamon on a ThinkPad T480 with dual batteries. Currently both batteries are shown in the panel applet, but only the internal one is shown in power management:
grafik

The reason for that is, as mentioned in #7745, that additional batteries are only shown if they are fully charged or empty. I see no point doing it like that. Adding the other charging states:
grafik

Now it's visible, but no percentage and it's impossible to rename it. Also can't see the point for doing it like this. So, after making every battery a "primary battery":
grafik

I've also added "PENDING_CHARGE" and "PENDING_DISCHARGE" as separate states, so it's possible to differentiate which battery is in use / being charged. Those will need translations.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py
@@ -320,16 +320,10 @@ class Module:
                 else:
                     primary_settings.add_row(self.set_device_ups_primary(device))
             elif device[UP_TYPE] == UPowerGlib.DeviceKind.BATTERY and not ups_as_primary:
-                if not have_primary:
-                    if not primary_settings:
-                        primary_settings = self.battery_page.add_section(_("Batteries"))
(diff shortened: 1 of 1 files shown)


libf-de:
Of course, can you just merge the single commit, or should I open another pull request?

The theme is Orchis-dark: [https://github.com/vinceliuice/Orchis-theme](https://github.com/vinceliuice/Orchis-theme)

mtwebster:
Ok merged the one commit - https://github.com/linuxmint/cinnamon/commit/b7f958ee0290abbe84c476fff03ee5826792f2a6

I'll leave this open until I get the other one in.

Thanks

mtwebster:
I had the additional commit in my local repo still.

https://github.com/linuxmint/cinnamon/commit/8214ed15927d59e4a7054404c2c584eeb4ba4a05

Thanks

PR #11801, opened 2023-07-30
Updated write-applet.xml for Cinnamon 4.6 to 5.8+

claudiux:
Changing example.

### DIFF
--- a/docs/reference/cinnamon-tutorials/write-applet.xml
+++ b/docs/reference/cinnamon-tutorials/write-applet.xml
@@ -272,7 +272,7 @@
     </informalexample>
 
     <para>
-      Note that <link linkend="cinnamon-js-ui-Applet-IconApplet-set_applet_icon_symbolic_name"><code>set_applet_icon_symbolic_name</code></link> is a function defined inside <code>Applet.IconApplet</code>, which makes the applet display the corresponding icon. By using the applet API, we have saved ourselves from the hassle of creating icons and putting them in the right place. (<code>bell-notif</code> is the name of an icon from the user's icon set. The icons available for use can be found in <code>/usr/share/icons/</code> or in the <code>icons</code> folder of this applet)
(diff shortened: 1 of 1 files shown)


docs/reference/cinnamon-tutorials/write-applet.xml
```diff
@@ -90,28 +95,36 @@
...
-                this.set_applet_tooltip(_("Click here to kill a window"));
-            },
-
-            on_applet_clicked: function() {
-                Util.spawn(['xkill']);
+            on_applet_clicked(event) {
+                // Please consult 'man notify-send'.
+                let command = `bash -c 'notify-send -u normal `;
+                command += `"Thank you ${user}" "for improving Cinnamon!"'`;
+                Util.spawnCommandLineAsync(command)
```
mtwebster, line 122: Maybe a comment emphasizing that async should always try to be used, and why.

docs/reference/cinnamon-tutorials/write-applet.xml
```diff
@@ -90,28 +95,36 @@
       <programlisting>
         const Applet = imports.ui.applet;
         const Util = imports.misc.util;
+        const GLib = imports.gi.GLib;
+
+        /**
+         * Variables (and other constants)
+         */
+        var user = GLib.getenv("USER"); // acts like the shell command: echo -n $USER
```
mtwebster, line 103: Maybe GLib.get_user_name()?

docs/reference/cinnamon-tutorials/write-applet.xml
```diff
@@ -90,28 +95,36 @@
...
+
+        user = user.charAt(0).toUpperCase() + user.slice(1); // Capitalize 'user'
```
mtwebster, line 105: we also have String.prototype.capitalize() (overrides.js)

docs/reference/cinnamon-tutorials/write-applet.xml
claudiux: Done!

PR #11823, opened 2023-08-11
[unfinished] disable touchegg support added

sulincix:
Users can configure touchegg with Touché.
I added extra config for disabling touchegg support. 
default disabled.

https://github.com/JoseExposito/touche

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -18,6 +18,11 @@ option('disable_networkmanager',
     value: false,
     description: 'Build without networkmanager'
 )
+option('disable_touchegg',
+    type: 'boolean',
+    value: false,
+    description: 'Build without touchegg support'
+)
 option('py3modules_dir',
     type : 'string',
     value : '',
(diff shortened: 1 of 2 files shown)


mtwebster:
This is redundant.  If you don't want Cinnamon to handle touchegg events, leave gestures disabled in Cinnamon's settings. This class you're removing from the build [isn't used until you enable it](https://github.com/linuxmint/Cinnamon/blob/master/js/ui/gestures/gesturesManager.js#L113C2-L151).

You can get the touchegg client to autostart again by removing it from the session blacklist:
Using dconf-editor:
```
org.cinnamon.cinnamon-session
```
or using the gsettings command:
```
org.cinnamon.SessionManager
```
and remove `touchegg` from the list.

sulincix:
touchegg killed by cinnamon and I cannot use.

mtwebster:
Remove touchegg from the blacklist like I explained, log out and back in, the touchegg client will auto-start like it used to.

Issue #11857, opened 2023-09-11
xdg-desktop-portal-gtk no longer used with x-d-p >= 1.17, x-cinnamon-portals.conf needed


smcv:
### Distribution

any, I used Debian testing (Debian 13 prerelease)

### Package version

5.6.8

### Graphics hardware in use

qemu virtual machine with virtio emulated GPU 

### Frequency

Always

### Bug description

In older versions of x-d-p, if x-d-p does not find a suitable backend for the current desktop environment, it would fall back to trying any backend. This meant that all backends needed to be prepared to run in an environment where their functionality cannot actually work, which is rarely tested, leading to undesired situations where a desktop environment's backend can cause bugs while running different desktop environments (like https://github.com/linuxmint/xdg-desktop-portal-xapp/issues/5 and https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/issues/74, each of which was a problem for "the other" desktop environment).

In the x-d-p 1.17.x prereleases and the upcoming 1.18.x stable release, this has been changed to give desktop environment developers more control over the backends they use. As described in https://github.com/flatpak/xdg-desktop-portal/blob/main/doc/portals-conf.rst, the recommendation is now that each desktop environment should provide a file listing its preferred backends, similar to the one added by https://gitlab.gnome.org/GNOME/gnome-session/-/merge_requests/95/diffs in GNOME.

This mechanism is similar to mimeapps.list, and it makes sense to configure it in some "reasonably central" desktop package - in GNOME, it's gnome-session that took responsibility for this. I don't normally use Cinnamon myself, but either cinnamon-common or cinnamon-session seems fairly appropriate.

As a result of not providing an `x-cinnamon-portals.conf` file, x-d-p falls back to only using the portals that have `UseIn=X-Cinnamon`. This includes x-d-p-xapp but not x-d-p-gtk. The result is that there is no portal for interfaces that are not implemented in x-d-p-xapp, notably FileChooser and OpenURI, which are important/basic functionality for sandboxed apps.

Another symptom of not having this configuration is that you will see warnings from `xdg-desktop-portal`:

```
xdg-desktop-portal-WARNING **: 16:14:51.348: Choosing xapp.portal for org.freedesktop.impl.portal.Inhibit via the deprecated UseIn key
```

and if debug output is enabled (`/usr/libexec/xdg-desktop-portal --verbose`), you'll also see that interfaces like `FileChooser` are not being provided.

### Steps to reproduce

[...]

PR #11863, opened 2023-09-16
add xdg-portal conf

leigh123linux:


mtwebster:
Can you add xdg-desktop-portal-gtk to debian/control Depends?

leigh123linux:
> Can you add xdg-desktop-portal-gtk to debian/control Depends?

Done.

Fantu:
Sorry, when I saw this PR I didn't noticed that x-cinnamon-portals.conf was added to cinnamon package, I added [on Debian 4 months ago](https://salsa.debian.org/cinnamon-team/cinnamon/-/commit/f87c64f8d35ba406eb11ad442989a0716f6620cf) in cinnamon-common instead as don't arch dependent and without need to stay on arch dep. package FWIK. I also saw it similar in -common of mate and gnome.
Is there a reason to have it in cinnamon instead cinnamon-common? In that case I would move it in Debian

mtwebster:
We can move it to cinnamon-common if that's better.

Fantu:
After https://github.com/linuxmint/cinnamon/issues/11857#issuecomment-1913783526 I decided to move it to Debian in cinnamon

PR #11871, opened 2023-09-25
Replace util.js polyfills with JS natives...

fredcw:
Replace all instances of `js/misc/util.js` polyfills (`findIndex(), find(), each(), filter() & map()`) with JS natives in cinnamon and delete these polyfills from `util.js`.

These were originally added for performance but are now slower in current cjs. They are no longer used by any xlets in linuxmint/cinnamon-spices-applets, linuxmint/cinnamon-spices-desklets, or linuxmint/cinnamon-spices-extensions. However, they are still used in the cinnamon menu applet so this PR depends on https://github.com/linuxmint/cinnamon/pull/11865 also being merged.

+grouped-window-list applet: Always add alert style to thumbnail when app window needs attention and implement same style removal when app is focused (not previously implemented)

+grouped-window-list applet: used `const` instead of `let` where possible.

### DIFF
--- a/js/misc/util.js
+++ b/js/misc/util.js
@@ -460,102 +460,6 @@ function queryCollection(collection, query, indexOnly = false) {
 }
 
 /**
- * findIndex:
- * @array (array): Array to be iterated.
- * @callback (function): The function to call on every iteration,
- * should return a boolean value.
- *
- * Returns (number): the index of @array, else -1.
- */
-function findIndex(array, callback) {
-    for (let i = 0, len = array.length; i < len; i++) {
-        if (array[i] && callback(array[i], i, array)) {
-            return i;
-        }
-    }
-    return -1;
-}
-
-/**
- * find:
- * @array (array): Array to be iterated.
- * @callback (function): The function to call on every iteration,
- * should return a boolean value.
- *
- * Returns (any): Returns the matched element, else null.
- */
-function find(arr, callback) {
-    for (let i = 0, len = arr.length; i < len; i++) {
-        if (callback(arr[i], i, arr)) {
-            return arr[i];
-        }
-    }
-    return null;
-};
-
-/**
- * each:
- * @array (array|object): Array or object to be iterated.
- * @callback (function): The function to call on every iteration.
- *
- * Iteratee functions may exit iteration early by explicitly returning false.
- */
-function each(obj, callback) {
-    if (Array.isArray(obj)) {
-        for (let i = 0, len = obj.length; i < len; i++) {
-            if (callback(obj[i], i) === false) {
-                return;
-            }
-        }
-    } else {
-        let keys = Object.keys(obj);
-        for (let i = 0, len = keys.length; i < len; i++) {
-            let key = keys[i];
-            callback(obj[key], key);
-        }
-    }
-};
-
-/**
- * filter:
- * @array (array): Array to be iterated.
- * @callback (function): The function to call on every iteration.
- *
- * Returns (array): Returns the new filtered array.
- */
-function filter(arr, callback) {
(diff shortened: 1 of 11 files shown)


js/misc/util.js
anaximeno, line 1: This might break extensions that depend on these utils.

js/misc/util.js
```diff
@@ -459,102 +459,6 @@ function queryCollection(collection, query, indexOnly = false) {
...
-    let len = arr.length;
-    let out = Array(len);
-
-    for (let i = 0; i < len; i++) {
-        out[i] = callback(arr[i], i, arr);
-    }
-
-    return out;
-};
-
```
anaximeno, line 557: This might break extensions that depend on these utils.

brownsr:
Applet fails to launch:

[grouped-window-list@cinnamon.org]: find is not a function
[grouped-window-list@cinnamon.org]: Failed to evaluate 'main' function on applet: grouped-window-list@cinnamon.org/16
Cjs-Message: 08:11:25.405: JS LOG: [LookingGlass/trace] 
<----------------
_connect@/usr/share/cinnamon/js/misc/state.js:250:24
connect@/usr/share/cinnamon/js/misc/state.js:271:25
GroupedWindowListApplet@/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/applet.js:211:20
main@/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/applet.js:1022:12
createApplet@/usr/share/cinnamon/js/ui/appletManager.js:597:25
addAppletToPanels@/usr/share/cinnamon/js/ui/appletManager.js:372:34
finishExtensionLoad@/usr/share/cinnamon/js/ui/appletManager.js:98:14
_init/<@/usr/share/cinnamon/js/ui/extension.js:275:32
---------------->
Cjs-Message: 08:11:25.406: JS LOG: [LookingGlass/error] 
[grouped-window-list@cinnamon.org]: Applet grouped-window-list@cinnamon.org: Could not create applet object.
[grouped-window-list@cinnamon.org]: Error importing applet.js from grouped-window-list@cinnamon.org
Cjs-Message: 08:11:25.406: JS LOG: [LookingGlass/trace]

fredcw:
@brownsr Thanks. Yes, I changed state.js and then forgot to add it to the commit 🤦‍♂️. I'll add it in a new PR

fredcw:
@mtwebster Unrelated to this PR but I was wondering: in GWL applet.js are the following comments:

line 205
```
// key-function pairs of actions that can be triggered from the store's callback queue. This allows the
// applet to avoid passing down the parent class down the constructor chain and creating circular references.

```
line 262
```
// Passing an empty object instead of `this` because its only used by SignalManager to bind the callback, which
// we already do here. Otherwise, it creates more circular references.
```

line 845
```
// TODO: Figure out exactly which properties on this applet constructor the Cinnamon APIs needs for all modes of
// DND, so we can kill the _delegate reference. Long term, a PR to Cinnamon should be opened fixing circular
// object reference structures for the applet and desklet classes.
```

Do you know of any reason why circular references should be avoided or perhaps this isn't necessary now?

mtwebster:
Seeing as the gwl applet is really the only place we worry about any of this (in Cinnamon proper, at least), I doubt this is as much of an issue as was being made here. Some googling also gives me the impression JS garbage collection is robust enough to handle it np.

We end up with circular refs all over the place now using _delegate in popup menus, xlets, etc... so even if it *is* an issue with GC it doesn't seem like a major one. We could always clear _delegates when we're cleaning up these things - that seems like it'd be easier than tying our hands ahead of time. That second snippet just seems like it'd be confusing.

PR #11878, opened 2023-10-02
Add Actions module

rcalixte:


fredcw:
Is there going to be a new nemo-actions spices webpage from which to download nemo-actions? Also, the "manage" tab in the nemo actions settings, will this be to enable/disable nemo-actions, and if so, won't this be a duplicate of the nemo preferences dialog in nemo (plugins->actions)?

rcalixte:
> Is there going to be a new nemo-actions spices webpage from which to download nemo-actions? Also, the "manage" tab in the nemo actions settings, will this be to enable/disable nemo-actions, and if so, won't this be a duplicate of the nemo preferences dialog in nemo (plugins->actions)?

A new page for Spices was my thinking. The Manage tab at the moment does not have the ability to enable or disable, just deleting downloaded Spices. (Previously installed Nemo Actions will be unaffected since they won't appear under the UI unless they follow the same directory structure.) I can maybe attach a mock screenshot.

fredcw:
Are you sure there's a need for this. There's already a few actions that come with Nemo and if anyone comes up with any new useful actions, can't these just be added to the nemo package?

rcalixte:
> Are you sure there's a need for this. There's already a few actions that come with Nemo and if anyone comes up with any new useful actions, can't these just be added to the nemo package?

I think the thought here is similar to other Spices. There are Applets, Desklets, and Themes which ship by default. Looking at linuxmint/discussions#130, linuxmint/discussions#149, and linuxmint/discussions#154, there's a lot of demand for greater customization and easier discoverability. Given the infinite possibilities, I think it makes sense to break it out and welcome contributions. Some of the current Nemo extensions could even be ported as well.

Personally, there are a few things I would love in the context menus for Nemo that there are already applets from other developers for and I feel like there's plenty of demand for it from users as well.

fredcw:
Now I think about it, people may start coming up with ideas for actions if the option is there to easily make them available to others. Could be a good idea. @smurphos has loads of them https://github.com/smurphos/nemo_actions_and_cinnamon_scripts

rcalixte:
Agreed. It would be great to have those ported for ease of use by users. If that's the plan, it might take some work to migrate, but it would ultimately be easier to maintain in the long run.

PR #11880, opened 2023-10-03
gestures: Allow the user to specify which phase of a gesture should trigger its action.

mtwebster:
ref: https://github.com/orgs/linuxmint/discussions/102

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_gestures.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_gestures.py
@@ -16,6 +16,64 @@ NON_GESTURE_KEYS = [
     "pinch-percent-threshold"
 ]
 
+ACTIONS = [
+    # Action, Label, Allow phase selection, extra widget type [entry|slider|none], default custom val
+    ["", _("Disabled"), False, "none", ""],
+    ["WORKSPACE_NEXT", _("Switch to right workspace"), True, "none", ""],
+    ["WORKSPACE_PREVIOUS", _("Switch to left workspace"), True, "none", ""],
+    # ["WORKSPACE_UP", _("Switch to the workspace above"), "none", ""],
+    # ["WORKSPACE_DOWN", _("Switch to the workspace below"), "none", ""],
+    ["TOGGLE_EXPO", _("Show the workspace selector (Expo)"), True, "none", ""],
+    ["TOGGLE_OVERVIEW", _("Show the window selector (Scale)"), True, "none", ""],
+    ["MINIMIZE", _("Minimize window"), True, "none", ""],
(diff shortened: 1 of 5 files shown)


anaximeno:
The 2 and 5 fingers options are being shown in the gesture swipe settings even if the screen does not support it, is this expected now?

Also, I've to confess that there are quite a number of configuration options in this module:

https://github.com/linuxmint/cinnamon/assets/63073056/2af38173-3728-40ce-b95d-157a1d733b56

I wonder if it would be better to bind that configuration to trigger at the start or the end, generally, to the swipe and pinch gestures or at least make that an option you choose for the swipe or pinch based on the number of fingers in general (like for the three fingers swipe you can choose for it to trigger at the start, but for the four fingers swipe, at the end). I probably would not have a reason to tune each specific action but, I am not against how this is implemented right now since someone might benefit from that, but I still believe it's a lot of settings for one section.

But, after all, it can be a bit more fluid by having some gestures set to trigger at the start. So this is a good update on gestures.

mtwebster:
Added desktop zoom support.

I made it so the phase combobox disappears if it's not applicable for its gesture type - if it's 'Disabled' or one of the types that forbid changing it (for now, zoom, volume changes only).

mtwebster:
I'll have a look at the custom command and making the zoom rate configurable (I sort of figured this would be needed anyhow), thanks.

floccus101:
Hi.
I installed Linux Mint a weeks ago and missed the touchpad-pinch-zoom that I used in Windows 10 browsers.
After searching for hours I found this conversation.

In the default gestures app there is "swipe", "pinch" and "settings". In the pinch-tab (under pinch with 2 fingers) is unfortunately no zoom option.

The [commit](https://github.com/linuxmint/cinnamon/pull/11880/commits/56cefee4ef35dcefad97846412b12090b79f48b3) of mtwebster should have added this.

I have followed the [instructions](https://github.com/orgs/linuxmint/discussions/102#discussioncomment-7179210) of mtwebster to enable the debug-mode for gestures.
And I also enabled the desktop-zoom accessibility feature, but the zoom-function did not appear.

Now my questions that came up:
1. Should the zoom function have been there by default (i.e. without any changes)?
2. If no: Is there a manual that explains how to get this function?
3. Do I need to pull something or manually change some files? If yes, please give me instructions.
4. Is it possible that my device thinks this is feature is not compatible? Could I turn this off?

I would be very grateful for answers.

specs:
Linux Mint 21.2 Cinnamon
Cinnamon 5.8.4
Kernel: 5.15.0-91-generic

anaximeno:
@floccus101 this feature will be available in the next version of Mint 21.3 which is almost ending beta testing and might be released in the following weeks.

PR #11884, opened 2023-10-09
Correction to the Brazilian translation of the word 'Layout' on the settings panel GUI

ricardoraposo:
'Leiaute' is not a Portuguese word, it's just the english word literally rewritten with a brazilian accent. The proper way to write it should be as it is in the Portugal's Portuguese version of the translation.


claudiux:
https://linuxmint-translation-guide.readthedocs.io/en/latest/translate.html

PR #11905, opened 2023-10-26
[network@cinnamon.org] - Try to fix #10225

### LINKED ISSUE #10225: [Feature request] Support WireGuard connections in the Network Manager applet (network@cinnamon.org)
```
 * Cinnamon version (cinnamon --version): Cinnamon 5.0.4
 * Distribution - Fedora 34 x86_64
 * Graphics hardware *and* driver used: Intel Corporation HD Graphics 620 (Dell XPS 13 9360), but I'm not sure what Xorg driver
  ```

**Issue**
The Network Manager applet doesn't support WireGuard connections, although current versions of NetworkManager do. Such connections aren't shown in the list of networks that you can select (under any category) and there's no sign if one is active or inactive.

NetworkManager reports WireGuard connections as a new type, "wireguard", instead of as type "vpn". Based on looking at the code in files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js, a simple solution might be to add a case for WireGuard connections to the this._ctypes variable that maps NetworkManager types to categories, as a 'vpn' connection. Probably most people using WireGuard connections (especially through the applet) are using them as a form of VPN, instead of more complex usage cases that might want them to be treated separately.

**Steps to reproduce**
Configure a WireGuard connection in NetworkManager, then attempt to see it in Cinnamon.

**Expected behaviour**
You can activate and deactivate a WireGuard connection.

### PR DESCRIPTION
**Pull request by claudiux**
Maybe fixes #10225 (at least partially).

### DIFF
--- a/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
@@ -22,7 +22,8 @@ const NMConnectionCategory = {
     WIRED: 'wired',
     WIRELESS: 'wireless',
     WWAN: 'wwan',
-    VPN: 'vpn'
+    VPN: 'vpn',
+    WIREGUARD: 'wireguard'
 };
 
 const NMAccessPointSecurity = {
@@ -260,7 +261,7 @@ NMWirelessSectionTitleMenuItem.prototype = {
     },
 
     updateForDevice: function(device) {
-		this._device = device;
+        this._device = device;
         // we show the switch
         // - if there not just one device
         // - if the switch is off
@@ -708,6 +709,7 @@ NMDevice.prototype = {
                     case NM.DeviceType.WIFI: return _("Wifi");
                     case NM.DeviceType.MODEM: return _("Modem");
                     case NM.DeviceType.BT: return _("Bluetooth");
+                    case NM.DeviceType.WIREGUARD: return _("WireGuard");
                     default: return "";
                 }
             }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by timur-g** [comment]
Tested .Seems to work, thanks. Some notes, though:
1. I cannot switch between tunnels with just clicking the other, need to turn of WG and turn on other tunnel (not grave)
2. I still cannot import WF conf from file, that can be imported via nmcli (says no IP address on line 3)

**Comment by claudiux** [comment]
I see. Thanx for the feedback.

**Comment by timur-g** [comment]
Plese see my edited 3., may be the bug.
Not related, but if you are willing to wok more on Network Manager, it has another annoying issue: does not show Hotspot to turn on/off with the same presets (so only command is possible).

**Comment by claudiux** [comment]
I've never seen "WireGuard Connections" ON at startup, except when a WireGuard connection is already activated.

**Comment by claudiux** [comment]
I have some problems (and I need help!):

  * even when a Wireguard connection is active, the only active connection detected is "wired". The icon therefore remains that of the wired connection;
  * clicking on another Wireguard connection doesn't deactivate the old one.

@timur-g Can you give me a .conf file (containing fake data) to understand the origin of your problem?

**Comment by timur-g** [comment]
Here it is:
```
[Interface]
PrivateKey = 12345=
Address = 1.2.3.4/32
MTU = 1280


[Peer]
PresharedKey = 67890=
PublicKey = 11111=
AllowedIPs = 11.22.33.44,55.66.77.88
Endpoint = end.point.com:51820
PersistentKeepalive = 25
```

claudiux:
@timur-g To import this .conf file, try to remove the `/32' part of Address.

timur-g:
Even if removed, import does not work, says Line 10 needs valid IP. It is where there are 2 IP addresses IP1,IP2.

claudiux:
@timur-g Try replacing comma by a space.

timur-g:
If comma is replaced by space, than GIU Import works, but puts that connection under VPN, which is expected, as it says "Import VPN". So it wouldbe good to have "Import Wireguard". Just saying. as import via nmcli works and it is one-time anyway. 
So this is an improvement used daily - only issue for me is that auto start on boot.

leigh123linux:
@claudiux  This could be ported to c-c-c 


https://github.com/GNOME/gnome-control-center/commit/475faaf26bccfd2bed0b3b5a79e314249697e9a3

claudiux:
The icon problem seems solved now!

claudiux:
@timur-g Thanks for the feedback. Good suggestion, I'll try to take it into account.

timur-g:
Hi @claudiux. Happy with this, just wondering will you work more or someone should check and merge PRs?
IDK how it goes here, should someone with permission be mentioned to look? 
It would be pitty that this just stays like this.

PR #11907, opened 2023-10-29
fix: incorrect icon for app running in a separate pid namespace

gaory1:
To reproduce:
With grouped-window-list, keep 'Group windows by application' off.
Run Chromium in a separate pid namespace. Chromium gets pid 2. (In my case firejail and 'ip netns' are used.)
Run another app in another pid namespace. It gets pid 2 too.
Sometimes the icon of this app in the grouped-window-list is replaced with the Chromium's icon, quite often after closing and opening a new window.

After examining the window properties of the two apps above, I find their _NET_WM_PID are both 2, and this makes get_app_from_window_pid() return the same CinnamonApp.

### DIFF
--- a/src/cinnamon-window-tracker.c
+++ b/src/cinnamon-window-tracker.c
@@ -567,9 +567,11 @@ get_app_for_window (CinnamonWindowTracker    *tracker,
   if (result != NULL)
     return result;
 
+#if 0
   result = get_app_from_window_pid (tracker, window);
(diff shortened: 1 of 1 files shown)


mtwebster:
I can't reproduce the issue described, and I'm concerned disabling this will affect other programs.

Please provide more info - what distribution, cinnamon and webapp-manager version mainly.

PR #11910, opened 2023-10-30
xlet-settings.py: Ensure settings window expands to fit toolbar

fredcw:
Fix bug where toolbar disappears if the window is too narrow.

As the settings window Toolbar only contains one ToolItem, this ToolItem can disappear if the window is too narrow and Toolbar:show-arrow property is set to true. Set Toolbar:show-arrow to false to ensure that the window will expand to accommodate the full width of the Toolbar.

ref. https://github.com/linuxmint/cinnamon-spices-applets/issues/5065

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/xlet-settings.py
@@ -194,6 +194,7 @@ class MainWindow(object):
 
         toolbar = Gtk.Toolbar()
         toolbar.get_style_context().add_class("primary-toolbar")
+        toolbar.set_show_arrow(False)
         main_box.add(toolbar)
 
         toolitem = Gtk.ToolItem()


claudiux:
This seems to work fine. What happens when, due to translation, the width of the toolbar becomes greater than that of the desktop?

fredcw:
>What happens when, due to translation, the width of the toolbar becomes greater than that of the desktop?

The window just becomes wider than the desktop so that you'll have to drag the window to see all of it. Better than the toolbar disappearing though I guess.

>It would be cool if you could create several toolbars when the number of buttons...

Yes, you'd think that the Gtk.StackSwitcher widget would do that automatically but it doesn't. Using two stackswitchers, one above the other, I think would look a bit odd though as they'd look separate from each other. An alternative is to detect if the stackswitcher is too wide and if so, make it a vertical switcher at the side of the window. In any case, it's easier to just make sure that applet configs don't use too many tabs.

claudiux:
You're right.

PR #11913, opened 2023-11-02
Adds a shift-middle click action to the sound applet.

sphh:
When discussing the MuteToggler@jebeaudet.com spice in https://github.com/linuxmint/cinnamon-spices-applets/pull/5090, we noticed that most of its functionality is already there. I noticed that it might be a good idea to be able to start/stop the player but also mute/unmute the microphone directly with the mouse without diving into the right-click menu. So I added a option for shift-middle-click very similar to the middle-click option.


claudiux:
I've included these changes in the sound150@claudiux applet and that works great! Thank you @sphh.

PR #11916, opened 2023-11-04
Added menu details option

kacperpaczos:
Work for me, it's usefull sometimes to checkout run command.

Please consider merging this Pull Request

image

image


anaximeno:
Just a minor suggestion here, "Properties" (or something like that) seems a better name than "Details" for that option since you can change/update those values in the launcher, also, it looks more in conformity with the title of the window that is open "Launcher Properties". You can change the symbol to something like the config symbolic icon (or something like that) if you want to make this change.

kacperpaczos:
I will do it soon

axel358:
I think something like ✏️ Edit would make more sense

kacperpaczos:
Now:
Icon: document-properties-symbolic
Option's name: Properties
image

PR #11933, opened 2023-11-19
Improve Timestamp for Notifications

cocker-cc:
1. display the whole Timestamp-String including the Date, as the pure Time of Day is useless when looking at a Notification from several Days ago

2. use the User's Locale instead of a Flag "use_24h", moreover as this Flag is not Part of the App's Configuration-Settings


claudiux:
`use_24h` is obtained from the "Date and time" settings in "Cinnamon Settings".

claudiux:
It seems that `orig_time.toLocaleFormat('%r')` returns a null string ???

mtwebster:
I like adding the original date but I want to maintain control of 24h mode in Date/Time settings, *not* by locale. It's a very common use case to want time format to be different than the local norm.

cocker-cc:
Feel free to change the PR.

PR #11951, opened 2023-11-28
Add integrated color picker

zehkira:
I don't know what I'm doing; see https://github.com/linuxmint/xdg-desktop-portal-xapp/pull/12


zehkira:
Build errror

mtwebster:
Ok, no worries. I'll take care of the Cinnamon work then we'll finish up the portal implementation. We were discussing things some more, and we may keep your binary picker as a fallback as well if we continue to support Xfce and MATE there.

I should be able to get this done fairly quickly.

Issue #11964, opened 2023-12-03
Actions update not working properly


anaximeno:
### Distribution

Mint 21.3 Alpha (romeo + newer version source)

### Package version

Cinnamon 6.0 unstable

### Frequency

Always

### Bug description

The spices update procedure is not working properly. After an update using mintupdate the spice would be uninstalled from the system instead of updating.

After some research, I found that the unzipped files were not being copied from the correct folder, instead of copying files from the dir `/tmp/<tmp-id>/` (that contains the subdir **uuid** and the action file **uuid.nemo_action**) it was only copying the files inside `/tmp/<tmp-id>/uuid/` to the install folder so in the end we would only have loose files originally from the directory `uuid/files/` (according to the actions structure) inside the `~/.local/share/nemo/actions/` folder. 

### Steps to reproduce

1. Install an action
2. Modify the **version** and **last-edited** attributes from the action metadata file to lower values so the mintupdate would show it for update
3. Update the action
4. Try to use the action under nemo
5. Go to the action module and you won't see it there as an installed action
6. List the files in the  `~/.local/share/nemo/actions/` folder and you'll find some loose files there

### Expected behavior

After an update, the action should still be installed on the system, and the files should be properly set in the  `~/.local/share/nemo/actions/` folder.

PR #11965, opened 2023-12-03
actions: fix update procedure

anaximeno:
Fixes https://github.com/linuxmint/cinnamon/issues/11964


anaximeno:
@clefebvre I suggest this merge before ending QA if it won't delay things.

PR #11976, opened 2023-12-09
Update README.Debian (fixed typo on line 26)

IlgazC:
<26> corage-->courage


Fantu:
Thanks for contributing, I suppose there are many others typo.
Add a codespell workflow to detect all them like what I did in 2 other components I think may be useful, for example see here: https://github.com/linuxmint/cinnamon-desktop/pull/237
I don't have time to do it for every cinnamon repositories and fix all the typos recently, so some help could be useful

PR #11977, opened 2023-12-10
[WIP] add Cinnamon Menu Editor to Menu by default

TranzistorP:
Currently, the menu editor by default only opens with 
'Right-click on the menu launch button - Edit menu.'

To figure out how to edit the Menu, I first spent about 20 minutes searching for this function in the Menu and System Settings. Then I googled it, but the search in Google in my language only suggested using third-party programs, and only the search in English gave a result. 
I think it would be useful to add the menu editor by default to the Menu and possibly to the System Settings.
image

This is my first pull request and I would appreciate help and feedback.
I have a few questions:
What do I need to do to properly localize this item in the Menu?
Do I need to add it to generate_additional_files.py or somewhere else, or is the translation to .desktop done manually?
Do I need to do the translation myself using Google Translate or are there separate people doing this?

If you agree, then I would also like to add this to the System Settings, I would be grateful if you could tell me where to start.

Thank you for your attention and sorry for my English.

### DIFF
--- a/files/usr/share/applications/cinnamon-menu-editor.desktop
+++ b/files/usr/share/applications/cinnamon-menu-editor.desktop
@@ -1,11 +1,10 @@
 [Desktop Entry]
-Name=Cinnamon Menu Editor
+Name=Menu Editor
 Comment=Edit the entries of the Cinnamon menu applet
 Exec=cinnamon-menu-editor
 Icon=alacarte
 OnlyShowIn=X-Cinnamon;
 Terminal=false
 Type=Application
-NoDisplay=true
-Categories=GNOME;GTK;Settings;DesktopSettings;
+Categories=Settings;
 StartupNotify=false


fredcw:
What's your use case for needing to edit the menu? I'm wondering if there may be some menu issues that might be better fixed elsewhere?

TranzistorP:
I was creating a shortcut for an application. 
If I could immediately specify the menu category, it would solve my issues.
image

anaximeno:
@fredcw do you know if menu configurations work for the other (third parties) menus as well? (Like Cinnamenu, and CinnVIIStarkMenu).

TranzistorP:
As I understand it, this is a general configuration because it is regulated directly in the .desktop files.
That is, this shortcut will be visible in any menu in the graphical shell with the name X-Cinnamon. 

I checked, everything works in Cinnamenu.

anaximeno:
Oh, ok.

fredcw:
@anaximeno Yes, menu categories are in `/usr/share/desktop-directories` and apps are in `/usr/share/applications` and user defined versions of these, which override the systems ones are in `~/.local/share/desktop-directories` and ~/.local/share/applications`

These files are read by cinnamon and all the menu applets query cinnamon appsystem for the menu structure.

Issue #11956, opened 2023-11-30
Cinnamon 6.0.0 not build on Debian SID


misterhsp:
### Distribution

Debian SID

### Package version

Cinnamon 6.0.0

### Frequency

Always

### Bug description

   dh_shlibdeps -a
dpkg-shlibdeps: warning: can't extract name and version from library name 'libst.so'
dpkg-shlibdeps: warning: can't extract name and version from library name 'libst.so'
dpkg-shlibdeps: warning: can't extract name and version from library name 'libcinnamon.so'
dpkg-shlibdeps: warning: can't extract name and version from library name 'libcinnamon.so'
dpkg-shlibdeps: warning: package could avoid a useless dependency if debian/cinnamon/usr/lib/x86_64-linux-gnu/cinnamon/libst.so was not linked against libGLESv2.so.2 (it uses none of the library's symbols)
   dh_girepository
dh_girepository: error: Could not find Clutter-0.gir dependency
make: *** [debian/rules:25: binary] Error 255
dpkg-buildpackage: error: debian/rules binary subprocess returned exit status 2

### Steps to reproduce

' '

### Expected behavior

' '

### Additional information

' '

Fantu:
This is related to gobject-introspection changes in debian testing/unstable and next ubuntu

to fix and also avoid additional *-dev deps in cinnamon (using next debian/ubuntu) needs to add in d/rules:
```
override_dh_girepository:
	dh_girepository -XCinnamon-0.1.gir -XSt-1.0.gir /usr/lib/$(DEB_HOST_MULTIARCH)/muffin
```

Fantu:
I think can be closed as solved with https://github.com/linuxmint/cinnamon/pull/11999

PR #11999, opened 2023-12-30
Move gir files to private path

Fantu:
After recent gobject-introspection changes cinnamon fails to build on next debian and ubuntu, as reported in https://github.com/linuxmint/cinnamon/issues/11956
A workaround to fix build and don't add additional "unwanted" -dev dependencies to cinnamon-common is:
```
override_dh_girepository:
	dh_girepository -XCinnamon-0.1.gir -XSt-1.0.gir /usr/lib/$(DEB_HOST_MULTIARCH)/muffin
```

But as reported by @smcv gir files should be private, same as library and typelib. It also to avoid possible confusion with St of gnome-shell (even if it is already in private path)

As there aren't other software using it but is rarely and manually used by only few developers (FWIK) put in new cinnamon-dev package (with all needed deps) should be not needed as developers that want to use them can manually add needed dependencies. Or I'm wrong? I suppose also that you don't want to add cinnamon-dev unless absolutely necessary, or am I wrong?
I'll do the same in debian unstable soon for fix fails to build, In debian packaging I'll also move the gir files from cinnamon to cinnamon-common to decrease differences/conflicts with upstream packaging (was done for public gir policy).


leigh123linux:
I delete these useless files as I don't see any reason to include them.

mtwebster:
It's useful for debugging but not much else.

Thanks, sorry for the time wasted over this..

Fantu:
@mtwebster thanks, I have a question about other difference/conflict I spotted with debian package: config.js in cinnamon instead of cinnamon-common.
It was done 10 years ago for values changing on arches using freebsd kernels, but it seems no longer necessary (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=733461). There was also another file with the architecture-specific path to the libraries, but it was later removed. If you give me confirmation that no more values will be added that could change in the builds between the different architectures like those that were there or like those in the other file that had been reported by arm users instead (if I remember correctly) I'll move it in debian to be the same of upstream.

mtwebster:
I can't predict the future, but there are other ways to detect hardware/system details at runtime - so I think it'd be ok to go back to cinnamon-common.

It looks like cinnamon-screensaver needs fixed as well - libcscreensaver should be private, not just the gir file (we don't need that installed at all).

Same with libcinnamon-control-center. This one could be a static library. I think back when I first wanted to fix it there was a bug in meson that prevented it. That may not be an issue any longer.

Fantu:
@mtwebster thanks for all
about cinnamon-screensaver I saw, and I had to do a package split to avoid open of RC bug for policy issue some time ago.
Recently I saw the problem of add of -dev deps caused by public gir present in a "runtime" package.
About it the problem of actual "mix" of public/private and @smcv opened a bug in debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1059341
I mentioned it in a message to @clefebvre and I preferred to know what you want to do to avoid doing something perhaps different in the debian packaging.
About cinnamon-control-center it don't have/had policy issue that can cause RC bugs, I don't checked from difference with upstream package. It don't have gir, so I have not checked it good in latest days as not affected by gobject-introspection changes.

mtwebster:
The .gir for cinnamon-screensaver doesn't need to be installed anywhere. I'm not sure why we started including it, git history is unclear.

PR #12001, opened 2023-12-31
calendar: allow to use line breaks on a custom date format

anaximeno:
This will make it possible to format the date as shown in the image below:

Captura de ecrã de 2023-12-31 19-10-57


rcalixte:
You can already do this with `%n`. That is the proper strftime format string for new lines. Check out the man page for `date` for supported options or [For a Good Strftime](https://www.foragoodstrftime.com/).

anaximeno:
Oh, I didn't know, it's in the reference section. Thanks!

PR #12012, opened 2024-01-10
Fix: messageTray table layout

### LINKED ISSUE #11961: Notification Header switches to RTL layout if title is RTL
### Distribution

Mint 21.1

### Package version

5.8.4

### Graphics hardware in use

_No response_

### Frequency

Always

### Bug description

I just listened to Spotify, installed via their official PPA: https://www.spotify.com/us/download/linux/
When the track switches I receive a notification with the now playing song.

My system is set to LTR layout. So usually the notifications look like this:
image

Now a song with a RTL title started playing and the notification header also appeared with an RTL layout.
image

### Steps to reproduce

Create a notification with an RTL title.
`notify-send 'این نیز بگذرد'`

If you want to verify a possible fix by using Spotify to test the behavior:
- Song with the RTL title: https://open.spotify.com/track/0QIfD7RzDtCawqruhGPUuu?si=9dea635d7edb4fdd
- Both songs are from the following album: https://open.spotify.com/album/3eTir1qeGrgiaFYFRn18vQ?si=mNjyXX-9S3a-an-zPDTgxw

(Hint: You can just copy-paste the URLs into the Spotify search bar and click Enter)

### Expected behavior

The notification header should stay in it's current layout, only the title text should use it's specified direction.

### Additional information

_No response_

sorenbryder:
Fixes #11961 

This is not really a bug. The way it currently works is by design. The idea is that the text direction of the title will determine the layout of the table with buttons, icons and title.

@clefebvre who originally committed the code should probably verify if it is time to remove it.


RafaelKr:
@sorenbryder @mtwebster Is this really the correct fix? I have no expecience with arabic written language but shouldn't the title be right aligned but the icons stay on their sides?

PR #12049, opened 2024-01-24
menu@cinnamon.org applet - Fixes #11760

### LINKED ISSUE #11760: Menu applet graphical vertical tearing on scroll
### Distribution

Mint 21.2

### Package version

Cinnamon 5.8.4

### Graphics hardware in use

Intel Haswell-ULT Integrated Graphics

### Frequency

Always

### Bug description

- Menu applet graphically tears vertically and stays long periods in teared state after scrolling
- Left side column has scroll even though it doesn't need it, also reacts to scrolling the menu's group listing

Occurred right after Mint 21.2 upgrade.

Disabled 3rd-party applets, rebooted cinnamon, rebooted system, still occurs.

https://github.com/linuxmint/cinnamon/assets/32353070/061cef61-3657-4e51-ab13-4cdc41364a7b

### Steps to reproduce

Open Menu applet, scroll menu groups

### Expected behavior

Menu graphics not tearing when scrolling, left side column scrollbar not visible

### Additional information

System: Lenovo Thinkpad X240
Display size: 1920x1080
Scaling: 1.0x

Theming (changing doesn't affect behavior, but just in case):

Applications: Mint-Y-Dark
Icons: Mint-Y-Yaru
Desktop: Mint-Y-Dark

claudiux:
Fixes #11760

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -2727,6 +2727,7 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
         this.categoriesBox = new St.BoxLayout({ style_class: 'menu-categories-box',
                                                 vertical: true,
                                                 accessible_role: Atk.Role.LIST });
+        this.categoriesBox.style = 'padding: 10px 30px 10px 0px;';
         this.categoriesScrollBox = new St.ScrollView({ style_class: 'vfade menu-applications-scrollbox' });
         this.categoriesScrollBox.add_actor(this.categoriesBox);
         this.categoriesScrollBox.set_policy(St.PolicyType.NEVER, St.PolicyType.AUTOMATIC);
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -2723,13 +2723,19 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
 
         this.categoriesApplicationsBox = new CategoriesApplicationsBox();
         this.right_box.add(this.categoriesApplicationsBox.actor, { expand: true, span: 1 });
+        
+        this.containerCategoriesBox = new St.BoxLayout({ style: 'padding: 0px;', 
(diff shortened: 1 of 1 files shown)


fredcw:
@claudiux This only seems to work on most themes because it reduces the left padding of the categories box. But of course, this also breaks all the themes because they've set the left padding of the categories box to how they think it looks good. And the problem would still exist on themes such as Sweet, Orchis, Qogir.

This is what I think seems to happening: If for instance the box is 100px wide and includes 10px of padding on the left and the right then the bit in the middle, where the text is, is 80px wide. But when you scroll with the mouse wheel or by dragging on the scroll bar, instead of scrolling the 80px in the middle where the text is, it's scrolling 80px from the left, including the padding. In other words, it's scrolling the 0 - 80 pixels instead of the 10 - 90 pixels, which means that some of the text on the right hand side gets left behind (depending on the size of the padding)

claudiux:
@fredcw 
Thanx for this review.

Idea: Category titles in an inner-box which is itself in a outer-box with l/r padding?

fredcw:
@claudiux Tbh, I think this is enough of a papercut bug to be not worth adding the extra bug workaround code to the menu applet. This would make the menu applet more complicated and harder to maintain in the future and it still doesn't fix the underlying bug. I think most people increase the height of the menu so that they don't have to scroll to see all the categories, so I doubt it affects a lot of people. Also, I think is better to try to fix the original bug rather than adding lots of workaround code in the applets. Unfortunately, I'm completely unfamiliar with the St library code.

claudiux:
@fredcw Can you test now, please?

fredcw:
menu-applications-box is not used in Mint-Y or system theme.

claudiux:
Thank you. Can't find the Sweet theme.

fredcw:
@claudiux This fixes the bug but it has the same problem as before, it breaks the themes. The padding defined in preexisting themes has to be preserved. The only solution I can think of is to put categoriesBox inside another St.BoxLayout (with 0 padding) and put that in categoriesScrollBox. I tried this but for some reason it causes the scrollbar to disappear.

claudiux:
@fredcw Thanks for this review. I'll try that later.

claudiux:
@fredcw 
You're quite right. I tested what you suggested.

PR #11744, opened 2023-07-06
inhibit@cinnamon.org: Add unique applet icons for notification toggles

rcalixte:
Closes #11228 

@BrunoNZ This completes the functionality originally proposed in #11228. I hope you're okay with the finished product.

PR #12061, opened 2024-01-31
inhibit applet - Allow power management inhibition on startup

claudiux:
Adds an option to the settings to enable inhibition of power management on start-up.

### DIFF
--- a/files/usr/share/cinnamon/applets/inhibit@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/inhibit@cinnamon.org/applet.js
@@ -350,6 +350,8 @@ class CinnamonInhibitApplet extends Applet.IconApplet {
         super(orientation, panel_height, instanceId);
 
         this.metadata = metadata;
+        
+        this.intervallId = -1;
 
         this.menuManager = new PopupMenu.PopupMenuManager(this);
         this.menu = new Applet.AppletPopupMenu(this, orientation);
@@ -419,11 +421,14 @@ class CinnamonInhibitApplet extends Applet.IconApplet {
     }
 
     on_applet_added_to_panel() {
-        if (this.inhibitPowerAtStartup) {
-            let id = setTimeout ( () => {
-                this.toggle_inhibit_power();
-                clearTimeout(id);
-            }, 1000); // Let 1000 ms to have this.sessionProxy not null.
+        if (this.inhibitPowerAtStartup && this.inhibitSwitch._switch.state) {
+            this.intervallId = setInterval ( () => {
+                if (this.inhibitSwitch.sessionProxy) {
+                    this.toggle_inhibit_power();
+                    clearTimeout(this.intervallId);
+                    this.intervallId = -1;
+                }
+            }, 300);
         }
     }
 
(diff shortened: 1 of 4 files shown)


files/usr/share/cinnamon/applets/inhibit@cinnamon.org/applet.js
```diff
@@ -416,6 +417,15 @@ class CinnamonInhibitApplet extends Applet.IconApplet {
...
     }
+    
+    on_applet_added_to_panel() {
+		if (this.inhibitPowerAtStartup) {
+			let id = setTimeout ( () => {
+				this.toggle_inhibit_power();
+				clearTimeout(id);
+			}, 1000); // Let 1000 ms to have this.sessionProxy not null.
+		}
+	}
```
rcalixte: I've noticed on this and other pull requests that you're using tabs instead of spaces. The coding guidelines highlight [here](https://linuxmint-developer-guide.readthedocs.io/en/latest/guidelines.html#consistency) and [here](https://linuxmint-developer-guide.readthedocs.io/en/latest/guidelines.html#indentation) about why not to do so.

files/usr/share/cinnamon/applets/inhibit@cinnamon.org/settings-schema.json
```diff
@@ -28,5 +28,14 @@
...
-    }
+    },
+    "section3": {
+        "type": "section",
+        "description": "Startup"
+    },
+    "inhibitPowerAtStartup": {
+		"type": "switch",
+		"default": false,
+		"description": "Inhibit power management at startup"
```
rcalixte: Same here except this is mixed spacing of what has been added.

rcalixte (changes requested):

files/usr/share/cinnamon/applets/inhibit@cinnamon.org/applet.js
claudiux: Fixed.

files/usr/share/cinnamon/applets/inhibit@cinnamon.org/settings-schema.json
claudiux: Also fixed.

rcalixte:
Coding-wise, it works. One issue I see is that this has a conflict with #11744. I'll leave it up to others if this should be merged or not though.

claudiux:
@rcalixte This applet already lets you disable notifications.

claudiux:
@rcalixte I'm the one who didn't understand. In this PR, I'll take your changes into account.

claudiux:
@rcalixte 
Your changes have been taken into account. That works well, good job!
I think there should be a specific icon when both are disabled.
Would you like me to do the same for notifications (disable at startup).

rcalixte:
> Your changes have been taken into account. That works well, good job!

I don't think that's what should have been done. That PR is still waiting on review and final feedback. You should be avoiding a conflict with that not creating one. Your proposed change and the changes in the already open PR should be independent and as conflict-free as possible.



> I think there should be a specific icon when both are disabled.

There already is. I've been running this locally for months.


> Would you like me to do the same for notifications (ability to disable at startup)?

There is no need for this. Notifications are disabled via gsettings which are not reset on Cinnamon restarts or Cinnamon reboots.

rcalixte (changes requested):
Please remove the other PR's proposed changes and try to avoid a conflict with that branch if possible.

rcalixte:
@claudiux You have the ability to communicate with me directly. Instead of operating on your own in a state of confusion, I implore you to make use of that so that these situations can be avoided in the future. Two minutes of conversation would have prevented hours of wasted time. If this isn't agreeable, please let me know and we can find another solution that works for you. This current state of affairs is not good.

PR #12074, opened 2024-02-07
Improve handling of invalid panel definitions

mtwebster:
Make sure only valid, and non-overlapping panel definitions are allowed to exist at runtime, and remove them if they're detected either by cinnamon-settings or Cinnamon during startup.


mtwebster:
already merged..

PR #12093, opened 2024-02-24
Allow user to set workspace OSD timeout duration

cmck-xyz:
Allow the user to set the duration of the workspace OSD timeout shown when switching workspaces.
Replaces a hardcoded const WORKSPACE_OSD_TIMEOUT.


fredcw:
Is this really necessary? I think that adding config options can have disadvantages and the pros and cons have to be considered carefully. The main problem is that the more options there are, the more complicated or complex the settings will appear to be and while some may welcome this, many others may be put off or even discouraged from even delving into the settings because of a perception of it being too complicated or time-consuming. And for each additional option, it makes it just that little bit harder to find the option you're looking for. We also have to remember that one of the objectives of Cinnamon is to be easy to use for non-techies and a perception that Mint is complicated or difficult to use could put some people off using Mint and/or Linux altogether.

I'm not saying that this option shouldn't be added, only that the cost/benefit needs to be considered as it does with all changes to the GUI.

JosephMcc:
I'm with @fredcw on this one. It's not even all about the GUI. Everything you add creates more complexity in the code as well. Making further changes and troubleshooting even more difficult. While this seems small, and it is, if you start adding a bunch of things like this it piles up quickly.

With this sort of thing, the better solution is just to adjust the hard coded timeout if it's needed. I've never seen a single issue or complaint personally about it being either too slow or too fast though.

cmck-xyz:
Thanks for the feedback.

@fredcw I see your point regarding adding potentially unnecessary / confusing options. However, this shows only in the workspace options, and would only be seen by users already looking to configure usage of their workspaces, as well as only showing when the workspace OSD is enabled.

@JosephMcc I've modified the changes to be more in line with the way other settings are handled.
Regarding just changing the hard-coded timeout: While this isn't directly a problem for most users, or a bug for that matter, it *is* a small quality-of-life change which can improve users' experience.

Regardless, it might be an idea to add a general OSD (brightness, volume, etc.) timeout setting (instead or as well), as they do stay for quite a time too, and some users might prefer a shorter, or longer, duration.

anaximeno:
Hi @cmck-xyz,

I agree with @fredcw @JosephMcc that it might not be a good idea to expose minor settings like this to the users because it can create precedence on this matter that could easily scale quickly through the system. Such kinds of settings in my view should be optimized for the best value for "everyone" instead of creating a configuration setting for them because few people want to change its value. I see the possible escalation of such type of settings as a problem because it will inherently increase the system's perceived complexity, which is one of the greatest disadvantages of KDE especially with new users, in my view.

The speed of some animations is something of concern to me as well (for example I would like the speed of the overview animation to be configurable), but I believe that maybe a "good" solution would be to update the effects module making the section where the user defines the "animation speed" a more generalized setting that could be used to increase or decrease the speed of some animations of some parts of the system (like overview and the workspace OCD if enable).

mtwebster:
I'm not really crazy about this one either - after trying it out, I'd rather just shorten the animation times to a better value, as it is a bit on the slow side.

Using 200ms for both animations made it quite a bit snappier, and it's still readable enough for most users I think. I'm not sure what values you were targeting for your personal case - faster still?

PR #12100, opened 2024-03-01
Add theme config gui for configurable themes

fredcw:
This is a way to enable gtk or cinnamon themes to have configurable options.

How it works:
A theme can make itself configurable by including a shell script or program that will alter the theme in place according to options passed to it as arguments.

A file called `config_options.json` will be included in the theme that lists the options available.

When a gtk or cinnamon theme is selected as the current theme in settings>themes, the themes module will check to see if the theme has a `config_options.json` file if so, show a "theme options" button.

Clicking the button causes cinnamon to read the `config_options.json` file and present the available options in a window. User changes options and clicks "Apply" button. Cinnamon calls the configure script with the options as arguments thereby altering the theme. Cinnamon then reloads the theme.

The configure script and `config_options.json` file should be in a directory called "config" in the theme directory and the script should have no dependencies other than what would very likely be found on any gtk based DE.

Test using "Fluent" theme in cinnamon-spices-themes repo.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/theme_config_gui.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/theme_config_gui.py
@@ -1,0 +1,270 @@
+#!/usr/bin/python3
+
+import gi
+gi.require_version('Gtk', '3.0')
+from gi.repository import Gtk
+from gi.repository import Gio, GLib
+import gettext
+import json
+import sys
+import os
+import subprocess
+import shutil
+import string
+import random
+import time
+
+gettext.install("cinnamon", "/usr/share/locale")
+
+desktop_environment = "Cinnamon"
+rename_theme_on_update = False
+
+class OptionsWindow(Gtk.Window):
+    def __init__(self, theme_path, desktop_environment, rename_theme_on_update):
+        super().__init__(title="")
+
+        self.desktop_environment = desktop_environment
+        self.rename_theme_on_update = rename_theme_on_update
+
+        self.current_theme_dir = theme_path
+        self.current_theme_name = os.path.basename(self.current_theme_dir)
+        
+        self.set_default_size(500, 100)
+        
+        box = Gtk.Box.new(Gtk.Orientation.VERTICAL, 5)
+        box.set_margin_top(15)
+        box.set_margin_bottom(15)
+        box.set_margin_start(25)
+        box.set_margin_end(25)
+        self.add(box)
+        
+        self.load_config_file()
+        self.set_title(self.options["theme_name"] + " theme options")
+        self.set_icon_name("applications-graphics")
+
+        self.widgets = {}
+        
+        for option in self.options["options"]:
+            if option["desktop"] == "all" or option["desktop"] == self.desktop_environment:
+                if option["type"] == "combo":
+                    combobox = Gtk.Box.new(Gtk.Orientation.HORIZONTAL, 10)
+                    combobox_label = Gtk.Label.new(option["label"])
+                    combobox_label.set_halign(Gtk.Align.START)
+                    combobox.pack_start(combobox_label, True, True, 0)
+                    combo_widget = Gtk.ComboBoxText()
+                    for label in option["labels"]:
+                        combo_widget.append_text(label)
+                    combo_widget.set_active(option["value"])
+                    combo_widget.set_halign(Gtk.Align.END)
+                    combo_widget.connect("changed", self.on_combo_changed)
+                    combobox.pack_start(combo_widget, True, True, 0)
+                    box.pack_start(combobox, False, True, 0)
+                    self.widgets[option["name"]] = combo_widget
+                elif option["type"] == "switch":
+                    switchbox = Gtk.Box.new(Gtk.Orientation.HORIZONTAL, 10)
+                    switchbox_label = Gtk.Label.new(option["label"])
+                    switchbox_label.set_halign(Gtk.Align.START)
+                    switchbox.pack_start(switchbox_label, True, True, 0)
+                    switch_widget = Gtk.Switch()
+                    switch_widget.set_active(option["value"])
+                    switch_widget.set_halign(Gtk.Align.END)
+                    switch_widget.set_valign(Gtk.Align.CENTER)
+                    switch_widget.connect("state-set", self.on_switch_changed)
+                    switchbox.pack_start(switch_widget, True, True, 0)
+                    box.pack_start(switchbox, False, True, 0)
+                    self.widgets[option["name"]] = switch_widget
+        #---------------------------
+        if gsettings_get("org.cinnamon.desktop.interface",
+                                "gtk-theme") == self.current_theme_name:
+            restart_label = Gtk.Label.new(_("Restart Cinnamon for titlebar changes to take effect. " \
+            "(Right click on panel>Troubleshoot>Restart Cinnamon.). Some apps may need to be restarted."))
+            restart_label.set_line_wrap(True)
+            box.pack_start(restart_label, False, True, 0)
+        #---------------------------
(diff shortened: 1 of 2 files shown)


Secret-chest:
Beautiful

fredcw:
@Secret-chest Not my theme.

fredcw:
I'll mark this [WIP] for now as I think it's more of a proposal than a PR. Thoughts/suggestions welcome.

I would prefer the "theme options" button to appear under selected theme button but I don't know how to do that though I guess having the button at the bottom of the page isn't too bad.

The idea of making this spec DE agnostic and including an gui config program in the theme is that it might encourage theme writers to distribute their theme like this?

Here's a possible spec for the `options_config.json` file:

All fields required unless stated otherwise.
```
{
    "major_version": <number> , Earliest version of this spec this file is compatible with.
    "minor_version": <number>,
    "script_name": <string>,
    "theme_name":  <string>,
    "adwaita_link_to_gtk4": <boolean> Optional, default=false
    "options": [ 
            <combo_type|switch_type>, ...
        ]
}
```
combo_type. 
```
{
            "name": <string>, Required. This is also the name of the argument passed to the script as  --<name> <value>
            "label": <string>, Required
            "type": "combo",
            "desktop": <string>, One of all|GNOME|Cinnamon|XFCE|MATE|Budgie|pop|Pantheon|Unity|<some other>
            "ids": [
                <string>, ...
            ],
            "labels": [
                <string>, ...
            ],
            "value": <number> A default value with distributed theme and subsequently used to store the current user value
        }
```
switch_type.
```
{
            "name": <string>, This is also the name of the argument passed to the script when the option is selected (true) eg. --name, no argument is passed when false
            "label": <string>,
            "type": "switch",
            "desktop": <string>,
            "value": <boolean>
        }
```
I suppose other option types could be added eg. a numerical value like Gtk.SpinButton and a color picker option.

Secret-chest:
Colour picker and numbers would be really useful.

JosephMcc:
Couple of initial thoughts on this.

First, with this part of Cinnamon, how many themes are really going to support it? If it's mostly about changing the highlight color, there is already a means in place for that in the theme chooser. So this feels a bit redundant and confusing the way it's presented to the user. On top of that, themes come and go. People get excited about and work on them and then just disappear and leave them unmaintained. This almost feels like something a theme dev could do themselves if they really want to.

The second is the handling of libadwaita. I only took a quick glance but what happens if I use this and then say just switch back to Mint-Y-Blue? Did you just leave a bunch of stuff hanging out in ~/.config? Not a good way to handle this as most people won't get why they are now stuck with that theming on libadwaita apps.

fredcw:
@JosephMcc 

>how many themes are really going to support it?

Yes, I guess that this is the crux of the matter. If it's not going to be adopted by theme devs then then this PR doesn't really serve any purpose. I guess it's also a bit of a chicken and egg situation, theme devs can't adopt it if it isn't implemented. Since the button only appears if a theme supports it, it would make no difference to the settings>themes GUI with themes that don't. Maybe it would be worth a try and if it isn't widely used then it could be removed again at a later date. I guess there's only really one way to find out.

>there is already a means in place for that in the theme chooser.

The idea of using it with Mint-Y, Mint-X, and Mint-L themes was just an afterthought, and having considered it, it's impractical. This idea was only really about 3rd party themes and as you say, there is already a way to select options for the Mint themes. The problem of course is that it only works for Mint themes and can't be used for 3rd party themes.

>If it's mostly about changing the highlight color

Well Fluent theme has 7 different things to configure and the Cinnamox theme variations in the repo for instance, mainly change the background color, and with the last theme uploaded to the repo, the author says he has variations in the gradient fill on panels.

>and work on them and then just disappear and leave them unmaintained

Yes, that's the current situation and I agree that that probably won't change much. But I think that an unmaintained configurable theme is still probably better than an unmaintained non-configurable theme.

>This almost feels like something a theme dev could do themselves if they really want to.

Yes and they do. On gnome-look.org for instance, there are many themes with configurable options. The problem with these though is that changing the options involves downloading the source repo, probably installing dependencies like sassc, and running some install script with various arguments. It's all a lot of hassle. The point of this PR is to take away the hassle and just present the user with options in a GUI. With cinnamon themes uploaded to the spices repo, maybe a GUI in cinnamon like this is exactly what is needed to make theme devs think it's worthwhile implementing configurable options.

>The second is the handling of libadwaita.

Yes, libadwaita is a problem. It seems to me that there are a few ways to deal with this.

1. Do nothing, which means that people can't theme libawaita apps except manually which is the current situation. And unthemeable libawaita apps defeats the purpose of theming. IMO.
2. What I've done with this PR - make it easier to link libadwaita to the current theme (if the theme supports it) and to easily remove the theming if wanted. Unfortunately, if libadwaita is to be themeable at all then there's no way around this. If you were to change to Mint-Y then there's no way to theme libadwaita to Mint-Y anyway except by using the command line. If you still have the theme installed that you linked to libadwaita then hopefully the user would remember that they used that to theme libadwaita and would think to use that to untheme libadwaita. And in any case, they still can't theme Mint-Y to libadwaita. (btw, if the theme that libadwaita is linked to is removed, thereby leaving broken links in `~/.config/gtk-4.0`, libadwaita automatically reverts to the default adwaita theme.)
3. This I think is the best option though I don't know how to go about implementing this: Perhaps in cinnamon settings deamon, the current gtk theme could be monitored and when changed, Cinnamon checks to see if the current gtk theme has a gtk-4.0 folder that appears suitable for linking to ~/.config/gtk-4.0 and if so, automatically installs links, otherwise remove links. That way, whenever the current theme "supports" libadwaita (including Mint-Y), libadwaita is automatically themed to it, otherwise it's set to the default adwaita theme. I don't think people would want to theme libadwaita differently to gtk3 any more than they'd want to theme gtk4 or gtk2 apps differently to gtk3.

So I agree, the libadwaita situation is not ideal but giving the option to theme libadwaita for some themes is at least better than nothing. In any case, this feature can be removed if necessary.

fredcw:
Added color-chooser and spinbutton option types and removed "link theme to adwaita" option. Themes fully update on wayland session without cinnamon restart.

Latest spec for `options_config.json` file here: https://github.com/fredcw/theme-config

Added configurable theme "Graphite" to cinnamon-spices-themes repo.

Secret-chest:
There is a program called Gradience for themeing LibAdwaita.

fredcw:
@Secret-chest Gradience allows you to painstakingly change some colors but it's not really theming in the same way that gtk2, 3, & 4, cinnamon, xfwm, etc are themed, i.e. install a theme and have nearly all your apps have a consistent theme.

Secret-chest:
But it does custom css.

fredcw:
I'll remove the WIP tag because I don't think there's much more I can add to this. There are 3 themes with configurable options in cinnamon-spices-themes now: Fluent, Graphite, and Material.

anaximeno:
@fredcw I believe you are aware of the recent developments on the KDE theme issue of a theme deleting user data? If I understood right, this will make it possible to add a script for extra customizations to be made, therefore making issues like that possible, which means that a greater level of caution would be necessary with pull request to the theme spices repository as well.

Some Refs:
- https://www.reddit.com/r/kde/comments/1bixmbx/do_not_install_global_themes_some_wipe_out_all/
- https://blog.davidedmundson.co.uk/blog/kde-store-content/

fredcw:
@anaximeno Thanks, this is something I hadn't considered. While themes uploaded to cinnamon spices would have to be checked for harmful code, this is the same with applets and desklets. But I think there is a scenario where a user could run untrusted code without realizing. If a user downloaded a theme from an untrusted source thinking that it's safe because it's only a theme, they could then find the theme options button in cinnamon, change the theme options and thereby run untrusted code without realizing that that's what they're doing.

Because of this, I can't recommend this PR be merged in its current state.

PR #12134, opened 2024-03-27
[WIP] Add System Monitor to right click panel when one is found

corbin-auriti:
This uses category heuristics similar to how we find Terminals for Preferred Applications setting
Also update system settings to use symbolic icon.
image

Purpose is quick access, feature parity with other desktop environments as a sane default for muscle memory

image
image

Forum Discussion and Poll:

https://forums.linuxmint.com/viewtopic.php?t=416628

### DIFF
--- a/js/ui/panel.js
+++ b/js/ui/panel.js
@@ -1681,19 +1681,19 @@ PanelCorner.prototype = {
     }
 }; // end of panel corner
 
-function SettingsLauncher(label, keyword, icon) {
-    this._init(label, keyword, icon);
+function SettingsLauncher(label, keyword, icon, command) {
+    this._init(label, keyword, icon, command);
 }
 
 SettingsLauncher.prototype = {
     __proto__: PopupMenu.PopupIconMenuItem.prototype,
 
-    _init: function (label, keyword, icon) {
+    _init: function (label, keyword, icon, command) {
         PopupMenu.PopupIconMenuItem.prototype._init.call(this, label, icon, St.IconType.SYMBOLIC);
 
         this._keyword = keyword;
         this.connect('activate', Lang.bind(this, function() {
-            Util.spawnCommandLine("cinnamon-settings " + this._keyword);
+            Util.spawnCommandLine(command + " " + this._keyword);
         }));
     },
 };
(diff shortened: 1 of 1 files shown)


corbin-auriti:
Purpose is feature parity with other desktop environments as a sane default for muscle memory
image
image

corbin-auriti:
Also for reference a similar PR I made was declined before which had hard-coded options which was less ideal https://github.com/linuxmint/cinnamon/pull/10352

mtwebster:
Speaking of muscle memory, what about Cinnamon users' muscle memories? :)

Maybe putting it in troubleshooting would be safer?  I know I use System Settings here constantly, others probably do as well - this makes it a bit crowded there.

corbin-auriti:
@mtwebster Done, fair compromise

js/ui/panel.js
```diff
@@ -1816,10 +1817,45 @@ PanelContextMenu.prototype = {
...
+                let settingsLauncher = new SettingsLauncher(systemMonitorInfo.name, "", "", systemMonitorInfo.command);
+                settingsLauncher.activate();
+            });
+        }
+
         menu.addMenuItem(new PopupMenu.PopupSeparatorMenuItem()); // separator line
         menu.addMenuItem(menu.troubleshootItem);
 
-        this.addMenuItem(new SettingsLauncher(_("System Settings"), "", "preferences-desktop"));
+        this.addMenuItem(new SettingsLauncher(_("System Settings"), "", "system-run", "cinnamon-settings"));
```
anaximeno, line 1834: Although working with the other icon themes I tested, the system-run icon presents something like a play icon for the breeze icon theme, so it might not be the best suited for System Settings there.

mtwebster:
This finds btop for me (which fails to launch because it's a terminal app). You're only using the executable, and ignoring the `Terminal=true` in the desktop file.

Why not return the appInfo, then [appInfo.launch()](https://docs.gtk.org/gio/method.AppInfo.launch.html)

js/ui/panel.js
mtwebster: The entry is in the Troubleshooting submenu now, so the icon was removed also.

clefebvre:
XApp should provide a system monitor, we might just default to that or nothing it's not present.

LinuxGuy-cyber:
> Speaking of muscle memory, what about Cinnamon users' muscle memories? :)
> 
> Maybe putting it in troubleshooting would be safer? I know I use System Settings here constantly, others probably do as well - this makes it a bit crowded there.

Why'd you require System Settings constantly?

MartinLeGrand:
Just wanted to post here that I would find this useful as well, I need system monitor way more often than the system settings, and if I need settings I usually search for the right category (display, mouse, theme etc.) in the start menu :)

PR #12144, opened 2024-04-03
Prevent window randomization on connecting a monitor

### LINKED ISSUE #12143: Window order in window-list is randomized when connecting a monitor
To replicate (same in a bunch of versions, up to & including `5.8.4`): 
* have several windows (some on 2nd monitor)
* remember the order in window-list
* disconnect 2nd monitor
* reconnect 2nd monitor
* order in window-list is now different

I think that the following line should be protected by `this.refresh = true;`...
https://github.com/linuxmint/cinnamon/blob/d7ca6a6b91a17afc180895be66b2f70790716060/files/usr/share/cinnamon/applets/window-list%40cinnamon.org/applet.js#L1168

Fry-kun:
Fixes #12143


Fry-kun:
Note: the build errors have nothing to do with the PR changes

PR #12171, opened 2024-04-27
cinnamon theme: fix expo-workspaces-name-entry

fredcw:
remove fixed width and adjust height.

before:

Screenshot from 2024-04-27 04-38-35

After:

Screenshot from 2024-04-27 04-41-07

### DIFF
--- a/data/theme/cinnamon.css
+++ b/data/theme/cinnamon.css
@@ -2009,6 +2009,7 @@ StScrollBar StButton#vhandle:hover {
     selected-color: black;
     caret-color: rgb(128, 128, 128);
     caret-size: 1px;
+    min-width: 80px;
     height: 1.4em;
     box-shadow: inset 0px 2px 4px rgba(0,0,0,0.6);
 }
--- a/data/theme/cinnamon.css
+++ b/data/theme/cinnamon.css
@@ -2009,10 +2009,8 @@ StScrollBar StButton#vhandle:hover {
     selected-color: black;
     caret-color: rgb(128, 128, 128);
     caret-size: 1px;
-    width: 250px;
-    height: 1em;
+    height: 1.4em;
     box-shadow: inset 0px 2px 4px rgba(0,0,0,0.6);
-    text-align: center;
 }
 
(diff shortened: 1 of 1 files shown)


anaximeno:
I believe the Cinnamon and the Linux Mint themes might need to update the tooltip style as well, I'm thinking of something similar to their notification style, what do you think about that?

fredcw:
@anaximeno IMO I think the tooltip should look sufficiently different from a popup menu so that they're not easily confused. However the yellow tooltip doesn't really seem in keeping with the rest of the theme, it stands out as a bit odd. Most themes seem to use the inverse colors of a popup menu for the tooltip. So not sure about this one. 🤷‍♂️

JosephMcc:
I think the current yellow is actually appropriate. It's easy to read and stands out the way a tooltip should. The way it's done in some of the other themes isn't ideal now. Some of those are very hard to read the text against the overly bright backgrounds.

anaximeno:
I would say that the main issue with tooltip theming is to have great overall contrast between the text and the tooltip background and the most common/recurrent colors (and brightness) that the user may face below the place the tooltip is being (or will be) shown, despite that I still do believe the current tooltip look on mint-y themes makes sense since it gives a more "cohesive" general look of the system (and some would say it is more modern), so probably the best solution would be something that is in the intersection of all these points.

I don't use the Cinnamon and the Linux Mint themes very frequently, so I wanted to hear more opinions to see if that change was relevant/necessary, so thank you for your input on this!

fredcw:
@A1exKH 
>Do you have any ideas about failed verification?
 
They can be ignored. It just false negatives from the spellchecker.

mtwebster:
Can you put a min-width there at least? Otherwise, it shrinks to nothing as you backspace, which is unexpected.

I tried `min-width: 100px` and got:

image

I know we're trying to get rid of excessive width, but there still needs to be some.

Thanks!

fredcw:
@mtwebster Good idea. How about 80px min as that's just about the width of the default name "Workspace 1"  with default font?

mtwebster:
Looks good to me

PR #12174, opened 2024-05-01
Spices.py: Preserve original file permissions of downloaded spices

fredcw:
Preserve original file permissions of all files in downloaded spice zipfile. This is an improvement on commit https://github.com/linuxmint/cinnamon/commit/42b1e53ef4ae0d26df7a8e8f588cdcfd47be2b12 which give all files the same permissions.


anaximeno:
I agree, I was thinking about this recently, since it could be a vulnerability.

PR #12192, opened 2024-05-12
grouped workspace list applet: move to clutter animations and undepend tweener

anaximeno:
Also improved the scale click animation a bit.

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -4,7 +4,6 @@ const Clutter = imports.gi.Clutter;
 const GLib = imports.gi.GLib;
 const St = imports.gi.St;
 const Main = imports.ui.main;
-const Tweener = imports.ui.tweener;
 const DND = imports.ui.dnd;
 const Tooltips = imports.ui.tooltips;
 const PopupMenu = imports.ui.popupMenu;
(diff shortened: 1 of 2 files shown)


JosephMcc:
I haven't tested these at all but do have one piece of feedback. Please use better commit messages. A bunch of "feat: Use clutter animations" is very unhelpful when browsing the history. Much better would be something along the lines of "gwl: Update appGroup to use clutter animations". It tells me both where you made a change and what you changed in one short title. Much more useful when browsing the commit history.

anaximeno:
> I haven't tested these at all but do have one piece of feedback. Please use better commit messages. A bunch of "feat: Use clutter animations" is very unhelpful when browsing the history. Much better would be something along the lines of "gwl: Update appGroup to use clutter animations". It tells me both where you made a change and what you changed in one short title. Much more useful when browsing the commit history.

Sure, thanks for that suggestion, I was following something called conventional commits, but the way you suggest seems more adequate for this project. I'll try to follow that pattern instead on my next commits here.

mtwebster:
I don't think this is working the way you think it is -

With easing,
- There is no `time` parameter, use `duration` instead, which expects ms, seconds.
- There is no `transition` parameter, use `mode` instead, which expects `Clutter.AnimationMode.EASE_OUT_QUAD`, for example.
- There is no `onCompleteScope` (this was completely breaking the app icon launch animation, and throwing an error into the log.

Have a look at Cinnamon's source, particularly js/ui/windowManager.js, where there are lots of examples of this.

and please test more thoroughly!

ref:
c5f76ff1a41fbd3bca9cae527cb6923a5b1a82d9
#12071
#12072 
#12073 
etc..

anaximeno:
I believe I saw someone using it like this on some applet so I thought it would work, I also didn't notice any issues while running and trying them locally, so thanks to that I didn't consider what you suggested before.

I will verify how this is being properly implemented inside Cinnamon (and also read some documentation) and then re-create these pull requests later. Thanks!

Issue #11731, opened 2023-06-28
Science category missed in menu but should be present based on the specification


Fantu:
### Distribution

Debian

### Package version

5.6.8

### Frequency

Always

### Bug description

Science category missed in menu but "must" be present based on the specification:
https://specifications.freedesktop.org/menu-spec/latest/apa.html
was reported in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1039691

### Steps to reproduce

Open the menu and Science category is missed, the application of this category is under "others"

### Expected behavior

Science category present in the menu

PR #12215, opened 2024-05-30
Add a Science category to the menu.

mtwebster:
fixes: #11731

Science is a mandatory category according to the fdo menu spec: https://specifications.freedesktop.org/menu-spec/latest/apa.html#main-category-registry


anaximeno:
Ping https://github.com/orgs/linuxmint/discussions/420 (@jose-navarro)

Issue #8802, opened 2019-08-18
Grouped window list - only list windows from the current monitor not working


jtlindsey:
```
 * Cinnamon version 4.0.10
   - Using default from Linux Mint 19.1
 * Distribution - Linux Mint 19.1
 * Graphics hardware *and* driver used
  - Using in vmware fusion pro 11 latest. 
 * 64 bit
 * /home/<username>/.xsession-errors

dbus-update-activation-environment: setting DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
dbus-update-activation-environment: setting DISPLAY=:0
dbus-update-activation-environment: setting XAUTHORITY=/home/travis/.Xauthority
dbus-update-activation-environment: setting GTK_MODULES=gail:atk-bridge
dbus-update-activation-environment: setting QT_ACCESSIBILITY=1
dbus-update-activation-environment: setting LANG=en_US.UTF-8
dbus-update-activation-environment: setting GDM_LANG=en_US
dbus-update-activation-environment: setting DISPLAY=:0
dbus-update-activation-environment: setting MANDATORY_PATH=/usr/share/gconf/cinnamon.mandatory.path
dbus-update-activation-environment: setting XDG_GREETER_DATA_DIR=/var/lib/lightdm-data/travis
dbus-update-activation-environment: setting USER=travis
dbus-update-activation-environment: setting DESKTOP_SESSION=cinnamon
dbus-update-activation-environment: setting DEFAULTS_PATH=/usr/share/gconf/cinnamon.default.path
dbus-update-activation-environment: setting PWD=/home/travis
dbus-update-activation-environment: setting HOME=/home/travis
dbus-update-activation-environment: setting QT_ACCESSIBILITY=1
dbus-update-activation-environment: setting XDG_SESSION_TYPE=x11
dbus-update-activation-environment: setting XDG_DATA_DIRS=/usr/share/cinnamon:/usr/share/gnome:/home/travis/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share:/usr/share
dbus-update-activation-environment: setting XDG_SESSION_DESKTOP=cinnamon
dbus-update-activation-environment: setting GTK_MODULES=gail:atk-bridge
dbus-update-activation-environment: setting SHELL=/bin/bash
dbus-update-activation-environment: setting XDG_SEAT_PATH=/org/freedesktop/DisplayManager/Seat0
[...]

PR #12225, opened 2024-06-08
Grouped window list: Add option to only list windows from the current monitor

jarierca:
Hi :)

I added an old feature to show only active applications on the current monitor which was deleted in [https://github.com/linuxmint/cinnamon/commit/ef4610cd8963c87eb12dc8dcf717a9e6ca1dbd78#diff-10f95e2348d1571d611014de1fa17308](commit)
Now allows to toggle between showing only current monitor applications or showing all apps from all monitors. This PR is related to this topic #8802

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/applet.js
@@ -328,7 +328,8 @@ class GroupedWindowListApplet extends Applet.Applet {
             {key: 'show-recent', value: 'showRecent', cb: null},
             {key: 'autostart-menu-item', value: 'autoStart', cb: null},
             {key: 'monitor-move-all-windows', value: 'monitorMoveAllWindows', cb: null},
-            {key: 'show-all-workspaces', value: 'showAllWorkspaces', cb: this.refreshAllAppLists}
+            {key: 'show-all-workspaces', value: 'showAllWorkspaces', cb: this.refreshAllAppLists},
+            {key: 'list-monitor-windows', value: 'listMonitorWindows', cb: this.reloadAllAppsMonitor}
         ];
 
         for (let i = 0, len = settingsProps.length; i < len; i++) {
@@ -503,6 +504,11 @@ class GroupedWindowListApplet extends Applet.Applet {
 
     }
 
+    reloadAllAppsMonitor(){
+        this.refreshAllAppLists();
+        this.updateMonitorWatchlist();
+    }
+
     updateMonitorWatchlist() {
         if (!this.numberOfMonitors) {
             this.numberOfMonitors = global.display.get_n_monitors();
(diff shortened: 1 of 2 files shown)


JosephMcc:
Imho, this functionality should never have been removed. It's one of the most useful features when using multiple monitors.

clefebvre:
How do you use it though?

As it is, it already does only show you your monitor apps if you have an applet on the other monitor.

jarierca:
Hello, I'm sorry for explaining so briefly.
As it is now, if you have 1 panel and for example 3 monitors, all applications are displayed in the single panel.

only-1-panel

However, if you want to have 3 panels (1 per monitor), applications from all monitors are not displayed.

only-list-apps-from-current-monitor

That's why I added the button "Only list windows in the current monitor" (when you have more than 1 panel) (defaulting to true, which would have the same effect as it currently is before this change).

grouped-window-list-config

If I set this button to false, for example on the central monitor it would look like this (if I tried to do the same as it is now, I wouldn't see all the apps open on the central monitor).

list-apps-from-all-monitor-mid-monitor

And if I mark it as false in the configuration for all panels, it would look like this:
list-apps-from-all-monitors

If you need more help or information, let me know. ;)

mtwebster:
Hi, seems to work fine, I'm just not sure about the wording of the setting - when you compare it to the current behavior it can be a bit ambiguous, depending on your current layout. I think maybe it could be better as a combo box:

Multiple instances:
- Track windows individually per monitor
- Always track windows from all monitors

The first being the default - it doesn't really describe the edge case of having two instances but three monitors, but I don't think that's common anyhow.

I'm not set on this, just throwing an idea out

jarierca:
Hi  @mtwebster 

Thank you for your feedback!. I have updated the settings to include a combo box with the two most common options: "Track windows individually per monitor" and "Track windows across all monitors." This change aims to improve clarity in the use of window display mode.

Thanks again for your input!

zolstarym:
I have 3 monitors. When I have one 'grouped window list' on one monitor and and the normal 'window list' on the other two monitors, 'grouped window list' will display icons from every monitor, ignoring the show windows from monitors without a window list setting, since the normal 'window list' applet already has those windows showing on those monitors.
   When I add a 'grouped window list' to a second monitor, all of a sudden both of my 'gwl' applets seem to only show windows from those monitors, even if I remove all of the other 'wl' applets from every monitor. In that case the 'gwl' still does not show the windows from my third now applet-less monitor. 
   So it seems like there could be a bug with 3 monitor setups, and that 'gwl' might also be ignoring 'wl'. 
   Toying around with it some more, when I move one of the 'gwl' applets to my middle (primary) monitor, it can see the windows from the left monitor, even though when it is on the left monitor it doesn't list windows from the middle (with another 'gwl' on the right monitor)

zolstarym:
window list itself seems to do the same thing. One window list on one monitor and it shows everything, but put a second one on the other monitor, and then they only show what is on their current monitor, leaving windows on the third unlisted.

PR #12227, opened 2024-06-09
window-list: Remove indicators for minimized and tiled windows from window names

j9d:
Hi!

This adds an option to remove the `[ ]` and `|` indicators on minimized/tiled windows (I'm not personally a fan, was also confusing figuring out why these were being added at first).

Setting:

settings

Cheers

### DIFF
--- a/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
@@ -524,12 +524,6 @@ class AppMenuButton {
         if (this._tooltip && this._applet.windowHover != "nothing" && this._tooltip.set_text)
             this._tooltip.set_text(title);
 
-        if (this.metaWindow.minimized) {
-            title = "["+ title +"]";
-        } else if (this.metaWindow.tile_mode != Meta.TileMode.NONE && this.metaWindow.tile_mode != Meta.TileMode.MAXIMIZED) {
-            title = "|"+ title;
-        }
-
         this._label.set_text(title);
     }
 


JosephMcc:
I'm with you on not being a big fan of these. I do think adding this option is overkill though. If they aren't liked or needed, they should just be removed.

j9d:
> If they aren't liked or needed, they should just be removed.

Fair enough. This has been in the codebase for a while so I didn't want to open with removing it, lol. I'll remove it later today if others are happy for me to.

Secret-chest:
This should also be changed in GWL.

mentha-spicata:
I was surprised to see this being completely removed. It was the only indicator left that tells you which windows are minimized.

I use (and switch between) multiple terminal windows every day on a laptop screen and when searching for a particular window I need to tell at a glance which ones are minimized. Same goes for any other program that one might have multiple windows of, like a file browser or a text editor.

I agree that brackets were never a very good indicator. In fact, years ago I have myself suggested their replacement with theming that would style minimized windows in a different way. But not having any indicator at all is even worse.

Focused windows already have their own style and theming, so minimized windows could use the same approach as well.

calestyo:
IMO that's a really unfortunate change... (well, especially not making it configurable)... brackets may not have been the most aesthetic way of indicting, but at least there was a way and now there's no way of differentiating minimised/non-minimised windows in the list, which is quite annoying.

mikaku:
How is supposed now I'll be able to know which windows are minimized in taskbar?

Secret-chest:
Maybe a fainter text would work?

calestyo:
> Maybe a fainter text would work?

May or may not work, e.g. for me, even the fainter grey for the window list item whose window has the focus was not really enough and I had to override it via CSS.<br>
One should also keep people with accessibility needs in mind.

Ideally, I think, the applet would allow a range of options, like:
- no indication
- smaller or greyed out icon (unless that could also be done via CSS)
- the "good old" `[...]`
- or any other arbitrary changes to text/background that can be done via some CSS selector

mikaku:
@j9d, my Cinnamon Desktop 6.2.9 looks different than yours:

Screenshot from 2024-11-19 09-08-18

What version you have there?

PR #12233, opened 2024-06-13
notifications@cinnamon.org : applet improvement

Hilyxx:
Hello

When users enable "Do not disturb" mod, icon isn't updated on the panel.
This commit improves notifications applet with these changes : 

- Add a new icon to panel when "Do not disturb" mod is enabled
- New menu label when "Do not disturb" mod is enabled
- Add a toogle switch to easily enable/disable notifications
- Tooltip now displays notification count

[Screencast 2024-06-14 04:42:26.webm](https://github.com/linuxmint/cinnamon/assets/21991448/cb9e2574-5d77-40eb-a52a-fd9657861131)


rcalixte:
This is what the inhibit@cinnamon.org applet does?

Hilyxx:
Hello @rcalixte 
Yes maybe but inhibit applet is not notification applet. Some features such as power management are not useful for desktop user. Having a notification applet with all functionality gives more choice for users. This avoids the need 2 applets to manage notifications.

rcalixte:
The inhibit applet has bindings to facilitate enabling and disabling notifications. The notification applet is also currently not visible unless there is at least one notification in the drawer as well. I would think changing that behavior would at least be a setting or an option for users who don't want it to change.

Hilyxx:
Do not disturb icon is visible only when notifications are disabled to advert the user, if this icon is hidden it's not useful... But I understand your position and I will add an option for users.

I close this commit and I restart a clean pull request.

PR #12234, opened 2024-06-14
Notifications@cinnamon.org : applet improvement

Hilyxx:
Hello

When users enable "Do not disturb" mod, icon isn't updated on the panel.
This commit improves notifications applet with these changes : 

- Add a new icon to panel when "Do not disturb" mod is enabled
- New menu label when  "Do not disturb" mod is enabled
- Add a toggle switch to easily enable/disable notifications
- Tooltip now displays notification count
- Add an option to hide "do not disturb" icon on panel when show empty tray option is enabled (@rcalixte request)

[Screencast 2024-06-14 04:42:26.webm](https://github.com/linuxmint/cinnamon/assets/21991448/2a5715b0-9b16-4813-ac65-96caca31d3ec)


rcalixte:
@Hilyxx Have you considered breaking this out into a separate applet? As a separate Spice, you would have much more flexibility for an implementation.

Hilyxx:
Hello @rcalixte 
Will this applet be approved if I submit it as a separate applet ?

rcalixte:
> Hello @rcalixte Will this applet be approved if I submit it as a separate applet ?

I don't see why not. It seems like the perfect solution actually. Similar to CinnaMenu as an alternative menu applet or the few window list ones. Having it separate means that you would have much more freedom with it and also fill a potential need if there are users that prefer an alternative.

Hilyxx:
Ok, thanks @rcalixte, I will do that.

PR #12247, opened 2024-06-20
Update appGroup.js

AchillesBoi:
Make taskbar app icon button width the same as the "Menu" applet button width, for consistency.

Ever since Mint 21.x the app icon buttons have slimmed down so much that the buttons look like upright rectangles and not the 4:3-ish square buttons one would expect, like the "Menu" applet. It's even more obvious since both the "Menu" and "Grouped window list" applets are sitting right next to each other.

I propose this change to make both applet buttons the exact same width for a more consistent and pleasant look.


JosephMcc:
I don't see any reason for this. The theme is in control of the width/padding of these buttons. That's how it should be. Your premise doesn't hold up here. In another theme, the padding on your menu applet may be different from your current theme, which will just break this again.

AchillesBoi:
Ah, my bad. My intention is to make the default button width the exact same as the "Menu" applet button. How can I go about proposing this change to the many possible variations of the Mint-Y theme? Some pointers would be great.

fredcw:
https://github.com/linuxmint/mint-themes/blob/master/src/Mint-Y/cinnamon/README.md

PR #12256, opened 2024-06-27
ExtensionCore.py: Check for display names in translations

rcalixte:
Some language codes are only available as translations via their long language code format. This will check for the more precise long language code first before looking for the shorter and more general variation. This will improve translation support for how Spices modules are displayed before being installed locally.

cc @aaasoft


aaasoft:
Well done.

PR #12260, opened 2024-06-29
make notification states immutable

hduelme:
Ensuring that notification states are immutable helps prevent any unwanted or accidental modifications.

### DIFF
--- a/js/ui/messageTray.js
+++ b/js/ui/messageTray.js
@@ -31,34 +31,34 @@ var LONGER_HIDE_TIMEOUT = 0.6;
 const NOTIFICATION_IMAGE_SIZE = 125;
 const NOTIFICATION_IMAGE_OPACITY = 230; // 0 - 255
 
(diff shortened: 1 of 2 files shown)


rcalixte:
Can you expand on the change here a little bit? What are the differences in the before/after in the user experience? Are there any examples that might be useful?

hduelme:
Currently anyone using the API (for example spices) could accidental modify the stats, for example `MessageTray.Urgency.CRITICAL = 0;` could accidental be executed, resulting in a wrong internal state for `CRITICAL`. This change prevents this. Any changes made to this would be ignored, as cinnamon don't run in strict mode. For the end-user nothing should change at all.

hduelme:
Sorry for my late response. I was able to reproduce the error. The problem occurs because I change `var` to `const`. The way the states are accessed requires them to be global scoped. I've reverted them to `var` to resolve the issue.

PR #12262, opened 2024-06-30
[power@cinnamon.org] Add support for more battery device labels and icons

garrettjstevens:
I have a device that shows up in UPower as a `HEADSET`, which wasn't one of the enumerated device types in the power applet so it showed up as "Unknown" with a generic battery icon. When I went to add support for the headset, I noticed this comment:

https://github.com/linuxmint/cinnamon/blob/8d97e5689f7bea670fb8882d54f48f91e33f04bf/files/usr/share/cinnamon/applets/power%40cinnamon.org/applet.js#L16

So instead of adding only the `HEADSET` device type, I've added `UPowerGlib` introspection and used `UPDevice.kind_to_string(kind)` so now any device type should get a reasonable label. I've also added icons for several more device types.

I've also renamed a few variables to be consistent with the `DeviceKind` terminology used by `UPowerGlib` (instead of `DeviceType`)

### DIFF
--- a/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
@@ -19,6 +19,7 @@ const {
     DeviceKind: UPDeviceKind,
     DeviceLevel: UPDeviceLevel,
     DeviceState: UPDeviceState,
+    Device: UPDevice
 } = UPowerGlib
 
 function deviceLevelToString(level) {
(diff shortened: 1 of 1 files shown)


files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
```diff
@@ -67,52 +39,69 @@ function deviceLevelToString(level) {
...
+        case UPDeviceKind.BLUETOOTH_GENERIC:
+            return _("Bluetooth device");
+        default: {
+            try {
+                return _(
+                    UPDevice
+                        .kind_to_string(kind)
+                        .replaceAll('-', ' ')
+                        .capitalize()
+                );
```
mtwebster, line 66: This would end up returning untranslated strings. I think we should keep the existing translated device types, then do this to handle any newly added device types.

PR #12272, opened 2024-07-03
Use Polkit agent for X11 sessions

SparkyBluefang:
This simplifies packaging since we don't need to worry if an external polkit agent is needed. This is currently a draft since I understand upstream may not be ready for this change yet. But I'm creating this now so I can reference it in Gentoo packaging.

Also, don't crash if the agent doesn't start properly. If another agent happens to be running, this will throw a DBus exception. I figured popping a notification would be most appropriate, since the user may have another agent configured to auto-start.
```
(cinnamon:9337): Gjs-CRITICAL **: 13:47:05.571: JS ERROR: Polkit.Error: GDBus.Error:org.freedesktop.PolicyKit1.Error.Failed: An authentication agent already exists for the given subject
_init@/usr/share/cinnamon/js/ui/polkitAuthenticationAgent.js:332:22
AuthenticationAgent@/usr/share/cinnamon/js/ui/polkitAuthenticationAgent.js:323:10
init@/usr/share/cinnamon/js/ui/polkitAuthenticationAgent.js:392:17
start@/usr/share/cinnamon/js/ui/main.js:433:31
@<main>:1:47
```

### DIFF
--- a/js/ui/polkitAuthenticationAgent.js
+++ b/js/ui/polkitAuthenticationAgent.js
@@ -35,6 +35,7 @@ const PolkitAgent = imports.gi.PolkitAgent;
 const ModalDialog = imports.ui.modalDialog;
 const CinnamonEntry = imports.ui.cinnamonEntry;
 const UserWidget = imports.ui.userWidget;
+const Main = imports.ui.main;
 
 const DIALOG_ICON_SIZE = 64;
 
@@ -389,5 +390,22 @@ AuthenticationAgent.prototype = {
 }
 
 function init() {
-    let agent = new AuthenticationAgent();
+    try {
+        let agent = new AuthenticationAgent();
+    } catch(err) {
+        if(!(err instanceof Error)) {
+            err = new Error(err);
+        }
+
+        log('polkitAuthenticationAgent: init error ' + err);
+
+        let icon = new St.Icon({ icon_name: 'dialog-warning',
+                                 icon_type: St.IconType.FULLCOLOR,
+                                 icon_size: 36 });
+
+        Main.warningNotify(_('Unable to start Cinnamon PolicyKit Agent'), err.message +
+                           "\n\n" +
+                           _("If you have another PolicyKit Agent configured to autostart, it should be disabled."),
+                           icon);
+    }
 }
--- a/debian/control
+++ b/debian/control
@@ -85,7 +85,6 @@ Depends:
  metacity,
  nemo,
  network-manager-gnome [linux-any],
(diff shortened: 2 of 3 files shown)


js/ui/polkitAuthenticationAgent.js
```diff
@@ -389,5 +390,22 @@ AuthenticationAgent.prototype = {
...
+        log('polkitAuthenticationAgent: init error ' + err);
+
+        let icon = new St.Icon({ icon_name: 'dialog-warning',
+                                 icon_type: St.IconType.FULLCOLOR,
+                                 icon_size: 36 });
+
+        Main.warningNotify(_('Unable to start Cinnamon PolicyKit Agent'), err.message +
+                           "\n\n" +
+                           _("If you have another PolicyKit Agent configured to autostart, it should be disabled."),
+                           icon);
```
SparkyBluefang, line 409: This was my best guess at how to handle this. I wasn't sure if it belonged here or in main.js

Fantu:
Fedora already switched to internal polkit agent also for xorg, probably @leigh123linux can tell you how it's going and if he has encountered any problems

leigh123linux:
No one has complained yet about the switch to internal polkit, 6.2.x was promoted to F40 stable the other day.

SparkyBluefang:
If that's the case, then maybe this is closer to prime time than I thought :)

JosephMcc:
This probably shouldn't be merged currently. Part of the reason we don't use this dialog is because it doesn't support multiple users yet. It's something I'm looking at adding this development cycle. Maybe should revisit the idea then.

Fantu:
Why close this? Will be default in 6.4?
I did a fast look and saw this that also remove the restrict to wayland but not the changes in packaging: https://github.com/linuxmint/cinnamon/commit/9945a8aca1550a19d377cdcb19f8addab994260e

JosephMcc:
It will be enabled by default for 6.4

JosephMcc:
> Remove of policykit-1-gnome from d/control is missed, if it will remain installed and will start at boot will make cinnamon polkit agent fails FWIK

I'm unsure about that part, I'll let clem remove dependencies he doesn't want. The polkit agent works fine here though.

SparkyBluefang:
Does https://github.com/linuxmint/cinnamon/commit/9945a8aca1550a19d377cdcb19f8addab994260e gracefully handle dbus errors? That was the majority of what this PR dealt with - making sure cinnamon didn't crash if the polkit bus was already claimed.

PR #12273, opened 2024-07-03
Optionally install wayland session files

SparkyBluefang:
Add options to disable installing wayland session and xdg-portal files. Cinnamon shouldn't advertise having wayland sessions if the rest of the packages are built without wayland support.

I didn't see a good option for handling `cinnamon-session-cinnamon` without a more significant refactor. But fortunately `cinnamon-session` aborts with an obvious error message
```
$ cinnamon-session-cinnamon --wayland
cinnamon-session-binary[25984]: CRITICAL: t+0.00369s: Unable to start session: Failed to load session "cinnamon-wayland"
```

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -16,11 +16,10 @@ desktop_conf.set('bindir', join_paths(prefix, bindir))
 desktop_conf.set('libexecdir', join_paths(prefix, libexecdir))
 desktop_conf.set('VERSION', meson.project_version())
 
-desktop_files = [
-  'cinnamon.desktop',
-  'cinnamon-wayland.desktop',
-  'cinnamon2d.desktop',
-]
+desktop_files = ['cinnamon.desktop', 'cinnamon2d.desktop']
+if get_option('wayland')
+    desktop_files += ['cinnamon-wayland.desktop']
+endif
 
 foreach desktop_file : desktop_files
     desktop = configure_file(
@@ -40,7 +39,9 @@ foreach desktop_file : desktop_files
 endforeach
 
 subdir('xdg-portal')
-subdir('wayland_sessions')
+if get_option('wayland')
+    subdir('wayland_sessions')
+endif
 subdir('xsessions')
 subdir('services')
 
--- a/meson.build
+++ b/meson.build
@@ -165,10 +165,15 @@ else
     session_conf.set('REQUIRED', '')
 endif
 
-foreach file : ['cinnamon.session', 'cinnamon2d.session', 'cinnamon-wayland.session']
+session_files = ['cinnamon.session', 'cinnamon2d.session']
+if get_option('wayland')
+    session_files += ['cinnamon-wayland.session']
+endif
+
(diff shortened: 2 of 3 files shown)


Fantu:
It seems clear to me that Wayland support is experimental and incomplete, but having the opportunity to try it can help you see the problems that will gradually be resolved.
Have you perhaps encountered problems with users who complained, or other problems?

SparkyBluefang:
For me, this is more about packaging consistency. All of the other cinnamon packages have optional wayland support, and if muffin (for example) is built without, it seems a little silly to be installing wayland session files.

I just checked and `cinnamon-session --session cinnamon-wayland` will happily launch a session even if muffin is built without. But that means it will be missing components like the screensaver daemon, which probably isn't ideal.

At least in Gentoo, wayland is not enabled by default unless you're specifically using a GNOME or KDE package manager profile. The xdg-portal support is also optional, though enabled by default. So it would be odd to install the config/session files if xdg-desktop-portal-xapp is not installed or wayland is disabled.

Side note - since cinnamon wayland support is experimental/incomplete, the Gentoo package does have an additional mask prevention users from enabling wayland builds. But a user is able to unmask and enable wayland, if they desire.

I'm not strongly pushing for this being accepted upstream, since this might be a bit specific to Gentoo packaging. If not, I'll just continue to conditionally remove these files in the package definition. But since I was going through and submitting a variety of PRs, I figured I'd give this one a chance too :)

leigh123linux:
I don't get the point to making xdg-portal conf optional, it is used in xorg session are well.
Are you going to deal with gentoo users who complain their apps (example gnome-online-accounts-gtk) don't have proper dark theme support.

SparkyBluefang:
At one point, the xdg portal stuff was just targeting flatpack and other app sandboxes. But, as you point out, it's evidentially becoming increasingly prevalent in normal desktop apps/envs. Annoying, considering `xdg-desktop-portal` pulls in stuff like `pipewire` and `fuse`, but maybe better in the long run since it standardizes some of the settings rat's nest that DEs have to deal with.

`libadwaita` does use the portal by default - it does support uing gsettings if `ADW_DISABLE_PORTAL=1` is in the environment, but that's hardly appropriate to set at the session/system level (and Cinnamon isn't copying the necessary gsettings keys anyways). More unexpectedly, it seems to break dark mode in Cinnamon itself, and I can't figure out where it's talking to either gsettings or the portal.

Regardless, the simple solution is to make the portal stuff required, since that seems to be the trend. I've dropped those changes from the patch.

mtwebster:
> More unexpectedly, it seems to break dark mode in Cinnamon itself, and I can't figure out where it's talking to either gsettings or the portal.

Many of our apps use [XAppDarkModeManager](https://projects.linuxmint.com/xapp/reference/XAppDarkModeManager.html) as a way of bringing Gtk3 in line with libadwaita and browsers, for setting dark mode. Maybe this is what you're running into?

PR #12278, opened 2024-07-06
Simplify 64bit path handling in cinnamon-settings

SparkyBluefang:
Platforms like  `arm64`/`aarch64` also use `/usr/lib64`, so don't restrict it to `x86_64`/`amd64`.

One additional path check is probably not worth enumerating all of the applicable machine names.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/capi.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/capi.py
@@ -27,20 +27,18 @@ class CManager:
         self.extension_point = Gio.io_extension_point_register ("cinnamon-control-center-1")
         self.modules = []
 
-        architecture = platform.machine()
         # get the arch-specific triplet, e.g. 'x86_64-linux-gnu' or 'arm-linux-gnueabihf'
         # see also: https://wiki.debian.org/Python/MultiArch
         triplet = sysconfig.get_config_var('MULTIARCH')
-        paths = ["/usr/lib", f"/usr/lib/{triplet}"]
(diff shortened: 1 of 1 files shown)


files/usr/share/cinnamon/cinnamon-settings/bin/capi.py
```diff
@@ -27,20 +27,18 @@ def __init__(self):
...
         # For instance, on a Mint i686 box, the path is actually /usr/lib/i386-linux-gnu
+        architecture = platform.machine()
         x86archs = ["i386", "i486", "i586", "i686"]
         if architecture in x86archs:
             for arch in x86archs:
                 paths += ["/usr/lib/%s" % arch]
-        elif architecture == "x86_64":
-            paths += ["/usr/lib/x86_64", "/usr/lib64"]
         else:
             paths += ["/usr/lib/%s" % architecture]
```
SparkyBluefang, line 43: Are these paths even valid? It seems like the [Ubuntu](https://wiki.ubuntu.com/MultiarchSpec) and [Debian](https://wiki.debian.org/Multiarch/Tuples) multiarch specs specially call out using the platform triplet and not just the plain machine/arch name.

leigh123linux:
It works fine on fedora

```
[leigh@mpd-pc ~]$ python
Python 3.13.0 (main, Oct  8 2024, 00:00:00) [GCC 14.2.1 20240912 (Red Hat 14.2.1-3)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform
>>> platform.machine()
'x86_64'
>>> 
```

PR #12300, opened 2024-07-21
cs_themes: Enable stack switcher when navigating out the theme module

anaximeno:
fix https://github.com/linuxmint/mint22-beta/issues/42

@mtwebster @clefebvre if you could give it a look at this, it would be nice if this was patched before the stable release.


anaximeno:
If this would have to be merged, the commit https://github.com/linuxmint/cinnamon/commit/da63f90c300d4479b0cb2b62b0a93c6236915cc2 can/should probably be reverted, because if the stack switcher at the top is not made insensitive, you still will be able to click on the invisible stack switcher in the module.

mtwebster:
Thanks, sorry about that.

PR #12302, opened 2024-07-22
added support for using numpad arrows to navigate the menu

Pjsrcool:
Currently, the real arrow keys work when navigating the Cinnamon Menu, but the numpad arrows do not.

This PR adds the ability to navigate the Cinnamon menu with the numpad arrows.

This is a quality of life change for those who
- use keyboards with a numpad but no arrows, or tiny laptop arrows
- use southpaw (left handed) keyboards
- use 60% keyboards and have external numpads

Question: For the section where it flips the direction for RTL layouts, does this also apply for numpad arrows too? I am unfamiliar with RTL layouts so I am not confident in making this decision myself.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1686,7 +1686,7 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
         }
 
         let minIndex = 0;
-        let goUp = symbol === Clutter.KEY_Up;
+        let goUp = symbol === Clutter.KEY_Up || symbol === Clutter.KEY_KP_Up;
         let nextActive = null;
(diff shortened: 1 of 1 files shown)


fredcw:
Looks OK and seems to work fine except yes, you do need to flip the direction for RTL layouts so that the keypad does the same as the arrow keys.

line 1759:
```
        if (St.Widget.get_default_direction() === St.TextDirection.RTL) {
            if (symbol === Clutter.KEY_Right || symbol === Clutter.KEY_KP_Right) {
                symbol = Clutter.KEY_Left;
            } else if (symbol === Clutter.KEY_Left || symbol === Clutter.KEY_KP_Left) {
                symbol = Clutter.KEY_Right;
            }
        }
```

Pjsrcool:
I've added the logic to account for RTL layout as recommended above. I'm happy with this unless there are more necessary changes.

Issue #12307, opened 2024-07-24
Grouped window list item become a difficult-to-see thin bar if the application specifies an unavailable icon


billyswong:
### Distribution

Mint 21.3 Cinnamon

### Package version

cinnamon 6.0.4

### Frequency

Always

### Bug description

It is a bug taken from https://forums.linuxmint.com/viewtopic.php?t=425061
If an application specifies an icon that doesn't exist in the icon theme, no placeholder icon is given, making the application become a thin almost invisible stripe within the grouped window list.

### Steps to reproduce

1. Select "Breeze" as the icon theme
2. Open Menu Editor via right click menu icon > Configure > Menu > Open the menu editor

### Expected behavior

Although Menu Editor specifies `menu-editor` to be its icon and "Breeze" as a KDE icon theme doesn't have such icon, the grouped window list should have put in a placeholder icon there to balloon / fill up the window entry, not leaving it as a extremely narrow stripe that's hard to see and hard to click.

fredcw:
The icon needed is called `alacarte`. A default icon should be added to cinnamon package. If someone cares to choose one, I guess we can add it here: https://github.com/linuxmint/cinnamon/tree/master/files/usr/share/icons/hicolor/scalable/apps

I haven't looked into added a placeholder icon in Grouped Window List but window-list@cinnamon.org applet has the same problem. Cassia Window List and CobiWindowList applets also have missing icons but a wider gap is used.

Since an app not having a default icon is a bug in itself and also extremely rare, I would say that adding an `alacarte` icon to cinnamon would be the easiest solution.

billyswong:
Thank you for correction in the icon name in use. I found the Breeze icon set contains an icon called `preferences-desktop-menu-edit`. Somebody ought to patch the `alacarte` to link to `preferences-desktop-menu-edit` in Breeze icon set.

PR #12339, opened 2024-08-06
GWL: Set reasonable min width for panel item ...

fredcw:
... in case application icon is missing.

Fixes: https://github.com/linuxmint/cinnamon/issues/12307

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -340,7 +340,7 @@ class AppGroup {
         const [labelMinSize, labelNaturalSize] = this.label.get_preferred_width(forHeight);
         // The label text starts in the center of the icon, so we should allocate the space
         // needed for the icon plus the space needed for(label - icon/2)
-        alloc.min_size = 1 * global.ui_scale;
+        alloc.min_size = this.iconSize * global.ui_scale;
 
         const {appId} = this.groupState;
 


mtwebster:
If the application icon is missing there should be a placeholder/missing-image icon, not an empty space - I don't think this is a good way to address that issue.

fredcw:
The problem is that app.create_icon_texture_for_window() and app.create_icon_texture() return an StIcon with no indication that the icon is missing and I don't see any obvious way to detect that. I have the same problem in Cinnamenu. I'll have another look later to see if I can find a way to do it.

fredcw:
Ah, it seems that both `app.create_icon_texture_for_window()` and `app.create_icon_texture()` should return the icon `application-x-executable` where no other icon is found. But these don't seem to be working reliably.

mtwebster:
I think since the example app/theme (breeze/alacarte) are defined and have a matching desktop file (so are not window-backed), we're [calling](https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js#L274) `cinnamon_app_create_icon_texture ()`.

That gets the icon info from the GAppInfo (the desktop file representation) and never check if it's available before returning a useless StIcon.

This could be fixed easily in StIcon (checking the theme for the icon before trying to use it) but we'd lose the ability to have different context-dependent fallback icons (default app icon for gwl, missing image icon in other places, etc...).

It makes the most sense to check this as part of `cinnamon_app_create_icon_texture ()` maybe using `gtk_icon_theme_lookup_by_gicon ()`, but desktop files can also use an absolute path for their Icon entry - now you're checking if a file exists, which is racy...

I don't have a good answer yet lol I was just digging a bit.

fredcw:
I think I have a solution. On line https://github.com/linuxmint/cinnamon/blob/master/src/cinnamon-app.c#L188 instead of `g_themed_icon_new`, use `g_themed_icon_new_from_names`.

I'm not sure how to do it in C but in JS I can use `Gio.ThemedIcon.new_from_names(['bob','application-x-executable'])` instead of `Gio.ThemedIcon.new('bob')` to get the 'application-x-executable' icon if `bob` doesn't exist.

fredcw:
Actually, this only solves the problem for `cinnamon_app_create_icon_texture_for_window` but not for `cinnamon_app_create_icon_texture`.

In `cinnamon_app_create_icon_texture`, line 303:
```
icon = g_app_info_get_icon (G_APP_INFO (app->info));
```
also returns an invalid gicon.

clefebvre:
Definitely a good issue to look into but the PR itself should be closed.

When the screen is too small to accommodate too many apps, the grouped window list icons get sliced/squashed:

image

It's not pretty but it means we see all apps.

Setting a minsize breaks that, it prevents the slicing/squashing so when there's no more space, we don't see the icon for newly added windows.

PR #12364, opened 2024-08-25
Replace the default Cinnamon theme

JosephMcc:
Replaces the very outdated default cinnamon theme with something more up-to-date. May still be a few issues here and there that more testing/using will turn up. A few things are also on the roadmap to update style wise. This will set the base for those to come next.


Initu-Castilhos:
❤️

Initu-Castilhos:
Will it be available for download so I can test it?

JosephMcc:
@Initu-Castilhos No it can't be downloaded anywhere. It's part of Cinnamon and I'm not developing it as anything else at the moment.

A couple screenshots of the new dialog styling.
Screenshot from 2024-09-15 20-56-40

Screenshot from 2024-09-15 20-57-08

JosephMcc:
Updated calendar applet styling

Screenshot from 2024-09-15 21-22-48

Initu-Castilhos:
Thanks for the reply!

Initu-Castilhos:
Will it have the gtk theme and icons too?

JosephMcc:
@Initu-Castilhos No. Cinnamon doesn't include Gtk and icon themes by default. This is just the default cinnamon theme which gets used when you install Cinnamon in distros outside of Mint. Many of those distros don't have a curated Cinnamon version like Mint does.

Initu-Castilhos:
Can you make the theme's cinnamon.css file available?

JosephMcc:
@Initu-Castilhos In this case it wouldn't do much good. There are actual code changes in the branch to redesign some things. Without those the theme won't work the same. You need to actually build and use the branch.

Initu-Castilhos:
> @Initu-Castilhos In this case it wouldn't do much good. There are actual code changes in the branch to redesign some things. Without those the theme won't work the same. You need to actually build and use the branch.

Ok!

clefebvre:
In the source the theme is written in SASS. In the OS once installed you can see the CSS in /usr/share/cinnamon/theme.

PR #12379, opened 2024-09-04
Update settings-schema.json - set enable-app-button-dragging to false

leigh123linux:
It's an annoying default.


anaximeno:
It's easy to disable on the settings, so why change the defaults?

leigh123linux:
> It's easy to disable on the settings, so why change the defaults?

I want to disable it for fedora, also it doesn't need to be enabled all the time.
If users want to rearrange the panel they can enable it.

PR #12394, opened 2024-09-16
environment.js: Add some convenience LayoutManager monkey patching

JosephMcc:
Add hookup_style() to bind layoutManager properties to CSS properties


JosephMcc:
Using this functionality in a new applet I've been working on.

PR #12395, opened 2024-09-18
docs: Remove recorder.xml.in

rcalixte:
This file references another for inclusion that no longer exists. With gtk-doc-tools 1.34.0, stricter checking is enabled that results in a failure which halts the build process due to this invalid file reference.


Fantu:
Thanks for this, tested on Debian and fixed failed build on Sid, will be used for next upload.
I think will be useful for other maintainers, for example @leigh123linux and @SparkyBluefang

PR #12396, opened 2024-09-18
cs_keyboard.py: Skip invalid bindings

rcalixte:
It is possible to remove Spices manually by removing the associated directories rather than using the appropriate module. This results in the bindings still being stored internally. When loading the Keyboard module and this scenario has occurred, it results in a failure due to the matching files having been removed. We leave the mappings internally so that if the matching Spice is reinstalled, the previous settings will be retained.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_keyboard.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_keyboard.py
@@ -535,9 +535,12 @@ class Module:
                     home = os.path.expanduser("~")
                     gettext.bindtextdomain(name, f"{home}/.local/share/locale")
                     gettext.textdomain(name)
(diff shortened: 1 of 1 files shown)


fredcw:
@rcalixte Bug still occurs but this time on line 531. Copy a default xlet to local directory in order to customise it and then rename it, e.g. `my_menu@cinnamon.org`. Also someone may put a 3rd party xlet into /usr, you never know.

Maybe the best way to do this is to just check if file exists in local and if not then check if it exists in /usr?

rcalixte:
And here I thought I was an edge case. 😅 

> Maybe the best way to do this is to just check if file exists in local and if not then check if it exists in /usr?

Anything just needs to be thought out and tested. I'm a fan of simple exception handling. Too many conditional logic paths eventually get tricky to debug.

PR #12408, opened 2024-09-23
Convert the force quit dialog to a cinnamon dialog

JosephMcc:


anaximeno:
Hello @JosephMcc, does this PR fix the issue below as well?

Untitled
*fig - force quit dialog*

That text in the list view was not getting the proper text theme according to the dark theme (which should be white I suppose), also I'm not sure if the translation is implemented for that (if it is possible).

JosephMcc:
Yes. It would no longer be a GTK dialog so that but would no longer be relevant.

clefebvre:
> Hello @JosephMcc, does this PR fix the issue below as well?
> 
> Untitled _pic. - force quit dialog_
> 
> That text in the list view was not getting the proper text theme according to the dark theme (which should be white I suppose), also I'm not sure if the translation is implemented for that (if it is possible).

No, that's a different dialog. That's one of the session quit dialog.

Alstruit:
It appears that the force quit dialog does not read the keyboard inputs (Tab,Left/Right keys). This may cause an issue if the application hides the mouse cursor.

JosephMcc:
An oversight on my part. Hope to PR a fix in the next day or two.

PR #12416, opened 2024-09-26
Respect `$GTK2_RC_FILES` when writing themes

RossBrunton:
`GTK2_RC_FILES` is an environment variable containing a colon seperated list of RC files that are used to theme GTK2 applications. This patch will write GTK2 theming to the first file in this list, if present, rather than ~/.gtkrc-2.0.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/CinnamonGtkSettings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/CinnamonGtkSettings.py
@@ -409,10 +409,10 @@ class PreviewWidget(SettingsWidget):
 class Gtk2ScrollbarSizeEditor:
     def __init__(self, ui_scale):
         rcpath = GLib.getenv("GTK2_RC_FILES")
-        if rcpath is None or rcpath == "":
-            self._path = os.path.join(GLib.get_home_dir(), ".gtkrc-2.0")
-        else:
+        if rcpath:
             self._path = rcpath.split(":")[0]
+        else:
+            self._path = os.path.join(GLib.get_home_dir(), ".gtkrc-2.0")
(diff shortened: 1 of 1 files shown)


files/usr/share/cinnamon/cinnamon-settings/bin/CinnamonGtkSettings.py
```diff
@@ -408,7 +408,11 @@ def on_overlay_scrollbars_changed(self, settings, key, data=None):
 
 class Gtk2ScrollbarSizeEditor:
     def __init__(self, ui_scale):
-        self._path = os.path.join(GLib.get_home_dir(), ".gtkrc-2.0")
+        rcpath = GLib.getenv("GTK2_RC_FILES")
+        if rcpath is None or rcpath == "":
```
rcalixte, line 412:
```suggestion
        if rcpath:
```

rcalixte, line 412:
```suggestion
        if not rcpath:
```

files/usr/share/cinnamon/cinnamon-settings/bin/CinnamonGtkSettings.py
```diff
@@ -408,7 +408,11 @@ def on_overlay_scrollbars_changed(self, settings, key, data=None):
...
+            self._path = os.path.join(GLib.get_home_dir(), ".gtkrc-2.0")
+        else:
+            self._path = rcpath.split(":")[0]
```
rcalixte, line 415:
```suggestion
        if rcpath:
            self._path = rcpath.split(":")[0]
        else:
            self._path = os.path.join(GLib.get_home_dir(), ".gtkrc-2.0")
```

PR #12430, opened 2024-10-04
Update classicSwitcher.js

trumenov:
Describe bug: When you click on the icon on pressed alt+tab - active became focused icon, but not you clicked.
This PR will fix this small bug.

### DIFF
--- a/js/ui/appSwitcher/classicSwitcher.js
+++ b/js/ui/appSwitcher/classicSwitcher.js
@@ -582,6 +582,7 @@ SwitcherList.prototype = {
     },
 
     _onItemClicked: function (index) {
+        this._itemEntered(index);
         this._itemActivated(index);
     },
 


mtwebster:
Can you describe the issue in more detail and steps to reproduce?

Currently:
- Alt-tab brings up the list
- Clicking on icon A makes that window A focused and closes the alt-tab switcher.
- Alt-tabbing brings up the list again, with icon A as the left-most icon, and the one next to it highlighted (because it will be the window made active if you release alt-tab immediately).

This seems correct to me. Is it possible when you alt-tab a second time your mouse pointer is (randomly) in a position that causes a different icon to be highlighted when the icons appear?

cleberzavadniak:
It seems to be reproducible this way:

1. open five or six windows
1. position the mouse pointer *at the center of the screen* -- or in an area you can predict the window switcher will appear
2. **without moving the pointer**, the window under it won't be selected! (it will if you move it, though)
3. pay attention to which window is actually selected: it'll be the second one
4. click on the window icon under the pointer (probably icon 3 or 4)
5. expected result: switch to the window whose icon was clicked; what happens: it switches to the second window (the selected one).

clefebvre:
I can reproduce this if I select an app in alt-tab with the mouse, then press Alt a few times to select another one without moving the mouse, then click the mouse.

The fix is good.

PR #12452, opened 2024-10-17
Convert the keyring unlock dialog to a clutter dialog

JosephMcc:


fredcw:
@JosephMcc Please excuse my ignorance but I'm curious about what the rationale is for moving from gtk dialogs to clutter dialogs. I thought that Clutter was in maintenance mode and has less compatibility with wayland, which it why GNOME it transitioning to gtk4 for it desktop UI? ChatGPT "informs" me:

>While Wayland support could be improved in Cinnamon over time, it would likely require moving away from Clutter to something like GTK 4 or another modern rendering library that supports Wayland natively.

Also, is [this](https://tree.taiga.io/project/linuxmint-linux-mint-21-1/us/11?kanban-status=2641776&kanban-swimlane=3950) an indication that @clefebvre planned to rewrite the cinnamon panel in gtk?

JosephMcc:
> I thought that Clutter was in maintenance mode and has less compatibility with wayland, which it why GNOME it transitioning to gtk4 for it desktop UI?

The standalone version of Clutter is pretty much dead. The versions that are internal to mutter and muffin certainly are not. GNOME is not moving it's desktop UI to Gtk4. It relies  very heavily on clutter for drawing all elements. Much more than Cinnamon does currently. This actually gives us a little more control over some UI elements ourselves instead of relying on dialogs that might be supplied by something else. And doing all of them keeps things consistent.

> ChatGPT "informs" me:

LOL

> Also, is [this](https://tree.taiga.io/project/linuxmint-linux-mint-21-1/us/11?kanban-status=2641776&kanban-swimlane=3950) an indication that @clefebvre planned to rewrite the cinnamon panel in gtk?

That's more a brainstorm than anything else. I've even experimented with it in the past. But that would put even more reliance on Gtk.

Secret-chest:
Wouldn't this make it no longer a window, so it blocks other apps, isn't moveable and so ob?

PR #12461, opened 2024-10-21
Add a clutter audio device selection dialog

JosephMcc:
Requires: https://github.com/linuxmint/cinnamon-settings-daemon/pull/401


clefebvre:
Easy test with d-feet.

Pass the following:  ['headphones','headset','microphone']

Issue #12478, opened 2024-11-02
Users and Groups app falsely claims that dots (.) are allowed in usernames


adrian-m-nguyen:
### Distribution

Linux Mint 22

### Package version

6.2.9

### Frequency

Always

### Bug description

Cinnamon's Users and Groups app falsely claims that dots (.) are allowed in usernames. They are not, or at least not by default.[[1]](https://forums.linuxmint.com/viewtopic.php?p=1535183#p1535183)[[2]](https://askubuntu.com/a/405668) Please fix Users and Groups so that it stops listing the dot as an accepted username character.

Screencap:
![screencap](https://github.com/user-attachments/assets/96df624b-0724-40ea-beaa-09309430b15e)

### Steps to reproduce

1. Click on the Mint menu icon. Type the word "use". Users and Groups should be highlighted. Hit the Enter key.  
4. Enter your password.  
5. Click "Add" on the bottom left of Users and Groups.  
6. View the misinformation.

### Expected behavior

Users and Groups should *not* list the dot as an allowed character in usernames.

### Additional information

If you try to add a new user with a dot in their username, nothing will happen. The system just won't create their account.

PR #12482, opened 2024-11-04
fix users and groups falsely claiming that dots are allowed in usernames

Joshiy13:
fix users and groups falsely claiming that dots are allowed in usernames

### DIFF
--- a/cinnamon.pot
+++ b/cinnamon.pot
@@ -2355,7 +2355,7 @@ msgid ""
 "The username must consist of only:\n"
 "    - lower case letters (a-z)\n"
 "    - numerals (0-9)\n"
-"    - '.', '-', and '_' characters"
+"    - '-' and '_' characters"
 msgstr ""
 
 #: files/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py:379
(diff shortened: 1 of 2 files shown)


Joshiy13:
https://github.com/linuxmint/cinnamon/issues/12478

fredcw:
The line you need to edit is [this one](https://github.com/linuxmint/cinnamon/blob/5b05d68226cd6c1b58709bc299e6946a83d9540a/files/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py#L376).

You then need to run `makepot` in the root directory to regenerate `Cinnamon.pot`. Actually, given that the only thing that will change in the `Cinnamon.pot` file when regenerated is the line you've already edited, you can probably get away with editing the .pot manually as you already have. Just make sure the string matches the one you change in `cinnamon-settings-users.py`

Joshiy13:
Done

Joshiy13:
Fixes https://github.com/linuxmint/cinnamon/issues/12478

clefebvre:
.pot and /po shouldn't be modified.

To properly fix this we need to forbid . chars in the regex.

The label itself isn't actually necessary since the entry/button give feedback on invalid characters as they are being typed.

Fixed on master with https://github.com/linuxmint/cinnamon/commit/02480c68f20e3ce5f5deedae7976c23fb8643304

PR #12486, opened 2024-11-06
Settings: Rounded sections

clefebvre:
Requires https://github.com/linuxmint/python3-xapp/pull/24

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_mouse.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_mouse.py
@@ -30,9 +30,10 @@ class Module:
 
             page = SettingsPage()
 
-            settings = page.add_section(_("General"))
+            settings = page.add_rounded_section(_("General"))
 
             switch = GSettingsSwitch(_("Left handed (mouse buttons inverted)"), "org.cinnamon.desktop.peripherals.mouse", "left-handed")
+            switch.set_tooltip_text(_("blah blah"))
             settings.add_row(switch)
 
             switch = GSettingsSwitch(_("Reverse scrolling direction"), "org.cinnamon.desktop.peripherals.mouse", "natural-scroll")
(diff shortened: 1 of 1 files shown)


clefebvre:
Here's a quick prototype to illustrate old vs new class structures and to troubleshoot with Glade:

[settings.tar.gz](https://github.com/user-attachments/files/17646188/settings.tar.gz)

To run it with Gtk Inspector:

```
gsettings set org.gtk.Settings.Debug enable-inspector-keybinding true
GTK_DEBUG=interactive python3 ./settings.py
```

clefebvre:
We can improve the default CSS.

clefebvre:
Here it is with no shadow.

image

clefebvre:
Here's an example in GNOME

image

Rounded section, shadows, subtitles and tooltip hint.

axel358:
For the Mint-Y style, I also agree that looks better with no shadow. I think subtitles should be preferred over tooltips when possible, as they expose the additional information/context right away to the user. It would also be nice to see a unified settings style between all OS components. xlets and system settings modules use the style shown above, but apps like the file manager use a completely different one

anaximeno:
Another point that might be worth looking out, is setting a maximum width for the sections, because on bigger screens the description of the option and their action button can be too far away from each other which could lead to more effort trying to understand and modify each setting option:

Captura de ecrã de 2024-11-29 06-50-30

GNOME already has a limit for the size of the sections in settings IIRC.

JosephMcc:
@anaximeno I'm pretty sure it was like that when we first implemented this style of settings. But people complained that it "didn't" expand to fill the window. I'm also not sure how many people maximize a settings window on a large screen. There is no reason to.

clefebvre:
Quick note on this: It was postponed to 22.2.

anaximeno:
> @anaximeno I'm pretty sure it was like that when we first implemented this style of settings. But people complained that it "didn't" expand to fill the window. 

I see.

> I'm also not sure how many people maximize a settings window on a large screen. There is no reason to.

I think it's not too uncommon, generally, you maximize a window if you want to focus on it and avoid whatever is in the background taking your attention, I can see this regularly being the case when you're looking for something in settings.

Issue #12393, opened 2024-09-15
Sound should display different icon when muted


benmccann:
### Distribution

Ubuntu 24.04.1 LTS

### Package version

Cinnamon 6.0.4

### Frequency

Always

### Bug description

When muting the sound it will show a muted symbol for a second:

![Screenshot from 2024-09-14 20-43-59](https://github.com/user-attachments/assets/145feb86-3201-4953-a10a-618710211c9e)

Then it will switch to a music symbol:

![Screenshot from 2024-09-14 20-43-47](https://github.com/user-attachments/assets/293de06d-1e6b-498c-b772-85291f710858)

### Steps to reproduce

Mute the sound

### Expected behavior

It should stay a muted symbol

### Additional information

When clicking the music icon it says "Chrome - Stopped"

ygerlach:
If you have music running (:musical_note:  icon shown) and then mute your system, it shows the muted icon for 3 seconds and then switches back. See this contant: https://github.com/linuxmint/cinnamon/blob/f32fbb04456aab1de7403e7954bf5a379a315bbb/files/usr/share/cinnamon/applets/sound%40cinnamon.org/applet.js#L24

So it looks to me, like this is a design decision. Is there a reason, why the current music player is more important than the mute icon?

benmccann:
Oh, I see. It seems that it is trying to display the current volume for three seconds when adjusting the volume and then switches to show the music player icon. This makes sense when adjusting the volume, but is a bit funny when the sound is muted

One of the things that's frustrating here is that even with the sound unmuted, my computer isn't actually playing any sound. I guess I must have some tab in Chrome which has a paused video or something, but it's super difficult to figure out which one that might be. It just says "Chrome - Stopped". After a lot of searching I found chrome://media-internals/, which helped me find one webpage with an ad containing a stopped video, but it shows two other stopped audio sources with not enough detail to figure out where they're coming from or what they are.

I find myself jumping on to zoom calls and not realizing the sound is muted, which frequently added an extra minute or two of frustration to the beginning of a call trying to figure out where the problem resides. Now I know it's one of the first things to check, but would still prefer to see it at a glance before hopping on a call instead of having to right-click the icon after something's gone wrong

There are a few potential changes that could be made here:
- the music note could only be shown if there's actually audio playing and not if there's stopped audio
- mute could be considered to be not the same as other volume changes such that the mute icon is always shown when muted
[...]

PR #12502, opened 2024-11-16
sound applet: player icon should not take priority over mute icon

benmccann:
closes https://github.com/linuxmint/cinnamon/issues/12393


claudiux (approved):

leigh123linux:
@claudiux  Why are you merging commits without discussing them with the team first!!

claudiux:
@leigh123linux I've tested these commits and they work well. It seems to me that the team had other problems to solve.

JosephMcc:
@claudiux but you don't even know if this was a change that was wanted without any discussion. That's the point. Maybe it was and this is the right solution. But typically that decision is something that is cleared first.

claudiux:
@JosephMcc When a PR solves a problem that has not been closed (cause: unplanned), it seems to me that this PR should be accepted.

PR #12508, opened 2024-11-18
[WIP] Add CinnamonBlurEffect.

mtwebster:
Test using ClutterEffect to apply blur.

sigma: spread/blurriness, how wide of a blur


pmanu93:
Hi,
Any updates on this? :)
By the way, if possible, don't forget to make it toggleable system-wide, so it can be disabled in case of performance issues, for example.

Milli189:
I would love to see this merged, I love blur and I would love to have it on Cinnamon too :)
It makes the desktop look more modern and beautiful

PR #12519, opened 2024-11-24
Add a clutter end-session dialog.

mtwebster:
Requires https://github.com/linuxmint/cinnamon-session/pull/177.

### DIFF
--- a/data/theme/cinnamon-sass/widgets/_dialogs.scss
+++ b/data/theme/cinnamon-sass/widgets/_dialogs.scss
@@ -27,6 +27,49 @@
   }
 }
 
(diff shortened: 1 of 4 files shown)


mtwebster:
image
image
image

Secret-chest:
What's the reason for migrating all these dialogues? Isn't Cinnamon a GTK desktop?

mtwebster:
Cinnamon is built on the Clutter library, not Gtk.

We prefer Gtk applications, and our tools use it, but Cinnamon itself does not, and the desktop will look more cohesive as system dialogs adopt that style.

PR #12549, opened 2024-12-03
Don't crash on start-up if polkit encounters a dbus error

SparkyBluefang:
From #12272

Don't crash if the agent doesn't start properly. If another agent happens to be running, this will throw a DBus exception. I figured popping a notification would be most appropriate, since the user may have another agent configured to auto-start.
```
(cinnamon:9337): Gjs-CRITICAL **: 13:47:05.571: JS ERROR: Polkit.Error: GDBus.Error:org.freedesktop.PolicyKit1.Error.Failed: An authentication agent already exists for the given subject
_init@/usr/share/cinnamon/js/ui/polkitAuthenticationAgent.js:332:22
AuthenticationAgent@/usr/share/cinnamon/js/ui/polkitAuthenticationAgent.js:323:10
init@/usr/share/cinnamon/js/ui/polkitAuthenticationAgent.js:392:17
start@/usr/share/cinnamon/js/ui/main.js:433:31
@<main>:1:47
```
This was my best guess at how to handle this. I wasn't sure if it belonged in `polkitAuthenticationAgent.js` or in `main.js`.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -89,7 +89,6 @@ Depends:
  metacity,
  nemo,
  network-manager-gnome [linux-any],
- policykit-1-gnome,
  python3,
  python3-dbus,
  python3-distro,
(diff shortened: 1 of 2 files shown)


clefebvre:
Do you have a quick way of reproducing this?

clefebvre:
Actually, I wonder if we should show a dialog or just log the issue...

This could be done on purpose by the user to force a GTK agent.

Issue #12354, opened 2024-08-15
cinnamon-settings : Can't update the list of extensions, applets, actions and themes


pzim-devdata:
### Distribution

Debian GNU/Linux trixie/sid

### Package version

6.0.4

### Graphics hardware in use

Intel Corporation TigerLake-LP GT2 [Iris Xe Graphics]

### Frequency

Always

### Bug description

when I run `cinnamon-settings themes` or `cinnamon-settings applets` or  `cinnamon-settings actions` or `cinnamon-settings extensions` when I click on the download/update tab nothing happen. 

### Steps to reproduce

```bash
neptune@Neptune:~$ cinnamon-settings themes
/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py:445: DeprecationWarning: Gtk.Window.set_wmclass is deprecated
  self.window.set_wmclass(wm_class, wm_class)
Loading Themes module
Downloading from https://cinnamon-spices.linuxmint.com/json/themes.json?time=1723705112
Traceback (most recent call last):
  File "/usr/share/cinnamon/cinnamon-settings/bin/Spices.py", line 842, in _ui_error_message
    markup += _("\n\nDetails:  %s") % (str(detail))
              ^
UnboundLocalError: cannot access local variable '_' where it is not associated with a value
Downloading from https://cinnamon-spices.linuxmint.com/json/themes.json?time=1723705514
Traceback (most recent call last):
  File "/usr/share/cinnamon/cinnamon-settings/bin/Spices.py", line 842, in _ui_error_message
    markup += _("\n\nDetails:  %s") % (str(detail))
              ^
UnboundLocalError: cannot access local variable '_' where it is not associated with a value
Downloading from https://cinnamon-spices.linuxmint.com/json/themes.json?time=1723705523
Traceback (most recent call last):
  File "/usr/share/cinnamon/cinnamon-settings/bin/Spices.py", line 842, in _ui_error_message
    markup += _("\n\nDetails:  %s") % (str(detail))
              ^
UnboundLocalError: cannot access local variable '_' where it is not associated with a value




neptune@Neptune:~$ cinnamon-settings applets
/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py:445: DeprecationWarning: Gtk.Window.set_wmclass is deprecated
  self.window.set_wmclass(wm_class, wm_class)
Loading Applets module
Downloading from https://cinnamon-spices.linuxmint.com/json/applets.json?time=1723706872
[...]

PR #12554, opened 2024-12-04
Fix crash when attempting to report an error in the UI with details

zahlman:
Ref #12354.

Assignment to `_` was almost certainly unintended in the line `_ = dialog.run()`, and has been removed.

As is, the code forced `_` to be recognized as a local (see https://stackoverflow.com/questions/370357); in the case where a non-`None` value is passed, this would mean that `_` is used before assignment.

Context implies that `_` is supposed to be a globally defined function, but further implies that the result from `dialog.run` *should not* replace that value.

Most likely, the original code was assigning to `_` for debugging purposes (i.e. so that the local value could be examined in a debugger), and the name collision was unintended.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
@@ -846,7 +846,7 @@ class Spice_Harvester(GObject.Object):
         esc = html.escape(markup)
         dialog.set_markup(esc)
         dialog.show_all()
-        _ = dialog.run()
+        dialog.run()
         dialog.destroy()
 
     def errorMessage(self, msg, detail=None):


rcalixte:
> Most likely, the original code was assigning to `_` for debugging purposes (i.e. so that the local value could be examined in a debugger), and the name collision was unintended.

This was an intentional reason that unfortunately escapes me at the moment. I remember debugging this issue when working on the code for this feature about a year ago. The single underscore would probably best be served to be replaced by the double underscore if at all. Considering the local scope of the assignment, it does not interact or interfere with the global assignment at all.

zahlman:
Assigning to `__` or something else should work fine but a proper investigation is needed to justify assigning the value at all. In any normal circumstance, such an assignment could not possibly have an effect beyond the lifetime of the function.

To clarify (I think you understood, but I got a bit confused by your response), while the assignment is local and would not impact on global scope, it does *prevent* global *lookup* of `_` previously (at line 845 in the current version; line 842 in the installed version where the problem was detected) in the function (where it *needs* to be looked up globally, because it's trying to find what appears to be a text localization function not defined locally).

Issue #10124, opened 2021-05-31
[Feature Request] Notification badges on dock app icons


Cantello:
In Cinnamon Desktop, apps that usually work with notification badges for unobtrusive indication of e.g. unread messages (see screenshot below) do not show these badges. Tried with Mailspring, Discord & Gmail via nativefier. With Ubuntu/GNOME they do work.

![image](https://user-images.githubusercontent.com/594700/120185441-7ef33500-c212-11eb-8d81-aefdc0deda8c.png)

My current configuration (some days ago) is as follows, in case it is important: https://termbin.com/qt8s

It would be great if these indicators could be implemented as in my regular workflow, notifications tend to be dismissed whereas icon badges stay visible.

Belaial:
Been searching far and wide for a solution to this but can only find this feature request, hopefully this is something that can be implemented in the future. I think this will become more important now when apps like Teams have moved to a PWA (Progressive Web App) which does not have it's own application icon tracking notifications, so today if you are using Teams PWA you need to open it up now and then to see if you have unread messages, notifications are displayed in the regular Cinnamon notifications but since everything is shown there the Teams PWA notifications are not clearly visible like they would be if the PWA icon would display a counter for unread messages.

(Just using Teams as an example, this would also be very handy for other applications)

Secret-chest:
Hope it's toggleable.

Mahi:
Literally the only reason I switched away from Cinnamon. Windows has this, Mac has this, Gnome has this, Android has this, even my car has this! It's crazy to me that Cinnamon doesn't have it yet.

Kaz-49:
Bump

eisbehr:
This is a feature common across all the big contemporary desktop environments (across operating systems), applications simply assume this feature exists for notifying the user about changes. Implementing it should have elevated priority.

LunarWatcher:
Windows doesn't show the full count IIRC. If there's multiple windows, it shows a short of overlap-effect. Windows 11 apparently has a dot too, according to a quick Google search, but it isn't used to reflect the count. It just shows that there are multiple windows, not many, from what I can tell. I've never used Win11 though (and I have no desire to), so might be wrong. But there is only one dot for three windows in an image I found on google, so it seems unlikely that it's used to count

Secret-chest:
Another option would be a 3-dots icon in the bottom-right, or just moving the count there and styling it differently

fredcw:
[...]

PR #12561, opened 2024-12-10
Add grouped window list notification count functionality

romario-07:
This feature allows users to see a notification badge on the grouped window list applet:
- current badge UI element was reused
- builtin message tray is used to be notified if an app got a notification
- "Show window count numbers" setting was changed to comobox
- extracted badge logic into a method where it's decided if the badge is shown, when is shown and what number to show

This is how settings combobox looks:
count_numbers_setting

And this indicates that I have a notification (because in settings I have "Show count numbers" = "Notification"):
image

Once I select "Show count numbers" = "Window" in settings combobox, it behaves as previous implementation, showing window count:
image

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -19,7 +19,8 @@ const {
     MAX_BUTTON_WIDTH,
     BUTTON_BOX_ANIMATION_TIME,
     RESERVE_KEYS,
-    TitleDisplay
+    TitleDisplay,
+    CountNumbers
 } = require('./constants');
 
 const _reLetterRtl = new RegExp("\\p{Script=Hebrew}|\\p{Script=Arabic}", "u");
@@ -70,6 +71,7 @@ class AppGroup {
             appInfo: params.app.get_app_info(),
             metaWindows: params.metaWindows || [],
             windowCount: params.metaWindows ? params.metaWindows.length : 0,
+            notificationCount: params.notificationCount ?? 0,
             lastFocused: params.metaWindow || null,
             isFavoriteApp: !params.metaWindow ? true : params.isFavoriteApp === true,
             autoStartIndex: this.state.autoStartApps.findIndex( app => app.id === params.appId),
@@ -1071,23 +1073,49 @@ class AppGroup {
         this.checkFocusStyle();
     }
 
+    notificationReceived(notification) {
+        if (this.groupState.willUnmount) return;
+
+        this.groupState.set({ notificationCount: this.groupState.notificationCount + 1 });
+        this.showBadge();
+    }
+
+    resetNotificationCount() {
+        if (this.groupState.willUnmount) return;
+    
+        this.groupState.set({ notificationCount : 0 });
+        this.showBadge();
+    }
+
     calcWindowNumber() {
         if (this.groupState.willUnmount) return;
 
-        const windowCount = this.groupState.metaWindows ? this.groupState.metaWindows.length : 0;
-        this.numberLabel.text = windowCount.toString();
+        this.groupState.set({windowCount: this.groupState.metaWindows ? this.groupState.metaWindows.length : 0});
+        this.showBadge();
+    }
 
-        this.groupState.set({windowCount});
-
-        if (this.state.settings.numDisplay) {
-            if (windowCount <= 1) {
-                this.badge.hide();
-            } else {
-                this.badge.show();
-
-            }
-        } else {
+    showBadge(){
+        if(this.state.settings.numDisplay == CountNumbers.None){
             this.badge.hide();
+            return;
+        }
+
+        let numberToShow = 0;
+        let minNumberToShow = 0;
+
+        if (this.state.settings.numDisplay == CountNumbers.Notification) {
+            numberToShow = this.groupState.notificationCount;
+        } else {
+            numberToShow = this.groupState.windowCount;
+            minNumberToShow = 1;
+        }
+
+        this.numberLabel.text = numberToShow.toString();
+
+        if (numberToShow <= minNumberToShow) {
+            this.badge.hide();
+        } else {
+            this.badge.show();
         }
     }
 
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -1083,7 +1083,7 @@ class AppGroup {
     resetNotificationCount() {
         if (this.groupState.willUnmount) return;
     
-        this.groupState.set({ notificationCount : 0 });
+        this.groupState.set({ notificationCount: 0 });
         this.showBadge();
     }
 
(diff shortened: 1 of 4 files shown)


fredcw:
Seems to work ok.

Reno-Sifana:
@romario-07 Hmm, I think you should add another option, like add Windows & Notifications grouped window list options (It's very useful when open multiple window e.g. Nemo File Manager and you have unread messages (WhatsApp, Telegram, Thunderbird and so on) at the same time), and turned on by default after installing the Linux Mint at these settings. 
(it's In My Opinion). 
You're though?

romario-07:
> @romario-07 Hmm, I think you should add another option, like add Windows & Notifications grouped window list options (It's very useful when open multiple window e.g. Nemo File Manager and you have unread messages (WhatsApp, Telegram, Thunderbird and so on) at the same time), and turned on by default after installing the Linux Mint at these settings. (it's just In My Opinion). You're though?

For showing both notification and window counts, I was thinking to do it the following way:
- badge icon background color to be changed to red
- bottom blue line which indicates that the app is active to be devided in two if there are 2 windows open and divided in three if there are 2+ windows open
375997442-99318257-38d8-4de3-82c2-92b13695a3e6


I did explain it a bit more in this comment: https://github.com/linuxmint/cinnamon/issues/10124#issuecomment-2408771142

I didn't make it part of this pr (the second part of the feature) because it was a bit challenging how to position correctly the dividers (based on left, right, top, bottom panel position) and to make sure it's working fine on different screen resolutions, different icon sizes, font size, screen scaling etc

On my personal day to day machine, I use the version from the image above (with the bottom blue line divided to hint if there are multiple windows open), but I can confirm works just for my specific machine settings.

Reno-Sifana:
> > @romario-07 Hmm, I think you should add another option, like add Windows & Notifications grouped window list options (It's very useful when open multiple window e.g. Nemo File Manager and you have unread messages (WhatsApp, Telegram, Thunderbird and so on) at the same time), and turned on by default after installing the Linux Mint at these settings. (it's just In My Opinion). You're though?
> 
> For showing both notification and window counts, I was thinking to do it the following way:
> 
>     * badge icon background color to be changed to red
> 
>     * bottom blue line which indicates that the app is active to be devided in two if there are 2 windows open and divided in three if there are 2+ windows open
>       375997442-99318257-38d8-4de3-82c2-92b13695a3e6
> 
> 
> I did explain it a bit more in this comment: [#10124 (comment)](https://github.com/linuxmint/cinnamon/issues/10124#issuecomment-2408771142)
> 
> I didn't make it part of this pr (the second part of the feature) because it was a bit challenging how to position correctly the dividers (based on left, right, top, bottom panel position) and to make sure it's working fine on different screen resolutions, different icon sizes, font size, screen scaling etc
> 
> On my personal day to day machine, I use the version from the image above (with the bottom blue line divided to hint if there are multiple windows open), but I can confirm works just for my specific machine settings.

Ah, nice idea. It's reminding behavior from Windows 10 and later. And it's up to you when you should decide want to make PR (Pull Request) now or later.
thanks.

fredcw:
There's an issue: if you change the current desktop theme while a notification badge is showing, it disappears.

Also, I don't think that this should be merged in its current form as asking users to choose between window count or notification badges is going to be unwanted choice for many people because they'd rather have both.

fredcw:
why not have a different color badge for each and have the size adjust automatically to panel size:

24px panel:
Screenshot from 2024-12-11 22-22-33
36px:
Screenshot from 2024-12-11 22-22-18
48px panel:
Screenshot from 2024-12-11 22-22-58

fredcw:
@Secret-chest You mean if you have multiple instances of an app that has notifications? All instances get the same notification indicator and focusing any of the app's instances removes the notification for all instances. Even the notifications applet doesn't distinguish between different instances of an app and think this would be difficult to change as they would all have the same app id.

fredcw:
@romario-07 This is what I've come up with: https://github.com/linuxmint/cinnamon/compare/master...fredcw:cinnamon:gwlnotifications

Do you want to create a new PR for this or shall I do it? Although, we may already too late for the 22.1 merge window.

romario-07:
> @romario-07 This is what I've come up with: [master...fredcw:cinnamon:gwlnotifications](https://github.com/linuxmint/cinnamon/compare/master...fredcw:cinnamon:gwlnotifications)
> 
> Do you want to create a new PR for this or shall I do it? Although, we may already too late for the 22.1 merge window.
> 
> Screenshot from 2024-12-12 07-06-41

You can do it. The reason why I didn't add an extra badge on the right is because it  was debated here https://github.com/linuxmint/cinnamon/issues/10124 about what happens when the icons are small (https://github.com/linuxmint/cinnamon/issues/10124#issuecomment-1641423700) and there is not enough place to put both. Also people said that it's not a good UI/UX to have two badges.

Other distros/OSes, will mainly use a bottom ui element to indicate there is a window open (like the blue bottom line in Cinnamon) and there should be a non intrusive way of indicating there are multiple windows open (some are adding dots under the icon to indicate if there are multiple window opened, some are adding a stack effect using shadows).

For me, knowing that I have multiple app windows open is relevant, but how many, is not that important. Switching from Windows 11 to Cinnamon, this version (from the image below) makes me happy and I don't miss any information (bottom line is divided in two if there are two windows, or divided in 3 if 2+ windows):
image

But the version from the image above is not production ready, and it works just for my specific system settings.

My plan was to have this feature broken in two:
1. first stage to have the same ui elements reused, but to have the notifications count information available and reuse current ui elements to display the information
2. find a good UI/UX way of showing both counters: open windows count and notifications count

fredcw:
@romario-07 How does the broken indicator line work with themes that don't have this line (e.g. Mint-X)? Do you override the theme?

Reno-Sifana:
> @romario-07 How does the broken indicator line work with themes that don't have this line (e.g. Mint-X)? Do you override the theme?
> 
> @Reno-Sifana The problem is that we need something that's going to work with all themes.

Ok. Noted.

romario-07:
> @romario-07 How does the broken indicator line work with themes that don't have this line (e.g. Mint-X)? Do you override the theme?
> 
> @Reno-Sifana The problem is that we need something that's going to work with all themes.

It won't work in that case (it will look strange).

rob-stewart:
I'm late to the game, but I wanted to suggest two things:
- Instead of "Show count numbers", make it _Show counts_ since _counts_ already implies numbers.
- To show more than one application instance open, change the solid line to one ending with an ellipsis: _____ . . . (I can't do it justice in text, but rather than the three segments, which is effectively a dashed/broken line, put an ellipsis on the right end and fill the rest with a solid line.) People are used to seeing an ellipsis and inferring that there was more.

fredcw:
@rob-stewart the line under the icon doesn't work with themes that don't have this line. Also, see https://github.com/linuxmint/cinnamon/pull/12569

rob-stewart:
> @rob-stewart the line under the icon doesn't work with themes that don't have this line. Also, see #12569

I had seen #12569 already, but I didn’t suppose that meant lines were out altogether. Why not provide an option to choose lines vs. badges? For themes that don’t support lines, users can choose badges. For those that support lines, users can choose whichever they prefer.

ImAwsumm (approved):

shankm:
@romario-07 why was this PR closed?

fredcw:
@shankm A new PR has been opened.

PR #12569, opened 2024-12-12
GWL: add notification badges to panel icons

fredcw:
Continuing on from the great work by @romario-07 in #12561 this adds a different color notification badge on the right of the icon and both badge sizes are adjusted automatically to the size of the icon. The "Show window count numbers" config option is removed as it seems fairly pointless.

@clefebvre @mtwebster Is this too late for the beta?

### DIFF
--- a/data/theme/cinnamon-sass/_colors.scss
+++ b/data/theme/cinnamon-sass/_colors.scss
@@ -20,6 +20,7 @@ $warning_bg_color: #cd9309;
 
 $accent_color: #78aeed;
 $accent_bg_color: #3584e4;
+$accent_bg_color_alternative: #d035e4;
 
 $large_icon_color: transparentize($fg_color, 0.6);
 $light_text_color: mix($fg_color, $bg_color, 65%);
--- a/data/theme/cinnamon-sass/widgets/_windowlist.scss
+++ b/data/theme/cinnamon-sass/widgets/_windowlist.scss
@@ -73,14 +73,24 @@
 
   &-button-label { padding-left: 4px;}
 
-  &-number-label {
-    font-size: 0.8em;
+  &-windows-badge {
+    border-radius: 9999px;
+    color: $fg_color;
+    background-color: $accent_bg_color;
+  }
+
+  &-windows-badge-label {
     z-index: 99;
   }
 
-  &-badge {
+  &-notifications-badge {
     border-radius: 9999px;
-    background-color: $bg_color;
+    color: $fg_color;
+    background-color: $accent_bg_color_alternative;
+  }
+
+  &-notifications-badge-label {
+    z-index: 99;
   }
 }
 
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -70,6 +70,7 @@ class AppGroup {
             appInfo: params.app.get_app_info(),
             metaWindows: params.metaWindows || [],
             windowCount: params.metaWindows ? params.metaWindows.length : 0,
+            notificationCount: params.notificationCount ?? 0,
             lastFocused: params.metaWindow || null,
             isFavoriteApp: !params.metaWindow ? true : params.isFavoriteApp === true,
             autoStartIndex: this.state.autoStartApps.findIndex( app => app.id === params.appId),
@@ -119,31 +120,50 @@ class AppGroup {
         });
         this.actor.add_child(this.progressOverlay);
 
-        // Create the app button icon, number label, and text label for titleDisplay
+        // Create the app button icon, window count and notification badges, and text label for titleDisplay
         this.iconBox = new Cinnamon.Slicer({name: 'appMenuIcon'});
         this.actor.add_child(this.iconBox);
         this.setActorAttributes(null, params.metaWindow);
 
-        this.badge = new St.BoxLayout({
-            style_class: 'grouped-window-list-badge',
+        this.windowsBadge = new St.BoxLayout({
+            style_class: 'grouped-window-list-windows-badge',
             important: true,
-            x_align: St.Align.START,
+            x_align: St.Align.MIDDLE,
             y_align: St.Align.MIDDLE,
             show_on_set_parent: false,
         });
-        this.numberLabel = new St.Label({
-            style_class: 'grouped-window-list-number-label',
+        this.windowsBadgeLabel = new St.Label({
+            style_class: 'grouped-window-list-windows-badge-label',
             important: true,
-            text: '',
-            anchor_x: -3 * global.ui_scale,
+            text: ''
         });
-        this.numberLabel.clutter_text.ellipsize = false;
-        this.badge.add(this.numberLabel, {
+        this.windowsBadgeLabel.clutter_text.ellipsize = false;
+        this.windowsBadge.add(this.windowsBadgeLabel, {
             x_align: St.Align.START,
             y_align: St.Align.START,
         });
-        this.actor.add_child(this.badge);
-        this.badge.set_text_direction(St.TextDirection.LTR);
+        this.actor.add_child(this.windowsBadge);
+        this.windowsBadge.set_text_direction(St.TextDirection.LTR);
+
+        this.notificationsBadge = new St.BoxLayout({
+            style_class: 'grouped-window-list-notifications-badge',
+            important: true,
+            x_align: St.Align.MIDDLE,
+            y_align: St.Align.MIDDLE,
+            show_on_set_parent: false,
+        });
+        this.notificationsBadgeLabel = new St.Label({
+            style_class: 'grouped-window-list-notifications-badge-label',
+            important: true,
+            text: ''
+        });
+        this.notificationsBadgeLabel.clutter_text.ellipsize = false;
+        this.notificationsBadge.add(this.notificationsBadgeLabel, {
+            x_align: St.Align.START,
+            y_align: St.Align.START,
+        });
+        this.actor.add_child(this.notificationsBadge);
+        this.notificationsBadge.set_text_direction(St.TextDirection.LTR);
 
         this.label = new St.Label({
             style_class: 'grouped-window-list-button-label',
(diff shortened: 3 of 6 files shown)


clefebvre:
Yes, new features don't go in past BETA.

anaximeno:
I don't know if I like the two numbers side by side (it excludes a big portion of the top of the icon and presents two conflicting pieces of information side by side, requiring more effort than necessary to interpret the values since they are very close, despite having different colors). We could try to think of a better way to present that information until the next merge period for the next release.

Secret-chest:
> The "Show window count numbers" config option is removed as it seems fairly pointless.

Why? I'm not using a phone OS

fredcw:
@anaximeno How about in opposite corners.

Screenshot from 2025-04-27 16-25-31

I can't think of anything else that will work with all themes except badges.

anaximeno:
> @anaximeno How about in opposite corners.
> 
> Screenshot from 2025-04-27 16-25-31

It looks better this way you did it now in my opinion. Good choice.

> I can't think of anything else that will work with all themes except badges.

Yeah, the options are limited thanks to that.

leigh123linux:
Is there an option to disable it completely?

fredcw:
@leigh123linux Not atm. Perhaps I should add that?

leigh123linux:
> @leigh123linux Not atm. Perhaps I should add that?

It would match the option to disable window count.

image

fredcw:
@leigh123linux Ah, I removed that as well. :P

fredcw:
@leigh123linux I've added the option to hide notification badges. Do you think the option to show window number should be added back? I don't have a strong opinion on this myself.

fredcw:
@clefebvre Why is there a thumbsdown on this PR by a member of your github team? Are my contributions no longer welcome?

Nekkowe:
> @leigh123linux I've added the option to hide notification badges. Do you think the option to show window number should be added back? I don't have a strong opinion on this myself.

I'd say so...! The more options for customization, the better :sparkles:   
There's no single one-size solution that'll fit everyone.  

Initially, you had both at the top, then you decided to move the window count to the bottom left and the notification count to the top right - I personally might prefer the window count in the top left and the notification count in the bottom right, while yet another person might prefer to have only the window count, only the badge number, or neither number showing up at all.

For my two cents, ideally both numbers should be optional, and the user should have a choice of which corner belongs to which number.  

How would you feel about having a dropdown for each corner in the settings, with the choices of "Window Count", "Notification Count", and "None"? That'd leave room for more corner options in the future, too.

fredcw:
Added back show window count config option.

clefebvre:
> @clefebvre Why is there a thumbsdown on this PR by a member of your github team? Are my contributions no longer welcome?

Of course they are. I don't know why @leigh123linux thumbed it down, but even if someone in the team or even if most of the team didn't like a particular PR, it wouldn't mean your contributions are not appreciated.

Nekkowe:
This is wonderful! Thank you for your work on this feature, I've been excited to see it implemented for a long time now (and so have others, judging by issue #10124)

> Yes, new features don't go in past BETA.

Is this the only thing this PR is waiting on, now? I'll confess I'm not sure how all the release timing stuff works, exactly... is it possible to merge now, or if not, around when could it be merged?

clefebvre:
> Is this the only thing this PR is waiting on, now? I'll confess I'm not sure how all the release timing stuff works, exactly... is it possible to merge now, or if not, around when could it be merged?

Yes, we're hoping to merge this.

clefebvre:
@fredcw I think this is great and it's working well.

Here's my feedback on what still requires change.

I don't like the use of `updateBadgesTextSize()`. The label was given a font-size in the SCSS and that worked well imo. Injecting CSS at runtime takes control away from the theme.

In the default theme, I'd like to keep the badges small, with their original color and original position (top-left). These numbers are only there as hints, they're not supposed to catch your attention.

image

There's plenty of space top-right for the notification badge if it's the same size. It has a noticeable color, and it complements the notification applet and the notification itself, so same thing here, there's no need for this number to be big.

It's a pity we store the notifications in an array and not just the number since that's all we need, but I understand it's more reliable that way. If we were to concurrently increase/decrease, we could run into bugs.

clefebvre:
Small detail, I noticed we're doing maths on the same childBox actor to allocate multiple things (icon and the 2 badges). We're modifying it as we go along so the math is relative to the changes made before.. it's a bit messy :)

It might cleaner to declare child boxes for the two badges and let them make their own maths from scratch, they'd look more similar that way in the code.

It doesn't really matter but I saw it so I thought I'd mention it.

fredcw:
@clefebvre In notificationDaemon.js, I've found that the only reliable way to determine the source app of a notification from a flatpak app is the desktop-entry hint. While this is set correctly for most flatpak apps, some set it incorrectly, hence the list of "exceptions". While I tested quite a few, I expect there to be some that are not listed. My thinking is that some of these will be found during the beta and can be added to the list. An alternative is to try to guess the app from the appName using a subString search of installed apps' names and ids. I can implement that if you think it's better.

clefebvre:
> @clefebvre In notificationDaemon.js, I've found that the only reliable way to determine the source app of a notification from a flatpak app is the desktop-entry hint. While this is set correctly for most flatpak apps, some set it incorrectly, hence the list of "exceptions". While I tested quite a few, I expect there to be some that are not listed. My thinking is that some of these will be found during the beta and can be added to the list. An alternative is to try to guess the app from the appName using a subString search of installed apps' names and ids. I can implement that if you think it's better.

Let's go with that imo. As you said we can always improve it when we get feedback.

clefebvre:
Almost there but there's something funky with the size now..

image

fredcw:
@clefebvre I fixed the badge size issue. Fixed in #13209 as well.

>It's a pity we store the notifications in an array and not just the number since that's all we need, but I understand it's more reliable that way. If we were to concurrently increase/decrease, we could run into bugs.

My main concern there was that the applet should be able to destroy any remaining notifications should it be removed from the panel whilst also being the last remaining applet handing notifications.

clefebvre:
It's more visible if you switch to Mint-X. Since we're using new CSS class names, the Mint-X theme doesn't interfere with testing, its light color just makes it easier to see the badge.

image

It should be round and placed in the corner of the icon.

I'll try to fix it this weekend. This is going in no matter what, I just want to nail the look and feel.

clefebvre:
This is master in comparison:

image

Not perfectly round when the number has 2 digits but pretty good otherwise. Well positioned, round with 1 digit.

The positioning and align code is pretty messy on master already. This should be much easier to do than it is :)

If I can't get this done this week-end I'll merge as is and we'll consider this a bug to be fixed during the ALPHA/BETA phases.

fredcw:
@clefebvre I think this is a bit better.

Here's cinnamon theme with ubuntu regular 10
Screenshot from 2025-11-28 18-35-33

Mint-X with ubuntu regular 13
Screenshot from 2025-11-28 18-41-06

I think some themes will need adjusting though

here's Mint-Y with ubuntu regular 13 font
Screenshot from 2025-11-28 18-45-59

JosephMcc:
Quick comment. Don't use that terrible color for the notification badge. You already have warning and error colors. Use those instead of just making up some color that doesn't fit at all.

Feel like this could have been simpler. Putting 2 badges with numbers is a bit much. Especially if you use a smaller panel height. There should only be one. You could show multiple windows another way. 
Screenshot from 2025-11-28 12-46-29

JosephMcc:
The font size problem is also a pretty easy fix. In most cases you want to use relative font sizes to respect user settings. But I don't think this is one of them. If a font size of  7 or 8 just work, then just use that in the theme. Using em here is just asking for a problem.

fredcw:
>If a font size of 7 or 8 just work, then just use that in the theme

Thanks, I didn't think of that xD.

>Don't use that terrible color for the notification badge.

I'm not bothered what the colour is, I only picked that colour because I thought it needed to stand out and that red looked too serious. But any colour is fine. I guess most themes will probably change it anyway.

>You could show multiple windows another way.

This is the problem though. We would need something that would work with all existing themes. Lines or boxes would conflict with themes that already use lines (Mint-Y) or boxes (Mint-X, Cinnamox) to show which apps have open windows, and I guess you'd have to shrink the icon to make room for it as well. Bear in mind that most apps won't have both badges at the same time, except maybe a browser, but notifications are normally fleeting anyway since the point of them is to get rid of them xD.

fredcw:
With fixed font size 10px, it now looks like this:

Mint-Y, panel size 31:
Screenshot from 2025-11-29 09-17-09

panel size 41:
Screenshot from 2025-11-29 09-14-25

Here's 8px font for comparison:
Screenshot from 2025-11-29 09-30-05

Here's 8px and 13 px notification, with window count badge moved closer to icon and using same red as iOS (#e74b37) for notification:
Screenshot from 2025-11-29 09-54-36

fredcw:
here's what I have now:

9px and 13px and iOS (#e74b37) for notification colour
Screenshot from 2025-11-29 13-41-25

Screenshot from 2025-11-29 14-25-38



----
Edit: Made sure buttons are always round:
Screenshot from 2025-11-30 05-38-25

fredcw:
@clefebvre Some themes still override the cinnamon theme font size:

Mint-Y theme, panel size 41, font: ubuntu regular 13:
Screenshot from 2025-11-30 18-34-27

~I can fix this in Mint-Y if you like. The only alternative is to ignore the theme and set the font size in the applet.~
Nevermind. We can just use the old class names, they already work and are used by 3rd party applets.

Added PRs for window count badge fixed font size: https://github.com/linuxmint/mint-themes/pull/524, https://github.com/linuxmint/mint-themes/pull/525, https://github.com/linuxmint/mint-l-theme/pull/20

Nekkowe:
> The only alternative is to ignore the theme and set the font size in the applet.

I like the sound of that. It'd be nice to not depend on the theme itself for the size of the badge. I think this is something that warrants the more granular control you'd get from an applet-specific setting.

clefebvre:
Thanks @fredcw 

It works really well. I just want to make the sizes the same, I have them in green here to see them better:

image

I'm going to merge this and tune it on master.

Nekkowe:
Thank you all very much for your work on this! Me and several friends I've helped dual boot Mint have been waiting for this eagerly :tada:

catscarlet:
I have a question.
Is there a way to disable it on certain app?
I like this feature but when using Nemo, the notification number is so confusing.
Is there a blacklist or something I can have to disable it on certain app like Nemo?

screenshot_on_b85m_by_flameshot_at_2026-01-28_20-40-37

fredcw:
@catscarlet There's no per app notification settings atm.

catscarlet:
> @catscarlet There's no per app notification settings atm.

OK. Understood.

JosephMcc:
I do have a branch I worked on awhile back that I resurrected recently that would allow per app notification settings. It will take some time to get done though.

fredcw:
@JosephMcc Do you plan on finishing it? If not, I might have a look at it myself. I thought maybe somewhere (an applet?) we could have a "Do not disturb" button as well to temporarily stop notifications from showing onscreen while still showing in the notifications applet?

JosephMcc:
I'd like to but it will take some time. It's a complete redo of notifications and still has a ways to go.

https://github.com/user-attachments/assets/47d6257c-1796-4829-b3ef-80aeb4be3b5d


https://github.com/user-attachments/assets/bde62ba4-4445-48d4-8ee7-4103721d672a


https://github.com/user-attachments/assets/ae61e918-6181-421f-9151-c1a12b563bfa

archisman-panigrahi:
The icons in the panel still doesn't show custom launcher counts (e.g. number of unread messages) from apps like telegram. For details, see https://github.com/orgs/linuxmint/discussions/1402.

PR #12576, opened 2024-12-15
panel: add padding for optimal icon sizes

okaestne:
When using the 'scale to panel size optimally' panel option, ensure that
there are some pixels of padding. Otherwise, the icons will fill the
whole height (or width for vertical panels) and thus may also flow over
the panel's border.

The padding has been set to 4px for each side, so that the sizes don't
change when using the default settings. Also, symbolic icons will be
addtionally shrunk to 87.5% to visually match the colored icons (which
also matches the existing effective default proportions (32px colored,
28px symbolic.)
By doing so, merge the two very similar size clamping functions.

With these changes, one can set the colored icons to 'scale optimally'
and crank up the symbolic icon size to max, to always have visually
matching sizes for both.

### DIFF
--- a/js/ui/panel.js
+++ b/js/ui/panel.js
@@ -43,6 +43,7 @@ const PANEL_PEEK_TIME = 1500;
 
 const EDIT_MODE_MIN_BOX_SIZE = 25;
 const VALID_ICON_SIZE_VALUES = [-1, 0, 16, 22, 24, 32, 48];
+const MIN_ICON_PADDING_PX = 4;
 
 /*** These are defaults for a new panel added */
 const DEFAULT_PANEL_VALUES = {"panels-autohide": "false",
@@ -58,7 +59,6 @@ const DEFAULT_SYMBOLIC_ICON_SIZE_VALUES = {"left":   28,
                                            "center": 28,
                                            "right":  28};
 const MIN_SYMBOLIC_SIZE_PX = 10;
-const MAX_SYMBOLIC_SIZE_PX = 50;
 
 const DEFAULT_TEXT_SIZE_VALUES = {"left":   0.0,
                                   "center": 0.0,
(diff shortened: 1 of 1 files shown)


okaestne:
The amount of the effective padding is debatable. I personally would increase it a little more. Also, what do you think about merging the icon size settings, i.e. always setting symbolic icon size to 0.875x colored icon size?

fredcw:
Issue: If you set your panel size to say 54 and then try to set the color icon size to 48, it pops it back down to 32. Same with symbolic icon size. This give users less control and makes some configurations impossible.

Some thoughts:

I think that the color icon size "scale to panel size optimally" option should be left as it is as users will just adjust the panel size so that it looks right to them anyway (i.e. add padding themselves). The real option that needs to change is "scale to panel exactly" as I doubt anyone thinks that full height icons with no padding at all looks good, so this option should add padding and the padding should be a percentage of the icon size rather than a fixed number of pixels, say 20% or something.

Also, I don't really see the need to have separate settings for left/centre/right zones. I doubt many people want different sized color icons or symbolic icons on different parts of the panel. So I think the zones option should be removed altogether as it's mainly just an inconvenience.

JosephMcc:
> Also, I don't really see the need to have separate settings for left/centre/right zones. I doubt many people want different sized color icons or symbolic icons on different parts of the panel. So I think the zones option should be removed altogether as it's mainly just an inconvenience.

You'd be completely wrong on that. I want all the applets on the right of the panel at a fairly small size which are mostly symbolic. The menu applet on the left of my panel also uses a symbolic icon and I don't want it that small.

Secret-chest:
Please no 5% increments. Just let it be set to any value. At least 0.5%.

fredcw:
@Secret-chest 
>Please no 5% increments. Just let it be set to any value. At least 0.5%.

Icon sizes would have to be rounded to the nearest integer anyway. So 0.5% wouldn't make any difference. And with a panel size of 40px, 74%, 75%, and 76% would all be rounded to 30px.

Secret-chest:
Then calculate the lowest increment based on the panel size? Or allow setting it in both % and px?

JosephMcc:
> Icon sizes would have to be rounded to the nearest integer anyway. So 0.5% wouldn't make any difference. And with a panel size of 40px, 74%, 75%, and 76% would all be rounded to 30px.

Well there are multiple issues here. One, color icons cannot be rounded to an arbitrary size without causing blurriness. So you might have a minimum padding value in there, but there is no reason for it to be user adjustable because it won't mean much once icons are adjusted to the needed sizes.

Also, the hard coded .0875 for symbolics is no good. That may work out on one icon theme but not the next. Some themes may design their icons to fill the entire canvas space while others will leaving spacing around the actual drawing. That makes the sizes appear different even when they aren't. So the thinking on that is faulty.

Secret-chest:
Please don't change the symbolic size!

1. The menu icon can be symbolic; it needs to be large like the apps.
2. I, and I'm sure many others, like large status icons.

Secret-chest:
> > Icon sizes would have to be rounded to the nearest integer anyway. So 0.5% wouldn't make any difference. And with a panel size of 40px, 74%, 75%, and 76% would all be rounded to 30px.
> 
> Well there are multiple issues here. One, color icons cannot be rounded to an arbitrary size without causing blurriness. So you might have a minimum padding value in there, but there is no reason for it to be user adjustable because it won't mean much once icons are adjusted to the needed sizes.
> 
> Also, the hard coded .0875 for symbolics is no good. That may work out on one icon theme but not the next. Some themes may design their icons to fill the entire canvas space while others will leaving spacing around the actual drawing. That makes the sizes appear different even when they aren't. So the thinking on that is faulty.
> 
> The best you can probably do here is take the panel size minus some constant value and then use that to determine your final icon size.

Well, that is what the other automatic setting does.

Secret-chest:
Also, when a value like this can so easily be made configurable (currently a constant, just add a gsetting for it) it really makes no sense for it not to be.

mtwebster:
I don't see an issue with how icons are handled currently, and I've read enough here to think this has the potential to raise more problems than it might fix.

PR #12584, opened 2024-12-23
modalDialog: Make sure we can use the on screen keyboard

JosephMcc:
We need to make sure the keyboard is stacked above our modal dialogs when they are open. Also restrict placement of the dialogs to the monitor work area. This ensures we don't end up with the dialog behind the keyboard in some situations.

Fixes: https://github.com/linuxmint/mint22.1-beta/issues/70

### DIFF
--- a/js/ui/modalDialog.js
+++ b/js/ui/modalDialog.js
@@ -84,6 +84,10 @@ var ModalDialog = GObject.registerClass({
         this._cinnamonReactive = params.cinnamonReactive;
         this._destroyOnClose = params.destroyOnClose;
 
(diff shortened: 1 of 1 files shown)


tutralex:
I downloaded and replaced the "modalDialog.js" file - it didn't work. I still can't use my TV using only the gyro mouse and the "Onboard" keyboard as before (Arch Linux).

tutralex:
I was able to disable the blackout of the entire desktop in modal dialogs and the control panel (task bar) is now accessible, but everything above it is not.

PR #12586, opened 2024-12-23
[Next] Redesign menu@cinnamon.org

JosephMcc:


Secret-chest:
How does it look like?

leigh123linux:
> How does it look like?

Try it instead of adding worthless noise!

clefebvre:
> How does it look like?

image

Secret-chest:
Is the default font being switched to Inter?

Initu-Castilhos:
It would be better to remove favorites in the sidebar and add favorites in categories such as whisker-menu and brisk-menu. Another suggestion: in the left sidebar remove the icons, so when hovering over the mouse pointer the icon appears with the user's image like CinnViistarkMenu, in categories hide the icons by default and in all applications the icons appear at 48px with the description below the icons same as Cinnamenu.

Qronikarz:
I really like the new action buttons (turn off, logout, lock) .
Will it be possible to disable the side menu and move the search to bottom section or will we need a new spice/applet for that?
Visualization made from above screenshot:
minimenu

JosephMcc:
Settings to hide the sidebar and bottom bar with the power buttons already there. No setting for moving the search bar and currently no plans to add one.

Secret-chest:
Looks nice, a suggestion is to have an option to hide names for the sidebar icons.

fredcw:
Here's the redesigned applet for testing purposes:
[menu_revamp.zip](https://github.com/user-attachments/files/18451523/menu_revamp.zip)
Extract to `~/.local/share/cinnamon/applets/`

contrarybaton60:
image
Looks quite nice! I do have some comments though.
image
The use of small symbolic icons in the session buttons (or what's called the bottom bar) kinda makes it hard to notice at a first glance, and an option to move it to the left is appreciated for some.

The large default icon sizes doesn't really make sense to me, I feel like it's only good idea if it's like that for a "Large icons" accessibility feature.

The emphasis of round corners make sense for Mint-Y and the default Cinnamon theme, though it looks out of place in any other (like Mint-X and L, yada yada)

An option to move the sidebar to the right Windows 7 style might be good to have, though there's already an applet for that, so I wouldn't complain too much.

Overall it's a nice redesign of a much neglected menu in my opinion.

JosephMcc:
Not really looking to add any additional options. If this is wanted and merged, that could be up for discussion.

Also not really worried about theming. The commit only add support for the default theme. Other themes would need to do some tweaking to match their own style. Just like any other new/updated thing.

Lurux:
May I suggest, as was said before, to move the left sidebar's content to their own categories ?

Rationale: this would make the menu simpler for newcomers, and less cluttered by default:
Categories on the left, favorite places and applications on the right.

Rough visualization based on the above screenshot:
| Old | Suggestion (rough estimate) |
|--------|--------|
| image | image |

This is similar the Whisker menu. I think it would make the menu simpler to understand and navigate.

---

Laying out the applications in a grid might be something to think about given the current trends and the fact that it gives the user a larger click area, however the addition of the app's description under their title makes the list style very appealing. Maybe a grid style could be used only and specifically for favorite applications, however I doubt this would be worth the hassle

jum690:
Hi. Would it be possible for the Favorites menu to be moved to the right and have a grid type display like the Mate menu? (configured by user; not requesting as a default)
MintMATEdesktop1-3791411977

clefebvre:
Continued here https://github.com/linuxmint/cinnamon/pull/13124

PR #12590, opened 2024-12-25
modalDialog.js: Add some support for old themes

fredcw:
@clefebvre @JosephMcc 

The new cinnamon modal dialogs break almost all existing cinnamon themes apart from cinnamon and Mint-Y, most of which will likely never be updated rendering most cinnamon themes effectively obsolete. To mitigate this, albeit somewhat imperfectly, some small support for old themes could be added. This could give more time for theme authors to update their themes (including Mint-L and Mint-X) and could be removed at a later time.

Here is a representative selection of themes showing how this added support would look including some of the most popular cinnamon themes in cinnamon spices. All these themes would otherwise have the default cinnamon theme dialog: 

Mint-X:
Mint-X
Mint-L
Mint-L
[Cinnamox-Aubergine
Cinnamox-Aubergine
Numix-Cinnamon-Transparent
Numix-Cinnamon-Transparent
CBlack
CBlack
New-Minty
New-Minty
qob
qob
Carta
Carta
Vertex-Maia (Manjaro default theme)
Vertex-Maia
Orchis-Light (dark background instead of light):
Orchis-Light
Eleganse-dark:
Eleganse-dark
Adapta-Nokto:
Adapta-Nokto
Adapta and Adapta-Nokto would otherwise have transparent backgrounds:
Adapta-Nokto_wo

### DIFF
--- a/js/ui/modalDialog.js
+++ b/js/ui/modalDialog.js
@@ -105,6 +105,8 @@ var ModalDialog = GObject.registerClass({
         this.dialogLayout = new Dialog.Dialog(this.backgroundStack, params.styleClass);
         this.contentLayout = this.dialogLayout.contentLayout;
         this.buttonLayout = this.dialogLayout.buttonLayout;
+        this.dialogLayout._dialog.add_style_class_name("modal-dialog"); // temporarily add some support for old themes
+        this.buttonLayout.add_style_class_name("modal-dialog-button-box"); // temporarily add some support for old themes
 
         this.openAndCloseTime = 100;
         let enableRadialEffect = true;
@@ -188,7 +190,9 @@ var ModalDialog = GObject.registerClass({
     }
 
     addButton(buttonInfo) {
-        return this.dialogLayout.addButton(buttonInfo);
+        const button = this.dialogLayout.addButton(buttonInfo);
+        button.add_style_class_name("modal-dialog-button"); // temporarily add some support for old themes
+        return button;
     }
 
     _fadeOpen() {


JosephMcc:
I'm not a fan of this. One, we already have a mechanism in place for these to be themed by Cinnamon's base theme if there is no current support in the 3rd party one. Two, this sort of stuff is never "temporary". If themes are unmaintained and not being updated, maybe they shouldn't still be on the spices site. There is also more stuff in the roadmap for Cinnamon that will need adaption by themes.

I don't think adding a bunch of hacky workarounds that keep people from properly updating things, or trying to keep alive things that no one wants to bother updating, is just a bad idea.

fredcw:
@JosephMcc So ruining every cinnamon theme in existence is ok if it avoids 4 lines of "hacky" code? We're not here to make pretty non-hacky code, we're here to give users what they want. If that's not what we're doing here then what we're doing here is pointless. Obviously things have to change if improvements are going to be made but backward compatibility is important and so are what users want and how they have styled their desktops so we have to at least make some effort in that regard. These are brilliant and impressive improvements you've made and plan to make no doubt, but we have to at least try to maintain backward compatibility as much as possible. As of Mint 22.1 (Cinnamon 6.4), cinnamon now has exactly one working theme (Mint-Y). How is this any different from libawaita? (and GNOME which doesn't care about backward compatibility and what users want and is the very reason for cinnamon's existence in the first place)

I'm exaggerating of course, as the themes aren't entirely broken, as it's just the modal dialogs and this fix is not perfect by any means - it doesn't address the osd (e.g. volume) theming. However, your proposed changes to the menu applet would indeed absolutely completely break all existing light cinnamon themes. Btw, I love the menu redesign, super impressed.

I just think we need to at least give a nod to user preferences and backward compatibility and for the sake of 4 lines of hacky code, this seems like a no-brainer to me.

Oh how I love technical disagreements. Merry Christmas, I love you. xD

JosephMcc:
>  what users want and how they have styled their desktops so we have to at least make some effort in that regard.

But this isn't the way to do it. As you pointed out yourself, this only fixes some new things. Part of the reason for moving the spices under the Mint umbrella is so it would be easier to help update them for new things. The proper way to do that would be to start help updating them. 

> I'm exaggerating of course, as the themes aren't entirely broken, as it's just the modal dialogs and this fix is not perfect by any means

Lol, just a bit. As I pointed out above, the proper fix is to update the themes to work. The one's that aren't unmaintained are going to rot if on one wants to do so.

> Oh how I love technical disagreements. Merry Christmas

No worries. You are never rude in these discussions, just giving a point of view. Merry Xmas to you!

clefebvre:
I completely understand the points made by @JosephMcc, but I agree with @fredcw.

Not even talking about 3rd party artists... I, myself, am not sure I have the resources to adapt Mint-L/Mint-X right now. I was happy enough to have these work with the default dialog styles for the BETA.

I adapted Mint-Y and the result looks good, but that took a while and I kinda cheated. I'm not happy with how I did this. I imported the whole SASS infra around button mixins and all. I took the shortest route, the result is good but the way I did this isn't great. I want to go back to this eventually and clean it up in Mint-Y.

When it comes to maintenance and spices, @JosephMcc is right, if something isn't maintained properly it should be deleted. How long should it take for them to be adapted before we delete them though? It would be completely unreasonable to expect them to adapt right now prior to STABLE. We also know we won't delete them that fast. So for users that means we're not pragmatic, we're aware of an issue and we can't take action.

A reason for deletion also should be that something is broken or unsupported. As they are right now, without support for the new dialogs, these themes are still functional though. Even if they didn't get updated, they probably wouldn't get deleted.

And then from a moral standpoint, it doesn't shock me for the new dialogs to support old style names. Sure, it will never look perfect until a theme actually gets updated. But until it actually does get updated or deleted, which could take a year or more, why not minimize the issue?

Joseph is right when he says that if we don't make this problem visible, it will never get solved by spice authors. But we've got more changes coming in 22.2. More visible changes, a new menu, a status applet. Updating and/or deleting themes in spices is going to be on the agenda no matter what. We will definitely get to a point where we're making spreadsheets, lists of TODO, and pinging people to get stuff done. I think when we get to that point, we'll want to make it easy for authors, but also put pressure.

Right now, we're not ready to tell authors what we want, how fast we want it, or to do anything if things don't get done. We don't need users to suffer just to keep some sort of pressure.

Secret-chest:
Status applet? Is that like GNOME with combined network, power and sound?

contrarybaton60:
I just want to state my opinion on the matter.

I think it's best to include this as a temporary solution, and get Mint Y/L/X updated, themes from Spices updated, tell the Ubuntu Cinnamon team to update their theme... basically what we can do to get this issue minimized but how long? that's something that y'all gonna have to know.

fredcw:
@JosephMcc Do you think it would be a good idea to add a "dialog-button-box" style class to dialog buttonLayout in dialog.js, as some theme authors may want to style the buttonbox differently (as in the Vertex Maia example above) and there seems to be no replacement for 'modal-dialog-button-box' style class. I tried `.dialog > StBoxLayout:last-child` but that doesn't work. It's probably a bit niche in theming terms, so not sure if it's a good idea or not? I can make PR if necessary.

fredcw:
I made a list of the style class names that have changed in cinnamon 6.4 which may be useful to some people updating themes: https://github.com/linuxmint/cinnamon-spices-themes/issues/843

PR #12601, opened 2025-01-01
dialog.js: add style class dialog-button-box

fredcw:
Allow themes to style the dialog buttonlayout. This is a replacement for the old modal-dialog-button-box class style. Without it, themes like Numix-Cinnamon-Transparent, New-Minty, and Vertex-Maia will not be able to style the button box separately.

Numix-Cinnamon-Transparent

Vertex-Maia


JosephMcc:
Looks good.

PR #12602, opened 2025-01-01
GWL: add custom labeled button width

expiredhotdog:
Adds an option to the Grouped Window List applet to set a custom width for labeled buttons.

Also fixes a minor bug where the button label would not update properly after changing the settings. Aside from that, default behavior is unchanged.


JosephMcc:
It would be nice to see the bug fix done separate from the new option. Quite a few options were already removed from GWL because it's a bit of an over complicated mess. Not sure this one is really needed.

expiredhotdog:
@JosephMcc I just realized that I probably misinterpreted what you meant. Would you like me to close this PR and open 2 new ones?

JosephMcc:
Not at the moment. Not sure what the feeling is for adding the setting, but the bug fix is pretty likely to get merged if it works :)

mtwebster:
Hi, I don't think this warrants another option. If there's an issue with refreshing after setting changes, we'd welcome that.

Also, please avoid whitespace fixes in areas of code that are *not* otherwise being modified.

PR #12629, opened 2025-01-13
osdWindow.js: adding text display of the level and a small increase in the icon

tutralex:
In my opinion, the volume percentage display in Cinnamon v.6.2.x was very convenient and I would like to bring it back. Also, media buttons are more often used on devices like TVs, so it would be better to make the icon a little bigger.

Voff
V25
V100

Before:
Vprev


tutralex:
There is no answer why the level is excluded? (https://github.com/linuxmint/mint22.1-beta/issues/4)
Moreover, it is taken locally and does not need to be passed as an additional parameter.

tutralex:
What nonsense sucked out of a finger, considering that the sound changes by +- 5%. You will achieve that the most devoted users will turn away from you...

PR #12672, opened 2025-01-22
Fix visual tearing in scroll-view (#11760)

CollinBartlam:
This should fix the visual tearing issue seen when scrolling or navigating in the categories box within menu@cinnamon as described in #11760. While this approach resolves the issue, I haven't worked with Cinnamon's codebase before, so feedback on improving the implementation is welcome.

### DIFF
--- a/src/st/st-scroll-view.c
+++ b/src/st/st-scroll-view.c
@@ -933,6 +933,16 @@ vfade_setting_changed_cb (GSettings *settings, gchar *key, gpointer user_data)
     st_scroll_view_style_changed (widget);
 }
 
+static void
+st_scroll_view_vadjustment_changed (GSettings *settings,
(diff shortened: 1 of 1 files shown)


fredcw:
I haven't tested this so I don't know if it works but in any case, this seems like a workaround rather than fixing the actual cause. The scrolling seems to be working ok except that the area being scrolled is being miscalculated. 

It seems that the width of the area on the right that is not being repainted (left behind) is equal to the left padding of the child actor. In other words, the width of the repainted area is ok but is positioned to ignore the left padding of the child actor.

Another thing to note is that this problem does not occur with autoscrolling (mouseover scrolling) and neither if the priv->vscrollbar_policy is set to ST_POLICY_NEVER

Issue #11983, opened 2023-12-17
Buggy handling of battery critical action setting by the cinnamon settings UI


Lanchon:
### Distribution

Mint 21.2

### Package version

5.8.4

### Graphics hardware in use

-

### Frequency

Always

### Bug description

the schema for "/org/cinnamon/settings-daemon/plugins/power/critical-battery-action" defines which settings are valid, not the UI.

when this setting is set to some values, eg 'suspend', the UI behaves in this erratic manner:
- the setting is incorrectly shown as blank in the UI.
- if you click on the drop-down box, the valid 'suspend' setting is inadvertently changed to anything else.
- coupled with the immediate updating of the backend, this means that the UI corrupts the current value when an action that should never change a setting is performed by the user (opening the drop-down).

this is beyond any reasonable doubt, a UI bug.

but there is more to this story...

i remember reading somewhere that some users were requesting for the ability to set 'suspend' as the critical battery action on their computers, only to be met by draconian answers by devs of the type "you are mistaken! suspend is never ever the right critical action! i will not add that option because i feel the urge to save you form yourselves!"

i could not find that discussion again, though i tried, so i am opening this bug.

to these devs, if they show up, i answer:

- you are wrong: suspend is the right action sometimes (more on this later), though changing the critical threshold is required for sane behavior in those cases.
- but if you are going to be draconian and forbid users from doing what they want or need, then **at least do it right and do not introduce bugs:** you should check if the current value is one of the anointed 'safe' values by those who know better, and if it is not you should avoid corrupting it: either add it now to the values normally populated in the drop-down, or disable the drop-down completely. but fix this buggy behavior please.

i would normally submit a pull request, but not in this case: those who want these restrictions up should maintain this silly code.

-----

finally... this is a bit off-topic, but given the previous threads i read, it seems necessary to explain why 'suspend' is the right critical action in some cases. i suppose there are many, but this is my case:

- i use a laptop exclusively as a desktop replacement.
- i need to access this machine remotely.
- i can only connect this machine to the network via wifi.
- thus, WoL (wake-on-lan) is not an option, so i need WoWLAN to always work.
- WoWLAN only works if the system is sleeping. **it does not work if it is shut down.**
- again, i need to access this machine remotely. always, no matter what! and i am sometimes away from home for months at a time.
- it follows that **this machine can never ever be powered off**. as such, hibernate and hybrid sleep systemd targets are disabled on this machine.
[...]

PR #12680, opened 2025-01-25
power: Add suspend option to battery-critical actions

Lanchon:
some say that including "suspend" in the set of available battery critical actions is dangerous: power might ran out during suspend and machine state would be lost.

however:

- the really dangerous action is "do nothing", and it is still allowed in the UI. suspend might loose state, but is way way safer than "do nothing": suspend will flush all caches to disks, making sure that all recently saved data is persistent. it also forwards filesystems to a point where -though still mounted- they are effectively clean: filesystems are guaranteed to be in a sane state on-disk, even if they do not journal. running out of power during suspend is infinitely more "graceful" and safe than running out of power while the CPU is running: the latter is allowed, but the former is not. go figure...
  - possible dangerous bugs in SSD firmware will also be sidestepped, as SSDs will be sleeping during the power off event.

- suspend can loose state, yes, but this can happen any time the user suspends. if suspending, the system will *not* wake up and do the critical action (eg, shutdown) when power is critical. by this token, suspend should never be allowed then. but we allow it because it makes sense: we assume users know what they are doing.

- the current UI has a bug by which, if the user chooses suspend in dconf (as i do; i also elevate the critical trigger percentage), the UI shows the setting as blank and corrupts it immediately if the user simply touches it. the bug was reported here: https://github.com/linuxmint/cinnamon/issues/11983. this PR fixes that issue (for suspend only, the most likely trigger by far). there are other ways to fix this (discussed in the bug report), but over a year passed and no fixes were proposed, so i propose merging this.

thank you!

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_power.py
@@ -642,6 +642,7 @@ def get_available_options(up_client):
     ]
 
     critical_options = [
+        ("suspend", _("Suspend")),
         ("shutdown", _("Shut down immediately")),
         ("hibernate", _("Hibernate")),
         ("nothing", _("Do nothing"))


titanohearn:
One stupid thing I discovered in new Cinnamon, is that if I run out of battery, it doesn't say anything, it doesn't notify me or suspend the laptop, my laptop just turns off

Lanchon:
@titanohearn use dconf Editor, look for where the power config is saved. change the critical warn and act levels to something reasonable for testing like 55% and 50% and ensure that battery % and not time-to-live is used for critical action. if the bad behavior is repeatable, then create an issue in linuxmint (not cinnamon; this would be a power daemon bug, not a UI bug).

titanohearn:
@Lanchon alright, gonna try it out

Lanchon:
thank you guys!!!

PR #12714, opened 2025-02-04
Fix nm-applet requirement in meson.build

timp87:
The logic was swapped in 16cf990 which is incorrect.
So require nm-applet when have_networkmanager is true, i. e. disable_networkmanager is false. Not vice versa.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -164,9 +164,9 @@ install_subdir(
 
 session_conf = configuration_data()
 if have_networkmanager
-    session_conf.set('REQUIRED', '')
-else
     session_conf.set('REQUIRED', 'nm-applet;')
+else
+    session_conf.set('REQUIRED', '')
 endif
 
 session_files = ['cinnamon.session', 'cinnamon2d.session']


timp87:
@JosephMcc and @clefebvre please take a look.

mtwebster:
The naming here just needs to be worded better, but the logic is correct.

This section controls whether *nm-applet* is required by the session. Now that cinnamon has an internal implementation of the authentication dialog we do *not* want nm-applet to be a required part of the session.

Fantu:
@mtwebster also with new commit seems to remain the regression to don't have the possibility to disable network manager (for example for build that don't want it or archs that don't have it).
I mean about this (if build_nm_agent require libnm and if not require nm-applet but no more option to not have network manager):
```
if build_nm_agent
    session_conf.set('REQUIRED', '')
else
    session_conf.set('REQUIRED', 'nm-applet;')
```

timp87:
Thanks @Fantu 
Exactly.
This was the reason why I thought the logic here is incorrect. I want to disable network manager, so why would it instead require nm-applet in such case.
I'm currently porting cinnamon to non-standard platform. So there are even more places where I have to delete nm artefacts in the build.

mtwebster:
Our point was we don't want nm-applet required in the session when using our own agent. We didn't consider 'no network manager' when repurposing this option. I guess it'd be best to make another meson option to control this?

Fantu:
One parameter with multiple-choice (like with_networkmanager=disabled|internal-agent|external-agent, or something better) or 2 parameters, for example return to have also disable_networkmanager and if true will disable also build_nm_agent.
Both option seems ok, do what you prefer.

timp87:
I would be happy if there was a way to fully disable network manager.
I can show more places where this full disablement should take place.

mtwebster:
What other places?

PR #12717, opened 2025-02-05
desklet-photoframe: change image size to guarantee keeping the ratio

watertrainer:
Currently, sometimes the images get squeezed, when in line 161 height is bigger than the height of the desklet and then the ratio isn't kept. Now this doesn't happen, as we always guarantee that the aspect ratio is kept and both height and width of the image stay smaller than the one of the desklet. See https://stackoverflow.com/questions/3971841/how-to-resize-images-proportionally-keeping-the-aspect-ratio for more details


mtwebster:
Can you fix the code style in your changes? Spaces between arguments/operators, etc... to match the rest of the codebase.

Thanks

PR #12764, opened 2025-03-03
Add gridmenu applet

fredcw:
Maybe this could be the new cinnamon default menu or users could be given a choice at install time between this and the classic list menu as Cinnamon seems to be the only DE without an icon menu by default.

This is basically a slimmed down version of Cinnamenu applet.

Features:
* Home directory file search
* Emoji search
* Drag and drop apps, files, and folders onto the sidebar
* Layout options: sidebar on top, bottom, left, or right of menu. Show/hide sidebar and categories.
* Backward compatible with existing cinnamon themes (no themes need updating)
* Built in file system browser + add directories as menu categories.
* Files and recently used files shown as thumbnails rather than icons + categorised recent files
* Built in simple calculator with JS math functions.
* Fuzzy search

Maybe the name should be changed to icon-menu or app-menu?
Emoji search terms are English only as there are about 2000 emoji with 5-6 search terms each so probably too much to translate.

Here's the settings-override.json file for inclusion in the applet folder for Mint distro only (Mint specific panel icons):
[settings-override.json](https://github.com/user-attachments/files/19056993/settings-override.json)

Here's the menu for testing purposes (including style changes to default theme):
[gridmenu@cinnamon.zip](https://github.com/user-attachments/files/19048317/gridmenu%40cinnamon.zip)

Edit: I'm currently keeping this code in sync with Cinnamenu, so it's very easy for me to swap in/swap out features to get the right balance between features and code complexity.

PR to make make this cinnamon default menu: https://github.com/linuxmint/cinnamon/pull/12906

### DIFF
--- a/data/theme/cinnamon-sass/widgets/_startmenu.scss
+++ b/data/theme/cinnamon-sass/widgets/_startmenu.scss
@@ -94,3 +94,83 @@ $menu_outer_border_radius: $base_border_radius * 1.25;
 .menu-search-entry-icon {
   icon-size: $scalable_icon_size;
 }
+
+// gridmenu (icon-menu)
+
+.menu-category-button:highlighted, // Used to highlight category containing newly installed apps.
+.menu-category-button-selected:highlighted {
+  font-weight: bold;
+}
+
+.menu-applications-header-text { // shows folder name, "no recent files", error messages etc.
+  font-weight: bold;
+  font-size: 110%;
+}
+
+.menu-applications-subheading {
+  font-weight: bold;
+  font-size: 100%;
+  padding-top: 0.3em;
+  padding-left: 0.3em;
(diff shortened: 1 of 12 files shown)


ImAwsumm (approved):

ImAwsumm (approved):

kacperpaczos:
Can you add an image?

fredcw:
@kacperpaczos It looks the same as Cinnamenu. It just has fewer features like list view option, web bookmarks search, etc.

clefebvre:
The new redesign is on master now. It's much improved compared to the previous menu, but it still doesn't bring an icon view and for this reason Cinnamenu will continue to be popular.

As a DE though, I don't think we should have multiple menus in the core system. It's good that something like Cinnamenu is a spice and that it's one of the most popular. It's a good alternative, it's different and because it's a spice it can afford to have more options/features.

PR #12786, opened 2025-03-13
Calender: Fix issue with with custom tooltip not updating seconds.

### LINKED ISSUE #12140: calendar@cinnamon.org if set to show seconds only in tooltip, the second doesn't tick properly
### Distribution

Mint 21.3 Cinnamon

### Package version

cinnamon 6.0.4

### Graphics hardware in use

_No response_

### Frequency

Always

### Bug description

Currently, the panel clock ticks per second if and only if the "date format" is set to contain %S in custom format. The "date format for tooltip" is ignored.

Thus if one want to hide second in panel and show it only within tooltip, the tooltip can't show a correct second.

The one who found this bug: https://forums.linuxmint.com/viewtopic.php?t=416598

### Steps to reproduce

Right click the panel clock > configure > toggle on "use a custom date format" > in "date format for tooltip", put "%A, %e %B %Y, %H:%M:%S".

Then move the mouse to the panel clock and see the tooltip misreporting the second.

### Expected behavior

The clock shall update itself per second if any one of the two custom date formats includes second.

### Additional information

I tried to fix this bug myself. The fix seems to work but I am not sure if it is quality enough for merging.
[applet.js.txt](https://github.com/linuxmint/cinnamon/files/14812198/applet.js.txt)

rmarker:
When the calender applet uses a custom format, and the tooltip contains seconds but the regular custom format doesn't,  the tooltip seconds wouldn't update properly.

This is because the update frequency is based on the regular custom format.

This has been fixed, so that the calender now checks whether the mouse has entered the applet. When the tooltip is shown, the format string used by the clock is updated to combine the tooltip format. This way it ensures that the update frequency is based on whichever has the finer resolution.

Fixes #12140

### DIFF
--- a/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
@@ -156,6 +156,17 @@ class CinnamonCalendarApplet extends Applet.TextApplet {
             } catch (e) {
                 this._upClient.connect('notify::resume', Lang.bind(this, this._updateClockAndDate));
             }
+
+            // Change the format string if the mouse is over the calender to account for the tooltip.
+            this._is_entered = false;
+            this.actor.connect('enter-event', Lang.bind(this, function (menu, event) {
+                this._is_entered = true;
+                this._updateFormatString();
+            }))
+            this.actor.connect('leave-event', Lang.bind(this, function (menu, event) {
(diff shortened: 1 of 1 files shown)


ImAwsumm (approved):

Kreijstal:
dumb question but how do we synchronize calendar@cinnamon with gnome calender or so?

PR #12899, opened 2025-05-21
cs_themes.py: Add accent color chooser for libadwaita >= 1.7.

mtwebster:
Only enabled when the current theme does *not* contain the 'accent_color' variable.


clefebvre:
That doesn't really make sense though. Most apps are GTK3 and they will not follow this. It's weird to have a setting like this which only works for certain apps.

PR #12902, opened 2025-05-24
cs_services.py: Add new module "Services" to cinnamon settings

Monn-i:
I got annoyed at having to manually skim some obscure systemd folders for services and using terminal for some really quick stuff. I wrote this at my leisure, would this be something you would consider adding to Cinnamon? I tried to adhere to existing code style and work with existing stuff as much as possible.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_services.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_services.py
@@ -1,0 +1,719 @@
+#!/usr/bin/python3
+
+# Service module for managing running systemd services
+#
+# Systemd provides services are divided to user and system services. These have different access
+# types and they correlate with different DBus connections.
+# (User services & session bus vs. system services & system bus)
+#
+# Changing root-level access services implicitly prompts the password without explicit code here.
+#
+# References:
+# [1]: https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.systemd1.html
+# [2]: man systemd.service - Command Lines
+
+import gi
+gi.require_version("Gtk", "3.0")
+gi.require_version('Notify', '0.7')
+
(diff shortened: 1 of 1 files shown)


leigh123linux:
You need to add python3-dbus to the debian control file.

https://github.com/linuxmint/cinnamon/blob/master/debian/control#L92

mtwebster:
No, please don't use the dbus module - instead use GLib/Gio. I just converted our last remaining user: 15b57237ea2ed40f.

I'm also not sure it's a good idea to expose this much to a typical (or even experienced) user. I'd definitely use a nice control panel like this for dev work, but I don't think I'd want to be responsible for one or fielding reports from users that decided to start flipping switches.

Monn-i:
> No, please don't use the dbus module - instead use GLib/Gio. I just converted our last remaining user: [15b5723](https://github.com/linuxmint/cinnamon/commit/15b57237ea2ed40f547df7def57ac3f10d568b5a).
> 
> I'm also not sure it's a good idea to expose this much to a typical (or even experienced) user. I'd definitely use a nice control panel like this for dev work, but I don't think I'd want to be responsible for one or fielding reports from users that decided to start flipping switches.

I agree that this is a tool that would allow making the system unstable. What if then this was simply a listing of services, their statuses and how the service file is built? (Together with refactoring to use GLib/Gio, of course)

This would not allow users to mess up their system, but would be nice GUI for even less experienced users to get to know what is happening and alive within the system. Services are somewhat fundamental part of Linux systems, so in my mind exposing them here would be useful to even inexperienced users e.g. to check if a certain service exists and is active, if asked.

Monn-i:
After some thinking, this does elicit some questions about if this fits the Linux Mint's ease-of-use narrative. Cinnamon extension would be still discoverable enough for users interested, so I think I'll investigate and probably pursue that route instead. Thanks for comments everyone!

PR #12909, opened 2025-05-30
Add a custom layout option for window buttons

Hilyxx:
- In the cs_windows module, users can now add a custom layout for window buttons.
- Error message if an invalid layout is added
- Tooltip with examples


mtwebster:
This is overkill for most users, and was previously removed.

https://github.com/linuxmint/cinnamon/commit/e1eaea39a0325bedf75522262bb8e0e689f2a15e

PR #12911, opened 2025-05-30
user@cinnamon.org - Add an option to resize user's avatar in the settings

Hilyxx:
Instead of using the cont declaration for avatar size, add an option in the settings to resize the user's avatar in the menu.

### DIFF
--- a/files/usr/share/cinnamon/applets/user@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/user@cinnamon.org/applet.js
@@ -11,7 +11,6 @@ const ScreenSaver = imports.misc.screenSaver;
 const Settings = imports.ui.settings;
 const UserWidget = imports.ui.userWidget;
 
-const DIALOG_ICON_SIZE = 64;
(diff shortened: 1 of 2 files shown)


anaximeno:
I'm not sure an option for this is needed directly in settings, at most this should be something defined by the theme (at most) if we're talking about the icon that appears in the popup menu, and the panel one by the panel icon sizes (for colored icons when the user image is used, and for symbolic icons when the symbolic one is used).

The concern here being, bringing unecessary additional complexity in comparison to the actual use potential.

Hilyxx:
"The concern here being, bringing unecessary additional complexity in comparison to the actual use potential."
Of course, I understand your point of view. 

The idea is not to make anything more complex, but just to provide an additional option to give users more choice, and there's nothing complex about the option.
As the avatar size (in the menu) is fixed at 64 px, some custom themes might use an incorrect size for the avatar (too big or too small). This would result in an incorrectly sized avatar at high / low screen resolutions.

mtwebster:
Overriding the theme isn't really a good case for adding a setting, especially something as minor as this.

PR #12921, opened 2025-06-04
add a new date and time display applet

rkarahan80:
this applet displays the current date and time on the panel. You can customize the date and time format strings through the applet's settings.

the applet consists of:
- metadata.json: defines the applet's properties.
- applet.js: contains the javaScript logic for displaying and updating the date/time.
- settings-schema.json: defines the configurable settings for date and time formats.
- icons/datetime_display_icon.svg: the icon for the applet.

### DIFF
--- a/files/usr/share/cinnamon/applets/datetime_display@cinnamon.org/.placeholder
+++ b/files/usr/share/cinnamon/applets/datetime_display@cinnamon.org/.placeholder
@@ -1,0 +1,2 @@
+# This is a placeholder file to create the directory.
+# It can be removed later if not needed.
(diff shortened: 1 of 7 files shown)


rkarahan80:
New features have been added.

rkarahan80:
New features have been added.

anaximeno:
There's already an applet with all these features https://github.com/linuxmint/cinnamon/tree/master/files/usr/share/cinnamon/applets/calendar%40cinnamon.org and much more.

So, I'm not sure if this PR makes sense.

mtwebster:
Hi,

We're not really interested in replacing the current time applet, as it does quite a lot more than just display the time.

If you've made an applet you think others might be interested in you should definitely submit it to Cinnamon's 'Spices' collection, which is specifically for this purpose.

https://cinnamon-spices.linuxmint.com/
https://github.com/linuxmint/cinnamon-spices-applets

Just for the heck of it, I installed this, and it doesn't appear to actually work - I think there may be some issue with your settings file. The settings page is empty, and so is the applet on the panel.
image

Thanks!

PR #12923, opened 2025-06-06
fix high cpu usage in shortcut view

### LINKED ISSUE #12807: Pressing space causes high CPU usage
In the keyboard settings in the "Shortcuts" tab, pressing space or enter while a category is selected, causes high CPU usage for some amount of seconds and freezing the settings window.

Here is a preview video (ignore the resources statistic at the end, OBS made the chart not really distinguishable but it still happens without OBS):

https://github.com/user-attachments/assets/ae5cebe5-6e16-47db-b5cf-8efc9111aa7e

Information:
Linux Mint 22.1
Cinnamon 6.4.8

hduelme:
Fixes  #12807 

I found a related log entry for the issue:
> TypeError: Module.onCategoryChanged() takes 2 positional arguments but 4 were given

The  [row-activated signal](https://docs.gtk.org/gtk3/signal.TreeView.row-activated.html)  expects a callback with four parameters. 
However, `onCategoryChanged()` only accepts two, which likely led to the unexpected behavior and resulting CPU spike.
After removing the signal connection, the issue no longer occurs.

Note: This bug is also present in Linux Mint 22.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_keyboard.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_keyboard.py
@@ -447,7 +447,6 @@ class Module:
             self.cat_tree.connect("cursor-changed", self.onCategoryChanged)
             self.cat_tree.connect("button-release-event", self.onCategorySelected)
             self.cat_tree.connect("key-release-event", self.onCategorySelected)
-            self.cat_tree.connect("row-activated", self.onCategoryChanged)
             self.cat_tree.connect("map", self.categoryHighlightOnMap)
             self.cat_tree.connect("focus-in-event", self.categoryHighlightOnMap)
             self.cat_tree.connect("unmap", self.categoryHighlightUnmap)


anaximeno:
The other option would be to add two new params to `Module.onCategoryChanged()` (even if dummy params) if that solves the issue also. That's assuming that callback is still necessary, otherwise might have to assure it won't lead to a downside just removing it.

hduelme:
@anaximeno Adding the two dummy parameters would break the call from the `cursor-changed` signal.
> TypeError: Module.onCategoryChanged() missing 2 required positional arguments: 'a' and 'b'

As the [row-activated signal](https://docs.gtk.org/gtk3/signal.TreeView.row-activated.html) never worked, I believe it's safe to remove it.

anaximeno:
> @anaximeno Adding the two dummy parameters would break the call from the `cursor-changed` signal.
> > TypeError: Module.onCategoryChanged() missing 2 required positional arguments: 'a' and 'b'

You can give default None values to these last two:
`a=None, b=None`

anaximeno:
Actually another more general solution is using the `*args, **kwargs` at the of the definition:

```python
def method(arg1, arg2, *args, **kwargs):
    ...
```

hduelme:
@anaximeno that would work, but the [cursor-changed signal](https://docs.gtk.org/gtk3/signal.TreeView.cursor-changed.html) always triggers, when the focused cell changes. This should cover the cases from the [row-activated signal](https://docs.gtk.org/gtk3/signal.TreeView.row-activated.html)

> The “row-activated” signal is emitted when the method gtk_tree_view_row_activated() is called, when the user double clicks a treeview row with the “activate-on-single-click” property set to FALSE, or when the user single clicks a row when the “activate-on-single-click” property set to TRUE. It is also emitted when a non-editable row is selected and one of the keys: Space, Shift+Space, Return or Enter is pressed.

Issue #7093, opened 2017-11-29
Multiple VPN connections are incorrectly handled by the NetworkManager applet


PauloGDPeixoto:
```
 * Cinnamon version (cinnamon --version) - 3.6.6
 * Distribution - Fedora 27
 * Graphics hardware *and* driver used - Not relevant
 * 64 bit
 ```

**Issue**
Currently, NetworkManager supports concurrent VPN connections, which the current design of the applet doesn't account for. When I connect to a second VPN, there is no visual indication in the applet of this connection and the on/off slider does not turn off this connection. The only way to see if the connection is active and disconnect, is to go into the network settings dialogue, which displays the correct status for each VPN connection.

**Steps to reproduce**
Connect to two VPNs simultaneously.

**Expected behaviour**
Have a correct display of the connection status in the NetworkManager applet list, VPN on/off slider should start (or terminate) all VPN connections, clicking on an already connected VPN in the NetworkManager applet list should disconnect that VPN.

icarter09:
@PauloGDPeixoto are you still experiencing this issue?

alankeny:
I'm running Cinnamon 4.2.4 on a fully updated Linux Mint 19.2 system.  I'm still experiencing all of the problems reported by @PauloGDPeixoto.  The applet doesn't handle multiple simultaneous VPN connections properly, while tools like Network Settings and nmtui do.

icarter09:
@alankeny thanks for the update

t-oster:
Just removed the applet and magically the original gnome NetworkManager applett appeared, which does handle VPNs as expected:

![Bildschirmfoto vom 2021-07-21 10-06-41](https://user-images.githubusercontent.com/1052720/126454379-063e6cd0-6bb3-4134-844f-9c5df5b9cc5f.png)

maybe checkboxes are the way to go

TheReptile:
@t-oster Or maybe something like this:
![image](https://user-images.githubusercontent.com/22954131/127062890-83b4704a-3652-4ebc-8621-ac0d6013e522.png)
Not my mockup, but it looks nice imho.

[Source](https://forums.linuxmint.com/viewtopic.php?p=177049)

uscmbrixton:
Applet bug still exist in 20.2.

renepavlik:
This is more likely a feature request, not a bug. It would be really appreciated if the support for multiple active VPNs simultaneously was added. The above mockup by @TheReptile is a way it should work in 2022. Thank you.

[...]

PR #12930, opened 2025-06-12
nm-applet: make visible multiple active vpn connection

ofixone:
I'm highly dissapointed about https://github.com/linuxmint/cinnamon/issues/7093 that never fixed in years so this is a kind of a try to do it. Just represent multiple vpn enabled vpn connections

Got some issue about toggle that deactivate all connections but on the other hand activate only last one. But still much better than now imho


ofixone:
Knock knock :)

Issue #8773, opened 2019-08-06
menu@cinnamon.org closes when adding menu items to favorites or desktop or panel


ChooChooAl:
```
 * Cinnamon version (cinnamon --version)
   - Please specify if you are using the daily builds PPA (https://launchpad.net/~linuxmint-daily-build-team/+archive/ubuntu/daily-builds).
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...)
 * Graphics hardware *and* driver used
 * 32 or 64 bit
 * Attach /home/<username>/.xsession-errors, or /var/log/syslog
 ```

**Issue**

**Steps to reproduce**

**Expected behaviour**

**Other information**

icarter09:
Linux Mint 19.2
Cinnamon 4.2.4

Was able to reproduce this issue.

PR #12941, opened 2025-06-17
Prevent context menu from closing when favorites add/remove is clicked

### LINKED ISSUE #12881: Linux Mint menu closes after adding a favorite icon
### Distribution

22.1

### Package version

6.4.8

### Graphics hardware in use

_No response_

### Frequency

Always

### Bug description

Linux Mint menu closes after adding a favorite icon. Then, one need to reopen Linux Mint menu to verify if it was added.

### Steps to reproduce

Open Linux Mint Menu. Choose an application and do a right-click on it. Add it to favorites. Linux Mint menu closes.

### Expected behavior

Linux Mint menu shouldn't close. One can't verify if the application icon added to favorites vertical menu, nor add another, without opening again Linux Mint menu.

### Additional information

_No response_

cagdasalagoz:
This partially fixes #12881 and #8773.

### Issue
menu@cinnamon closes whenever something is selected from the `contextMenu`. Including `add/remove favorite` button.

menufavclosebug

### Considerations
As well as I believe the menu should not be closed when something is added or removed from the favorites, I am not so sure about the other actions like `add to desktop` or `add to panel`. Should we keep the experience consistent and close the menu for all options, or should we have special cases for `favorite`, just like I've added here?

This is more like a design choice. I would be happy to implement whatever the decision is.
Another point to consider if we should keep `contextMenu` open or close it after every operation. I left it as it is, so it closes, but this requires some discussion with the design team, I suppose.

### Fix
Basically, wrapped the close action with an if statement to prevent it from closing the menu for favorite operations.
menufavclosebugfix

I would appreciate any feedback.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -406,7 +406,9 @@ class ApplicationContextMenuItem extends PopupMenu.PopupBaseMenuItem {
                 } else return true;
         }
         this._appButton.applet.toggleContextMenu(this._appButton);
-        this._appButton.applet.menu.close();
+        if (this._action !== "add_to_favorites" && this._action !== "remove_from_favorites") {
+            this._appButton.applet.menu.close();
+        }
         return false;
     }
 


PurpleSmurf2:
It definitely feels more consistent when the menu remains open. I would find it even better if the expandable menu remains open until I manually close it again and the selected options, such as adding to favourites, give feedback in the form of a small check mark.

fredcw:
@PurpleSmurf2 
>if the expandable menu remains open...

As it is atm, the option "add to favourites" would have to change to "remove from favourites". You could just use "Favourite" instead I guess but it would still be programmatically more complex.

Edit: Actually, I guess the text changing is also ok so you're probably right, it would be more intuitive behaviour if the context menu didn't close.

samw-26:
Thank you for this, this has been bothering me since I started using Mint. My opinion is that the menu should stay open until manually closed for the Add/remove to panel, desktop, and favorite controls. I find it much more user friendly than to repeatedly reopen the menu when I want to add multiple apps to favorites.

cagdasalagoz:
Okay, updated the code so ContextMenu also will stay open, and the text will get updated.

Although I had to duplicate the `_("Remove from favorites")` for this. I thought of creating a static definition, but wasn't able to see any examples of it. So I'm not sure what's preferred by the maintainers. I can update accordingly.

menufavbugfix

This still closes the menu for any other action other than favorite actions. I am still not convinced about them. 

1. Properties -> opens a properties window. This definitely should close the menu.
2. Add to desktop -> We don't have a way of notifying the user if the operation completed or not. If we leave it as it is, they might think the click did not register. Maybe we can send a notification about it? I have no idea how the experience should be with that. So I'm leaving it as it is.
3. Add to panel -> Same as above.
4. Uninstall -> Opens another window. It's normal for the menu to be closed.

PurpleSmurf2:
It looks great so far! But in your animation when you open the Applets and the expandable menu appears, it first is shown bigger like a zoom and gets smaller when your mouse leaves the Applet menu. It looks a bit weird. The other expandable menus keep their size.

Edit
Your code is not the cause of this behavior. It is already a "problem" in the main version. Probably the fontsize, icon or padding (or all together) of the expandable menu is too big. This is especially a problem with words in different languages which need more letters because the menusize is limited. But it doesn't look good. 

you can also see this behavior in you first unmodified animation when you open the expandable menu of Accessibility where the vertical separator of the startmenu is moving its position to the left and reduce the space of the application section. I think this is not a wanted behavior.

cagdasalagoz:
Thank you for the feedback @PurpleSmurf2
I'm having a hard time noticing the resize issue you've mentioned. If you can create another issue for it, we can take a look at it after this one. It's a separate issue, so we shouldn't combine them in one pull request.

PurpleSmurf2:
I have tried to extract my points from your animations so that you know what I am talking about.

I created a Bugreport #12946. Let us see what will happen.

**In the first animation you can see the dancing separator:**

output

**In the second animation you can see your zoom-dancing expandable menu:**

output zoom

In this way, the start menu appears inconsistent. To make it look professional the separator needs to stop dancing and the expandable menu zoom have to disappear.

**Edit:**

At least for the first case, I have probably found out why this happens. If you look closely, you can see that as soon as the menu is expanded, the “Advanced Network Configur...” entry changes to its full name “Advanced Network Configuration” and thus takes up all of its space. 

This happens when the start menu is too small. However, you can see here that the start menu adjusts its size in the wrong direction and collapses other entries, making them partly unreadable, so it should correctly expand to the right when space is tight and not to the left.

samw-26:
@cagdasalagoz Here are [my modifications](https://github.com/samw-26/cinnamon/commit/6197103ae352a6f6f2fce5abdade0d884b0caa49) which I find make the most sense.  This makes all options just close the context menu except for properties, run with dedicated gpu, and uninstall which close the menu.

This makes it easy to add multiple favorites, panel pins, and desktop shortcuts without needing to reopen the menu.

In response to your points:

> 2. Add to desktop -> We don't have a way of notifying the user if the operation completed or not. If we leave it as it is, they > might think the click did not register. Maybe we can send a notification about it? I have no idea how the experience should be with that. So I'm leaving it as it is.
> 3. Add to panel -> Same as above.

When the user adds something to the desktop, they get feedback of their click registering by the context menu closing. They also would likely be looking at their desktop when they are using the button which will also confirm that it was successful.

When adding to panel, the user can see it was successful by the context menu closing and the icon being added to the window list.

Issue #9593, opened 2020-09-19
"Backgrounds" settings app does not accept legal filename


jwillia3:
```
 * 4.6.7-1
 * Arch Linux
 * AMD Raven Ridge (amdgpu)
 * 64 bit
 * (no relevant errors in log)
 ```

**Issue**
"Backgrounds" settings app does not accept legal filename. Specifically, a filename with %20 in it.

**Steps to reproduce**
1. Download https://raw.githubusercontent.com/elementary/wallpapers/c5a5e8208a851069bf93b3ccdab7bac158efa11f/Pablo%20Garcia%20Saldana.jpg
2. Name the file `Pablo%20Garcia%20Saldana.jpg` with the `%20`
3. Move to a directory that the "Background" app sources for images
4. Select the image thumbnail
5. The background will not change to that image; (for me, it changes to Adwaita 7680x4320)
6. Replace the `%20` in the filenames with ` `
7. Re-open the background app
8. Select the thumbnail
9. The background will be set correctly

**Expected behaviour**
Any filename works correctly

TechnologyMan101:
I just tested on Linux Mint 20 Cinnamon. It ends up reverting to the default wallpaper that came with the distro. Same as what you experienced.

IvoAndonov:
It's the percentage symbol that causes this, not the 20 in it. This also happens if you put the e-mail symbol in the name - '@'.
There's also another symbol that mixes up things: name one file whatever you want but at the end put an ampersand (&) - for example: "car&.png". Refresh the folder where your wallpapers are in the background app and either wait a minute or two with that folder open and watch what happens, or if you don't wanna wait, simply hover over the '@' image several times. This is best seen when there are at least 5 other images aside from the '@' image.

TechnologyMan101:
I tested that and I noticed that it cycles the names in my Pictures folder (I have about 30-50 images). It does apply the picture though.

IvoAndonov:
> I tested that and I noticed that it cycles the names in my Pictures folder (I have about 30-50 images). It does apply the picture though.

Exactly! Such a cycling shouldn't be happening, I just didn't know how better to describe it, that's why I mentioned it. Every symbol that is BEFORE the file extension should be perceived as text, not as "cheat codes" for the background app. The last sentence is to the devs who read this post.

cagdasalagoz:
This issue persists for the Backgrounds app. I can confirm. However, if you right-click and choose `Set as Wallpaper` in the Nemo, it works.

PR #12945, opened 2025-06-19
cs_backgrounds.py: properly encode filename URI

fredcw:
fixes: https://github.com/linuxmint/cinnamon/issues/9593


cagdasalagoz:
Thank you for looking into this.
I have just applied the patch to my local installation, and I can confirm that it resolves the issue.

PR #12958, opened 2025-06-28
Don't hide cinnamon's chrome if fullscreened app isn't focused

fredcw:
The current behaviour is to hide cinnamon's chrome on a monitor if the monitor has a fullscreened app. This commit changes that behaviour so that chrome isn't hidden unless the fullscreened app is focused on that monitor or the currently focused window is on another monitor.

Ideally, the chrome would still be visible on a monitor with a fullscreened app if that app is not the top level window even if the currently focused window is on another monitor. But I don't know how to do that other than to keep track of the most recently focused window on each monitor which would be a lot more complicated. In any case, this is a fairly minor inconsistency.

### DIFF
--- a/js/ui/layout.js
+++ b/js/ui/layout.js
@@ -637,6 +637,8 @@ Chrome.prototype = {
         global.display.connect('restacked',
                               Lang.bind(this, this._windowsRestacked));
         global.display.connect('in-fullscreen-changed', Lang.bind(this, this._updateVisibility));
+        global.display.connect('notify::focus-window', Lang.bind(this, this._updateVisibility));
+        global.display.connect('window-monitor-changed', Lang.bind(this, this._updateVisibility));
         global.window_manager.connect('switch-workspace', Lang.bind(this, this._queueUpdateRegions));
 
         // Need to update struts on new workspaces when they are added
@@ -773,18 +775,17 @@ Chrome.prototype = {
     },
 
     _updateVisibility: function() {
-        for (let i = 0; i < this._trackedActors.length; i++) {
-            let actorData = this._trackedActors[i], visible;
+        this._trackedActors.forEach( actorData => {
+            const monitor = this.findMonitorForActor(actorData.actor);
+            let visible = false;
             if (!actorData.isToplevel)
-                continue;
+                return;
             else if (global.stage_input_mode == Cinnamon.StageInputMode.FULLSCREEN) {
-                let monitor = this.findMonitorForActor(actorData.actor);
-
                 if (global.display.get_n_monitors() == 1 || !monitor.inFullscreen) {
                     visible = true;
                 } else {
                     if (Main.modalActorFocusStack.length > 0) {
-                        let modalActor = Main.modalActorFocusStack[Main.modalActorFocusStack.length - 1].actor;
+                        const modalActor = Main.modalActorFocusStack[Main.modalActorFocusStack.length - 1].actor;
 
                         if (this.findMonitorForActor(modalActor) == monitor) {
                             visible = true;
@@ -794,15 +795,20 @@ Chrome.prototype = {
             } else if (this._inOverview)
                 visible = true;
             else {
-                let monitor = this.findMonitorForActor(actorData.actor);
-                
-                if (!actorData.visibleInFullscreen && monitor && monitor.inFullscreen)
+                const focusedWindow = global.display.get_focus_window();
+                if (!actorData.visibleInFullscreen && focusedWindow && focusedWindow.is_fullscreen()
+                    && focusedWindow.get_monitor() === monitor.index) {
                     visible = false;
-                else
+                } else if (!actorData.visibleInFullscreen && monitor.inFullscreen
+                    && focusedWindow && focusedWindow.get_monitor() != monitor.index) {
+                    visible = false;
+                } else {
                     visible = true;
+                }
             }
             Main.uiGroup.set_skip_paint(actorData.actor, !visible);
-        }
+        });
+
         this._queueUpdateRegions();
     },
 


mtwebster:
I don't know about this.. at the end of the day you're now saying 'I want panels to be visible in fullscreen mode', which we definitely don't want.

I had a look at muffin, and how it determines a monitor to be in 'fullscreen' mode (we reference this with Monitor.inFullscreen in Cinnamon. It considers fullscreen 'broken' only if a *maximized* window covers the fullscreen one. That distinction doesn't make a lot of sense to me - you should have your panels regardless of whether or not a window is maximized.

I'm going to change it so that any normal window can break fullscreen mode. Maybe our layout.js code can be simplified as a result, but for now it seems to work fairly well (and our panel logic in Cinnamon is maintained).

Try it out if you can and update this PR or close it if you're satisfied with the new behavior as is but we might be able to simplify things little bit.

fredcw:
@mtwebster Yes, your muffin change is good. But the main motive for this PR is to enable the panel to be usable on a monitor when the monitor's top window is fullscreened. This is so that users can press the meta key to open the app menu and then continue to use the panel until the fullscreened window is again focused by the user. In PR #12959, this is done by focusing the desktop window (even though the desktop window is hidden by the fullscreened window) when the menu is opened over a fullscreened window so that in this PR, the fullscreened window is no longer focused and the panel therefore remains usable. This works because in this PR, the "top" window is determined not by the z order but rather by the currently focused window or, if the currently focused window is on a different monitor, the most recently focused window the monitor. While this is admittedly a sightly hacky way to do it (focus the desktop window to keep the panel usable,) I can't think of a better way to do it.

>at the end of the day you're now saying 'I want panels to be visible in fullscreen mode'

Yes, but only if the fullscreened window is unfocused and wasn't the most recently focused window on the monitor.

Now with your above change in muffin, I guess PR #12959 can be left as is (focuses the desktop window when the app menu is opened with the meta key) and this PR can be changed to ensure that cinnamon's chrome is visible in the specific circumstance of the desktop window being focused regardless of whether or not the monitor is fullscreened.

This is a seemingly small usability improvement that makes a big difference because of the convenience of being in the middle of a fullscreened game and being able to press just the meta key to use the panel while keeping the game visible, do whatever with the panel, and then click on the fullscreened game again to refocus it, causing the panel and everything else to disappear. Cinnamon's chrome would never normally be visible over a fullscreened window because it would only happen in the specific circumstance of pressing the meta key to open the app menu.

fredcw:
@mtwebster I've updated this to reflect your change in muffin. Btw, showing the panel over a fullscreened window by pressing the meta key is the default behaviour in M$ Windows I'm told.

clefebvre:
I'm going to mark this (and github.com/linuxmint/cinnamon/pull/12959) as "Next" for now, so they won't get in this release but they'll be looked at again for the release after that.

This is quite an old bug, and it definitely requires attention but Michael isn't 100% happy with the solution here and we'd like more time to look into this and explore alternative solutions.

fredcw:
Something else that may be worth considering: with just this PR, and not https://github.com/linuxmint/cinnamon/pull/12959, the user could set up a hotkey that would run a script (using wmctrl) to focus the desktop, thereby showing the panel.

<details>
<summary>here's an example script written by gemini:</summary>

```bash
#!/bin/bash

# USAGE: ./focus_desktop.sh [MONITOR_INDEX]
#        0 = Leftmost monitor
#        1 = Next monitor to the right
#        2 = Third monitor, etc.

TARGET_IDX=${1:-0} # Default to 0 (first monitor) if no number given

# 1. Get the raw list of desktop windows using -lxG to see the class name
WINDOW_LIST=$(wmctrl -lxG | grep -i "nemo-desktop")

# 2. Identify the distinct Monitor X-coordinates
#    We extract column 3 (X pos), sort them numerically (-n) and uniquely (-u).
#    This creates a clean list of monitor starting positions (e.g., "0", "2560", "5120").
#    We then pick the one corresponding to the requested index.
TARGET_X=$(echo "$WINDOW_LIST" | awk '{print $3}' | sort -nu | sed -n "$((TARGET_IDX + 1))p")

# 3. Check if a monitor exists at that index
if [ -z "$TARGET_X" ]; then
    echo "Desktop window for monitor $TARGET_IDX not found."
    exit 1
fi

# 4. Get the specific Window ID for that X coordinate
#    We look through the list again for the window that matches our Target X.
#    'exit' ensures we take the first valid match (ignoring any ghost windows).
WID=$(echo "$WINDOW_LIST" | awk -v tx="$TARGET_X" '$3 == tx {print $1; exit}')

# 5. Activate the window
if [ -n "$WID" ]; then
    wmctrl -ia "$WID"
fi
```
</details>

Edit: Alternatively, with just this PR, a menu applet could detect when it is being opened over a fullscreened monitor and if so, focus the desktop window.

mtwebster:
I was playing around with this for a bit yesterday, and testing out the behavior in Windows.

I noticed there that many fullscreen apps go minimized if you hit the windows key (especially games). That's pretty easy to implement here, and side-steps all the panel/modal garbage we're trying to deal with here.

Unfortunately it's not as nice for something like a browser, or video-player-type app, though usually these apps behave a bit better with window managers than games do. I'm sure Windows has a better system in place for deciding how to 'interrupt' fullscreened-apps of various types (game vs browser vs 'player' vs ...). It's more difficult for us. Apps can do whatever they want, especially in x11 mode.

Using 'windows that can't be maximized' as a proxy for 'poorly-behaved fullscreen games', I managed to get some reasonable behavior when the menu key is pressed by:
- Minimizing the games entirely (this matches what I saw in Windows as well)
- Un-fullscreening other apps. The drawback to this is re-focusing won't be enough to re-fullscreen them.
- Waiting for either operation to complete plus a small timeout to let focus stabilize before trying to `pushModal` and opening the menu.

I'm happy with 'game' behavior, but I'm not sure complete de-fullscreening is the greatest behavior for browsers, but minimizing isn't acceptable either. But overall I am happy because it's not depending on a lot of different conditions, and it seemed to work for every case I threw at it (this PR does not - it works ok for celluloid, hypnotix and browsers, but not many games).

Either way we're going to hold off for 6.6, but maybe this can provide some ideas - I'm not opposed to changing muffin also if it means less headache in Cinnamon. I'm going to mark the other new PRs as next also - they're pretty uncommon situations anyhow.

```diff
diff --git a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
index 9d3d1db5c..ecb8d152d 100644
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1240,8 +1240,25 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
     _updateKeybinding() {
         Main.keybindingManager.addXletHotKey(this, "overlay-key", this.overlayKey, () => {
             if (!Main.overview.visible && !Main.expo.visible) {
-                if (this.forceShowPanel && !this.isOpen) {
-                    this.panel.peekPanel();
+                const focusWindow = global.display.focus_window;
+                if (focusWindow && focusWindow.is_fullscreen()) {
+                    let sigName, windowFunc;
+
+                    if (focusWindow.requested_bypass_compositor()) {
+                        sigName = "notify::minimized";
+                        windowFunc = () => focusWindow.minimize();
+                    } else {
+                        sigName = "notify::fullscreen";
+                        windowFunc = () => focusWindow.unmake_fullscreen();
+                    }
+
+                    let notifyId = focusWindow.connect(sigName, () => {
+                        GLib.timeout_add(GLib.PRIORITY_DEFAULT, 100, () => this.menu.toggle_with_options(false));
+                        focusWindow.disconnect(notifyId);
+                    })
+
+                    windowFunc();
+                    return;
                 }
                 this.menu.toggle_with_options(this.enableAnimation);
             }
```

PR #12963, opened 2025-06-29
Add "Share pinned windows with all window lists" option for Grouped Window List applet

samw-26:
This optional feature (disabled by default) allows the user to share pinned windows in a window list with all other window lists.
- When enabled, all other window lists will gain the pinned windows of the sharing window list. 
- Adding or removing pins from the shared window list will update all other window lists.
- Pins in other window lists that are not in a sharing window list are unaffected.

image

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/applet.js
@@ -25,6 +25,11 @@ const {
 class PinnedFavs {
     constructor(params) {
         this.params = params;
+        this.params.settings.bind(
+            'share-pinned-windows', // key
+            'sharePinnedWindows', // value
+            () => this.saveFavorites() // cb
+        );
         this.reload();
     }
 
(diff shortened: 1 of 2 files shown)


fredcw:
I don't think there's any great demand for this. And every extra setting has a cost in code complexity/maintainability. It makes it just a little bit harder to find the option you're looking for and the more options there are, the higher the chance someone will be put off even looking at the options bc it's looks too daunting. IMHO, there's way too many options in this applet already.

I'm impressed with your programming skills though tbf, given that this is your first PR.

PurpleSmurf2:
I do believe that there is a need for this. If the entries become too confusing due to their quantity, I would suggest introducing a general switch that only makes the options visible to advanced users. Then Linux beginners won't feel overwhelmed and advanced users who also work productively with the system will have additional options that they can use.

JosephMcc:
I'm completely with @fredcw on this one. A ton of options have been removed from this applet in the past and they don't need to come back. Hiding under "Advanced" is not needed.

mtwebster:
My opinion, pinning apps isn't going to be an every day thing. Once you're set up with your workflow and the apps you want, you stop messing with it.

Issue #9993, opened 2021-03-20
A bug with using Cinnamon and Coverflow (3D) OR Timeline (3D) on multiple screens


86ul:
* Cinnamon version (cinnamon --version): 4.8.6
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...): Mint 20.1
 * Graphics hardware *and* driver used: AMD Radeon 5500
 * 32 or 64 bit: 64b
 * Extra: A selection of inxi -Fxxxrz run in terminal:
**System**:
  Kernel: 5.4.0-67-generic x86_64 bits: 64 compiler: gcc v: 9.3.0 Desktop: Cinnamon 4.8.6 wm: muffin 4.8.1 dm: LightDM 1.30.0 Distro: Linux Mint 20.1 Ulyssa base: Ubuntu 20.04 focal 
**Machine**:
  Type: Desktop System: Gigabyte product: X570 AORUS PRO v: -CF serial: <filter> Mobo: Gigabyte model: X570 AORUS PRO v: x.x serial: <filter> UEFI: American Megatrends v: F32 date: 01/18/2021 
**CPU**:
  Topology: 8-Core model: AMD Ryzen 7 3700X bits: 64 type: MT MCP arch: Zen L2 cache: 4096 KiB flags: avx avx2 lm nx pae sse sse2 sse3 sse4_1 sse4_2 sse4a ssse3 svm bogomips: 114985 Speed: 2158 MHz min/max: 2200/3600 MHz boost: enabled Core speeds (MHz): 1: 2158 2: 2158 3: 4314 4: 2158 5: 2195 6: 2195 7: 2196 8: 2196 9: 4316 10: 2158 11: 2196 12: 2196 13: 2196 14: 2196 15: 2158 16: 2158 
**Graphics**:
  Device-1: AMD Navi 14 [Radeon RX 5500/5500M / Pro 5500M] vendor: Gigabyte driver: amdgpu v: kernel bus ID: 0a:00.0 chip ID: 1002:7340 Display: x11 server: X.Org 1.20.9 driver: amdgpu,ati unloaded: fbdev,modesetting,radeon,vesa resolution: 3440x1440 75Hz, 3440x1440 75Hz, 1920x1080 60Hz OpenGL: renderer: Radeon RX 5500 XT (NAVI14 DRM 3.35.0 5.4.0-67-generic LLVM 11.0.0) v: 4.6 Mesa 20.2.6 direct render: Yes 

**Issue**
[...]

PR #12968, opened 2025-07-02
Add a warning message about monitor freezing with Coverflow and Timeline

gfsheppard:
Use of multiple screens with Coverflow (3D) or Timeline (3D) settings may cause screen freezing (Bug #9993). Other forum posts also suggest a similar issue.
This commit introduces a note to the System Settings > Windows setting under the Alt-Tab tab that use of Coverflow (3D) or Timeline (3D) may cause monitor freezing.

[This post](https://forums.linuxmint.com/viewtopic.php?t=341496) outlines that the bug likely originates in Coverflow, the problems with troubleshooting it, and the bug report (#9993) which has been closed as not planned.

I had the exact same thing happen with monitor freezing while using Coverflow, and no freezing when it is turned off. I propose a warning note in the settings configuration to alert new users that this may occur.


mtwebster:
I'd rather get whatever is causing freezes to be fixed, not tell people something's buggy.

PR #12980, opened 2025-07-14
chore: Rename 'Linux Kernel' to 'Kernel' in system info

ykla:
Given that Cinnamon already supports various types of UNIX operating systems, it should not be restricted to depend solely on the Linux kernel.


kacperpaczos:
Maybe it should detect different operating system kernels (e.g., Unix, Linux, Hurd) and display them correctly?

ykla:
> Maybe it should detect different operating system kernels (e.g., Unix, Linux, Hurd) and display them correctly?

That’s too troublesome, let’s leave it like this.

Issue #9801, opened 2020-12-28
Cinnamon DE does NOT remember the position (left) of second monitor when I deactivate and reactivate it


tethragon:
```
 * Cinnamon version (cinnamon --version)
   - default PPA from Linux Mint 20
     (https://launchpad.net/~linuxmint-daily-build-team/+archive/ubuntu/daily-builds).
 * Linux mint 20
 * AMD Radeon 5600XT with default AMDGPU open source driver
 * 64 bit
  ```

**Issue**
I use a dual monitor setup (please see attached image)
![6a66e52dd13faca10952dd8ef960696f38c61709 png 560f43a25f7544b8136a7d2fa44b887e](https://user-images.githubusercontent.com/39238674/103234006-4095d900-4947-11eb-82e9-1810ab1de66f.png)

My second monitor is positioned left and my main monitor right. Sometimes I don't use my second monitor and I deactivate it clicking the "activate" toggle switch. The problem is than when I activate it again, the display manager (cinnamon-settings display) does NOT remember the previous position and it defaults the position of my second monitor again to the right. So, each time I deactivate and reactivate my second monitor, I need to position it again to the left. This is counter productive and I think it should be fixed.  

**Steps to reproduce**

1) open cinnamon-settings display
2) deactivate 2nd monitor
3) click apply
4) reactivate second monitor. Now the second monitor is positioned to the right, even though it was previously set to the left.

**Expected behaviour**

cinnamon-settings display should remember the last set monitors positions.

**Other information**

I managed to overcome this issue using a hand-made script

`xrandr --listactivemonitors | grep 'DisplayPort-0' && xrandr --output DisplayPort-0 --off || xrandr --output DisplayPort-0 --left-of HDMI-A-0 --auto
`

ItzSwirlz:
I have a feeling that somewhere Cinnamon clears the monitor configuration and then applies. So here's what I want you to do.

Find ~/.cinnamon-monitors.xml. It may be in a subdirectory but find it.

First, ensure that the configuration is the way you want it. Then attach the file.

Now that the first file is uploaded, disable your monitor. This way I'll know what happens when you deactivate it.

Rename the cinnamon-monitors.xmlf ile to cinnamon-monitors-2.xml or some other way I can know this is after and then upload it.

zaatas:
[...]

PR #12995, opened 2025-07-23
Workaround proposal for monitor layout bug (#9801)

tethragon:
This PR adds a markdown file describing a workaround using udev + xrandr to restore monitor layout when a secondary display is re-enabled. It may serve as inspiration for a more permanent solution to issue #9801.


mtwebster:
This is a DisplayPort bug isn't it? 

I honestly don't see the point putting this in the source tree. It's an old and pretty widely documented issue, as are its workarounds.

PR #13025, opened 2025-08-08
Make notifications applet multi instance

fredcw:
depends on https://github.com/linuxmint/cinnamon/pull/12569

Edit: I guess it makes sense to merge these two commits separately as they do different things.

PR #13034, opened 2025-08-10
Fixed bug which causes this._panelZoneSizes is null error

samw-26:
Occurred when toggling panel edit mode when these applets were removed from a panel and that panel was removed.

Applets were still listening for panel-edit-mode changed signal and not garbage collected.


fredcw:
Nice fix. I noticed this myself in notifications applet and included a fix in PR #13025

Issue #5352, opened 2016-06-01
Mirror Panels for Multiple Displays


eygraber:
It would be awesome if I could just setup the panel on my main display and have it mirrored to all other displays.

Borillion:
You can right click on the main default panel, Select Panel Settings and add more. 
Some changes to it would be helpful, many desktops are lacking even on other desktops. We can already add our own panels. It would be nice to start off with a clone of what is set on the primary screen and then allow it to be customized. A checkbox at the simplest. Also support on the window list for knowing what screen its on as well as corresponding windows.

Definitely able to code this up here, not sure what branch to code on.

JosephMcc:
The window list already works that way. Just add another instance to your other monitor.

leigh123linux:
Doesn't the the display clone setting already do this?

Minipada:
On linux mint 19, do:

On the main panel:
Right click -> Modify panel -> Copy Applet configuration

On the second one:
Right click -> Modify panel -> Paste Applet configuration

ManIVIctorious:
I think a major limitation at the moment is, that many stock applets do not support multiple instances.
E.g. Sound, network and systray are the ones bugging me the most

jlaustill:
I think this is a very important feature.  Some of us have different monitor setups at home and at work.  I love having the ability to copy a panel and create new ones, but it is very annoying having to do that every time I plug a new monitor in.  I would like the option on any given panel to simply say "show this panel on all monitors."  Plug a new one in, it shows up, unplug it, it is destroyed.  

The mac top bar and dock do this, and it is a really important feature for the future of multi-monitor support.  I would point out that no other DE supports this that I am aware of(or I would switch to it). 

Lastly, I am a developer with decades of experience.  I would be more than willing to do the work for this panel feature if someone would be willing to help me get my dev environment for cinnamon setup.  I've never contributed to Linux Mint, but I would be glad to try :)

rphenix:
Agreed I was surprised this was an issue.  Panel copy doesn't work when majority of the system applets wont work on multiple panels you end up with a second panel that looks nothing like the primary.

icarter09:
Many thanks for submitting your feature request for Cinnamon. Your feature request has been documented and will be reviewed by the Linux Mint team.

For more information on our workflow and feature requests, read https://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/faq.html.

aaronnewcomb:
[...]

PR #13035, opened 2025-08-10
[WIP] Shared panel feature

samw-26:
# Shared panel feature

Depends on #13055, #13057 

Implementation for feature request [Mirror Panels for Multiple Displays](https://github.com/linuxmint/cinnamon/issues/5352).

The purpose of this feature is to provide an option to create shared panels which shares panel settings, applets, and applet configurations.

## Adding a shared panel

1. Right click on an existing panel and click "Add a shared panel".
2. Click on an area to add the shared panel.

You can add as many shared panels as you like by right clicking on existing shared panels and clicking "Add a shared panel" again. The option is disabled on regular panels if there are already existing shared panels.

After adding a shared panel, panel settings, applets, and applet settings will be duplicated onto the new panel.

## Adding/Removing/Moving applets

Shared panels will respond to adding, removing and moving applets. Adding/removing an applet from a shared panel will add/remove the corresponding applets on all other shared panels. Moving an applet on a shared panel will move the corresponding applets on the other shared panels as well. Moving an applet off of a shared panel is akin to removing it.

The cinnamon-settings applet menu has been changed to reflect this behavior. All shared panels will be highlighted at once and they will behave as one panel when clicking the next and previous panel buttons. This also means that the next and previous panel buttons will be hidden if you only have shared panels.

## Changing Settings

The panel settings and applet configurations on all shared panels will be kept in sync.

### Panel settings

In the panel settings menu, selecting a shared panel will highlight all shared panels. The next and previous panel buttons will treat all shared panels as one panel. The next and previous panel buttons will be disabled if only shared panels exist.

### Applet configuration

Opening the configuration menu for an applet on a shared panel will highlight all corresponding applets on all shared panels, and they will behave as one applet when going to the next and previous instances. The next and previous instance arrows will be hidden if there is only one applet of that uuid on a shared panel.

Any settings changes in the applet configuration menu of a shared panel will update the corresponding applets on all the other shared panels. If an applet setting that is not listed in the configuration menu of an applet changes (Such as the pinned favorites setting of the grouped window list applet), the corresponding applets on the shared panels will be re-instantiated to show the updated setting.

### DIFF
--- a/data/org.cinnamon.gschema.xml
+++ b/data/org.cinnamon.gschema.xml
@@ -124,6 +124,14 @@
       </description>
     </key>
 
+    <key name="shared-panels" type="s">
+      <default>'[]'</default>
+      <summary>JSON string of shared panel information</summary>
+      <description>
+        Stores the panel ids that are being shared.
+      </description>
+    </key>
+
     <key name="panels-autohide" type="as">
       <default>['1:false']</default>
       <summary>Auto-hide panel</summary>
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/JsonSettingsWidgets.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/JsonSettingsWidgets.py
@@ -147,6 +147,13 @@ class JSONSettingsHandler(object):
                 for callback in callback_list:
                     callback(key, new_value)
 
+        for key in self.settings:
+            if ("value" in self.settings[key]
+                and self.settings[key]["value"] != old_settings[key]["value"]
+                and self.notify_callback
+            ):
+                self.notify_callback(self, key, new_value)
+
     def get_settings(self):
         try:
             file = open(self.filepath)
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
@@ -14,6 +14,7 @@ try:
     from PIL import Image
     import datetime
     import time
+    import json
 except Exception as error_message:
     print(error_message)
     sys.exit(1)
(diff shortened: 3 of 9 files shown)


fredcw:
This requires all applets to be multi-instance compatible right? The following default applets are single instance:

inhibit@cinnamon.org
network@cinnamon.org
notifications@cinnamon.org
on-screen-keyboard@cinnamon.org
power@cinnamon.org
slideshow@cinnamon.org
sound@cinnamon.org
systray@cinnamon.org
xrandr@cinnamon.org

I'm not sure of the reasons each of these are single instance or how much work would be needed to make them them multi-instance. For instance, with notifications@cinnamon.org, the notifications will only show in one of the instances. I've fixed this in PR #13025 but I don't know what issues there are with the others.

My idea was to make each of these applets multi-instance first (if possible) and then implement a panel copy feature. This of course wouldn't be the same as mirroring the panels as making changes to either panel would not affect the other but it would be a lot simpler and, perhaps to some degree, it would scratch the same itch. While this is very impressive, it's also pretty complex from a maintainability perspective.

samw-26:
@fredcw Thanks for the feedback. I didn't know about the issue with notifications. I haven't had any other issues with the rest of the single instance applets so far. I agree that this feature is pretty complex, but for me at least, just a copying panel feature would still be burdensome when changing something on a panel because I would either have to manually do the change on the other panels or remove those panels and copy and paste the panels again.

Lurux:
Talking specifically about the single-instance applet problem here, I'm willing to attempt to dig through this with the goal of eventually making a "clone panel" feasible, if the team is interested.

See also https://github.com/orgs/linuxmint/discussions/1429

PR #13036, opened 2025-08-12
Menu applet: add workaround for tearing bug on categories scroll

fredcw:
This isn't a fix but a workaround for #11760

I tried the same for the applications box which has the same problem (though much less common) but it causes the app buttons to be too short which I don't know how to fix.


fredcw:
No longer necessary as category scrollbars removed in redesigned menu (bug doesn't occur when scrollbars are not shown.)

clefebvre:
It might still happen... there's a scrollbar but it's hidden.

fredcw:
Yes, it still occurs.

clefebvre:
Yes, I saw it a couple of times. I'm not sure exactly how to reproduce it yet but I've a feeling this fix will fix it.

fredcw:
No longer needed due to https://github.com/linuxmint/cinnamon/commit/8a4860a00e1f89247006fc682c29c4d06283bcad

PR #13051, opened 2025-08-21
expo.js: Force expo to show on the monitor that the mouse is on

rlatosky:
This PR helps users with multiple monitors. Currently, the expo mode only shows on the dedicated primary monitor. In my opinion it should follow the mouse for multi-monitor users, since it's a good assumption that the user is looking at the monitor where their mouse cursor is currently at. Tested and works with multiple monitors and behaves properly with only one monitor.

Below is a video from VirtualBox (with two virtual monitors) where I showcase the expo functionality launching on the screen where the mouse is:

https://github.com/user-attachments/assets/d746f9a8-859b-4a4c-bacd-413afaa95f11

Below is only a single monitor being used (via VirtualBox):

https://github.com/user-attachments/assets/ad243f58-ce44-4226-a672-c51a92fa888f

### DIFF
--- a/js/ui/expo.js
+++ b/js/ui/expo.js
@@ -195,22 +195,22 @@ Expo.prototype = {
         // when it is next shown.
         this.hide();
 
-        let primary = Main.layoutManager.primaryMonitor;
+        let current = Main.layoutManager.currentMonitor;
         let rtl = (St.Widget.get_default_direction () == St.TextDirection.RTL);
 
         let contentY = 0;
-        let contentHeight = primary.height;
+        let contentHeight = current.height;
 
-        this._group.set_position(primary.x, primary.y);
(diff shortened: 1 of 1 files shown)


mtwebster:
Hi, I don't think this is a bad change to have, and it works well, but in this case I think it should be configurable.

I use three monitors, but the center one is quite a bit larger than the others and the primary focus for me always - I personally want all exciting things to happen on *it*, and not the auxiliary monitors, no matter where my pointer happens to be.

Probably something in cs_workspaces.py - "Always show Expo on the primary monitor" - and keep the default behavior the same as currently?

rlatosky:
I added some code to configure the option whether to have Expo show on primary or the current monitor - I set it up so that, by default, Expo shows on primary:

image

PR #13054, opened 2025-08-24
sound-applet: Fix mute/unmute toggle when clicking volume icon (#13053)

### LINKED ISSUE #13053: Sound applet: Volume icon in popup only mutes, does not unmute
### Distribution

Linux Mint 22.1 (Xia)

### Package version

Cinnamon 6.4.8

### Graphics hardware in use

Device-1: Intel CometLake-H GT2 [UHD Graphics] driver: i915 v: kernel Device-2: NVIDIA GA107M [GeForce RTX 3050 Ti Mobile] driver: N/A Display: x11 server: X.Org v: 21.1.11 with: Xwayland v: 23.2.6 Drivers: modesetting, iris (gpu: i915) Renderer: Mesa Intel UHD Graphics (CML GT2

### Frequency

Always

### Bug description

Clicking the sound icon in the Cinnamon panel’s sound popup correctly mutes audio, but clicking it again does not restore/unmute sound. Expected behavior: clicking the icon should toggle between mute and unmute.

### Steps to reproduce

Click the sound applet in the Cinnamon panel.

In the popup, click the speaker icon next to the volume slider.

Observe: sound is muted.

Click the speaker icon again.

it does not unmute or restore the previous volume

### Expected behavior

Clicking the speaker icon should toggle mute/unmute (like in other desktop environments and in the full Sound Settings dialog).

### Additional information

_No response_

JayadityaGit:
This PR fixes an issue where clicking the speaker icon in the Cinnamon sound applet popup would only mute audio but not restore/unmute it.

Before: Clicking the icon muted sound, but subsequent clicks had no effect.

After: Clicking the icon now correctly toggles between mute and unmute.

Steps to test

Build and restart Cinnamon (Alt+F2 → r → Enter).

Open the sound applet from the panel.

Click the speaker icon next to the volume slider:

First click → mutes audio.

Second click → unmutes audio.

Related issue

Fixes #13053

### DIFF
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -107,6 +107,16 @@ class VolumeSlider extends PopupMenu.PopupSliderMenuItem {
             this.icon = new St.Icon({icon_name: this.app_icon, icon_type: St.IconType.FULLCOLOR, icon_size: 16});
         }
 
+        this.icon.reactive = true;
+        this.icon.track_hover = true;
+        this.icon.connect('button-press-event', (actor, event) => {
+            if (this.stream && event.get_button() === 1) {
+                this.stream.change_is_muted(!this.stream.is_muted);
+                return Clutter.EVENT_STOP;
+            }
+            return Clutter.EVENT_PROPAGATE;
+        });
+
         this.removeActor(this._slider);
         this.addActor(this.icon, {span: 0});
         this.addActor(this._slider, {span: -1, expand: true});


JayadityaGit:
I’ve noticed that while this fix works in most cases, there are rare situations where clicking the sound icon doesn’t unmute as expected. I’m still trying to identify the exact trigger, but wanted to flag it here so reviewers are aware

JayadityaGit:
https://github.com/user-attachments/assets/9ea52123-ff4b-4ad8-a7a9-c71910cdb4c7

JayadityaGit:
Hi @mtwebster , just wondering if this fix is valid and if there’s any feedback I should take into account. I’d really like to understand if I’m approaching this the right way, so I can improve my future contributions. Thanks!

mtwebster:
Hi, this looks ok - it appears the original behavior is due to the fact that you're clicking on the left end of the slider widget. The way the slider popup 'widget' is implemented, any click on that row is fed to the slider itself. And in this case, with the volume all the way down, the muted icon is used. So we're not really muting here, just turning the volume 'off'.

Even with your fix, if you're just a tiny bit off that icon, the underlying handler can still get the event (and make the volume 'off' rather than just 'muted')

I'm going to fix the implementation for the slider itself, so only clicks on the visible slider part will affect its position. This should make your icon handler more reliable.

mtwebster:
One thing to note, actors are reactive by default (you didn't have to explicitly set it - no big deal).

PR #13059, opened 2025-08-25
Feature/calendar advanced formatting

kacperpaczos:
**Status:** WIP – ready for review. So far tested only in VirtualBox.  

**Changes:**  
* added an advanced date/time formatting system for the Cinnamon calendar applet  
* integrated with system date/time settings (two-way synchronization via GSettings)  
* moved screensaver time formatting from `cs_screensaver` to `cs_calendar`  
* new `Use custom date format` switch with two modes:  
  * system presets (Linux Mint, macOS, Windows 10/11, Ubuntu, GNOME, KDE, with options for 24h, seconds, separators)  
  * custom format (text field with live preview, `strftime` support)  
* dynamic option reveal/hide (`SettingsRevealer` in `cs_calendar.py`)  
* consistent logic and ordering of options between the applet configuration and system settings  
* immediate refresh on changes (`fast-path listeners`, `GSettings.sync()`)  
* tooltip with custom format and live preview (`strftime`)  
* visual alignment of week numbers with day rows  

**Plans:**  
* add preview widgets in the applet configuration for `applet-format` and `tooltip-format`  
* review and fix remaining logic issues  
* ensure identical UI between the date applet config and `cs_calendar`  
* validate presets for different systems  

**Modified files:**  
* `files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js`  
* `files/usr/share/cinnamon/applets/calendar@cinnamon.org/settings-schema.json`  
* `files/usr/share/cinnamon/cinnamon-settings/modules/cs_calendar.py`  
* `data/org.cinnamon.gschema.xml`  
* `data/theme/cinnamon-sass/widgets/_calendar.scss`  

Requesting review – especially regarding UX, compatibility, and functionality.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_calendar.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_calendar.py
@@ -111,6 +111,9 @@ class Module:
             os_revealer_format = SettingsRevealer()
             self.format_combo = GSettingsComboBox(_("Date format style"), "org.cinnamon.applets.calendar", "os-format-type", format_style_options)
             applet_format.add_reveal_row(self.format_combo, revealer=os_revealer_format)
+            
+            # Formaty które wspierają separatory (tylko te z datą w formacie DD/MM/YYYY)
+            separator_supported_formats = ["Windows 7", "Linux Mint", "Ubuntu"]
 
             os_revealer_time = SettingsRevealer()
             self.time_format_switch = GSettingsSwitch(_("Use 24-hour time format"), "org.cinnamon.applets.calendar", "use-24h-format")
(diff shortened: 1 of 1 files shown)


fredcw:
I don't see the point of making all these settings global. If anything, "Display the date" and "display seconds" should be removed from global settings (gsettings). "Use 24h clock" and "first day of the week" kind of make sense as global settings and should be kept in "Date & Time" settings, but I think everything else should be regarded as specific to either the calendar applet or the screensaver as there's no reason to think that people would want the same display on both given the space limitation of the panel. They seem to me to be very different use cases. Also, if people want to configure the date/time in the screensaver, the intuitive place for that is in the screensaver settings as it currently is, not in Date & Time. Same with the calendar applet IMO.

mtwebster:
Honestly I'm not sure where to start with this...

- Why have cs_calendar_backup.py?
- Why are you adding a schema file for the screensaver here? Are you missing PRs for other packages?
- Events don't appear in the calendar applet for quite a while, not sure what triggers them, but they should be available immediately.
- Calendar formats are a locale/region thing, *not* static formats - your format won't be the same as someone from another country or region, regardless of the OS platform.
- I don't agree with making applet settings 'global' - it's an applet, it should keep its own settings, and the default format should be drawn by the system-defined (locale-based) values.
- Same with screensaver settings - there shouldn't really be two different 'screensaver settings' locations.
- Common practice for code comments to be in English.
- Applet 'configure' doesn't do anything now.

Overall I think this moves a lot of stuff around, but the real differences like custom formats just won't work. There's also just way too much going on here and it also seems incomplete.

PR #13074, opened 2025-09-04
mod(cinnamon-settings): use fstrings

allendema:
Modify files under /cinnamon-settings/ to use fstrings.
Did some tests locally but more tests are welcome.

Note: 
- no logic changes.
- some edge cases are still left.

Versions:
- [Python 3.12.3](https://www.python.org/downloads/release/python-3123/)
- [LM 22.2 Zara](https://blog.linuxmint.com/?p=4881)

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_themes.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_themes.py
@@ -826,7 +826,7 @@ class Module:
 
                 with open(icon_cache_path, 'w') as cache_file:
                     for theme_name, icon_path_val in icon_paths.items(): # Renamed icon_path to avoid conflict
-                        cache_file.write(f'{theme_name}:{icon_path_val}\\n')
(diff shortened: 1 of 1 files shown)


mtwebster:
Hi, this is ok for non-translated strings, but the old format needs to be maintained for anything with gettext (`_()`, etc...)

mtwebster:
For the record, I had copilot do a review not because I value its coding knowledge and prowess (I don't, really) but it helps me when I do large sets of changes (especially [repetitive things, refactoring](https://github.com/linuxmint/cinnamon/pull/13182)) to catch careless detail things.

It looks like it found a number of things (I would have skimmed right past!). You should be able to get a re-review from it after you fix things (upper right corner of the page). If it's not offering you that option, let me know I'll trigger it.

allendema:
That button seems to not be available for enduser.
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review#requesting-reviews-from-collaborators-and-organization-members

PR #13079, opened 2025-09-07
menu@cinnamon.org: If multiple menus share an overlay keybinding, prioritize the menu on the current monitor

samw-26:
When using the menu overlay keybinding, if there are menus with the same keybinding, then it will prioritize opening the menu on the current monitor (monitor with the mouse cursor).

Unfortunately, this requires a change to menu settings to only allow 1 keybind to be set to open / close the menu. With 2 possible keybinds, I could not find a way to get which key was used to trigger the callback in order to properly check for other menus with the same keybind.


mtwebster:
Just a heads-up, since 95e9d25b7c3ccc, keybinding settings will complain about duplicate keybindings *anywhere*. You can dismiss the popup without being forced to take action.

Actual behavior while using them is unchanged, but we would have to make a determination to 'ignore cases of a keybinding in multiple applet instances' or something...

mtwebster:
I don't really like that this only works for the menu applet. A number of default cinnamon applets support shortcuts, and there are lots of third-party spices that do also.

I'd prefer to consider a solution that would address it globally.
- applet registers shortcut by calling a special `KeybindingManager.addAppletKeybinding()` method (instead of the typical addHotKeyArray or whatever). This could detect a duplicate uuid and store additional `{uuid:callback}` sets, but stop short of trying to add the keybinding to muffin again.
- It binds the shortcut to its own internal callback, which will, when triggered, do the work of figuring out which applet instance's real callback should be used

mtwebster:
Implemented what I talked about above in https://github.com/linuxmint/cinnamon/commit/40e360dc20f070e95db3807f203531e4c5fdc6db

Third-party xlets will need to make minor, simple changes to take advantage of this.

PR #13091, opened 2025-09-11
Prepare for GLib 2.86.0 typelib break

3v1n0:
In newest glib GioUnix symbols are not anymore exposed in GioUnix, while cjs should use https://github.com/linuxmint/cjs/pull/131 it's still better to use GioUnix all the times.

~~More importantly though, when scanning the cinnamon headers, GioUnix-2.0 should be used or the APIs exposing `GDesktopAppInfo` won't work or they will be broken.~~ (this should not be  a problem here since you actually use `GMenuDesktopAppInfo` instead which is not exposed by `GioUnix`)

~~It's not included here, but it's *highly* recommended to now depend on GLib 2.86.0, or at least communicate clearly to distributors that they **must** recompile cinnamon on GLib 2.86.0 (and with these patches) to get it working with newer GLib version (or well with the newest gio introspection typelibs).~~

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_default.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_default.py
@@ -307,10 +307,10 @@ class DefaultTerminalButton(Gtk.AppChooserButton):
 
         while self.this_item is not None and count_up < len(apps):
             self.this_item = apps[count_up]
-            cat_val = Gio.DesktopAppInfo.get_categories(self.this_item)
-            exec_val = Gio.DesktopAppInfo.get_string(self.this_item, "Exec")
-            name_val = Gio.DesktopAppInfo.get_string(self.this_item, "Name")
-            icon_val = Gio.DesktopAppInfo.get_string(self.this_item, "Icon")
+            cat_val = GioUnix.DesktopAppInfo.get_categories(self.this_item)
+            exec_val = GioUnix.DesktopAppInfo.get_string(self.this_item, "Exec")
+            name_val = GioUnix.DesktopAppInfo.get_string(self.this_item, "Name")
+            icon_val = GioUnix.DesktopAppInfo.get_string(self.this_item, "Icon")
             # terminals don't have mime types, so we check for "TerminalEmulator" under the "Category" key in desktop files
             if cat_val is not None and "TerminalEmulator" in cat_val:
                 # this crazy if statement makes sure remaining desktop file info is not empty, then prevents root terminals from showing, then prevents repeating terminals from trying to being added which leave a blank space and Gtk-WARNING's
@@ -364,11 +364,11 @@ class DefaultCalculatorButton(Gtk.AppChooserButton):
 
         while self.this_item is not None and count_up < len(apps):
             self.this_item = apps[count_up]
-            cat_val = Gio.DesktopAppInfo.get_categories(self.this_item)
-            exec_val = Gio.DesktopAppInfo.get_string(self.this_item, "Exec")
-            name_val = Gio.DesktopAppInfo.get_string(self.this_item, "Name")
-            icon_val = Gio.DesktopAppInfo.get_string(self.this_item, "Icon")
-            comment_val = Gio.DesktopAppInfo.get_string(self.this_item, "Comment")
+            cat_val = GioUnix.DesktopAppInfo.get_categories(self.this_item)
+            exec_val = GioUnix.DesktopAppInfo.get_string(self.this_item, "Exec")
+            name_val = GioUnix.DesktopAppInfo.get_string(self.this_item, "Name")
+            icon_val = GioUnix.DesktopAppInfo.get_string(self.this_item, "Icon")
+            comment_val = GioUnix.DesktopAppInfo.get_string(self.this_item, "Comment")
             #calculators don't have mime types, so we check for "Calculator" under the "Category" key in desktop files
             if (cat_val is not None and "Calculator" in cat_val) or \
                (exec_val is not None and "alculator" in exec_val.lower()) or \
--- a/js/ui/main.js
+++ b/js/ui/main.js
@@ -78,6 +78,7 @@
 
 const Clutter = imports.gi.Clutter;
 const Gio = imports.gi.Gio;
+const GioUnix = imports.gi.GioUnix;
 const GLib = imports.gi.GLib;
 const Gtk = imports.gi.Gtk;
 const Mainloop = imports.mainloop;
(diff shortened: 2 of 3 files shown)


3v1n0:
Note that the  `Gio` -> `GioUnix` rename commit is optional here, assuming you have https://github.com/linuxmint/cjs/pull/131. ~~While it's not optional the fact that GioUnix should be used to introspect.~~

Fantu:
@3v1n0 Thanks for your work.
There are some things I can't quite understand.
https://github.com/linuxmint/cjs/pull/131 seems a "workaround" to keep retrocompatibility while with this patch you use the new APIs which would be better, or I'm wrong?
Cjs PR can be needed also with this patch if other cinnamon components or spices (applets and extensions) use removed API, right?
Searching if needed to update the glib min version for build, GioUnix seems to have been around for a long time but the documentation seems to have just changed for example from Gio.DesktopAppInfo to GioUnix.DesktopAppInfo only recently, from what I thought I saw, so I have a doubt if there was some problem with GioUnix or if it was just not well known what had been moved and deprecated.
Sorry if my doubt is stupid because I didn't understand or find it well.

3v1n0:
> https://github.com/linuxmint/cjs/pull/131 seems a "workaround" to keep retrocompatibility while with this patch you use the new APIs which would be better, or I'm wrong?

Yeah, it's more a wrapper to keep deprecated Gio.Unix* stuff to work as it used to do, so it allows all the cjs related stuff to continue working, since [Gio unix is used in other places too](https://github.com/search?q=org%3Alinuxmint%20Gio.DesktopAppInfo&type=code).

We have the same in upstream gjs.

> Cjs PR can be needed also with this patch if other cinnamon components or spices (applets and extensions) use removed API, right?

Yes, this patch is enough to get cinnamon running, but not for all the applets (that indeed requires having an updated cjs, or a workaround to inject `GioUnix.DesktopAppInfo` into `Gio` in cinnamon itself).

3v1n0:
> Searching if needed to update the glib min version for build, GioUnix seems to have been around for a long time but the documentation seems to have just changed for example from Gio.DesktopAppInfo to GioUnix.DesktopAppInfo only recently, from what I thought I saw, so I have a doubt if there was some problem with GioUnix or if it was just not well known what had been moved and deprecated.

As per [gjs change](https://gitlab.gnome.org/GNOME/gjs/-/commit/ce06bd7891bcc9c1a5c7aec78514474681d591d2) this is deprecated for a while (so I'd switch the dependency on glib 2.79.2), but we did not advertise it properly for `Gio.DesktopAppInfo`

Fantu:
Thanks for all the replies

3v1n0:
Ah, btw I saw all the dependencies using `Gio.DesktopAppInfo` in linuxmint, they're all in python and that's handled via pygobject (https://gitlab.gnome.org/GNOME/pygobject/-/merge_requests/451) already.

mtwebster:
> if other cinnamon components or spices (applets and extensions) use removed API, right?

I won't mind doing a mass edit to all third-party applets and extensions to fix this stuff when necessary.

bobby285271:
Hi, I package cinnamon on NixOS and was trying this patch with GLib 2.84.4 and pygobject 3.50.0 **(i.e. still GNOME 48)** and got this when trying to open "Preferred Applications" in cinnamon-settings:

```
  File "/nix/store/2x06wciyzljz8lfq06zkhkifz4mvan1j-cinnamon-6.4.12/share/cinnamon/cinnamon-settings/cinnamon-settings.py", line 751, in button_press
    self.side_view_nav(widget, None, category)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/2x06wciyzljz8lfq06zkhkifz4mvan1j-cinnamon-6.4.12/share/cinnamon/cinnamon-settings/cinnamon-settings.py", line 176, in side_view_nav
    self.go_to_sidepage(sidePage, user_action=True)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/2x06wciyzljz8lfq06zkhkifz4mvan1j-cinnamon-6.4.12/share/cinnamon/cinnamon-settings/cinnamon-settings.py", line 185, in go_to_sidepage
    sidePage.build()
    ~~~~~~~~~~~~~~^^
  File "/nix/store/2x06wciyzljz8lfq06zkhkifz4mvan1j-cinnamon-6.4.12/share/cinnamon/cinnamon-settings/bin/SettingsWidgets.py", line 212, in build
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/nix/store/2x06wciyzljz8lfq06zkhkifz4mvan1j-cinnamon-6.4.12/share/cinnamon/cinnamon-settings/modules/cs_default.py", line 627, in on_module_selected
    button = DefaultCalculatorButton()
  File "/nix/store/2x06wciyzljz8lfq06zkhkifz4mvan1j-cinnamon-6.4.12/share/cinnamon/cinnamon-settings/modules/cs_default.py", line 367, in __init__
    cat_val = GioUnix.DesktopAppInfo.get_categories(self.this_item)
              ^^^^^^^
NameError: name 'GioUnix' is not defined
```

After apply the following patch it works. Is that expected?

```diff
diff --git a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_default.py b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_default.py
index 4c69191d3..8763511bf 100755
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_default.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_default.py
@@ -4,7 +4,10 @@ import os
 
 from SettingsWidgets import SidePage
 from xapp.GSettingsWidgets import *
+
+gi.require_version('GioUnix', '2.0')
 from gi.repository import *
+from gi.repository import GioUnix
 
 PREF_MEDIA_AUTORUN_NEVER = "autorun-never"
 PREF_MEDIA_AUTORUN_X_CONTENT_START_APP = "autorun-x-content-start-app"
```

mtwebster:
That's odd...

I fixed the imports - https://github.com/linuxmint/cinnamon/commit/330b9ff19f33ec1e94e36048ca46011404f796b4

I'm not sure why it would fail to bring in GioUnix - maybe related to the namespace workarounds that had been in place?

We should always be explicit with imports anyhow. (... *ignores line 6*...)

Fantu:
`from gi.repository import *` is bad
@mtwebster thanks for https://github.com/linuxmint/cinnamon/commit/330b9ff19f33ec1e94e36048ca46011404f796b4

PR #13094, opened 2025-09-13
fix crash in new Gio due to API change

joequant:
This fixes a crash due to deprecated calls being
removed.


joequant:
Ah.  You've seen the problem already.  Good job.

PR #13103, opened 2025-09-17
Call on_applet_reloaded() for all instances of an Applet

klangman:
When reloading an Applet that has more than one instance, the on_applet_reloaded() function is only called on the first Applet in the definitions array that matches the UUID. This fix allows the loop to continue through all Applet definitions and call on_applet_reloaded() for all Applet instances matching the UUID.

### DIFF
--- a/js/ui/appletManager.js
+++ b/js/ui/appletManager.js
@@ -122,7 +122,6 @@ function prepareExtensionReload(extension) {
             if (!applet) continue;
             global.log(`Reloading applet: ${extension.uuid}/${applet_id}`);
             applet.on_applet_reloaded();
-            return;
         }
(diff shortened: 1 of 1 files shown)


js/ui/appletManager.js
```diff
@@ -122,7 +122,7 @@ function prepareExtensionReload(extension) {
             if (!applet) continue;
             global.log(`Reloading applet: ${extension.uuid}/${applet_id}`);
             applet.on_applet_reloaded();
-            return;
+            //return;
```
mtwebster: Just delete the line please.

klangman: Done.

PR #13153, opened 2025-10-21
desklet: Fix open config window menu action not working

anaximeno:


js/ui/desklet.js
```diff
@@ -233,7 +233,7 @@ var Desklet = class Desklet {
 
         if (!this._meta["hide-configuration"] && GLib.file_test(this._meta["path"] + "/settings-schema.json", GLib.FileTest.EXISTS)) {
             this.context_menu_item_configure = new PopupMenu.PopupMenuItem(_("Configure..."));
-            this.context_menu_item_configure.connect("activate", Lang.bind(this, this.configureDesklet));
+            this.context_menu_item_configure.connect("activate", Lang.bind(this, (e) => this.configureDesklet()));
```
mtwebster: You can remove the `Lang.bind()` wrapper entirely here. The `() => {}` handles that on its own.

See:
https://github.com/linuxmint/cinnamon/commit/8bde87c8867045b2abfb5a08b588258857277670
(I should have checked here also when I fixed applets).

anaximeno: Done!

PR #13157, opened 2025-10-22
build: drop intltool requirement

sgn:
Since Gettext 0.19.7, gettext is fully capable of working with desktop and xml file. Intltool wrapper is no longer necessary.


Fantu:
intltool is still present in debian/control

mtwebster:
Thanks, fixed.

PR #13163, opened 2025-10-24
photoframe@cinnamon.org: fix broken aspect ratio

KopfdesDaemons:
The aspect ratio of the images in the photoframe desklet has not been correct for some time.

The problem has already been discussed by users in the [forum](https://forums.linuxmint.com/viewtopic.php?t=424783).

Previously:
broken

Fixed:
fixed


anaximeno:
I wonder if there should be a settings for this, like the previews style as expanded, and this as in scale option.

mtwebster:
Fixed as part of 3d57bc7cbc9410b9a05a5c43447d690731e1f44f

PR #13167, opened 2025-10-26
fixed .desktop names for Account Details, Virtual Keyboard

### LINKED ISSUE #13123: App names not capitalized correctly
**Distribution**

Mint 22.2

**Package version**

6.4.8

**Extremely minor issue**

Every default app in Linux Mint is capitalized in Title Case. The only exceptions are the "Account details" and "Virtual keyboard" apps, which should be named "Account Details" and "Virtual Keyboard" instead.

timfeierabend:
All app names have significant words in Title Case. Two apps did follow this - "Account Details" and "Virtual Keyboard". This PR corrects this.

resolves #13123


mtwebster:
Hi, good catch - unfortunately it's probably better to leave these as they are. If we merge this, these strings will need to be re-translated.

Thanks

PR #13209, opened 2025-11-21
window-list applet: Add notification badges

fredcw:
Depends on #12569


mtwebster:
I don't know about this:
image

I ejected a usb stick, and this is the result of the 'it is safe to remove your device' notification.

For the GWL it works out when using app-grouping, but here, I'm not so sure, since we'll have this with any app that supports multiple windows.

fredcw:
@mtwebster I'm not sure there's anything we can do about this. notificationDaemon.js only creates new notification sources based on pid. I'm guessing there's not enough info in a notification to distinguish between different windows of the same app.

fredcw:
@mtwebster It appears there's no reliable or consistent way to associate a notification with any particular window. The only other options I can think of: only put the badge on the most recently active window (though this would probably be misleading more often than not and wouldn't account for multiple monitors and workspaces) or only put a badge on the leftmost or rightmost icon in the panel but this too may be potentially confusing.

PR #13220, opened 2025-11-26
menu: Add context menu actions for favorites/recents

rlbxku1r:
Add `Remove from favorites` and `Open containing folder` action for favorites/recents, because you can't do this without using Nemo.

image

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -807,7 +807,17 @@ class PathContextMenuItem extends ContextMenuItem {
         return true;
     }
 
+    static _useDBus = true;
+
     _openContainingFolder() {
+        if (!PathContextMenuItem._useDBus || !this._openContainingFolderViaDBus()) {
+            // Do not attempt to use DBus again once it's failed.
+            PathContextMenuItem._useDBus = false;
+            this._openContainingFolderViaMimeApp();
+        }
+    }
+
+    _openContainingFolderViaDBus() {
         try {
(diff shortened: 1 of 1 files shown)


mtwebster:
I think this should be in the context menu (similar to what you get when right-clicking an application). This impacts the available width of the filename and path labels, and isn't consistent with the rest of the applet.

rlbxku1r:
> I think this should be in the context menu (similar to what you get when right-clicking an application). This impacts the available width of the filename and path labels, and isn't consistent with the rest of the applet.

Changed to the context menu action.

hthienloc:
The official menu applet should be Cinnamenu. It already has this feature and is far better than the default one.

image

files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -788,15 +791,53 @@ class RecentButton extends SimpleMenuItem {
...
+                ]),
+                null,
+                Gio.DBusCallFlags.NONE,
+                1000,
+                null
+            );
+        } catch (e) {
+            global.logError(`Could not open containing folder: ${e}`);
+        }
+    }
```
mtwebster, line 855: This is nice, but only if your default file manager supports the FileManager service.

You could add a fallback that uses `Gio.app_info_launch_default_for_uri()` on the parent uri (you can use Gio.File methods to reliably get the parent uri). This won't 'select' the file but will at least follow through with opening the containing folder.

rlbxku1r: Fixed.

PR #13229, opened 2025-11-30
cs_startup: fix custom commands with spaces in path

okaestne:
Automatically add quotes if selected filepath contains spaces.

fixes https://github.com/linuxmint/cinnamon-settings-daemon/issues/363

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_startup.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_startup.py
@@ -899,6 +899,8 @@ class AppDialog(Gtk.Dialog):
 
         if response == Gtk.ResponseType.ACCEPT:
             name = chooser.get_filename()
+            if " " in name:
+                name = '"' + name + '"'
             self.command_entry.set_text(name)
 
         chooser.destroy()


files/usr/share/cinnamon/cinnamon-settings/modules/cs_startup.py
```diff
@@ -899,6 +899,8 @@ def on_browse_button_clicked(self, button):
 
         if response == Gtk.ResponseType.ACCEPT:
             name = chooser.get_filename()
+            if " " in name:
+                name = '"' + name + '"'
```
mtwebster, line 903: Can we just use GLib.shell_quote() here?

okaestne: As far as I understand the fd.o specs, we need to use double quotes. I didn't use `GLib.shell_quote` or `shlex.quote` because they both returned single quoted strings in my testing.

mtwebster: Both work, but single quotes are safer, as their contents are taken literally.

okaestne: I spent quite some time to read the fd.o spec [1], which is pretty clear:

> Arguments may be quoted in whole. If an argument contains a reserved character the argument must be quoted. The rules for quoting of arguments is also applicable to the executable name or path of the executable program as provided. 
> 
> Quoting must be done by enclosing the argument between double quotes and escaping the double quote character, backtick character ("\`"), dollar sign ("$") and backslash character ("\") by preceding it with an additional backslash character. Implementations must undo quoting before expanding field codes and before passing the argument to the executable program. Reserved characters are space (" "), tab, newline, double quote, single quote ("'"), backslash character ("\"), greater-than sign (">"), less-than sign ("<"), tilde ("~"), vertical bar ("|"), ampersand ("&"), semicolon (";"), dollar sign ("$"), asterisk ("*"), question mark ("?"), hash mark ("#"), parenthesis ("(") and (")") and backtick character ("\`"). 

This is why I explicitly tried to avoid single quotes. All `.desktop` files I found used double quotes if they used quotes.

[1] https://specifications.freedesktop.org/desktop-entry/latest/exec-variables.html

PR #13239, opened 2025-12-05
cs_user.py: Switch from pam passwd to system-auth or common-auth with login fallback

leigh123linux:
Ref:
https://bugzilla.redhat.com/show_bug.cgi?id=2418600


leigh123linux:
Confirmed to fix the issue https://bugzilla.redhat.com/show_bug.cgi?id=2418600#c15

Issue #12664, opened 2025-01-21
(regression) 6.4 allows to accidentally hibernate twice, and hibernation takes much longer


calestyo:
### Distribution

Debin unstable

### Package version

 6.4.5

### Graphics hardware in use

Intel Alder Lake

### Frequency

Always

### Bug description

Hey.

I've just upgraded on Debian from previously Cinnamon 6.2 to 6.4 and quite quickly stumbled over the following:

The "shutdown dialogue" is all new (and unfortunately not only better[0]):

Previously when I clicked "Hibernate", it went immediately into hibernation... not for some unknown reason it takes 10-20 seconds till it starts (not till it finishes) with the hibernation (which one can easily recognise on my system as the screen goes off once and CPU goes up)

Because of this, I clicked several times on "Hibernate",... eventually it did hibernate, but when I resumed this morning, it went right back into hibernation after a few seconds.<br>
First I though it crashed, but when I booted again, it resumed again.

So the bugs here are:
1) It takes considerably longer to even start with the hibernation process
2) It allows to queue hibernations, whereas it should discard any further clicks.

Thanks,
Chris.

[0] Why the GNOME style?! It was much better before when it was a proper window that could be left open, while one could still interact with other windows, and that had a title bar with which it could be moved.

### Steps to reproduce

see above

### Expected behavior

see above

fnetX:
<details><summary>Aside / reply to your footnote remark</summary>

Hi there,

while I am not involved with the Cinnamon or Linux Mint development, maybe you want to consider this from a usability perspective to understand why this behaviour change was added. I initially thought that I'd agree to your comment and that having a proper window had advantages, but upon thinking a little, I see little reason for this. On the contrary.

> It was much better before when it was a proper window that could be left open, while one could still interact with other windows, and that had a title bar with which it could be moved.

Feel free to elaborate (we could switch to a private chat or a discussion to not clutter this conversation) about your need to interact **while** the dialog is open.

In general, it is very useful to let dialogs draw all the attention, because otherwise, you end up with typical issues like:

- A dialog is opened but hides below another window (or another window pops over it) so the user doesn't find it.
- Users inadvertently open a dialog without noticing, which is especially dangerous if there is e.g. a shutdown timer, or if they inadvertently press "Enter" on the wrong focused window.
[...]

PR #13243, opened 2025-12-06
endSessionDialog: Fix hibernate action

anaximeno:
closes https://github.com/linuxmint/cinnamon/issues/12664

### Before the fix

https://github.com/user-attachments/assets/c06dbff1-8742-4d47-a732-032622af5cfd

Another issue that happens due to this, since the dialog stays open for longer and active is that as you can see in the video, I can click the other options and even the hibernate option more than once, which since the hibernation stores the state can lead to the system firing these actions (like a reboot) unecessarily (and unatended too likely) after the hibernation is resumed, like happened here:

https://github.com/user-attachments/assets/c014f0e2-cb30-4256-9018-7d6d522edabb

### After the fix

https://github.com/user-attachments/assets/a6e84f4c-ff32-4451-b17d-d00d549b9dc7

This was tested in a computer with LMDE 7.

### DIFF
--- a/js/ui/endSessionDialog.js
+++ b/js/ui/endSessionDialog.js
@@ -165,7 +165,10 @@ class EndSessionDialog extends ModalDialog.ModalDialog {
                 if (canHibernate) {
                     this.addButton({
                         label: _("Hibernate"),
-                        action: this._dialogProxy.HibernateRemote.bind(this._dialogProxy)
+                        action: () => {
+                            this._dialogProxy.HibernateRemote();
+                            this.close();
+                        }
                     });
                 }
 


anaximeno:
Also not necessarily related to this PR (this should be good enough to fix the main issue happening in my view), is that during the hibernation something that can be confusing happens which is the screensaver appearing - then the screen goes dark (which gives the impression the hibernation has happened) - then the screensaver appearing again (which gives the impression the system has had some issue during the hibernation), to then in the end the hibernation taking the effect. This is something that happens regardless of my fix, and happened before iirc, however I'm not sure what causes it, so I just wanted to leave this here as a note.

calestyo:
@anaximeno Good to see that merged… last time I tried to improve upon thee end session dialog, contribution wasn’t that welcomed.

I guess that the screen goes dark might simply be a kernel thing?!

mtwebster:
I think it's an xserver timing/race issue - I've seen this on some machines and not others, nothing really consistent about it.

PR #13247, opened 2025-12-11
notifications@cinnamon.org: Do not recreate menu on orientation change.

fredcw:
as it produces a lot of errors in ~/.xsession-errors when changing orientation.

Also disconnect signals and finalize settings on applet removal.

### DIFF
--- a/files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js
@@ -12,6 +12,7 @@ const NotificationDestroyedReason = imports.ui.messageTray.NotificationDestroyed
 const Settings = imports.ui.settings;
 const Gettext = imports.gettext.domain("cinnamon-applets");
 const Util = imports.misc.util;
+const SignalManager = imports.misc.signalManager;
 
 const PANEL_EDIT_MODE_KEY = "panel-edit-mode";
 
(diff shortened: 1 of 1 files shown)


files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js
```diff
@@ -122,7 +124,26 @@ class CinnamonNotificationsApplet extends Applet.TextIconApplet {
...
+        if (this._orientation == St.Side.BOTTOM) {
+            this.menu.addActor(this.menu_label.actor);
+            this.menu.addActor(this._maincontainer);
+            this.menu.addActor(this.clear_separator.actor);
+            this.menu.addActor(this.clear_action.actor);
+        } else {
+            this.menu.addActor(this.clear_action.actor);
+            this.menu.addActor(this.clear_separator.actor);
+            this.menu.addActor(this.menu_label.actor);
+            this.menu.addActor(this._maincontainer);
```
mtwebster, line 143: Is it ok we're using addActor for everything here but initially we use addMenuItem for some? (maybe affects cleanup/signals from the PopupMenuBase?)

fredcw: AFAICT, this doesn't seem to affect any signals. `this.menu.addActor()` is the same as `this.menu.box.add()` which leads to `this.menu.box.add_actor()` (`clutter_container_add_actor`). I was originally going to use `this.menu.box.add()` but since `addActor` was already in PopupMenuBase, I thought I may as well use that. So it seems to just reposition the actor without affecting any signals.

fredcw: Maybe I should just call the clutter method directly:
```javascript
this.menu.box.add_actor(this.menu_label.actor);
this.menu.box.add_actor(this._maincontainer);
this.menu.box.add_actor(this.clear_separator.actor);
etc.
```
It won't make any difference to signals, but maybe more readable?

mtwebster: It's ok, I was just worried about leftover signal connections, etc...

PR #13248, opened 2025-12-11
Make notifications applet multi-instance

fredcw:


mtwebster:
I'm for making this applet multi-instance, but maybe it's time we start having the applet use copies of notifications instead of handing them off (This is my lazy fault for doing it originally).

- `notify-applet-update` emitted with original notification
- applet instance calls `let myCopy = notification.clone()`, adds *that* to its list
- messageTray destroys original notification.

The Notification class already keeps the `source` and `title`, it could keep a copy of the original `body` and `params` also. The clone method could just be `return new Notification(this.source, this.title, this.body, this.params)`. It could probably be more efficient than this, just giving this as an example.

There might be some technical reason this won't work as easily as I think, but I'd feel better in the long run about having multiple instances that we didn't have to worry about juggling actors around.

fredcw:
@mtwebster That does sound more logical. I'll look into it.

fredcw:
@mtwebster The problem is that a notification's Source can't be destroyed while there are still notifications or notification clones for that Source because the DBus session needs to be kept open so that any remaining notifications can still be interactive or replaceable. So the Source or the notification itself would have to keep count of the cloned notifications so that the Source would not consider the notification dismissed until the notification's clone count is zero.

I think the current method works OK, as long as we assume that a notification will never need to be shown in two different places at the same time. However, this would not be the case if someone were to create a notifications desklet for instance.

Another method would be to separate the notification's logic from the notification's UI (actors, buttons), and just clone the UI bit.

PR #13253, opened 2025-12-12
formatted the python files using ruff + some autofixes

Noam173:


mtwebster:
Hi, this isn't something we want - it just ends up being extra noise when we're looking at code diffs with git.

Also, I'll *never* merge something with this many changes. If you want to clean up things, fine, a couple of files at a time, but I do not care what kind of quotes someone's using in python code. Other things, including, but not limited to, line lengths, we fix as we go.

Issue #13257, opened 2025-12-13
Menu-applet not works (cinnamon 6.6 Debian sid)


misterhsp:
I built Cinnamon 6.6 for Debian Sid. Everything seems to be fine as far as I can see. Only the menu applet does not work in the panel. It cannot be added again either.
This does not occur at lower resolutions, which is strange.

Looking Glass
https://pastebin.com/0HVKMySA

fredcw:
@misterhsp Can you try the above fix and see which category icons are missing?

misterhsp:
I've already done that. For me, that's fixed.

...

fredcw:
@misterhsp Do all your category names now have symbolic icons next to them? I'm just wondering what the cause of the issue may have been.

misterhsp:
Yes, the symbols are present.

...

PR #13258, opened 2025-12-13
menu@cinnamon.org: check category icon exists

fredcw:
as user defined categories can have no defined icon.

Fixes: #13257


mtwebster:
This won't happen when using the menu editor, it gives a generic icon.

Issue #13261, opened 2025-12-14
power applet: it is hard to read the battery percentage


baronleonardo:
It is very hard to read battery percentage within power applet
currently the percentage is **after** the battery name, and my battery name is long and only one space between the name and the percentage

JosephMcc:
There is an issue template for a reason. Opened issues that ignore it, will be closed.

PR #13262, opened 2025-12-14
power_applet: rearrange battery label, so the percentage is on left

baronleonardo:
old: `<battery logo> <battery name> <percentage>`
new: `<battery logo> <percentage>\t<battery name>`

example for the new result:
image

This is make it easier to see the percentage as it is beside the battery logo

solved #13261


JosephMcc:
Thinks this looks good except for the weird space between the percentage and name. No reason for that.

baronleonardo:
it was a tab I put it for extra readability, anyway I removed it
image

Issue #13270, opened 2025-12-15
Fail to load svg icons from pixmaps folder since 6.6.x


fulalas:
Old:
<img width="1290" height="866" alt="Image" src="https://github.com/user-attachments/assets/e3e430ee-0187-4b61-8fea-ebb62fda94e2" />

New:
<img width="1290" height="866" alt="Image" src="https://github.com/user-attachments/assets/6efb79bd-4084-4828-a502-4f63d3c4af9a" />

As you can see, the new application menu fails to load .svg icons. Maybe a new dependency is needed?

EDIT: it fails to load .svg (and maybe other formats) when they're in `/usr/share/pixmaps`. If they're present in `/usr/share/icons/hicolor` they load just fine.

Tested on PorteuX 2.5 dev
All Cinnamon projects build in their latest version from Github.

fredcw:
svg icons are working fine in the new menu applet. Must be something specific to PorteuX.

fulalas:
> svg icons are working fine in the new menu applet. Must be something specific to PorteuX.

Thanks for the feedback. After digging a bit further I found out that it fails to load .svg (and maybe other formats) when they're in `/usr/share/pixmaps`. If they're present in `/usr/share/icons/hicolor` they load just fine.

Can you confirm that please?

leigh123linux:
@fulalas  Report the broken app to the maintainer, svg support is optional according to the spec.


https://specifications.freedesktop.org/icon-theme/latest/#directory_layout

leigh123linux:
> Can you confirm that please?

No apps use the pixmap directory  for svg  desktop  file icons in fedora.

```
leigh@mpd-pc:~:$ dnf provides /usr/share/pixmaps/*.svg
Updating and loading repositories:
Repositories loaded.
fedora-logos-42.0.1-2.fc43.noarch : Fedora-related icons and pictures
Repo         : @System
Matched From : 
Filename     : /usr/share/pixmaps/fedora-logo-sprite.svg

fedora-logos-42.0.1-2.fc43.noarch : Fedora-related icons and pictures
Repo         : fedora
Matched From : 
Filename     : /usr/share/pixmaps/fedora-logo-sprite.svg

generic-logos-18.0.0-26.fc43.noarch : Icons and pictures
Repo         : fedora
Matched From : 
Filename     : /usr/share/pixmaps/fedora-logo-sprite.svg

[...]

PR #13272, opened 2025-12-16
Menu: Removes the icon validation logic that only checked Gtk.IconTheme

leigh123linux:
Removes the icon validation logic that only checked Gtk.IconTheme


leigh123linux:
Fixes #13270

PR #13277, opened 2025-12-17
messageTray.js: Check for pending notifications after showing a ...

fredcw:
... notification.

Notifications received while a notification was being shown were not being shown and were left pending. 

Fixes https://github.com/linuxmint/cinnamon/issues/13293


fredcw:
@apkvmrmob7-netizen I don't understand what you mean.

PR #13304, opened 2025-12-24
menu: Add placeholders for empty categories

JosephMcc:
Gives something a little nicer when Favorites or Recents are empty.

image

### DIFF
--- a/data/theme/cinnamon-sass/widgets/_base.scss
+++ b/data/theme/cinnamon-sass/widgets/_base.scss
@@ -50,6 +50,26 @@
   -pie-background-color: rgba(140, 140, 140, 0.6);
 }
 
+// placeholder
+
+.placeholder {
+  spacing: 8px;
+  padding: $base_padding * 2;
+
+  .placeholder-icon { color: $insensitive_fg_color; }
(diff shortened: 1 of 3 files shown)


clefebvre:
Very nice!

Do we want to add an extra line of text to explain what favorites are, or how to favorite something?

JosephMcc:
We could, but it seems a bit much. I'll play with that idea a little.

JosephMcc:
Now with description

image

JosephMcc:
Squashed this down and corrected a couple small errors. Still looks the same.

mtwebster:
If I have the menu a bit too narrow, the 'No Favorite Documents' gets ellipsized, but if I then mouse over that content, it is 'reactive' and triggers a relayout (which still doesn't fix anything, it's weird)... I wonder if that primary label could be multi-line instead, and make it non-reactive.

JosephMcc:
@mtwebster  both issues fixed. Title wraps properly and stays center aligned. The weird relayout thing was do to the parent container in the menu still being reactive.

Rebased and force-pushed. Should work properly now.

PR #13317, opened 2025-12-27
menu: Fix input method key event handling

rlbxku1r:
When typing something into the search box, key events should be propagated to the input method.

Fixes: linuxmint/mint22.3-beta#60


mtwebster:
This seems to work, but not consistently. Japanese (mozc) is fine but most others seem broken - it turns out their pre-edit text isn't being properly reported (we receive only an empty string, which is considered 'false' to `has_preedit()`.

I'm considering trying to work around this in inputMethod.js (which is our interface with ibus)...

mtwebster:
It looks like only sunpinyin doesn't report pre_edit, and I can't completely un-break it in Cinnamon. Most others seem to work, that I tested. Thanks!

PR #13320, opened 2025-12-28
[Next] Panel cleanup

JosephMcc:
Our panel code could use a good bit of cleanup and modernizing. There are no actual functionality changes here outside the removal of the panel corner related code, which is unused.


JosephMcc:
This creates a bug in accounting of highlighted panels. Closing temporarily until I get that fixed.

PR #13323, opened 2025-12-31
menu@cinnamon.org: hide sidebar config options when sidebar is hidden

fredcw:
Fixes: "when hiding the sidebar, the options for Avatar, Favourites and Recents don't get hidden correctly." mentioned in https://github.com/linuxmint/mint22.3-beta/issues/71


anaximeno:
The `show-favorites` and the `show-recents` aren't the options on the category menu instead of the sidebar?

fredcw:
@anaximeno You're right, thanks. I've updated the PR.

PR #13328, opened 2026-01-02
Menu applet: fix menu sidebar radius

fredcw:
$appmenu_radius (used for sidebar only) should match popup-menu-content radius ($base_border_radius * 2) so that left and right sides of menu applet are the same.

Fixes https://github.com/linuxmint/mint22.3-beta/issues/80

@JosephMcc


JosephMcc:
Well they should definitely match. I wouldn't call it window_border_radius. That doesn't even make sense. Calling larger_border_radius probably makes more sense and fits other variable naming.

fredcw:
Changed variable name to larger_border_radius.

PR #13338, opened 2026-01-05
Fix OSD level animation lag when using gesture volume control

### LINKED ISSUE #12760: Volume OSD is very laggy
### Distribution

Mint 22.1

### Package version

6.4.8

### Graphics hardware in use

Intel Corporation Raptor Lake-P [Iris Xe Graphics]

### Frequency

Always

### Bug description

After issue #12646 it seems like the volume OSD has gotten very laggy while using the touchpad gesture for changing volume, in the bottom right volume applet it moves smoothly but in the OSD it seems to move very laggy (could be because of the animation), the icon does seem to change correctly though.

### Steps to reproduce

Use the touchpad gesture for volumes with OSD enabled.

Video attached:
https://github.com/user-attachments/assets/849a0429-3164-40c6-8f99-3122621fb89d

### Expected behavior

To not lag and move smoothly just like the volume applet.

### Additional information

_No response_

screem02:
Fixes visual laggy appearance of level animation in OSD when using touchpad gesture volume control, from #12760.

### Need
The previous behavior forces 100ms LEVEL_ANIMATION_TIME transition for all level changes, even minuscule ones, causes extreme lag when using gestures. (addressable by throttling alone for now)

### Fix
- Throttled osdWindowManager.show() call in similar fashion as in zoom control class, prevent event flooding.

- Initially made other changes, but realised they were not really necessary and felt bloaty, throttling works fine.

Thanks for reviewing.

Closes #12760

### DIFF
--- a/js/ui/gestures/actions.js
+++ b/js/ui/gestures/actions.js
@@ -17,46 +17,46 @@ var make_action = (settings, definition, device) => {
         threshold = settings.get_uint("swipe-percent-threshold");
     }
     else
-        if (definition.type === GestureType.PINCH) {
-            threshold = settings.get_uint("pinch-percent-threshold");
-        }
+    if (definition.type === GestureType.PINCH) {
+        threshold = settings.get_uint("pinch-percent-threshold");
+    }
 
-        switch (definition.action) {
-            case "WORKSPACE_NEXT":
-            case "WORKSPACE_PREVIOUS":
-            case "WORKSPACE_UP":
-            case "WORKSPACE_DOWN":
-                return new WorkspaceSwitchAction(definition, device, threshold);
-            case "TOGGLE_EXPO":
-            case "TOGGLE_OVERVIEW":
-            case "TOGGLE_DESKTOP":
-                return new GlobalDesktopAction(definition, device, threshold);
-            case "MINIMIZE":
-            case "MAXIMIZE":
-            case "CLOSE":
-            case "WINDOW_WORKSPACE_NEXT":
-            case "WINDOW_WORKSPACE_PREVIOUS":
-            case "FULLSCREEN":
-            case "UNFULLSCREEN":
-            case "PUSH_TILE_UP":
-            case "PUSH_TILE_DOWN":
-            case "PUSH_TILE_LEFT":
-            case "PUSH_TILE_RIGHT":
-                return new WindowOpAction(definition, device, threshold);
-            case "VOLUME_UP":
-            case "VOLUME_DOWN":
-            case "TOGGLE_MUTE":
-                return new VolumeAction(definition, device, threshold);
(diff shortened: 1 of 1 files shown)


screem02:
hey @mtwebster , sorry for the bother, but [c8cb4ec](https://github.com/linuxmint/cinnamon/pull/13338/commits/c8cb4ecc064ffb1bfdaa55c9e236c5ac4653698e) is failing on mint22 build but the later commit is passing, also since I'm new can you check if the code is following good standards.
Thank you!

mtwebster:
Hi sorry I missed your comment, don't worry about those builds - they're meaningless for commits like this that don't touch compiled code, and there are many reasons unrelated to a pull request that can cause failures as well.

screem02:
I made a typo in my code and crashed my ui, thought something was wrong with the code so closed the pull request and checked the code, no issue here.
We can add some more good looking ui stuff but I removed the extra funcs, vars and conts I added, they felt bloaty.

js/ui/gestures/actions.js
mtwebster, line 407: Missing `this.last_time = 0`? (refer to ZoomAction for comparison)

js/ui/gestures/actions.js
```diff
@@ -320,9 +320,12 @@ var VolumeAction = class extends BaseAction {
             this.max_volume = mixer.get_vol_max_norm();
 
         this.pct_step = Math.ceil(this.max_volume / 100);
+
+        this.last_time = 0;
+        this.poll_interval = 50 * 1000;
```
mtwebster: Since the poll interval is used in two places, maybe it should just be a constant for the entire file and we get rid of poll_interval altogether. 

like `const CONTINUOUS_ACTION_POLL_INTERVAL = 50 * 1000;`

js/ui/gestures/actions.js
mtwebster, line 438: Should there be a final `Main.osdWindowManager.show()` in `end()`? It may have been skipped during the update, and the volume ends up not being the same as what the user last saw in an Osd.

screem02: added this.last_time = 0;

screem02: added the show() fnc to end, I lazily thought the visual and actual volume diff in these cases would be very small, but yeah we should try to present it cleanly.

screem02: is the change correct or should I have gone with this:
`const percentage = Math.ceil(sink.volume / this.pct_step).clamp(0, 100);
Main.osdWindowManager.show(-1, this._get_volume_icon(percentage), null, percent, false);`

mtwebster: this is ok i think.

PR #13341, opened 2026-01-06
grouped-window-list: Add PID validation in utils.py (hardening improvement)

rgfegegeegege:
### Summary

This pull request adds explicit validation of the PID argument in the helper script `utils.py` used by the grouped-window-list@cinnamon.org applet.

The script is invoked by the applet to retrieve process information and generate desktop entries for Wine applications and other window-backed programs. It receives a process ID (PID) as a command-line argument (`sys.argv[2]`) and uses it directly in string formatting to construct system commands.

While the PID is expected to be a positive integer provided by Cinnamon's internal window tracking mechanism (making malicious injection unlikely in normal operation), the current code does not perform any validation on the input.

### Security Consideration

Although no known exploitation path exists in the current architecture (Cinnamon always passes a valid numeric PID), relying solely on the caller for input correctness is not considered a security best practice. Explicit validation defends against:
- Potential future changes in how the applet invokes the script
- Unforeseen edge cases or malformed data from window manager events
- Theoretical issues if the script were reused or invoked in different contexts

This follows the principle of defense in depth and aligns with secure coding guidelines for handling external input, even when the source is generally trusted.

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/utils.py
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/utils.py
@@ -22,38 +22,43 @@ def spawn(command):
     out = process.stdout.decode('utf-8')
     return out
 
+def get_valid_pid():
+    if len(CLI) < 3 or not CLI[2].isdigit():
+        print("Invalid PID provided")
+        sys.exit(1)
+    return int(CLI[2])
+
+def read_process_cmdline(pid):
+    return spawn(f"cat /proc/{pid}/cmdline")
+
+def normalize_process(process):
+    if '.exe' in process:
+        if 'Z:' in process:
+            process = process.split('Z:')[1]
+        process = process.replace('\\', '/')
+        process = process.split('.exe')[0] + '.exe'
+        process = 'wine ' + process.replace(' ', r'\ ')
+    else:
+        process = process[:-1]  # remove trailing null byte
+
+    if process == 'python mainwindow.py':
+        process = 'playonlinux'
+
+    return process
+
 # Utility script that creates GDesktop files for Wine and other window backed applications.
 def handle_cli():
     if len(CLI) < 2 or CLI[1] != 'get_process':
         return
 
-    if len(CLI) < 3 or not CLI[2].isdigit():
-        print("Invalid PID provided")
-        sys.exit(1)
-
-    pid = int(CLI[2])
-    process = spawn(f"cat /proc/{pid}/cmdline")
-
-    if '.exe' in process:
-        if 'Z:' in process:
-            process = process.split('Z:')[1]
-
-        process = process.replace('\\', '/')
-        process = process.split('.exe')[0] + '.exe'
-        process = 'wine ' + process.replace(' ', r'\ ')
-
-    # Remove trailing null byte from /proc/.../cmdline if not a Wine app
-    if not process.endswith('.exe'):
-        process = process[:-1]
-
-    if process == 'python mainwindow.py':
-        process = 'playonlinux'
+    pid = get_valid_pid()
+    process = read_process_cmdline(pid)
+    process = normalize_process(process)
 
     try:
         proc_array = process.split('/')
         process_name = proc_array[-1].title()
 
-        # Since this is a window backed app, make sure it has an icon association.
         icons_dir = os.path.join(os.getenv('HOME'), '.local/share/icons/hicolor/48x48/apps/')
 
         if '\\ ' in process_name:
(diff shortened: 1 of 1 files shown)


mtwebster:
There's no danger here. It's all internal to Cinnamon.

I don't want to be fielding AI-generated non-issue pull requests.

rgfegegeegege:
Hey @mtwebster,

I totally get that this isn’t a real-world exploitable bug the PID comes from Cinnamon itself so there’s no way for an attacker to inject anything.

I opened the PR purely as a small hardening improvement It’s just a couple of lines to explicitly validate that the argument is actually a number which is considered best practice when dealing with command-line input (even internal). It prevents potential future headaches if the calling code ever changes or if someone reuses the script in a different context.

I wasn’t trying to push an “AI-generated non-issue” I actually reviewed the code manually understood the context and thought it was a harmless low-effort way to make things a bit more robust.

The contributing guidelines encourage community contributions so when I saw a tiny opportunity to improve something I figured a PR was the easiest way to offer it (instead of just opening an issue and making you do the work).

No hard feelings if you don’t want it I just don’t quite understand the strong reaction when it’s such a minor, safe change. Anyway thanks for maintaining Cinnamon

mtwebster:
I admit this was a bit knee-jerk, but we're already in a defensive posture over AI.

We get AI-generated pull requests, AI-generated bug reports, AI-generated *responses to comments on said bug reports and feedback I give*... I'm getting drafted into extra work because I *know* something shouldn't require the 5,000 lines of javascript to perform some trivial task, but now I have to explain and be reasonable about it.

Has this been tested via cinnamon (not just running utils.py from a terminal)?

I'll merge it during the next dev cycle.

anaximeno:
Are you sure you "actually reviewed the code manually understood the context"?

This is not proper Python syntax:
```python
if len(CLI) < 3 or not CLI[2].isdigit():
print("Invalid PID provided")
sys.exit(1)

pid = int(CLI[2])
process = spawn(f"cat /proc/{pid}/cmdline")
 ```
ref: https://github.com/linuxmint/cinnamon/pull/13341/changes#diff-17fd473951bc7e674608af98513f15cf729cbf835189029df0809bd54b651178R29-R34

---

I'm not against the use of AI in development, it's a pretty good helper and accelerator, but should be used responsibly for stuff going to production and to be used by real people: https://medium.com/@addyosmani/vibe-coding-is-not-the-same-as-ai-assisted-engineering-3f81088d5b98

rgfegegeegege:
hey @mtwebster @anaximeno
thanks again for the feedback and for planning to merge it really appreciate it
yeah the indentation got messed up in the diff totally my fault i didn't double check
i'm out right now and not at my pc so i can't fix the syntax thing immediately but i'll sort it tomorrow no worries
once the indentation is proper the CI should pass fine
thanks for being patient and for keeping cinnamon great happy to help with small stuff like this

mtwebster:
I get notifications to all PR updates... this is already getting out of hand. Your first fix would have been sufficient, though the indentation was wrong. Honestly I'm surprised AI screwed that up, more than likely you copy/pasted it wrong maybe?

But it was small and targeted.

The more you change, the more likely you'll end up creating regressions in what is, by the fact it hasn't needed to be changed in years, stable code. Sure it hasn't kept up with moving targets like python string formatting, etc... but none of that is related to 'add PID validation...'. You're not quite at 5000 lines yet but you're well on your way :).

If you want to update format strings, that's fine, but that's a separate PR, and again it should be as small as possible, targeted, and above all, *easy to review*. This PR is no longer easy to review, it basically would need to be treated as 100+ lines of new code.

rgfegegeegege:
just pushed the fix for the indentation sorry it took a bit i'm not at my main setup right now so it was a bit tricky to get it perfect

i also cleaned up a few small things while i was in there (os.path.join for paths, proc_array[-1] instead of calculating length combined the except blocks) nothing that changes behavior just makes it a bit es and brings the complexity score down so codefactor stops complaining

about @anaximeno @mtwebster and  comment from yesterday you're totally right i want to help and i'm putting real time into reviewing the code properly

i don't just throw random stuff at the repo in this project alone there were over 30  flags and i went through every single one manually

i only opened PRs for stuff that looked like actual non-best-practice patterns worth improving

in other mint repos i've seen 550+ flags and i'm doing the same manual review not just dumping noise

anyway thanks for the patience hope the new version works smooth and the checks all pass this time

cheers

mtwebster:
I understand and appreciate what you're doing, but from a maintainer standpoint, I'd rather have slightly smelly code that works and is stable than shiny perfection.

I'm sure not every project is like this, but it works for us and our smallish team, and prevents surprises later. There's nothing wrong with modernizing syntax, but I don't really want it mixed in with bugfixes unless there's no way around it.

rgfegegeegege:
im really sorry for pushing those extra commits it got out of hand and i understand it makes reviewing harder
next time ill keep it to the absolute minimum
for future stuff like potential vulnerabilities or bad practices should i open an Issue first instead of a direct PR
i prefer PRs because i can provide the exact code and it saves you work but if youd rather have issues to discuss first thats totally fine ill do that
just let me know how you prefer it so i dont create extra work

Issue #10581, opened 2022-01-07
Cinnamon menu cut and paste issues


GretasMotherInLaw:
```
 Cinnamon version 5.0.7+uma
   - Please specify if you are using the daily builds PPA - no (as far as I know - im just a humble user of mint)
 Distribution - 20.2

 **issue** - when you cut and paste in menu, it is as if you said copy and paste. When you delete the old item, it deletes both the old item and the copy!
 
steps to reproduce:
right click on the menu and you get an option to "configure"
Click "menu" up top and then click "open the menu editor"
In that menu editor, if you try to cut something within a "folder" (eg: internet) it disappears (as expected) then you go to another folder (eg office) and click paste it appears there (as expected). So far so good. Go back to internet and you will see that it has reappeared there too! Delete that and it disappears in both places forever and ever too!

Lurux:
Related: #10139
I really hope this gets fixed

Issue #10140, opened 2021-06-07
Menu editor: moving an application to a new category also pastes it into the category it had been copied from


Lurux:
* Cinnamon version: 4.8.6 (No PPA)
 * Distribution: Linux Mint 20.1
 * Graphics hardware: Nvidia GM204 [GeForce GTX 970]
 * Graphics driver: Nvidia v460.80
 * Attach [~/.xsession-errors](https://termbin.com/rvxc)

---

**Context**

Trying to move a badly placed application from one category to another

**Issue**

- When using the "cut' and "paste" buttons to move an application from one category to another, the application still appears in the category it was previously found in.

  Because of #10139, deleting the application from the older category would make it disappear forever. You need to _uncheck_ it from there

- When using the "copy' and "paste" buttons to copy an application into a new category, the application gets duplicated, now appearing once in the new category and twice in the category it had been copied from.

**Expected behavior**

- When using "cut" and "paste" to move an application from one category to another, the application should actually be deleted from the previous category

- When using "copy" and "paste" to copy an application into a new category, the application should not get duplicated

**Additional info**

Originally found while trying to move the "Freetube" Flatpak app from the "accessories" to the "internet" category.

Lurux:
Still applicable. Cut and paste duplicated the entry.

Issue #10139, opened 2021-06-07
Menu editor: Delete button is unclear and destructive


Lurux:
* Cinnamon version: 4.8.6 (No PPA)
 * Distribution: Linux Mint 20.1
 * Graphics hardware: Nvidia GM204 [GeForce GTX 970]
 * Graphics driver: Nvidia v460.80
 * Attach [~/.xsession-errors](https://termbin.com/rvxc)

---

**Context**

Trying to move a badly placed application from one category to another

**Issue**

When a single desktop file points to several categories, clicking the "delete" button in one of the categories removes it from all categories, with no option to recover other than restoring the default configuration.

This does not affect applications that got duplicated into multiple desktop files (see related issues)

**Steps to reproduce**

Cut and paste an application into a new category. Notice how the application still appears in the previous category (see related issue). Now delete it from one of the categories, and try to recover it.

**Expected behavior**

- The "delete" button should only delete the entry from the selected category, or even just uncheck it from there

- When an application has been removed from all categories, it should probably be silently moved to the "other" category, so the user can still recover it from there

**Additional info**

Originally found while trying to move the "Freetube" Flatpak app from the "accessories" to the "internet" category.

Lurux:
Delete option doesn't work on 22.3, which ig is an improvement. Would be nice to more easily be able to move applications between categories though. See also #10140

JosephMcc:
No one else really seems to have an issue with this. No plans currently to change it.

Lurux:
Probably because most people don't actually use the menu editor I reckon. Feel like it should either be disabled or set "for advanced users only" because it has many bugs and shortcomings

fredcw:
There's an open PR to fix this linked above.

Issue #9293, opened 2020-05-10
Menu Editor Bugs


cnm2020:
```
 * Mint 19.3, updated at 10th May 2020
 * 64 bit system

**Issue**

1. If using the Cut a Menu Item and Paste it into another Menu, the Cut one reappears in its original Menu, this creates a duplicate.

2. If there are two instances of a Menu Item in two different Menus, either from using Cut or Copy, if then Delete one instance of the Menu Item the other one also removes.

3. If use Delete on a Menu Item, when the Restore System Configuration is used it moves all items back to original places and reticks unticked ones but does not restore Menu Items that were deleted using Delete, ie System Info (app, menu item)

4. Some Menu Items can appear in the Menu called Other when they have been unticked if they were previously put in a new Menu and renamed in Properties, ie Firewall Configuration (app, menu item)

**Steps to reproduce**
Right click Menu, select Config, in Menu settings- Menu, Open the Menu Editor.....

**Expected behaviour**
Cut a Menu Item should remove it from the Menu when its Paste into another Menu. 

If Delete is used on a Menu Item it should not delete all other instances of that Item.

Restore System Configuration should restore any Menu Items removed by Delete.

**Comment by socketbox** [comment]
Just to update this issue, there still exists a problem with the menu editor's functionality in Cinnamon `6.4.8`:
```
[2025-07-13 17:53:47]» inxi -S
System:
  Host: Hostname Kernel: 6.11.0-29-generic arch: x86_64 bits: 64
  Desktop: Cinnamon v: 6.4.8 Distro: Linux Mint 22.1 Xia
```

When I attempt to use the _Copy_ button, it doesn't appear to do anything (based on the fact that the _Paste_ button doesn't then become available). Also, the _Edit Desktop File_ button generates an Xfce-related error:

<img width="729" height="165" alt="Image" src="https://github.com/user-attachments/assets/a58235d9-c720-486a-a7d9-d065cbd3e99f" />

PR #13346, opened 2026-01-08
Menu editor fixes

NotSirius-A:
Hello, this PR fixes some issues (see bottom) related to cutting/pasting/deleting items in the `cinnamon-menu-editor`.

Sorry for the very long description, but it took me a long time to (sort of) understand this application, so I think I should provide a detailed description so that more people can easier understand/review these changes. 

#### Commit 4b0f5cb9e5f9c7aa1cad2669275f7e52e54d5464:

This commit fixes bad UX when copying items. Whenever user clicks on the "Copy" button no visual changes occur, despite the fact that the application places the item into a buffer. This might lead the user to think that copying "doesn't work". Pasting will only activate when user selects a category from the left menu. 

To fix this I've added one line to the `on_edit_copy_activate()` function. This line simulates user reopening the category that is currently open -> the one where copying occured. This will cause the currently selected items to deselect and will activate the "Paste" button, which will provide necessary UI feedback to the user. This issue was mention in the https://github.com/linuxmint/cinnamon/issues/9293#issuecomment-3067325479

#### Commit dbc7fe6fc2c7249e65b7589ec4346509453db1d7

This commit fixes two cutting/pasting bugs. First of all the current "cut" functionality is not really a cut. It just copies an item and immediately removes it, which is not how cutting should work. Items shouldn't be deleted until they're pasted somewhere. This commit fixes this through the introduction of a new variable `self.item_to_cut`, which indicates that cutting is (not) in progress.

This commit also fixed issues with duplicate cutting, by that I mean the incorrect behavior that occurs whenever cutting non-user-made items, which causes the same item to appear where its supposed to be pasted (correct), but also in its original category (incorrect duplicate). This is only worsened by fact that deleting either of these duplicated causes both to be deleted. This is caused by the fact that the pasted item is XML `<Include>`ed into the target category, but it is also displayed in the original one, because of the `Categories` `.desktop` property.

My fix https://github.com/linuxmint/cinnamon/commit/dbc7fe6fc2c7249e65b7589ec4346509453db1d7#diff-c98f55b82ed859a581afcbd203640e9e8dd29811a0cfd3db9d133c99f1becbdcR257 to this works by altering how items are pasted. Instead of just pasting the item I think it's also required to alter its `Categories` attribute in the `.desktop` file. My code basically just overrides the default `Categories` (which cause the duplication), with the category name of where the item was pasted into. 

For example, the calculator app is by default in `Categories=GNOME;GTK;Utility;Calculator;`, after the paste into "Graphics" my code would transform `.desktop` of the pasted item to `Categories=Graphics;`. 

Unfortunately recent changes from https://github.com/linuxmint/cinnamon/commit/a692d520ce79a536e088e7aaa1aefce63d83c94b#diff-04e819d30a3cd74ec4ed5faf25823f7121c8043dc21dbe9ecfa9231b2e7e19e7R54 are unexpectedly causing more issues here, because this also makes the editor add an additional (unneeded) localized property for ex. `Categories[en_US]=Graphics;`. While this is incorrect behavior I don't think it would cause any issues for now. I figured I should leave this alone and let maintainers make a decision, because I'm sure you know better than me. 

### Issues which this PR is based on
In my opinion this PR closes
* https://github.com/linuxmint/cinnamon/issues/10581
* https://github.com/linuxmint/cinnamon/issues/10140
* https://github.com/linuxmint/cinnamon/issues/10139

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-menu-editor/cme/MainWindow.py
+++ b/files/usr/share/cinnamon/cinnamon-menu-editor/cme/MainWindow.py
@@ -77,6 +77,7 @@ class MainWindow(object):
         self.popup_menu.show_all()
 
         self.cut_copy_buffer = None
+        self.item_to_cut = None
         self.file_id = None
         self.last_tree = None
         self.main_window = self.tree.get_object('mainwindow')
@@ -352,7 +353,11 @@ class MainWindow(object):
             return
         if not isinstance(item, CMenu.TreeEntry):
             return
-        (self.cut_copy_buffer, self.file_id) = self.editor.cutItem(item)
+        (self.cut_copy_buffer, self.file_id) = self.editor.copyItem(item)
+
+        # Update focus to menu tree view, which will activate paste button and deselect copied item
+        self.on_menu_tree_cursor_changed(self.tree.get_object('menu_tree'))
+        self.item_to_cut = item
 
     def on_edit_copy_activate(self, menu):
         item_tree = self.tree.get_object('item_tree')
@@ -366,6 +371,7 @@ class MainWindow(object):
 
         # Update focus to menu tree view, which will activate paste button and deselect copied item
         self.on_menu_tree_cursor_changed(self.tree.get_object('menu_tree'))
+        self.item_to_cut = None
 
     def on_edit_paste_activate(self, menu):
         item_tree = self.tree.get_object('item_tree')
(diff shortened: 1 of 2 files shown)


fredcw:
I looked at the code a few years ago, and to me it looked half finished. It also seems unnecessarily complicated in the way it makes changes to the user's desktop-directory XML files rather than just the user's .desktop files. I think it would be better to get rid of it altogether and replace it with something else. Mozo (MATE) would be a lot of work to convert to cinnamon and also looks unnecessarily complicated, although the UI is better. ~I think the way to go is to fork MenuLibre and slim down the UI a bit so it isn't too technical.~

Edit: On second thoughts, the issue is the cut/copy/paste buttons and how they work. We could allow the user to select which categories to show an app in in the properties dialog and this would change only `Categories=` string in the .desktop file instead of adding stuff to `~.config/menus`. That way a lot of the complexity in the menu-editor code could be removed. The ability to add nested categories is also not needed since all the menu applets ignore this anyway.

PR #13347, opened 2026-01-08
[Next] util.js splitByGlyph: Use new JS native method

fredcw:
It's does exactly the same thing, just using native JS.


mtwebster:
This isn't available in mozjs 115 (Mint 22). I'll wait until we hop versions for 23.

fredcw:
@mtwebstar you're right, sorry. Requires mozjs 125. I'll close for now, and reopen when appropriate. Funny thing is, I tested it and it doesn't report any errors, but on further testing, it doesn't get all the emoji correct.

PR #13350, opened 2026-01-09
Fix SyntaxWarning in GtkSettingsEditor._get_keyfile

leigh123linux:
Move return statement out of finally block to eliminate SyntaxWarning. 
The finally block served no cleanup purpose and has been removed.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/CinnamonGtkSettings.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/CinnamonGtkSettings.py
@@ -40,7 +40,7 @@ class GtkSettingsEditor:
             keyfile.load_from_file(self._path, 0)
         except:
             pass
-        finally:
+
             return keyfile
 
     def get_boolean(self, key):


leigh123linux:
```
$ cinnamon-settings
KeybindingWidgets session type: x11
/usr/share/cinnamon/cinnamon-settings/bin/CinnamonGtkSettings.py:44: SyntaxWarning: 'return' in a 'finally' block
  return keyfile
Using pam module (python3-pampy)
```

files/usr/share/cinnamon/cinnamon-settings/bin/CinnamonGtkSettings.py
```diff
@@ -40,7 +40,7 @@ def _get_keyfile(self):
             keyfile.load_from_file(self._path, 0)
         except:
             pass
-        finally:
+
             return keyfile
```
mtwebster, line 44: I think you need to unindent this line - otherwise this method will only ever return None (on error) or fall through and  also return nothing.

mtwebster:
Fixed as part of https://github.com/linuxmint/cinnamon/commit/127fe2d5ceb411f981257ab47d9ee5e13007c052

PR #13365, opened 2026-01-12
gwl: Refresh all workspaces when the title display setting is changed

anaximeno:


anaximeno:
Replaced by https://github.com/linuxmint/cinnamon/pull/13599

PR #13417, opened 2026-01-16
Panel: Add 'Intelligent (Dodge All Windows)' autohide option

ner216:
This PR introduces a new panel visibility option called 'Intelligent (dodge all windows)'. It also renames the previous 'Intelligently hide panel' to 'Intelligent (dodge active window)'. The new option will make the panel hide if any window overlaps it, even if that window is not the active window. This option will stop the panel from popping up and down if the user is clicking between windows that may or may not cover the panel(such as if one window covers the panel and the other does not).

Changes:
- `cs_panel.py`: Added the 'dodgeall' visibility option and label and renamed the 'intel' visibility option label.
- `panel.js`: Implemented logic for the 'Intelligent (dodge all windows)' option in the `_updatePanelVisibility` function.

Testing:
- I tested the new feature on a bare-metal system running Mint 22.3 Zena.
- I tested functionality across multiple desktops as well as multiple monitors.
- I verified that all options in the panel visibility drop-down menu still function and switch properly.

I copied the code to calculate window x/y as well as the code for determining if a window overlaps the panel from the original intelligent auto-hide case. I did this with the intention of keeping the code simple. Please let me know if there is a preferable or simpler approach. This is my first contribution to Cinnamon. I have tested this on bare metal and have not encountered any bugs, but I would appreciate a closer look at the logic to ensure that it aligns with the project's standards such as performance and code formatting.


mtwebster:
Why a whole new mode? How about just a boolean option to switch between active and all windows?

ner216:
I could do that! Just to make sure I am understanding, would you like for the switch statement to be condensed so that there are only the two cases and the default like before, or would you like for the settings menu to be altered to have a toggle switch when Intelligent is selected, or both?

fredcw:
This is not an option that I've seen people request. That doesn't necessarily mean is not a good idea but we have to be careful with adding new options for a number of reasons: code complexity/maintainability, too many options can leave users confused or make the options seem too complicated or time consuming to deal with. I think we should only add options that are either vital for a few people or we're sure would be popular with a lot of people. 

Perhaps intelli hide should be replaced with dodge all?, tbh, I have no idea.

ner216:
I completely understand, no hard feelings if this feature is not good for the project. I enjoy the learning experience of making it and I understand that having more options is not always better.

ner216:
I’ve updated the PR to address both concerns. I added a toggle that appears only when 'Intelligently hide panel' is active, labeled 'Include non-focused windows' for clarity. Does this align with the boolean option you had in mind, @mtwebster . I will attach a screenshot to show how it looks.

Settings

mtwebster:
I just tried this out... can you be a bit more the 'why' for this?

If I enable that switch, but have any maximized windows on that panel's monitor, the panel never shows, even when I focus a non-maximized window there. That seems to defeat the purpose of intelligent hide I think.

ner216:
@mtwebster  (I am not sure if you will see my reply if I do not use the @, please let me know if I should not use the mention feature for this in the future.)

Of course! That is by design to keep it hidden. My intention is to prefer seeing the full content of all windows over the panel. When I have many un-maximized windows on the screen, clicking between different windows on the screen when some are overlapping the panel, the panel may 'unnecessarily' show and hide which can be distracting or bothersome if the panel covers a part of a window I would like to use/see. 

For example, if you are to move/add a panel to the top of the screen and have intelligent hide only consider the focused window, then clicking an un-maximized window will cause the panel to block the title-bar of the maximized window.

Or, If I have a Firefox window that is maximized and I am using Ctrl+f text search, then clicking on a window that does not overlap the panel will make the panel show and block the search bar.

On the contrary, when no windows overlap the panel, it will always show. This is nice to be able to see things like time or applet status at a glance without needing to mouse over to the panel.

PR #13421, opened 2026-01-16
[cinnamon/windowManager] PERF: Batch workspace-switch animations

carsteneu:
Prevent compositor "overload" when switching workspaces with many windows.

- Added animation queue with MAX_CONCURRENT_ANIMATIONS = 10 limit
- Process animations in batches instead of starting all simultaneously
- Each completed animation triggers the next from queue

Without batching, 100+ windows create 100+ parallel Clutter timelines, causing frame drops and compositor stutter.

I had this problem because I often have more than 100 windows open across 12 workspaces. (Really? Yes, really.)

With the fix, switching between workspaces is noticeably faster and smoother.

Cheers, Carsten

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,9 +14,12 @@ on:
         description: 'Start an SSH server on failure.'
         required: false
         default: false
+  issue_comment:
+    types: [created]
 
 jobs:
   build:
+    if: github.event_name != 'issue_comment'
     uses: linuxmint/github-actions/.github/workflows/do-builds.yml@master
     with:
       commit_id: master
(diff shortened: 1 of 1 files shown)


mtwebster:
Your use-case is exceptional - let's just skip the animation if there are more than so many windows open. I don't like this added complexity here. This won't ever be an issue for most users.

carsteneu:
Thats true, not so special if you think about 10 or 20, but it is also more snappy if there are only a few windows on standard 4 workspaces. I think it will also count on older hardware.

mtwebster:
I don't like the added complexity, or the effect you get with this - it's distracting to change workspaces and then have to wait extra time for the overflow to catch up.

Again I think the best answer here is to skip the animation in extreme situations - get the window list immediately in `_switchWorkspace` and add a length check to the early return for skipping the animation. I think the max could be increased to at least 20 also.

Issue #4324, opened 2015-06-16
desktop unlocked after resume


igordovgaluk:
desktop unlocked for a few seconds after resume then locks.
cinnamon 2.4.8, version Linux Mint 17.1 (with all level 3 and above updates), kernel 3.13.0-37-generic

igordovgaluk:
similar issue http://forums.linuxmint.com/viewtopic.php?f=49&t=197000

dalcde:
Do you happen to have popup menus open before locking?

zerbons:
I have this issue as well, no popup menus are open, Cinnamon 2.6.11.

linux-modder:
@dalcde, @OP, look at  the Main Settings  menu   Is   'Lock before Sleep'  enabled  if not  enable it and  retry.   with enabled this is no longer an issue  on Fedora Cinnamon   also  in xscreensaver there is  setting to  blank out  on / return form lock  could also try those.

rapgro:
Reproducible for me. After a resume from standby, the screen gets locked but after some long delay. so the evil user is able to enter some commands still. IMHO that's a security issue.
RedHat security response team requests that the screen gets locked on an standby|suspend event.
https://bugzilla.redhat.com/show_bug.cgi?id=614608
https://bugzilla.redhat.com/show_bug.cgi?id=632978
https://bugzilla.redhat.com/show_bug.cgi?id=1225229#c11

shnplr:
happens for me on 2.6.13 LM 17.2
i can usually start my web browser (i.e. click a panel shortcut) before the screen locks. I have never been able to do more than one thing before it locks.  When I unlock the browser has indeed been started.

ion201:
Issue still in Cinnamon 2.8.6

Related issues:
#3060 
#4601 

As far as I can tell, the screen is not actually locking on suspend, but instead locking after waking up. I'm not experiencing this on my desktop, but am on my dual-core pentium laptop when manually suspending or closing the lid. The time it is unlocked for is ~1-2 seconds at most. Perhaps the issue is present but not noticeable on faster hardware?

igordovgaluk:
Perhaps ... my i5 is a slow hardware

cindyq:
Also seeing this on kernel 4.2, Cinnamon - 2.6.13 -- on resume, the desktop contents flash for a couple of seconds then the lock screen shows up

gerow:
[...]

PR #13423, opened 2026-01-16
Feature/suspend lock - Ensure Desktop is locked after resume

carsteneu:
This PR solves a problem that has been annoying me for a long time—when the load is too high when opening my laptop, you could sometimes see the content for a few seconds. The problem is apparently well known. The PR solves the problem at the JavaScript level. It works pretty well on my laptop. Details are summarized here in a more officially tone:

### Summary
This module fixes a race condition where the desktop can be briefly visible after resume when the system is under heavy CPU load. It ensures the screen is always locked before suspend completes, regardless of which screensaver is installed.

### Problem
When suspending under heavy CPU load, there's a timing window where:

1. The suspend signal arrives
2. The screensaver hasn't finished rendering the lock screen yet
3. The system suspends with an unlocked display
4. On resume, the desktop is visible for a moment before the lock screen appears
5. This is a security issue as it can expose sensitive information.

### Solution
The SuspendLockManager uses systemd-logind's delay inhibitor mechanism:

1. Hold a delay inhibitor - Prevents suspend from proceeding immediately
2. Listen for PrepareForSleep signal - Detects when suspend is about to happen
3. Trigger screen lock - Calls loginctl lock-session (universal, works with any screensaver)
4. Wait for confirmation - Monitors LockedHint property via D-Bus PropertiesChanged signal
5. Release inhibitor - Allows suspend to proceed only after lock is confirmed

### Features

- Universal compatibility - Works with cinnamon-screensaver, gnome-screensaver, light-locker, xscreensaver+xss-lock, i3lock+xss-lock, swaylock, and any other compliant screensaver
- Non-blocking - Uses async D-Bus calls and subprocess handling to avoid freezing the UI
- Robust session detection - Falls back to GetSessionByPID when XDG_SESSION_ID is unavailable
- Respects user settings - Only activates when lock-on-suspend is enabled in power settings
- Timeout protection - Releases inhibitor after 5 seconds to prevent blocking suspend forever
- Clean shutdown - Properly disconnects all signal handlers in destroy()

## Defense in Depth
This acts as a safety net alongside cinnamon-settings-daemon's existing lock-on-suspend functionality, ensuring the lock always completes before suspend regardless of system load or timing issues.

### DIFF
--- a/js/ui/suspendLock.js
+++ b/js/ui/suspendLock.js
@@ -44,6 +44,10 @@ const LoginManagerIface = '\
             <arg type="s" name="session_id" direction="in"/> \
             <arg type="o" name="session_path" direction="out"/> \
         </method> \
+        <method name="GetSessionByPID"> \
+            <arg type="u" name="pid" direction="in"/> \
+            <arg type="o" name="session_path" direction="out"/> \
+        </method> \
         <signal name="PrepareForSleep"> \
             <arg type="b" name="start" direction="out"/> \
         </signal> \
@@ -75,6 +79,8 @@ var SuspendLockManager = class SuspendLockManager {
         this._sessionPath = null;
         this._powerSettings = null;
         this._propertiesChangedId = 0;
+        this._prepareForSleepId = 0;
+        this._settingsChangedId = 0;
         this._lockTimeoutId = 0;
 
         // Inhibitor file descriptor - when closed, inhibitor is released
@@ -87,6 +93,11 @@ var SuspendLockManager = class SuspendLockManager {
         // Get power settings to check lock-on-suspend
         try {
             this._powerSettings = new Gio.Settings({ schema_id: POWER_SETTINGS_SCHEMA });
+            // Listen for settings changes to manage inhibitor accordingly
+            this._settingsChangedId = this._powerSettings.connect(
+                'changed::' + LOCK_ON_SUSPEND_KEY,
+                this._onLockOnSuspendChanged.bind(this)
+            );
         } catch (e) {
             global.logWarning('SuspendLockManager: Could not load power settings: ' + e.message);
             return;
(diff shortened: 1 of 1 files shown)


carsteneu:
I found the related Bug Report #4324 i think.

jb137:
Thank you.

carsteneu:
## Related Issues

I think this PR fixes the following cinnamon-screensaver security issues:

- Fixes linuxmint/cinnamon-screensaver#373 - Screen not immediately locked after standby
- Fixes linuxmint/cinnamon-screensaver#440 - Desktop visible briefly after wake
- Fixes linuxmint/cinnamon-screensaver#406 - Flashing shows unlocked desktop after external monitor reconnect

Maybe some more

clefebvre:
Thanks for your work on this.

We found a lead yesterday to a very similar issue: https://github.com/linuxmint/cinnamon-settings-daemon/pull/442

On laptop lid-close, we're currently turning off the monitor right after locking the screen. The screen getting turned off prevents Xorg from rendering the lock screen even though it is locked. When the lid opens, the screen is already locked, but the rendering hasn't updated yet, so the screen comes back to life and we see the session windows prior to seeing the lock screen.

Suspend/hibernate already take care of turning off the screen at HW level, so we can remove the call to turning off the screen there. This fixes the issue for us for these two events.

When it comes to just locking the screen on lid-close, we'll probably need a delay or to do it async().

carsteneu:
Thank you very much for looking into this. I really appreciate your feedback and the detailed explanation. The insight that it's a rendering issue caused by the premature DPMS off is really valuable. I also thought it may be a race condition.

I independently investigated the same area while working on this and ended up approaching it from the JS side as a workaround. Your fix here addresses the actual root cause, which is for sure much cleaner.

As you mentioned, there are some edge cases, but I think > 99.9% should be solved with the new patch.

I'm happy to close my PR — your fix makes the JS-level approach largely unnecessary. I really appreciate the great work on this!

PR #13439, opened 2026-01-18
Blacklist icon themes only

City-busz:
While the Adwaita icon theme is incomplete and doesn't support Cinnamon, the Adwaita GTK theme and the cursor theme has no such issues. Also, while the HighContrast icon theme is incomplete, the HighContrast GTK theme works fine, there is no reason to hide it. The other blacklisted themes are icon-only themes.


Initu-Castilhos:
On Arch Linux and its derivatives, the Adwaita GTK theme is available in the Cinnamon theme settings.

Issue #12721, opened 2025-02-06
Auto-hiding panel is temporarily stuck down, on wake


LinuxOnTheDesktop:
### Distribution

Mint 22.1

### Package version

6.4.6

### Graphics hardware in use

Onboad AMD; Mesa

### Frequency

Only occasionally

### Bug description

My computer (a Framework laptop - I can give screen details if necessary) awoke and I unlocked it by entering my password. Subsequently the desktop appeared, but my auto-hiding top panel was down and - for details see below - stayed down.

### Steps to reproduce

**Prerequisites**

- Possibly: have a particular monitor.

- Have an autohiding top panel (but panels on other sides might be affected, and so might 'intelligently hiding' panels).

**Actual steps**

1) Wake the computer from sleep. Enter password.

2) Observe that the panel is down - even though the pointer is far from it - and does not go up until some ten seconds have elapsed and/or (I am unsure what restored normality) one move's the pointer (in some particular way?). (When I experience the problem, the computer was responding; for, my dock (Plank) was working. But putting the pointer on the panel and then removing the pointer did not, I think cause the panel to hide, or if it did then it did not do so quickly.)

### Expected behavior

If an auto-hiding panel must drop down on wake (and why must it?) then it should quickly go up again, if the pointer is not within the trigger zone.

### Additional information

The situation made me think that the computer had hung.

**Looking Glass log**

```
info t=2025-01-17T23:18:02Z Cinnamon.AppSystem.get_default() started in 0 ms
info t=2025-01-17T23:18:02Z loading user theme: /usr/share/themes/Mint-Y-Dark-Teal/cinnamon/cinnamon.css
info t=2025-01-17T23:18:02Z added icon directory: /usr/share/themes/Mint-Y-Dark-Teal/cinnamon
info t=2025-01-17T23:18:02Z NetworkManager agent: enabled
info t=2025-01-17T23:18:02Z Polkit agent: enabled
info t=2025-01-17T23:18:02Z SSH agent: enabled
info t=2025-01-17T23:18:02Z loaded at Fri Jan 17 2025 23:18:02 GMT+0000 (Greenwich Mean Time)
info t=2025-01-17T23:18:02Z ExtensionSystem started in 0 ms
info t=2025-01-17T23:18:02Z DeskletManager started in 0 ms
info t=2025-01-17T23:18:02Z SearchProviderManager started in 0 ms
[...]

PR #13444, opened 2026-01-19
Fix: Prevent TypeError in network applet when devices is undefined

utilisateursrichard:
This PR fixes a TypeError that appears in logs (and potentially affects UI rendering) when a connection is removed while `devices` is undefined.

**The issue:**
In `_connectionRemoved`, the code accesses `devices.length` without checking if `devices` exists. This can happen during specific wake-from-sleep scenarios or network resets.

**The fix:**
Added a null check (`if (devices)`) before iterating over the devices list.

**Verification:**
Tested locally on Linux Mint. The error no longer appears in logs when network state changes.

issue found here : https://github.com/linuxmint/cinnamon/issues/12721

### DIFF
--- a/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
@@ -2297,19 +2297,16 @@ CinnamonNetworkApplet.prototype = {
 
         this._updateIcon();
     },
-
     _connectionRemoved: function(client, connection) {
         let pos = this._connections.indexOf(connection);
         if (pos != -1)
             this._connections.splice(pos);
-
         let section = connection._section;
-
         if (section == NMConnectionCategory.VPN) {
             this._devices.vpn.device.removeConnection(connection);
             if (this._devices.vpn.device.empty)
                 this._devices.vpn.section.actor.hide();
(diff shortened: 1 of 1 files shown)


utilisateursrichard:
the issue was found in https://github.com/linuxmint/cinnamon/issues/12721 btw

files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
```diff
@@ -2309,10 +2309,13 @@ CinnamonNetworkApplet.prototype = {
             this._devices.vpn.device.removeConnection(connection);
             if (this._devices.vpn.device.empty)
                 this._devices.vpn.section.actor.hide();
-        } else if (section != NMConnectionCategory.INVALID) {
-            let devices = this._devices[section].devices;
-            for (let i = 0; i < devices.length; i++)
-                devices[i].removeConnection(connection);
+            } else if (section != NMConnectionCategory.INVALID) {
```
mtwebster: Please fix the indentation of this line, it should be aligned with the `if` block for the VPN category.

utilisateursrichard: This should be fixed now. Sincerely sorry about that  I'm still fairly new to this.

Issue #13440, opened 2026-01-18
Recent files "Open with" functionality missing in 22.3


JChristensen:
### Distribution

Mint 22.3

### Package version

6.6.5

### Graphics hardware in use

Radeon RX 580

### Frequency

Always

### Bug description

Prior to 22.3, one could go to "Recent files" in the application menu, then right-click to reveal an "Open with" menu which allowed selecting the application to open the file with. This functionality is missing in 22.3 with Cinnamon 6.6.5. I found this feature to be very useful.

### Steps to reproduce

1. Click to open application menu.
2. Hover mouse over "Recent Files".
3. Hover mouse over one of the files.
4. Right-click.
5. Observe no "Open with" menu.

### Expected behavior

"Open with" menu appears as in the attached.

![Image](https://github.com/user-attachments/assets/3109ed40-9a4d-425e-9851-db9107c1b9c5)

PR #13454, opened 2026-01-20
FIX: Restore "Open with" context menu for recent/favorite files

carsteneu:
The "Open with" functionality was lost when commit d1ea12a27 ("menu: Simplify path buttons") refactored RecentButton/FavoriteButton into the generic PathButton class.

This fix adds context menu support to PathButton:
- Enable withMenu for type='recent' and type='favorite'
- Add populateMenu() method with "Open with" functionality
- Add hasLocalPath() helper for URI validation

The menu provides:
- Default application option
- List of compatible applications
- "Other application..." (launches nemo-open-with)

Fixes: linuxmint/cinnamon#13440


mtwebster:
Needs refactor

Issue #13414, opened 2026-01-16
mintmenu: unable to hide 'Templates' folder


Pol853:
### Distribution

Mint 22.3

### Package version

6.6.4

### Graphics hardware in use

Integrated AMD graphics

### Frequency

Always

### Bug description

In the sidebar of the new menu (since 22.3), it is not possible to hide the 'Templates' folder. I have the templates folder pinned to the sidebar in Nemo, which makes it show up in the menu sidebar (see screenshot 1). The menu configuration utility does not have any toggle to show/hide the templates, unlike it does for downloads, pictures, etc. That means the only way to not show templates in the menu is to hide 'Other bookmarks', which also hides all my other nemo bookmarks (which I do want), or to remove 'Templates' from the sidebar in nemo (screenshot 2). I would suggest adding a toggle to the menu configurator to allow showing/hiding templates.

Shot 1:
<img width="2580" height="1720" alt="Image" src="https://github.com/user-attachments/assets/ae852481-62ff-406b-8b45-c79851998a73" />

Shot 2:
<img width="2580" height="1720" alt="Image" src="https://github.com/user-attachments/assets/71914c35-7031-4f77-a175-cd43657894cd" />

### Steps to reproduce

N/A

### Expected behavior

I should be able to toggle show/hide for the templates folder like I can for the other default home folders.

### Additional information

Copied from https://github.com/linuxmint/mint-themes/issues/530 which I filed on the wrong repo.

PR #13456, opened 2026-01-20
UPD: Clarify bookmarks toggle description in menu applet

carsteneu:
Changed "Other bookmarks" to "Custom folders and bookmarks" with tooltip explaining it includes all user-pinned folders from Nemo (including Templates, if pinned).

This makes it clearer that the toggle controls ALL custom bookmarks, not just "other" ones, addressing user confusion (including me).

Related: linuxmint/cinnamon#13414

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/settings-schema.json
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/settings-schema.json
@@ -154,7 +154,8 @@
  "show-bookmarks" : {
     "type" : "switch",
     "default" : true,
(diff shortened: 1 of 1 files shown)


mtwebster:
These are just bookmarks though, all of them. They also technically have nothing to do with Nemo (and we usually try to stick with pronouns - 'file manager', etc...)

Why don't we just add Templates as a toggle in settings that defaults to disabled? We've already included every other XDG special directory besides Public Share, and we exposed templates a bit more in the last release, it could be useful here to some.

Then we can leave the description alone.

Issue #13422, opened 2026-01-16
User image in menu is displaying incorrectly


Davethepitt:
### Distribution

Mint 22.3

### Package version

6.6.5

### Frequency

Always

### Bug description

If the user image is in portrait mode (higher than it is wider) then the image is displayed incorrectly in the menu. The top and bottom of the images leak above the menu and below the username. 

### Steps to reproduce

Set user image to a picture which is higher than it is wider (ie 1000 X 1500) and open the cinnamon menu.

### Expected behavior

The image either needs to be cropped to a square, fixed to a height with a scalable width or the menu needs to grow to accomodate the image.

PR #13458, opened 2026-01-20
FIX: Add overflow hidden to user avatar to prevent clipping issues

carsteneu:
Prevents portrait images from leaking outside the circular avatar bounds.

Fixes: linuxmint/cinnamon#13422


mtwebster:
Did you test this?  The 'overflow' property isn't supported. I've yet to see any overflow, either. I think the theme rendering is handling this fine already.

PR #13459, opened 2026-01-20
Fix: Panel shifts when clock seconds change (#13336)

### LINKED ISSUE #13336: The panel shifts left and right every time the seconds change on the clock + panel flickers when switching keyboard layouts
### Distribution

EndeavourOS

### Package version

Cinnamon 6.6.3

### Graphics hardware in use

_No response_

### Frequency

Always

### Bug description

There are two related visual issues with the Cinnamon panel regarding element positioning and redrawing:

1. **Clock shifts left and right**: When "Show seconds" is enabled in the Clock applet, causes adjacent panel applets and icons to shift left and right every second. The panel elements do not stay in a fixed position.
2. **Keyboard Layout Flicker**: When switching between keyboard input layouts (e.g., English to another language), the panel elements located to the left of the Keyboard Layout applet flicker momentarily during the switch.

Both issues cause noticeable visual instability of the panel and negatively affect the user experience.

### Steps to reproduce

Issue 1 (Clock shifts left and right):

1. Left-click on the Clock applet in the panel and select "Date and Time Settings".
2. Enable the option "Display seconds".
3. Observe the panel icons adjacent to the clock as the time changes.

Issue 2 (Layout Flicker):

1. Ensure the "Keyboard" applet is added to the panel.
2. Ensure you have at least two input sources configured.
3. Switch the input language (either by clicking the icon or using a keyboard shortcut).
4. Observe the panel icons positioned after (to the left of) the keyboard indicator.

### Expected behavior

1. **Clock applet**: Enabling seconds in the clock should not cause the panel to move or shift.
2. **Keyboard applet**: Switching the input layout should only update the language icon/text without causing the surrounding panel elements to flicker or redraw visibly.

The panel should remain visually stable in all situations.

### Additional information

https://github.com/user-attachments/assets/b741fd3b-7ef3-4a72-8c32-ba53bdac0ede

carsteneu:
Use tabular figures (font-feature-settings: "tnum") for the clock label to ensure all digits have equal width. This prevents the panel from shifting when the time changes (e.g., from "9:59" to "10:00" or when seconds update). I think its 

Fixes linuxmint/cinnamon#13336

### DIFF
--- a/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
@@ -22,6 +22,10 @@ class CinnamonCalendarApplet extends Applet.TextApplet {
     constructor(orientation, panel_height, instance_id) {
         super(orientation, panel_height, instance_id);
 
+        // Use tabular figures to prevent panel shifting when clock updates (issue #13336)
+        // This ensures all digits have equal width (e.g., "1" same width as "0")
+        this._applet_label.set_style('font-feature-settings: "tnum"');
+
         this.setAllowedLayout(Applet.AllowedLayout.BOTH);
 
         try {


mtwebster:
This doesn't do anything for me. I think it's already enabled for most fonts.

But also, it seems like what you really want is to revert this change:
https://github.com/linuxmint/cinnamon-desktop/commit/f1f82b35b48c2ebeca000caa2ef47d3b4d6b210c

I don't disagree that having everything on your panel shift can be annoying, but to me, being forced to stare at extra spacing within the clock label is worse.

image
vs

image

An ideal fix here would be for the label to have fixed size larger than maximum width the label could ever be based on the current format.

example:
https://github.com/mtwebster/Cinnamonitor/blob/master/applet.js#L102-L113

Then the layout could resize itself internally, without affecting the rest of the panel. This would be a lot more complex than my example, however - you need to account for locale, days of the week, etc...

IT-Tool:
The fix has still not been implemented, or it is not working.
https://github.com/linuxmint/cinnamon/issues/13336#issuecomment-4291131532

PR #13466, opened 2026-01-22
FIX: Prevent crash when window icon texture is null (#13461)

### LINKED ISSUE #13461: Installed WebApp from browser crash after a while
### Distribution

Linux Mint 22.3

### Package version

Cinnamon 6.6.5

### Graphics hardware in use

CPU Integrated: AMD Ryzen 5 7530U

### Frequency

Quite often

### Bug description

After using WhatsApp Web installed from Brave Browser for a while, the desktop environment crashes. The same thing happens if I use it from the “WebApps” app included in Linux Mint. Apparently, it happens when trying to open a chat after having the app open for a while.

### Steps to reproduce

Install WhatsApp from Brave Browser.
Keep the app running in the background for a while.
Open a chat (sometimes opening an image causes the desktop environment to crash).

### Expected behavior

Web apps installed from the browser should not suddenly crash after a while.

### Additional information

I recently updated Linux Mint to version 22.3, including the desktop environment.
I previously had Linux Mint 22.2.
I have already disabled all third-party applets that do not come with Cinnamon by default.
This is the log obtained from the Crash Reports section in the System Information app: https://termbin.com/pg66

carsteneu:
Fixes #13461

Cinnamon crashes when using WebApps (e.g. WhatsApp installed from Brave Browser) after the app has been running for a while.

When muffin emits a notify::icon signal during idle_update_icon(), the window-list applets call create_icon_texture_for_window() which can return null in race conditions (e.g., when a WebApp window is being destroyed or its icon is unavailable).

Previously, this null value was passed directly to set_child(), causing a SIGSEGV in _clutter_actor_queue_only_relayout(self=0x0).

This fix adds a null check after create_icon_texture_for_window() and falls back to a default icon, preventing the crash. 

Tried it with that fix - no crash detected.

Affected applets:

[window-list@cinnamon.org](mailto:window-list@cinnamon.org)
[grouped-window-list@cinnamon.org](mailto:grouped-window-list@cinnamon.org)

### DIFF
--- a/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
+++ b/files/usr/share/cinnamon/applets/grouped-window-list@cinnamon.org/appGroup.js
@@ -293,7 +293,9 @@ class AppGroup {
             } else {
                 icon = this.groupState.app.create_icon_texture(this.iconSize);
             }
-        } else {
+        }
+
+        if (!icon) {
             icon = new St.Icon({
                 icon_name: 'application-default-icon',
                 icon_type: St.IconType.FULLCOLOR,
--- a/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/window-list@cinnamon.org/applet.js
@@ -818,9 +818,13 @@ class AppMenuButton {
 
         let icon = this.app ?
             this.app.create_icon_texture_for_window(this.icon_size, this.metaWindow) :
-            new St.Icon({ icon_name: 'application-default-icon',
+            null;
+
+        if (!icon) {
+            icon = new St.Icon({ icon_name: 'application-default-icon',
                 icon_type: St.IconType.FULLCOLOR,
                 icon_size: this.icon_size });
+        }
 
         let old_child = this._iconBox.get_child();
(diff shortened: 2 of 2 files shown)


mtwebster:
Fixed as part of https://github.com/linuxmint/cinnamon/commit/d4166a12f33852e0af8b0bf1106201fd34e669cc

edit: please double-check for me - this was mainly to fix wayland sessions, but may not completely have fixed the issue in x11.

carsteneu:
Thank you very much for your Feedback - I tested your fix in X11 (Linux Mint 22.3 VM) with aggressive stress tests:

Tests:

30x rapid Firefox launch/kill with spiegel.de (heavy content page)
30x multi-app stress (calculator + nemo simultaneous)
Logs: 0 segfaults, 0 crashes
Result: ✅ Your fix works perfectly in X11.

My PR added a defensive null-check after `create_icon_texture_for_window()`. Your architectural fix (avoiding the call for window-backed apps) addresses the root cause and is from my point of view absolutly sufficient.

Happy to close PR #13466 since your fix covers both Wayland and X11. Let me know if you prefer we should keep the additional null-check as defense-in-depth.

mtwebster:
I don't think we need this now

PR #13467, opened 2026-01-22
FIX: Fix clock rendering artifacts on time update

### LINKED ISSUE #13441: clock is often renderred incorrectly
### Distribution

22.3

### Package version

6.6.5

### Graphics hardware in use

Advanced Micro Devices, Inc. [AMD/ATI] Navi 33 [Radeon RX 7700S/7600/7600S/7600M XT/PRO W7600]

### Frequency

Quite often

### Bug description

The last digit of the clock is often renderred incorrectly and appears with extra pixels on the right side.

Image

### Steps to reproduce

Remove the "show desktop" thing next to the clock and wait til the time jumps to the next minute, maybe two minutes.

### Expected behavior

clock without extra pixel

### Additional information

When I hover over the clock is good again or when I take a screenshot.

carsteneu:
Forces actor relayout after clock label update to prevent rendering artifacts 
(ghost pixels from previous digits remaining on screen).

**Tests**
- Tested in VM with Linux Mint 22.3 Cinnamon 6.6.5
- Observed clock transitions before/after fix
- Fix successfully prevents visual artifacts during time updates

The fix calls `this.actor.queue_relayout()` after `set_applet_label()` 
to ensure complete redraw of the clock actor.

Fixes #13441

### DIFF
--- a/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js
@@ -274,6 +274,8 @@ class CinnamonCalendarApplet extends Applet.TextApplet {
         }
 
         this.set_applet_label(label_string);
+        // Force relayout to fix rendering artifacts when clock updates (issue #13441)
+        this.actor.queue_relayout();
 
         let dateFormattedTooltip = this.clock.get_clock_for_format(DATE_FORMAT_FULL).capitalize();
         if (this.use_custom_format) {


mtwebster:
I can't reproduce the issue, can you?

This seems like an issue with rendering/shadows. Queuing up this extra layout call any time something changes is likely redundant.

djphazer:
@carsteneu Did this actually fix the issue for you?

I and several others are still having this problem. Perhaps this PR should be reopened and tested.

Issue #12803, opened 2025-03-22
Menu very small after clicking resize edge of menu


chocolateimage:
### Distribution

Mint 22.1

### Package version

6.4.8

### Graphics hardware in use

Intel Corporation Raptor Lake-P [Iris Xe Graphics]

### Frequency

Always

### Bug description

When clicking on the menu edge without resizing, then after closing and opening the menu, the menu would be very small. 

### Steps to reproduce

1. Restart Cinnamon because the bug always happens once per cinnamon lifetime
2. Open the start menu
3. Click on the corner of the menu where you would usually resize, but do not actually resize
4. Close the menu
5. Open it back up
6. Menu is very small

### Expected behavior

That the menu would stay the same size.

### Additional information

I provided a video:

https://github.com/user-attachments/assets/bc28deb5-0d8a-4646-9bfa-0ab2c73f4e60

fredcw:
Cause found. I'm on it. Thanks.

Zerophase:
I'm having that issue too. I did not resize it though. After updating to 6.4.9 it randomly shrank.

PR #13469, opened 2026-01-22
FIX: Menu applet size initialization order

### LINKED ISSUE #12972: menu@cinnamon.org - shrunk itself out of nowhere...
### Distribution

22.1

### Package version

6.4.8

### Graphics hardware in use

NVIDIA GeForce GTX 1050 Ti

### Frequency

Always

### Bug description

Just a regular day, started up my machine, unsuspectingly, and then this happens:
Image

### Steps to reproduce

click the menu button, and ta-da...

### Expected behavior

Mint and Cinnamon in general should be more reliable than this.

### Additional information

Image

carsteneu:
Fixes #12972

The menu applet wasn't properly sized on first open because popup-width and popup-height settings were bound AFTER the PopupResizeHandler was created. This caused the resizer's get_user_width/height callbacks to be invoked before the values were defined.

The fix moves the settings bindings for popup-width and popup-height before the PopupResizeHandler creation, ensuring the values are available when needed.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1146,6 +1146,12 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
             this._size_dirty = true;
         })
 
+        // Bind popup size settings BEFORE creating the resizer, so that
+        // popup_width and popup_height are defined when the resizer's
+        // get_user_width/height callbacks are invoked. Fixes #12972.
+        this.settings.bind("popup-width", "popup_width");
+        this.settings.bind("popup-height", "popup_height");
+
         this._resizer = new Applet.PopupResizeHandler(this.menu.actor,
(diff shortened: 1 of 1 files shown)


fredcw:
I'm confused by this. It looks like the PopupResizeHandler doesn't call it's `this._get_user_width` or `this._get_user_height` until the popup (`this.menu` in the applet) is shown. But that can't happen until the applet is initialised, right?

carsteneu:
Thanks for taking the time to review this, @fredcw! You're absolutely right - the callbacks are arrow functions that get evaluated lazily when called, not at construction time. I appreciate the correction.

After digging deeper into #12972, it looks like it's actually a duplicate of #12803. I noticed that your fix from PR #12804 was merged to master but never backported to the 6.4-maintenance branch - so users on Linux Mint 22.x still experience this issue.

I'll close this PR and submit a backport of #12804 to 6.4-maintenance instead. That should actually help the affected users.

Thanks again for pointing me in the right direction!

PR #13471, opened 2026-01-22
FIX: Add default home folders to search results (#13402)

### LINKED ISSUE #13402: Menu search no longer return default home folders nor allow to directly navigate to a folder
### Distribution

22.3

### Package version

6.6.4

### Graphics hardware in use

Advanced Micro Devices, Inc. [AMD/ATI] Phoenix1

### Frequency

Always

### Bug description

I am not sure if this regression was intended but before 6.6.4 we could navigate to folders from the menu search bar in 2 ways

1) Searching for default home folders would return the result and direct navigation in the menu, it used to return all the default home folders ex: Downloads, Pictures, Videos... now it only happens to folders that are added to the favorites

2) There was the ability to navigate to any folder in the system (option searchFilesystem) when starting the search query with `/` or `~` it would return folders 
ex: search string `/lib` would return 
```
/lib
/lib32
/lib64
```

The search does seem to return some files in the home folder (unsure about the criteria, maybe it is recent files), but it does not return the results for the folder itself

### Steps to reproduce

1) Start typing "Downloads" the default Downloads folder does not appear in the menu

2) No option for searchFilesystem, starting the query with `/` or `~` does not return folder completion

### Expected behavior

1) Start typing "Downloads" (or any other default home folder) and it would appear in the search without the need for this folder to be added to the favorites

2) Either with or without the searchFilesystem option, the user should be able to navigate to any folder on the filesystem via the menu search

### Additional information

_No response_

carsteneu:
Fixes #13402

Users expect to find default home folders (Downloads, Documents, Pictures, Music, Videos, Desktop, Home) when searching in the menu, but currently they only appear if they're enabled in the sidebar settings.

This fix separates search functionality from sidebar visibility. Default home folders are now always searchable regardless of the show* settings. The settings only control whether folders appear in the sidebar, not whether they're searchable.

Changes:

- Added PlaceSearchResultButton class for place search results
- Added _searchablePlaces array containing all default places
- Modified _refreshPlaces to populate searchable places independently of sidebar visibility
- Modified _searchAppSystem to search through _searchablePlaces and display results

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -712,6 +712,35 @@ class SearchProviderResultButton extends SimpleMenuItem {
     }
 }
 
+class PlaceSearchResultButton extends SimpleMenuItem {
+    constructor(applet, place) {
+        super(applet, {
+            name: place.name,
+            type: 'search-result',
+            styleClass: 'appmenu-application-button',
+            place: place,
+        });
+
+        this.icon = place.iconFactory(applet.applicationIconSize);
+        if (this.icon)
+            this.addActor(this.icon);
+        else
+            this.addIcon(applet.applicationIconSize, 'folder');
+
+        this.addLabel(this.name, 'appmenu-application-button-label');
+    }
+
+    activate() {
+        this.place.launch();
+        this.applet.menu.close();
+    }
+
+    destroy() {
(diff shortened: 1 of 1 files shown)


mtwebster:
I think in the old menu these results were from the Places category (bookmarks + devices).

This only restores a fixed set of XDG locations - it should include the 'Other bookmarks' locations also if you're going to do this. Though it's still odd either way returning results from our of no-where like this.

PR #13472, opened 2026-01-22
FIX: Add right-click context menu to search entry

### LINKED ISSUE #13416: Cannot insert content by right-clicking in the Start menu input field
### Distribution

LMDE7

### Package version

Cinnamon 6.6.5

### Graphics hardware in use

Intel I7 CPU own graphics

### Frequency

Always

### Bug description

Cannot insert content by right-clicking in the Start menu input field

Input in the Start menu input field works via keyboard and also via Ctrl+v.

### Steps to reproduce

Try to insert text by right-clicking in the Start menu input field

### Expected behavior

Input in the Start menu input field works via keyboard and also via Ctrl+v.

### Additional information

_No response_

carsteneu:
Fixes #13416

The menu search field didn't support right-click copy/paste. Users could only paste with Ctrl+V keyboard shortcut.

Added CinnamonEntry.addContextMenu() to the search entry, enabling the standard right-click context menu with copy/paste/select all options.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -2580,6 +2580,55 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
         });
         this.searchEntry.add_accessible_state(Atk.StateType.EDITABLE);
         CinnamonEntry.addContextMenu(this.searchEntry);
+        this.searchEntry._menuManager.shouldGrab = false;
+
+        // Close the search entry context menu when main menu closes
+        this.menu.connect('open-state-changed', (menu, open) => {
+            if (!open && this.searchEntry._menu.isOpen) {
+                this.searchEntry._menu.close();
+            }
+        });
+
+        // Extend _eventIsOnActiveMenu to include the search entry context menu
+        // This prevents the main menu from closing when clicking on the context menu
+        let originalEventIsOnActiveMenu = this.menuManager._eventIsOnActiveMenu.bind(this.menuManager);
+        this.menuManager._eventIsOnActiveMenu = (event) => {
+            if (this.searchEntry._menu.isOpen &&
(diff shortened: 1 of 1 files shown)


fredcw:
This doesn't seem to work well for me. When I click paste, the applet just closes. And copy seems always greyed out even if there's text in the search entry.

carsteneu:
Thank you very much @fredcw for testing and the valuable feedback!

I've updated the PR to fix the issues you reported:

What's fixed:

Clicking "Paste" no longer closes the menu - it now works as expected. I tested before, but I think I only checked that the context menu appeared, not the actual paste action. My apologies for that oversight!
Hover highlighting on menu items now works properly. This turned out to be quite complex to achieve - I found no other way than implementing it manually with mouse position polling.
A note on "Copy" being greyed out:
I believe this is the expected behavior - "Copy" should only be enabled when text is selected in the search field. Without a selection, there's nothing to copy. Similarly, "Paste" is only active when there's something in the clipboard.

Technical details:
The fix required extending the MenuManager's event handling to treat the context menu as part of the active menu, plus a manual hover implementation since the MenuManager blocks normal hover events.

Please test again when you have a chance - I really appreciate your help!

fredcw:
This works well now.

fredcw:
This seems a little complicated for what it does. It seems either `CinnamonEntry.addContextMenu` needs fixing itself or it wasn't designed for this purpose. It's not something I would add to Cinnamenu for instance unless it was much simpler.

mtwebster:
I agree - there's some issue in the context menu implementation. It works ok in the runDialog, but here, something must be calling popModal or the main menu is closing due to loss of focus.

This is a lot of code here as a workaround, I'd rather investigate the root cause in cinnamonEntry.js. The whole point is that it should be an easy way to add a context menu to any entry.

mtwebster:
See: https://github.com/linuxmint/cinnamon/commit/14c83ba7e77405c17a8c80d84bdd729681167f8d

The original commit here should work now.

Issue #13407, opened 2026-01-15
switching the keyboard layout on Mint 22.3


sergDash:
### Distribution

Mint 22.3

### Package version

new

### Graphics hardware in use

amd

### Frequency

Always

### Bug description

When I open the main menu, the keyboard layout cannot be switched by any method.
First, the Super key is used both for switching keyboard layouts and for opening the menu. If I remove it from the menu shortcut, layout switching works but also inserts a space character, so all applications receive a space key press, including media players (play/pause).

Second, there is no way to add an additional key combination for switching the keyboard layout.

### Steps to reproduce

---

### Expected behavior

---

b0t0s:
@sergDash , Thanks for workaround. 
Changed hotkey from "win + space" to "alt + shift" and at least can continue to work.

Using latest LMDE version, faced this issue too

sergDash:
I set the keyboard layout switch to Caps Lock in two places: where it is defined in the UI and in the XKB parameters. After that, the layout started switching even when a menu was open.

In addition, in the game Rust there were severe lags — a kind of running inertia where the character kept moving for several seconds after releasing the key. The only thing that helped was this:

sudo chmod -x /usr/bin/ibus* /usr/libexec/ibus*
reboot

I think there really should be some kind of visible switch or toggle for ibus, especially if it’s impossible to get rid of these lags otherwise.

mtwebster:
https://github.com/linuxmint/cinnamon/commit/58e1c478d71e78f9902fddf7b031aad4f97f3f8a

PR #13473, opened 2026-01-22
FIX: keyboard layout switching while search is focused (#13428)

### LINKED ISSUE #13428: [Regression] Keyboard layout switching latency (Alt+Shift) and failure to switch within Mint Menu search
### Distribution

Mint 22.3

### Package version

6.6.5

### Graphics hardware in use

Advanced Micro Devices, Inc. [AMD/ATI] Lucienne

### Frequency

Always

### Bug description

After updating to Linux Mint 22.3 (Cinnamon 6.6.5), two major issues have emerged regarding keyboard layout switching:

1.Input Latency: There is a noticeable delay (several milliseconds) when using the Alt+Shift shortcut. The switch is no longer instantaneous, which disrupts fast typing. It seems the system waits for a key release or a potential sequence instead of triggering on key press.

2.Mint Menu Focus Issue: While the Mint Menu is open and the search bar is focused, the Alt+Shift shortcut fails to work entirely. The user must switch the language before opening the menu to search in a different language.

### Steps to reproduce

1.Use the default keyboard layout switching method in Mint 22.3.

2.Press Alt+Shift quickly while typing: Observe the lag.

3.Open the Mint Menu and try to change the language while the search bar is active: The layout remains unchanged.

### Expected behavior

1.Keyboard layout switching should be instantaneous (as it was in previous versions or when using XKB options).

2.Layout switching shortcuts should be globally intercepted, even when the Mint Menu or its search bar is focused.

### Additional information

The latency issue can be bypassed by using the "XKB Options" tab in Keyboard settings, but this hides a core functionality behind a "legacy" menu and doesn't solve the Mint Menu focus bug.

carsteneu:
Keyboard layout switching doesn't work when Cinnamon text fields are focused. Keybindings like Alt+Shift or Super+Space have no effect because text entry key event handlers stop event propagation.

**Extended scope:** 
This fix now covers ALL (hopefully) Cinnamon text entries, not just the menu:
- Menu search field
- Alt+F2 run dialog
- Expo workspace renaming
- Polkit authentication dialog (sudo passwords)
- Network password dialog (WiFi/VPN)
- Keyring password prompt

But please this should be double checked - these are system critical fields, i dont feel enough experienced to be shure i checked every possible caveat here.

The fix has two parts:

1. **For custom keybindings** (like switch-input-source), invoke the keybinding callback before stopping event propagation
2. **For modifier-only keys** (Alt, Shift, Ctrl, Super), allow the event to propagate so XKB-based layout switching can work

The Implementation
- Created reusable helper function `handleKeyboardLayoutSwitchingInTextEntry()` in `util.js`
- Respects Input Methods (has_preedit check for CJK languages)
- Performance optimized (Set-based lookup)
- Extended modifier list (Hyper, Meta keys)

This should allow users to switch keyboard layouts while typing in any Cinnamon text field.

#### Root Cause in my point of view

- Commit 6a5105420 (Dec 15) moved `switch-input-source` from `global.display` to `KeybindingManager`
- Commit 05e68dd2c (Dec 12) blocked keybindings in text entries to prevent duplicate events
- Components must now explicitly invoke KeybindingManager for layout switching

Tests done:

Tested on Linux Mint 22.3 Cinnamon 6.6.6 with US + DE layouts:

- ✅ Menu search field
- ✅ Alt+F2 run dialog
- ✅ Expo workspace renaming
- ✅ Polkit authentication (sudo)
- ✅ Keyring prompt

Fixes #13428, #13513, #13407

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1717,37 +1717,15 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
             }
         }
 
-        /* check for a keybinding and quit early, otherwise we get a double hit
-           of the keybinding callback */
-        let action = global.display.get_keybinding_action(keyCode, modifierState);
-
-        if (action == Meta.KeyBindingAction.CUSTOM) {
-            // Invoke the keybinding callback (e.g., for keyboard layout switching)
-            // before stopping event propagation. This ensures keybindings like
-            // switch-input-source work even when the menu search is focused.
-            Main.keybindingManager.invoke_keybinding_action_by_id(action);
-            return Clutter.EVENT_STOP;
-        }
-
-        // Allow modifier-only key combinations (Alt, Shift, Ctrl) to propagate
-        // so XKB-based keyboard layout switching can work when no custom
-        // keybinding is registered for the combo.
-        if (action == Meta.KeyBindingAction.NONE) {
-            const isModifierOnly = (
-                symbol == Clutter.KEY_Shift_L || symbol == Clutter.KEY_Shift_R ||
-                symbol == Clutter.KEY_Alt_L || symbol == Clutter.KEY_Alt_R ||
-                symbol == Clutter.KEY_Control_L || symbol == Clutter.KEY_Control_R ||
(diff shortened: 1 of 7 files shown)


mtwebster:
Addressed as part of https://github.com/linuxmint/cinnamon/commit/58e1c478d71e78f9902fddf7b031aad4f97f3f8a

PR #13477, opened 2026-01-23
Fix: PopupResizeHandler work area size calculation

carsteneu:
While analyzing issue #13469 and  #12972 (menu shrinking unexpectedly), I noticed a subtle 
calculation issue in `PopupResizeHandler._on_actor_show()` that could contribute 
to sizing problems under certain conditions.

The previous code compared the outer actor size against the work area but then 
adjusted the inner content size. This works correctly most of the time, but can 
lead to incorrect results when there's padding between the actor and its content 
(e.g., from theme CSS) and timing issues occur during layout.

This fix explicitly calculates the padding offset between the outer actor and 
inner content, then uses it to determine the correct maximum content size that 
fits within the work area.

This is admittedly an edge case, but it becomes more relevant with UI scaling. 
For example, on a 1366×768 laptop with 150% scaling and the menu resized to 
maximum (900px logical = 1350px physical + ~30px padding = 1380px), the menu 
actor exceeds the work area (1366px), triggering this code path with potentially 
incorrect calculations.

## Changes

- Calculate padding between outer actor and inner content
- Use padding-aware calculation for work area constraints
- Ensures mathematically correct sizing regardless of layout timing

## Tested

- Tested in VM with Linux Mint 22.3 / Cinnamon 6.6.5
- Menu opens/closes correctly
- No JavaScript errors in logs
- Multiple open/close cycles work as expected

## Related

- #12972 (menu shrinking issue - analysis led to finding this)
- #12803, #12804 (related resize handler fixes)

### DIFF
--- a/js/ui/applet.js
+++ b/js/ui/applet.js
@@ -1246,15 +1246,20 @@ var PopupResizeHandler = class PopupResizeHandler {
         const user_w = this._get_user_width();
         const user_h = this._get_user_height();
 
+        // Calculate padding between outer actor and inner content
+        const padding_w = Math.max(0, cur_w - user_w);
+        const padding_h = Math.max(0, cur_h - user_h);
+
         let new_w = user_w;
         let new_h = user_h;
 
-        if (cur_w > this._workAreaWidth) {
-            new_w -= cur_w - this._workAreaWidth;
+        // Check if content + padding exceeds work area
+        if (user_w + padding_w > this._workAreaWidth) {
+            new_w = Math.max(0, this._workAreaWidth - padding_w);
         }
 
-        if (cur_h > this._workAreaHeight) {
-            new_h -= cur_h - this._workAreaHeight;
+        if (user_h + padding_h > this._workAreaHeight) {
+            new_h = Math.max(0, this._workAreaHeight - padding_h);
         }
 
         if (new_w !== user_w || new_h !== user_h) {


fredcw:
This is the same calculation except for the two Max() conditions. The first one `Math.max(0, cur_w - user_w)` only takes effect if the user size is greater than both the menu size and the workarea size, in which case it sets the new user size to the workarea size, which is wrong. The second Max() `Math.max(0, this._workAreaHeight - padding_h)` prevents the user size from going negative but that makes no difference because the applet obviously has to set a minimum size anyway. 

You're obviously putting a lot of work into finding and fixing bugs and many or most of your PRs may very well be good and correct fixes so I certainly wouldn't want to say anything to discourage your efforts. But in this particular case, this PR and explanation don't make much sense and I suspect that's because it's almost 100% LLM generated. One thing's for sure, LLMs are only going to get better. In two years time, I doubt any of us will barely ever write a single line of code again. :o

carsteneu:
You are absolutly right, I reviewed this and the calculations are obiously equivalent. I will double check this in the future. The padding is already implicitly accounted for. I will close this.

Regarding LLM usage: Yes, I use CC Opus as a tool to explore the codebase, generate ideas, potential solutions, and also check potential solutions. However, I always test changes in a dedicated VM to verify they actually work and don't cause regressions. I dont want to pollute the timeline with unneccessary work.

 Thank you very much for the detailed review!

Issue #13490, opened 2026-01-26
Mint Linux 22.3 with Cinnamon no Preferences


rsauve:
### Distribution

Mint Linux 22.3

### Package version

Cinnamon 6.6.5

### Graphics hardware in use

Intel Corporation Skylake GT2 [HD Graphics 520]

### Frequency

Always

### Bug description

I updated my HP Spectre laptop from Mint Linux 22.2 to 22.3. Since updating I have not been able the open Preferences.
Here is some system information:

    NAME="Linux Mint"
    VERSION="22.3 (Zena)"
    ID=linuxmint
    ID_LIKE="ubuntu debian"
    PRETTY_NAME="Linux Mint 22.3"
    VERSION_ID="22.3"
    HOME_URL=["https://www.linuxmint.com/"](https://www.linuxmint.com/)
    SUPPORT_URL=["https://forums.linuxmint.com/"](https://forums.linuxmint.com/)
BUG_REPORT_URL=["http://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/"](http://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/)
    PRIVACY_POLICY_URL=["https://www.linuxmint.com/"](https://www.linuxmint.com/)
    VERSION_CODENAME=zena
    UBUNTU_CODENAME=noble
    -----------------------
    Base Board Information
        Manufacturer: HP
        Product Name: 804E
        Version: 33.24
        Serial Number: PFLJK038J9N04S
        Asset Tag: Base Board Asset Tag
        Features:
                Board is a hosting board
                Board is replaceable
        Location In Chassis: Base Board Chassis Location
        Type: Motherboard

Here is what happens when I open Preferences from the Gnome terminal:

    robert@Mint-HP:~$ cinnamon-settings
    Traceback (most recent call last):
      File "/usr/share/cinnamon/cinnamon-settings/cinnamon-settings.py", line 34, in <module>
        from bin import SettingsWidgets
      File "/usr/share/cinnamon/cinnamon-settings/bin/SettingsWidgets.py", line 13, in <module>
        from KeybindingWidgets import ButtonKeybinding
      File "/usr/share/cinnamon/cinnamon-settings/bin/KeybindingWidgets.py", line 6, in <module>
        print("KeybindingWidgets session type: %s" % util.get_session_type())
 ^^^^^^^^^^^^^^^^^^^^^
    AttributeError: module 'util' has no attribute 'get_session_type'

or

[...]

PR #13492, opened 2026-01-26
FIX: util module import to prevent name collisions

carsteneu:
This PR addresses issue #13490 where cinnamon-settings fails to start with an AttributeError: module 'util' has no attribute 'get_session_type'.

The root cause appears to be the use of bare import util statements in four files, which can be shadowed by any util.py file present in the user's working directory or Python path.

I was able to reproduce this issue by placing an empty util.py in the home directory – this caused the exact error reported in the issue. After applying the fix, cinnamon-settings launched correctly even with the conflicting file present.

**Changes**
Changed` import util` to `from bin import util` in:

- bin/KeybindingWidgets.py
- bin/SettingsWidgets.py
- bin/XkbSettings.py
- modules/cs_sound.py

This aligns these files with the import pattern already used throughout the codebase (e.g., cinnamon-settings.py, cs_keyboard.py, cs_info.py, cs_general.py, etc.).

**Tested**
Tested in Linux Mint 22.3 VM:

- cinnamon-settings launches without errors
- Sound settings module loads correctly
- Keyboard settings with XKB options tab works

While this issue may be relatively rare in practice (requiring a conflicting util.py in the user's environment), using explicit imports is more robust in my point of view and i think consistent with the rest of the codebase.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/KeybindingWidgets.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/KeybindingWidgets.py
@@ -1,7 +1,7 @@
 #!/usr/bin/python3
 
 from gi.repository import Gtk, Gdk, GObject
-import util
+from bin import util
 
 print("KeybindingWidgets session type: %s" % util.get_session_type())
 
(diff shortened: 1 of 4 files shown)


mtwebster:
I'm not able to reproduce this - placing a util.py in my home directory makes no difference unless that directory has been added to PYTHONPATH for some reason.

I'm not arguing against the PR but I like to be able to confirm a before/after behavior, and I'm having trouble coming up with a real-world test case - we already filter out /usr/local and ~/.local imports via `strip_syspath_locals()`

Running:
```sh
touch ~/util.py
PYTHONPATH=~ cinnamon-settings
```
is the only way I can get an error.

Issue #13489, opened 2026-01-25
Cinnamon menu profile picture does not update when selecting custom image file (PNG/JPG)


tarunsamal:
### Distribution

Linux Mint 22.3 - Cinnamon 64-bit

### Package version

6.6.5

### Graphics hardware in use

Intel Corporation Raptor Lake-P [Intel Graphics]

### Frequency

Always

### Bug description

Issue while updating profile pictures on new menu
Working : 
1. pre existing icons
2. photo taken on camera

Not Working :
1. Selected photo from files

### Steps to reproduce

1. click on menu 

<img width="803" height="406" alt="Image" src="https://github.com/user-attachments/assets/eb7f286f-43b0-4dfd-8c1b-1abc39c52c49" />

2. select profile picture

<img width="1240" height="749" alt="Image" src="https://github.com/user-attachments/assets/5f54ca58-eaf2-41f7-9fb5-00a02ffcf660" />

3. Click to select Picture

<img width="1503" height="383" alt="Image" src="https://github.com/user-attachments/assets/08c3db10-a6aa-4294-88db-23dacc2f28a1" />

4.  Broswe for more pictures
5.  Select a png picture from home/pictures 

### Expected behavior

Selected(choosen from file) picture should be displayed as profile picture

JosephMcc:
Reproduced

tarunsamal:
Issue could be with higher file size, looks like lower size images are working

PR #13494, opened 2026-01-26
Fix: User avatar not updating in menu (#13489)

carsteneu:
After some research i think this fixes two related issues when changing user profile pictures:

1. **Image caching problem**: When selecting a custom image from files, the menu avatar wouldn't update without restarting Cinnamon. This was caused by St's CSS background-image caching - AccountsService copies all custom images to the same path (/var/lib/AccountsService/icons/<user>), so the URL stays identical while only the file content changes.

2. **Size limit**: AccountsService rejects images larger than ~1MB with a silent error, preventing large photos from being set as avatars.

**Solution:**

- userWidget.js: Implements cache-busting by copying AccountsService icons to ~/.cache/cinnamon/avatars/<user>-<mtime> with unique timestamps, forcing St to reload the image. Also resizes oversized images as fallback.

- cs_user.py: Resizes large images (>512px) before passing to AccountsService, preventing rejection. Also fixes existing bug where original path was used instead of the resized .face file.

Tested with images up to 9600x6000 (5.4MB) - now displays correctly without Cinnamon restart. I hope this will help. 
Looking for Feedback.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_user.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_user.py
@@ -212,10 +212,29 @@ class Module:
 
             face_path = os.path.join(self.accountService.get_home_dir(), ".face")
 
-            pixbuf = GdkPixbuf.Pixbuf.new_from_file_at_size(path, 255, -1)
-            pixbuf.savev(face_path, "png")
-            self.accountService.set_icon_file(path)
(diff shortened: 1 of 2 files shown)


mtwebster:
Fixed (finally) in https://github.com/linuxmint/cinnamon/commit/8f73a54d14dc400a27e7bba121c34efe7e171aed

PR #13498, opened 2026-01-27
appSwitcher: Add option to group windows by application

b3rs3rk1ng:
This adds a new configurable setting `alttab-switcher-group-by-app` that allows users to group windows by application in the Alt+Tab switcher.

## What this does

When enabled:
- Alt+Tab shows one icon per application (not per window)
- Alt+` cycles directly between windows of the same app
- Application names are displayed instead of window titles

When disabled, Cinnamon behaves exactly as before.

## Configuration

The setting can be toggled in **System Settings → Windows → Alt-Tab** or via:
```bash
gsettings set org.cinnamon alttab-switcher-group-by-app true
```

## Files Modified

- `js/ui/appSwitcher/appSwitcher.js` - Groups windows by wm_class
- `js/ui/appSwitcher/classicSwitcher.js` - Shows app name instead of window title
- `js/ui/windowManager.js` - Enables instant window cycling
- `files/.../cs_windows.py` - Adds UI toggle
- `data/org.cinnamon.gschema.xml` - Defines the setting

## Related Issues

This addresses long-standing requests in #2883 and #4 by providing a configurable option that satisfies both preferences.

### DIFF
--- a/data/org.cinnamon.gschema.xml
+++ b/data/org.cinnamon.gschema.xml
@@ -444,6 +444,12 @@
       <summary>Show all windows from all workspaces</summary>
     </key>
 
+    <key type="b" name="alttab-switcher-group-by-app">
+      <default>false</default>
+      <summary>Group windows by application</summary>
+      <description>When enabled, Alt+Tab will show only one entry per application instead of one per window. The most recently used window of each application will be shown.</description>
+    </key>
+
     <key type="b" name="alttab-switcher-warp-mouse-pointer">
       <default>false</default>
       <summary>Warp mouse pointer to the new focused window</summary>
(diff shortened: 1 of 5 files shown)


mtwebster:
Alt+` already does what you're describing, without this PR.

I also disagree with what the stated purpose of this is - you're not grouping apps, you're making alt-tab only show the most recently used window for each application.

Depending on which alt-tab mode you use, this can be confusing - 
- Thumbnails mode still shows window *titles*
- Icon+Thumbnail mode shows a single thumbnail - I would probably expect to see thumbs for all windows below this...
- The 3d switchers still show the window *title*, not the app name.

I'd be more inclined to include a new 'mode' based on Icon+Thumbnail that groups primarily by app but shows all thumbnails and allows sub-navigation between them. I don't think this global option works very well.

PR #13500, opened 2026-01-27
menu@cinnamon.org: Avoid using absolute path to launch ...

fredcw:
... screensaver as this doesn't work on NixOS


bobby285271:
Thank you! For the record we worked around this issue in our packaging [here](https://github.com/NixOS/nixpkgs/blob/8ba5eb22ba5bb4ae2df737a84a545c038694027f/pkgs/by-name/ci/cinnamon/package.nix#L157).

PR #13508, opened 2026-01-28
FIX: Grey out format options when calendar applet uses…

### LINKED ISSUE #13447: Date/time applet doesn't update properly on settings change in Mint 22.3
### Distribution

22.3

### Package version

6.6.5

### Graphics hardware in use

_No response_

### Frequency

Always

### Bug description

When I toggle on "Display seconds" or "Display the date" in Date/Time settings, nothing happens.

### Steps to reproduce

Toggle "Display seconds" or "Display the date" in Date/Time settings

### Expected behavior

Seconds should show up/vanish immediately, or the date should show up/vanish immediately. This was the behavior in previous versions of Mint (e.g. 22.2).

### Additional information

Even if I enable "Display seconds" it does not display seconds even after restarting Cinnamon.

carsteneu:
Fixes #13447 - Grey out date/time format settings when calendar applet uses custom format

When the calendar applet has "Use a custom date format" enabled, the system date/time format settings in cinnamon-settings are now disabled with an info banner explaining why they don't apply.

## Problem reported in #13447

Users were confused when toggling "Display the date" or "Display seconds" in Date/Time Settings had no effect. This happened because the calendar applet's "Use a custom date format" option overrides all system format settings, but there was no visual indication of this in the settings UI.

## Solution

- Added a new "Custom Format Active" section that appears when any calendar applet instance has custom format enabled
- Shows a Gtk.InfoBar explaining the situation and how to resolve it
- Disables the affected format switches:
  - Use 24h clock
  - Display the date
  - Display seconds
- Uses the same pattern as KeybindingTable.py for reading applet configurations from `~/.config/cinnamon/spices/`

**Screenshots**

before
issue-13447-before

after 
issue-13447-after

### Tested in VM with Linux Mint 22.3, Cinnamon 6.6.6:

- [x] Settings display normally when custom format is disabled
- [x] Info section appears when custom format is enabled
- [x] All three affected switches are greyed out
- [x] "First day of week" remains active (not affected by custom format as far as i know)
- [x] Graceful handling when no calendar applet is present
- [x] Should work with multiple calendar applet instances - but normally there should only be one

### Technical Details

- Uses standard Gtk.InfoBar widget (consistent with cs_startup.py, cs_user.py)
- Should be a robust error handling for missing config files or malformed JSON
- No changes to calendar applet itself - pure settings UI enhancement

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_calendar.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_calendar.py
@@ -7,6 +7,8 @@ from zoneinfo import ZoneInfo, available_timezones
 import gi
 import datetime
 import locale
+import json
+from pathlib import Path
 try:
     from babel import Locale as BabelLocale
 except:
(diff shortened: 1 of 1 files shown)


mtwebster:
This is wrong - see my comment in the other issue.

carsteneu:
Ah yes, maybe a hint above the settings may be an option, that was already my first idea.

PR #13522, opened 2026-02-01
Improvements in printers@cinnamon.org

hans-fritz-pommes:
I encountered several small issues with the printers applet.
I modified the `applet.js` in `/usr/share/cinnamon/applets/printers@cinnamon.org` and tested printing documents.
The things I did correct:
 - The "Cancel all jobs" command didn't work for an user who had no permisisons to run `cancel -a` for a specific printer, *even if all of the jobs on this printer belonged to him*.
 -> Now the "Cancel all jobs" command tries to cancel every single job, not `cancel -a` for every printer
 - The appearance of the listed jobs was really bad: e.g. when you had more than one job, the description for **every** job was "<desc. first job><desc. second job><desc..." and so on.
 - The job description consisted of: `(job number) lpstat -o` (lpstat several times, if there was more than one job - see point before)
 -> The description is now: `(job number) 'document_name' on <printer_name> by <user>` (using `lpq -a`).

I removed the stuff with the bytes.

### DIFF
--- a/cinnamon.pot
+++ b/cinnamon.pot
@@ -2039,28 +2039,28 @@ msgstr ""
 #. Label                           Executable
 #. Icon                        Category      Keywords for filter
 #. printers@cinnamon.org->metadata.json->name
-#: files/usr/share/cinnamon/applets/printers@cinnamon.org/applet.js:22
-#: files/usr/share/cinnamon/applets/printers@cinnamon.org/applet.js:133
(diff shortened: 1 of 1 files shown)


JosephMcc:
Do not include changes to the POT and translation files in your commits. Those are handled on our side.

hans-fritz-pommes:
Nice! Opened new PR without `pot`-changes

Issue #9265, opened 2020-04-21
Printer applet is shown when cups service is down


vdcow:
```
 * Cinnamon version 4.4.8
 * Distribution - Alt Linux Sisyphus
 ```

**Issue**
When cups service is stopped/disabled Cinnamon always shows printer applet regardless option 
"Show printer icon". Left click on this applet (when cups is down) do nothing that confuses user.

**Steps to reproduce**
1. Stop cups service
2. Restart Cinnamon

**Expected behaviour**
Printer applet will not be shown if "Show printer icon" has value "When printer exists" or "When printing". If "Always" value is selected then probably left click on applet should notify user about cups service not running.

**Comment by JulianGro** [comment]
I wonder if this is related to: https://github.com/linuxmint/cinnamon/pull/13039

**Comment by mendhak** [comment]
> I wonder if this is related to: [#13039](https://github.com/linuxmint/cinnamon/pull/13039)

I can see this merged line in my local printers@cinnamon applet code, but it seems to have no effect. The printer is always visible, when the service is off, and even when the option is set to 'when printing'. 

(I am not an expert, sorry for speculation) I _think_ this is because the `lpstat -o` command returns a non-empty output when the CUPS service is off. The code counts that as 1 line, and therefore thinks that there is a print job available. 

https://github.com/linuxmint/cinnamon/blob/bebaf4cf54d6c2740f37af35ba086b7654fae9b6/files/usr/share/cinnamon/applets/printers%40cinnamon.org/applet.js#L177-L179


When CUPS is off:

```
$ lpstat -o
lpstat: Scheduler is not running.
$
```

When CUPS is on, empty.

```
$ lpstat -o
$
```

mendhak:
But I also see that applet has [been reworked completely](https://github.com/linuxmint/cinnamon/commit/769757bd88da65d54144cc5f1e24ee00ae5de78b), so possibly fixed in a future release?

PR #13524, opened 2026-02-01
Fixes in "printers@cinnamon.org"

hans-fritz-pommes:
I encountered several small issues with the printers applet.
I modified the `applet.js` in `/usr/share/cinnamon/applets/printers@cinnamon.org` and tested printing documents.
The things I did correct:

 - The "Cancel all jobs" command didn't work for an user who had no permisisons to run `cancel -a` for a specific printer, even if all of the jobs on this printer belonged to him.
 -> Now the "Cancel all jobs" command tries to cancel every single job, not `cancel -a` for every printer
 - The appearance of the listed jobs was really bad: e.g. when you had more than one job, the description for every job was "<desc. first job><desc. second job><desc..." and so on.
 - The job description consisted of: `(job number) lpstat -o` (lpstat several times, if there was more than one job - see point before)
 -> The description is now: `(job number) 'document_name' on <printer_name> (<size in Bytes/MB/KB>) by <user>` (using lpq -a).

~The following problem could occur:~
~If a job has the same number as byte-size like the job-id of the following one, the username and the filesize of the first one will be strange (job-id and functions stay working).~
~I would say this will happen that seldom  - we should ignore it.~
~Which user will open the menu exactly in that minute and have a problem with a strange username or filesize?~

EDIT: fixed in new commit

### DIFF
--- a/files/usr/share/cinnamon/applets/printers@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/printers@cinnamon.org/applet.js
@@ -187,28 +187,6 @@ class CinnamonPrintersApplet extends Applet.TextIconApplet {
                 this.out1 = stdout.trim().split('\n');
             }
 
-            //Update icon
-            Util.spawnCommandLineAsyncIO('/usr/bin/lpstat -l', (stdout, stderr, exitCode) => {
-                if(exitCode) {
-                    this.handleError(stderr);
-                    this.updating = false;
-                    return;
-                }
-
-                if(stdout.trim()) {
-                    this.set_applet_tooltip(stdout.trim().split('\n')[1].trim());
-                } else {
-                    this.set_applet_tooltip(_("Printers"));
-                }
-                if(this.printWarning) {
-                    this.set_applet_icon_symbolic_name('xsi-printer-warning');
-                } else if(this.jobs.length > 0) {
-                    this.set_applet_icon_symbolic_name('xsi-printer-printing');
-                } else {
-                    this.set_applet_icon_symbolic_name('xsi-printer');
-                }
-            });
-
             //Check default printer and add printers
             Util.spawnCommandLineAsyncIO('/usr/bin/lpstat -d', (stdout, stderr, exitCode) => {
                 if(exitCode) {
@@ -328,6 +306,30 @@ class CinnamonPrintersApplet extends Applet.TextIconApplet {
                         }
 
                         //Update icon
+                        Util.spawnCommandLineAsyncIO('/usr/bin/lpstat -l', (stdout, stderr, exitCode) => {
+                            if(exitCode) {
+                                if(stderr.length>70) {
+                                    stderr = stderr.slice(0, 70) + "...";
+                                }
+                                this.set_applet_tooltip(stderr.trim());
+                                return;
+                            }
+
+                            if(stdout.trim()) {
+                                this.set_applet_tooltip(stdout.trim().split('\n')[1].trim());
+                            } else {
+                                this.set_applet_tooltip(_("Printers"));
+                            }
+                            if(this.printWarning) {
+                                this.set_applet_icon_symbolic_name('xsi-printer-warning');
+                            } else if(this.jobs.length > 0) {
+                                this.set_applet_icon_symbolic_name('xsi-printer-printing');
+                            } else {
+                                this.set_applet_icon_symbolic_name('xsi-printer');
+                            }
+                        });
+                        
+                        //Update job count
                         if(this.jobs.length > 0) {
                             this.set_applet_label(this.jobs.length.toString());
                         } else {


hans-fritz-pommes:
Oh and I changed icons. If you think they're ugly, just revert it

hans-fritz-pommes:
I have an idea how to fix this mini-problem (identical bytes & job-number). But later

hans-fritz-pommes:
Ok, now the identical bytes/document-name & job-number-problem is fixed.

hans-fritz-pommes:
Sorry that I'm so chaotic
I forgot to take care of document names with whitespaces

files/usr/share/cinnamon/applets/printers@cinnamon.org/applet.js
```diff
@@ -11,6 +11,43 @@ const PANEL_EDIT_MODE_KEY = "panel-edit-mode";
...
+    const k = 1024;
+    const sizes = ["Bytes", "KB", "MB", "GB", "TB"];
+    const i = Math.floor(Math.log(bytes) / Math.log(k));
+    const value = bytes / Math.pow(k, i);
+
+    return value.toLocaleString(undefined, {
+        minimumFractionDigits: 1,
+        maximumFractionDigits: 1
+    }) + " " + sizes[i];
+}
```
mtwebster: Consider using `GLib.format_size()` or `GLib.format_size_full()` instead here

files/usr/share/cinnamon/applets/printers@cinnamon.org/applet.js
```diff
@@ -176,33 +215,34 @@ class CinnamonPrintersApplet extends Applet.TextIconApplet {
...
-                                } else {
-                                    break;
-                                }
-                            }
+                            let doc = jobInfo[job][2];
+                            let user = jobInfo[job][1];
+                            let size = formatBytes(jobInfo[job][3]);
+
                             if(doc.length > 30) {
                                 doc = doc + '...';
```
mtwebster: This isn't part of your changes, but maybe fix this - it detects a long string but doesn't actually shorten it before adding ellipses.

files/usr/share/cinnamon/applets/printers@cinnamon.org/applet.js
```diff
@@ -176,33 +215,34 @@ class CinnamonPrintersApplet extends Applet.TextIconApplet {
...
                             }
+                            
                             let text = '(' + job + ') ' + _("'%s' on %s").format(doc, printer);
+                            text = text + ' (' + size + ')' + _(" - by %s").format(user);
```
mtwebster: Can't this be a single translatable string?
```
# Translators: strings are job #, document name, printer name, size, username
# example: (23) 'test-doc.md' on HP_Smart_Tank_5100_series (44.0 KB) - by mtwebster
text = "(%s) '%s' on %s (%s) - by %s".format(job, doc, printer, size, user);
```
When 'Translators' is detected, these comments are included in the potfile to aid in translations

files/usr/share/cinnamon/applets/printers@cinnamon.org/applet.js
hans-fritz-pommes: Nice function!

hans-fritz-pommes:
> ## Best-practices scanner
> 
> This is a regex-based check for API usage that can pose security, performance or maintainability issues, or that may already be provided by Cinnamon. Having code flagged by it doesn't automatically disqualify a pull request.
> ### This check is not perfect will not replace a normal review.
> 
> Found 1 potential issue(s):
> ### ⚠️ lang_bind
> 
> **[files/usr/share/cinnamon/applets/printers@cinnamon.org/applet.js:218](https://github.com/linuxmint/cinnamon/pull/13524/files#diff-8be7fbee5740c468b22f4cefae4cc752cd6d49a4cd07a36c20f3eb15b4eb442fR218)**
> 
> ```
> Util.spawn_async(['/usr/bin/lpq', '-a'], Lang.bind(this, function(out2) {
> ```
> 
> `Lang.bind()` is deprecated. Use arrow functions (`() => {}`) or `Function.prototype.bind()` instead.
> 
> _Automated pattern check._

What should be done with that?

leigh123linux:
Please squash the commits into one.

mtwebster:
> What should be done with that?

What it suggests - use an arrow function instead of using Lang.bind(). Look at other applets, it's a pretty straightforward change.

hans-fritz-pommes:
By the way:
I was wondering why ever `lpstat` was used for gathering the job-data.
[This pull request](https://github.com/linuxmint/cinnamon/pull/13039) removed `lpq`. I thought every system having `lpstat` also has `lpq` since it's a part of cups?

files/usr/share/cinnamon/applets/printers@cinnamon.org/applet.js
hans-fritz-pommes: Is corrected now but the document-output of `lpq -a` seems to be always shorter than 31 characters

files/usr/share/cinnamon/applets/printers@cinnamon.org/applet.js
hans-fritz-pommes: You're right

hans-fritz-pommes:
Those new commits intend to make it compatible with a turned-off CUPS-service

hans-fritz-pommes:
> Please squash the commits into one.

~I will  do that when merging~

🤦oups merging is not my task, right?

hans-fritz-pommes:
Now the guy from [https://github.com/linuxmint/cinnamon/pull/13039](https://github.com/linuxmint/cinnamon/pull/13039) shouldn't have problems anymore with missing `lpq` and there shouldn't be a problem with missing / not running `cups` ([https://github.com/linuxmint/cinnamon/issues/9265](https://github.com/linuxmint/cinnamon/issues/9265)).

files/usr/share/cinnamon/applets/printers@cinnamon.org/applet.js
```diff
@@ -102,113 +126,214 @@ class CinnamonPrintersApplet extends Applet.TextIconApplet {
...
-                            jobItem.connect('activate', Lang.bind(jobItem, this.onCancelJobClicked));
+                            jobItem.connect('activate', () => this.onCancelJobClicked(jobItem));
                             this.cancelSubMenu.addMenuItem(jobItem);
-                            if(out2[out2.indexOf(job) - 2] != 'active' && out2[out2.indexOf(job) - 2] != '1st') {
+                            this.jobs.push(jobItem);
+                            if(lpq_present && jobInfo[job][0] != 'active' && jobInfo[job][0] != '1st') {
                                 sendJobs.push(new PopupMenu.PopupIconMenuItem(text, 'xsi-go-up', St.IconType.SYMBOLIC));
                                 sendJobs[sendJobs.length - 1].job = job;
-                                sendJobs[sendJobs.length - 1].connect('activate', Lang.bind(sendJobs[sendJobs.length - 1], this.onSendToFrontClicked));
+                                sendJobs[sendJobs.length - 1].connect('activate', () => this.onSendToFrontClicked(sendJobs[sendJobs.length - 1]));
```
mtwebster: This might not do what you expect:
  
The callback will evaluate `sendJobs[sendJobs.length - 1]` at the time of activation, *not* when you created the callback. You should probably just do something less clever like:

```
let sendjob = new PopupMenu.PopupIconMenuuItem.....
sendJobs.push(sendjob)
...
...
```
then reference `sendJob` as the callback argument.

mtwebster:
Can we drop the python script? `util.spawnCommandLineAsyncIO()` should work fine in this case - they're all fixed arguments. Sorry I should have caught this earlier...

files/usr/share/cinnamon/applets/printers@cinnamon.org/applet.js
```diff
@@ -102,113 +126,214 @@ class CinnamonPrintersApplet extends Applet.TextIconApplet {
...
-                                } else {
-                                    break;
-                                }
+                            let doc = "unknown";
+                            let user = "unknown";
+                            let size = "unknown";
+                            if (lpq_present) {
+                                doc = jobInfo[job][2];
+                                user = jobInfo[job][1];
+                                size = GLib.format_size_for_display(jobInfo[job][3]);
```
mtwebster: `format_size_for_display()` is deprecated - use `format_size()` instead (same arguments).

hans-fritz-pommes: ok

hans-fritz-pommes:
> Can we drop the python script? `util.spawnCommandLineAsyncIO()` should work fine in this case - they're all fixed arguments. Sorry I should have caught this earlier...

Why would it be insecure if there were user inputs?
And why is there no function with `argv` _and_ the possibility to catch `stdout`, `stderr` and `exitCode`?

(It's just about understanding)

hans-fritz-pommes:
Ah and I replaced the second python script - this means now we don't have Gtk-dialogs anymore. Instead: A Cinnamon modal dialog.

files/usr/share/cinnamon/applets/printers@cinnamon.org/applet.js
hans-fritz-pommes: ok

mtwebster:
> Why would it be insecure if there were user inputs?

You can get a lot more info with Google or asking AI to explain shell injection vulnerabilities... but basically,

This is safe - it will always run exactly that command, you're using a static string.
```
Util.spawnCommandLineAsyncIO('/usr/bin/lpq -a', ()=>{});
```
but, for example filename restrictions are fairly lax...

image

```js
let file = file.get_basename() // '; dangerous-command ~'
let cmd = "ls -la " + file;
Util.spawnCommandLineAsyncIO(file, ()=>{}); // ls -la ; dangerous-command ~
```

That example is pretty extreme and unlikely, but it's a real concern, and we're all better off being in the habit of using argv variants in the first place, to prevent misinterpretation like this.

> And why is there no function with argv and the possibility to catch stdout, stderr and exitCode?

Because I didn't notice it was missing before implementing our pattern checker here - I'll be adding one for sure. You could use `GLib.spawn_async_with_pipes` or Gio.Subprocess api directly (which is what spawnCommandLineAsyncIO uses internally), but they're much less convenient to setup and use. For your purposes here, it's ok.

hans-fritz-pommes:
Should I port to `spawnAsyncIO` or just leave it?

mtwebster:
> Should I port to spawnAsyncIO or just leave it?

You can leave as is for now, unless you're already running the latest Cinnamon git master yourself... it's an easy fix down the road. I'm a bit more concerned about the `update()` flow, between the `lpstat -l` and `-d` calls.

I had a graphic generated:

```mermaid
 flowchart TD
      Start["update()"] --> Guard{"updating OR<br/>menu.isOpen OR<br/>removed?"}
      Guard -->|Yes| Bail["return (skip)"]
      Guard -->|No| Init["updating = true<br/>menu.removeAll()<br/>Add 'Printers' item<br/>out1=[] out2='' out3=[]"]

      Init --> LpstatA["<b>lpstat -a</b><br/>(get available printers)"]

      LpstatA -->|error| ErrA["handleError(stderr)<br/>⚠️  menu.removeAll()<br/>updating = false"]
      LpstatA -->|success| StoreOut1["this.out1 = printer list"]

      StoreOut1 --> Fork["<b>⚡ TWO ASYNC CALLS<br/>LAUNCHED IN PARALLEL</b>"]

      Fork --> BranchA["<b>BRANCH A</b><br/>lpstat -l<br/>(tooltip/icon)"]
      Fork --> BranchB["<b>BRANCH B</b><br/>lpstat -d<br/>(default printer)"]

      %% Branch A
      BranchA -->|error| ErrL["handleError(stderr)<br/>⚠️  menu.removeAll()<br/>⚠️  updating = false"]
      BranchA -->|success| IconUpdate["Set tooltip<br/>Set icon based on<br/>⚠️  this.jobs.length<br/>(STALE - not yet
  populated!)"]

      %% Branch B - main chain
      BranchB -->|error| ErrD["handleError(stderr)<br/>⚠️  menu.removeAll()<br/>⚠️  updating = false"]
      BranchB -->|success| Printers["Add separator<br/>Add printer items to menu<br/>Store this.printers"]

      Printers --> LpstatO["<b>lpstat -o</b><br/>(get job list)"]
      LpstatO -->|error| ErrO["handleError(stderr)<br/>updating = false"]
      LpstatO -->|success| Jobs["Store this.out3<br/>Add 'Cancel all' menu<br/>Add cancel submenu"]

      Jobs --> LpqA["<b>lpq -a</b><br/>(get job details)"]
      LpqA -->|error| LpqFail["lpq_present = false<br/>(graceful, no handleError)"]
      LpqA -->|success| LpqOk["lpq_present = true<br/>Parse job info"]

      LpqFail --> BuildJobs
      LpqOk --> BuildJobs

      BuildJobs["Build job menu items<br/>Populate this.jobs[]<br/>⚠️  jobInfo#91;job#93; may be undefined!"] -->
  FinalUI["Set label (job count)<br/>Set visibility<br/>Open submenus<br/><b>updating = false</b>"]

      %% Highlight race conditions
      style Fork fill:#ff6b6b,color:#fff
      style ErrL fill:#ff6b6b,color:#fff
      style IconUpdate fill:#ffaa00,color:#000
      style BuildJobs fill:#ffaa00,color:#000
      style ErrA fill:#ff6b6b,color:#fff
      style ErrD fill:#ff6b6b,color:#fff
      style ErrO fill:#ff6b6b,color:#fff
```
##### The Problems Illustrated

  Race condition (red nodes): After lpstat -a succeeds, lpstat -l (Branch A) and lpstat -d (Branch B) are launched
  simultaneously with no ordering guarantee. Three things can go wrong:

  1. Branch A error destroys Branch B's work: If lpstat -l fails, handleError() calls menu.removeAll() — wiping out
  everything Branch B is building or has built. It also sets updating = false, allowing a new update() to start while Branch
   B is still running.
  2. Branch A reads stale this.jobs: Branch A sets the icon based on this.jobs.length, but this.jobs is only populated much
  later at the end of Branch B's chain (lpq -a callback). So it always uses the previous update cycle's job count.
  3. Double updating = false: If Branch A errors out (setting updating = false) while Branch B is still mid-chain, a new
  update() could start, calling menu.removeAll() and clobbering Branch B.

  The fix would be to make lpstat -l part of the sequential chain rather than parallel — e.g., call it at the very end after
   jobs are populated, which is where the original code had it.

hans-fritz-pommes:
Yes, I was intending to make it faster - which is a very small advantage compared to the possible problems...

hans-fritz-pommes:
Sadly, I don't know how to test the error handling of `lpstat -l` now. The icon update and tooltip update in case of no errors when calling `lpstat -l` are working.

mtwebster:
Hi,

I hate to do this, but I've been testing this out for a while and I'm not real crazy about continuing to rely on shell commands here.

It turns out cups-daemon dbus interface (`org.cups.cupsd.Notifier`) that we currently only subscribe to to trigger updates can also provide pretty much all the information we need for printers and jobs - we can just react to individual signals and the information they provide, no parsing shell output. We also get the benefit of clear status messages if something is wrong (out of paper, etc...)

The only thing we need to run commands for is actually canceling or re-ordering a job.

I'm working on refactoring the applet to work off the cups notifier - I'll include the fixes you introduce here regarding 'cancel-all' and job descriptions.

PR #13527, opened 2026-02-03
menu@cinnamon.org: Replace "Uninstall" context menu item with "App Info"

fredcw:
"Uninstall" calls cinnamon-remove-application, a simple mint only script. Replace with "App Info" to open mintinstall or pamac-manager (many arch based distros) on the app details page so user can view an app's details and optionally uninstall using the OS's software manager. Other distro's GUI software managers could optionally be added later.

Requires: https://github.com/linuxmint/mintinstall/pull/492


fredcw:
oops, messed up the merge conflict.

PR #13582, opened 2026-02-17
gwl: Avoid adding the window again after removal

anaximeno:
This was triggering when using the show window in all workspaces option and was leading the app group representation to an irregular state where removed windows would still be counted and show irresponsible thumbnails.


anaximeno:
Replaced by https://github.com/linuxmint/cinnamon/pull/13599

PR #13604, opened 2026-02-26
Center menu on monitor if applet in center zone of horizontal panel

fredcw:
Fixes https://github.com/orgs/linuxmint/discussions/555

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1243,6 +1243,7 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
     constructor(orientation, panel_height, instance_id) {
         super(orientation, panel_height, instance_id);
 
+        this.orientation = orientation;
         this.setAllowedLayout(Applet.AllowedLayout.BOTH);
 
         this.set_applet_tooltip(_("Menu"));
@@ -1507,6 +1508,7 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
     }
 
     on_orientation_changed (orientation) {
+        this.orientation = orientation;
         this._updateIconAndLabel();
         this._size_dirty = true;
     }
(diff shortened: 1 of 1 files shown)


anaximeno:
Shouldn't it be centered over the applet in general, then, instead of the monitor? (seems to already be the behavior, probably better setting this as a settings option then imo)

fredcw:
Normally it's centered over the applet, yes. But with Cinnamenu, a few people requested that it be centered on the monitor when the applet is in the center zone of the panel and it's been this way in Cinnamenu for 4 years and nobody's complained yet so I don't think it needs to be an option. This also matches the windows 11 menu behaviour.

anaximeno:
I see, sounds like a potential UX improvement for that case. I see it as being useful for users who rely on opening the menu mostly using the keyboard shortcut in cases when the menu is in the center area, since, unlike when you use the cursor, you don't have complete spatial awareness for the menu applet button (to figure were the menu will appear), but on the other side it can be a bit weird for those using the mouse cursor, I imagine. However, I don't have a strong position on this (I mean it being a settings option or just the default behavior in that); so I'm ok with whatever the Mint team arbitrates on that :-)

Just one last point here, you might want to update `this.orientation` in the `on_orientation_changed` method as well, to handle when the menu is moved from a horizontal panel to a vertical one.

I also found this issue (that was there before) while testing the point above. Also, add `this.menu.setOrientation(orientation)` in the `on_orientation_changed` as well, so the menu recognizes that in that case (otherwise it'll appear displaced on the screen when you move the panel from horizontal to vertical).

```js
on_orientation_changed (orientation) {
    this.orientation = orientation;
    this.menu.setOrientation(orientation);
    this._updateIconAndLabel();
    this._size_dirty = true;
}
```

JosephMcc:
I don't really get why the menu should behave different from any other applet. Changing it to this makes no sense at all to me.

How Cinnamenu behave has not bearing on this. It's a 3rd party spice. If someone didn't like the behavior, they would simply disable it. People treat the built in applets much differently.

fredcw:
@anaximeno Nice catch, thanks. PR updated.

fredcw:
@JosephMcc 
>I don't really get why the menu should behave different from any other applet.

I think because it's a big applet (usually the biggest) and well, it's main menu. Small applets you'd expect to be centered on the applet but it just doesn't look as tidy if you have the big main menu slightly off center IMO. It also seems to be a *relatively* popular request: https://github.com/orgs/linuxmint/discussions/555

mtwebster:
A couple of issues:
- The first time I open the menu, it's not centered.
- It's never centered on vertical panels.

Note: I'm testing on Cinnamon git master, though, not 6.6.*

fredcw:
@mtwebster 

>The first time I open the menu, it's not centered....on Cinnamon git master

I can't reproduce this.

>It's never centered on vertical panels.

This didn't really occur to me as nobody requested it and it's mainly just to emulate the behaviour in windows, but can add that easily enough. But now that you mention it, another thing occurs to me: would it not be better if all applets are centered on their icons on vertical panels just as they are on horizontal panels?

PR #13613, opened 2026-03-01
Panel cleanup 2

JosephMcc:
This ports a few more things over to classes and GObject. It removes two more uses of CinnamonGenericContainer in the PanelDummy and Panel objects. This could be taken further but doing the best I can to minimize any possible breakage.

The entire file is a bit of a mess in styling. Cleaned up a bit but it could use some more work.

Tested this quite a bit but it's possible there are a few lingering issues.

### DIFF
--- a/js/ui/appletManager.js
+++ b/js/ui/appletManager.js
@@ -597,7 +597,7 @@ function createApplet(extension, appletDefinition, panel = null) {
         // FIXME: Panel height is now available before an applet is initialized,
         // so we don't need to pass it to the constructor anymore, but would
         // require a compatibility clean-up effort.
-        applet = extension.module.main(extension.meta, orientation, panel.heightForZone, applet_id);
+        applet = extension.module.main(extension.meta, orientation, panel.heightForZones, applet_id);
     } catch (e) {
         Extension.logError(`Failed to evaluate 'main' function on applet: ${uuid}/${applet_id}`, uuid, e);
         return null;


js/ui/appletManager.js
```diff
@@ -597,7 +597,7 @@ function createApplet(extension, appletDefinition, panel = null) {
         // FIXME: Panel height is now available before an applet is initialized,
         // so we don't need to pass it to the constructor anymore, but would
         // require a compatibility clean-up effort.
-        applet = extension.module.main(extension.meta, orientation, panel.height, applet_id);
+        applet = extension.module.main(extension.meta, orientation, panel.heightForZone, applet_id);
```
mtwebster: should be Zones not Zone?

mtwebster:
Quite a few spices check `panel.height` - they'll probably be ok except in hidpi, or does it even matter?

I wonder if I should flag it in our PR scanner..

edit: It looks like panel.height was the 'real' height already, though I guess vertical panels would break this... so we probably need to flag them in the scanner...

js/ui/appletManager.js
JosephMcc: Good catch. Fixed now.

JosephMcc:
It probably does need to be flagged. I almost left that commit out to do in a separate PR because I knew it might cause issues. Previously, panel.height() returned either the panels height or width based on the orientation. Using a variable name that shadowed an objects actual property wasn't a good choice to begin with. Now that it's an actual object, panel.height will return the panels real height regardless of orientation. That's probably not what's wanted in most cases. 

In summary, it had to be renamed to something that wasn't already one of the widgets actual properties.

PR #13616, opened 2026-03-03
keyboardManager: don't append locale layout if already present

Fizmatik:
## Problem

`_buildGroupStrings()` in `js/ui/keyboardManager.js` unconditionally appends the locale-derived layout (from `_getLocaleLayout()`) to every XKB group via `.concat(this._localeLayoutInfo)`.

When the user's configured layouts already include the locale layout (e.g. `ru` layout with `ru_RU.UTF-8` locale), this produces duplicates — `setxkbmap` reports `us,ru,ru` instead of `us,ru`, adding a phantom third layout to the switcher.

This affects all users whose system locale matches one of their configured keyboard layouts (common for Russian, Ukrainian, German, French, and many other non-English locales).

### Steps to reproduce

1. Set system locale to `ru_RU.UTF-8` (or any non-English locale)
2. Configure keyboard layouts: English (US) + Russian
3. Check with `setxkbmap -query`

**Expected:** `layout: us,ru`
**Actual:** `layout: us,ru,ru`

## Fix

Check whether the locale layout is already present in the group before appending it. This preserves the original intent (ensuring a locale layout slot is always available for toolkit mnemonics) while avoiding duplication.

## Testing

Tested on Linux Mint 22 Cinnamon (cinnamon 6.6.7, locale `ru_RU.UTF-8`). After the fix, `setxkbmap -query` correctly reports `us,ru` and the layout switcher shows exactly two layouts.

### DIFF
--- a/js/ui/keyboardManager.js
+++ b/js/ui/keyboardManager.js
@@ -159,9 +159,9 @@ var KeyboardManager = class {
     }
 
     _buildGroupStrings(_group) {
-        let dominated = _group.some(g => g.layout === this._localeLayoutInfo.layout &&
-                                         g.variant === this._localeLayoutInfo.variant);
-        let group = dominated ? _group : _group.concat(this._localeLayoutInfo);
+        let alreadyIncluded = _group.some(g => g.layout === this._localeLayoutInfo.layout &&
+                                                g.variant === this._localeLayoutInfo.variant);
+        let group = alreadyIncluded ? _group : _group.concat(this._localeLayoutInfo);
         let layouts = group.map(g => g.layout).join(',');
         let variants = group.map(g => g.variant).join(',');
         return [layouts, variants];


js/ui/keyboardManager.js
```diff
@@ -159,7 +159,9 @@ var KeyboardManager = class {
     }
 
     _buildGroupStrings(_group) {
-        let group = _group.concat(this._localeLayoutInfo);
+        let dominated = _group.some(g => g.layout === this._localeLayoutInfo.layout &&
```
mtwebster: Can you use a more self-descriptive variable name here? `alreadyIncluded` or even `exists` would be better.

Fizmatik: Done, renamed to `alreadyIncluded` in eabd0f8.

mtwebster:
Thanks, Claude likes using that variable name for some reason... I can't imagine what content it trained on to get that impression.

PR #13624, opened 2026-03-07
screensheild.js: Listen to changes to allow floating setting

JosephMcc:
We were only reading this setting at creation time and not updating on settings changes. The timer to start widgets floating also wasn't getting started on first show.


mtwebster:
Why don't we just check it in `activate()` (instead of at startup) - then we don't really need a listener.

JosephMcc:
Sure, I can do that. Is there a reason one way is better than the other?

JosephMcc:
Done. Updated and force pushed.

mtwebster:
Just so you don't need a signal handler - the setting will never change while the screensaver's active.

PR #13639, opened 2026-03-13
notification: Limit min/max height and enable vscroll

anaximeno:
Avoids cases where notifications would grow beyond the monitor screen height.

### Before
Captura de ecrã de 2026-03-12 20-49-43

### After

image


anaximeno:
replaced by https://github.com/linuxmint/cinnamon/pull/13640

PR #13640, opened 2026-03-13
[WIP] notification: Limit max height and enable vscroll

anaximeno:
Avoids cases where notifications would grow beyond the monitor screen height.

### Before
Captura de ecrã de 2026-03-12 20-49-43

### After

image

### DIFF
--- a/files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js
@@ -74,7 +74,7 @@ class CinnamonNotificationsApplet extends Applet.TextIconApplet {
     }
 
     _openMenu() {
-        this._update_timestamp();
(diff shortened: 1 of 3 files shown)


mtwebster:
I think this is good for a notification's initial display, but maybe the max height can be increased, or even based on some percentage of the monitor height.

Once it's in the notification *applet*, however, it should revert to its full size and remove the individual scrolling altogether - I think the parent being scrollable there is enough, and having both is confusing (scroll works for the outer but not the inner).

anaximeno:
> I think this is good for a notification's initial display, but maybe the max height can be increased, or even based on some percentage of the monitor height.
> 
> Once it's in the notification _applet_, however, it should revert to its full size and remove the individual scrolling altogether - I think the parent being scrollable there is enough, and having both is confusing (scroll works for the outer but not the inner).

Done.

Issue #11427, opened 2023-01-16
Missing icon, with modified *.desktop file


Nightwing0815:
When you modify *.desktop files with a new entry, the icon in context menu is missing, but only in start menu.
The linked file in panel will show an icon, but not the right one (better, than nothing).
Two examples to reproduce:

firefox.desktop (new entry: open a new tab)

```
[Desktop Entry]
Version=1.0
Name=Firefox Web Browser
Comment=Browse the World Wide Web
GenericName=Web Browser
Keywords=Internet;WWW;Browser;Web;Explorer
Keywords[de]=Internet;WWW;Browser;Web;Explorer;Webseite;Site;surfen;online;browsen
Exec=firefox %u
Terminal=false
X-MultipleArgs=false
Type=Application
Icon=firefox
Categories=GNOME;GTK;Network;WebBrowser;
MimeType=text/html;text/xml;application/xhtml+xml;application/xml;application/rss+xml;application/rdf+xml;image/gif;image/jpeg;image/png;x-scheme-handler/http;x-scheme-handler/https;x-scheme-handler/ftp;x-scheme-handler/chrome;video/webm;application/x-xpinstall;
StartupNotify=true
NoDisplay=false
Actions=new-tab;new-window;new-private-window;

[Desktop Action new-tab]
Name=Open a new tab
Name[de]=Einen neuen Tab öffnen
Exec=firefox -new-tab about:newtab

[Desktop Action new-window]
Name=Open a New Window
Name[de]=Ein neues Fenster öffnen
Exec=firefox -new-window

[Desktop Action new-private-window]
Name=Open a New Private Window
Name[de]=Ein neues privates Fenster öffnen
Exec=firefox -private-window
```

nemo.desktop (open as root)

```
#!/usr/bin/env xdg-open
[Desktop Entry]
Name=Files
Name[de]=Dateien
Comment=Access and organize files
Comment[de]=Dateien aufrufen und organisieren
Exec=nemo %U
Icon=system-file-manager
# Translators: these are keywords of the file manager
Keywords=folders;filesystem;explorer;
Terminal=false
Type=Application
StartupNotify=false
Categories=GNOME;GTK;Utility;Core;
MimeType=inode/directory;application/x-gnome-saved-search;
Actions=open-home;open-root;

[Desktop Action open-home]
Name=Home
Exec=nemo %U

[Desktop Action open-root]
Name=Root
Name[de]=Systemverwaltung
# Exec=sudo -H nemo %U
[...]

PR #13641, opened 2026-03-13
extend DESKTOP_ACTION_ICON_NAMES

AntVil:
I saw that it wasn't possible to add icons manually through the `.desktop` files. After some digging i found this file which seems to decide what icon to use based on the action identifier.

I think a nice option would be to have the power to customize this entirely through the the `.desktop` file, but a far easier approach (for now) would be to just add more identifiers.

I tried to only add icons which are general purpose for various people and not specific to my use-case. If it weren't intended for general purpose i would have added more like a mapping coding-projects.


AntVil:
i think this is related to https://github.com/linuxmint/cinnamon/issues/11427

PR #13643, opened 2026-03-14
Fix several bugs: regex flags, implicit globals, crashes, ngettext, undefined icon

KAMI911:
## Summary

This PR contains 8 small, independent bug fixes found during a code review. Each fix is a separate commit.

---

### 1. `util: fixupPCIDescription: fix missing global flag in regex replacements`

Two regex replacements in `fixupPCIDescription()` were missing the `/g` flag, causing only the **first** occurrence to be replaced:
- `/[_,]/` → `/[_,]/g` — replace all underscores/commas with spaces
- `/\([\s\S][^\(\)]{2,}\)/` → same with `/g` — strip **all** parenthesized segments, not just the first

---

### 2. `keyboardManager: fix implicit global variable declarations`

Five constants at the top of the file (`DESKTOP_INPUT_SOURCES_SCHEMA`, `KEY_INPUT_SOURCES`, `KEY_KEYBOARD_OPTIONS`, `KEY_PER_WINDOW`, `KEY_SOURCE_LAYOUTS`) were assigned without `var`, making them implicit globals that pollute the global scope. Added `var` to match the style used elsewhere in the file.

---

### 3. `notificationDaemon: add default case for unknown urgency in icon switch`

The `switch (hints.urgency)` in `_iconForNotificationData()` had no `default` branch, leaving `stockIcon` as `undefined` for any urgency value outside LOW/NORMAL/CRITICAL. This silently created a broken `St.Icon` with `icon_name: undefined`. Added a `default` case falling back to the information icon.

---

### 4. `notificationDaemon: fix crash in _expireNotification when notification is null`

`NotifyAsync()` adds `ndata` to `_expireNotifications` and starts the expire timer **before** the async `GetConnectionUnixProcessID` D-Bus call completes. `ndata.notification` is only assigned later inside `_notifyForSource()`. If the timer fires during that window, calling `ndata.notification.destroy()` on `undefined` throws a `TypeError`. Added a null guard; if the notification object doesn't exist yet, the orphaned entry is removed and the timer is restarted for the next item.

---

### 5. `windowManager: fix ngettext missing count argument in display-change dialog`

`ngettext(singular, plural)` requires a third argument — the count used to select the correct plural form. Without it the function always returns the singular string (e.g. _"Reverting in 5 **second**."_ instead of _"Reverting in 5 **seconds**."_). Fixed to match the pattern used in `endSessionDialog.js`.

---

### 6. `cinnamonDBus: fix crash in activateCallback when xlet is not found`

`_getXletObject()` returns `null` when no applet/desklet/extension matches the given uuid/instance_id. The next line called `null[callback].bind(null)`, throwing a `TypeError`. This happens when cinnamon-settings calls a callback for an xlet that was removed without a shell restart. Added a null guard with a warning log.

---

### 7. `cinnamonDBus: fix crash in updateSetting when instance_id is not found`

`updateSetting()` already guarded against an unknown `uuid`, but not against an unknown `instance_id` within a valid `uuid`. Accessing `uuids[uuid][instance_id].remoteUpdate()` when `instance_id` is absent throws _"Cannot read property 'remoteUpdate' of undefined"_. Added a second guard with a warning log.

---

### 8. `cinnamonDBus: remove leftover debug comment in GetInputSources`

Removed a stray `// global.log(source.preferences);` debug line accidentally left in `GetInputSources()`.

---

## Test plan

- [ ] PCI device description cleanup in network manager shows all underscores/commas replaced and all parenthesised segments stripped
- [ ] Keyboard layout switching works correctly (no regressions from `var` fix)
- [ ] Notifications with non-standard urgency values display the information icon
- [ ] No crash when a short-expiry notification fires before its source PID is resolved
- [ ] Display-change countdown dialog shows correct plural form ("5 seconds", "1 second")
- [ ] No crash in cinnamon-settings when calling callbacks or updating settings for a removed xlet
- [ ] `GetInputSources` D-Bus method works correctly

### DIFF
--- a/js/ui/cinnamonDBus.js
+++ b/js/ui/cinnamonDBus.js
@@ -560,7 +560,6 @@ var CinnamonDBus = class {
 
         for (let idx in sources) {
             const source = sources[idx];
-            // global.log(source.preferences);
             ret.push([
                 source.type,
                 source.id,
--- a/js/ui/cinnamonDBus.js
+++ b/js/ui/cinnamonDBus.js
@@ -374,6 +374,10 @@ var CinnamonDBus = class {
 
     activateCallback(callback, uuid, instance_id) {
         let obj = this._getXletObject(uuid, instance_id);
+        if (!obj) {
+            global.logWarning(`[CinnamonDBus] activateCallback: xlet not found (uuid=${uuid}, instance_id=${instance_id})`);
+            return;
+        }
         let cb = obj[callback].bind(obj);
         cb();
     }
--- a/js/ui/cinnamonDBus.js
+++ b/js/ui/cinnamonDBus.js
@@ -390,6 +390,13 @@ var CinnamonDBus = class {
             );
             return;
         }
+        if (!Main.settingsManager.uuids[uuid][instance_id]) {
+            global.logWarning(
+                `[CinnamonDBus] [${uuid}] Unable to find instance '${instance_id}' from SettingsManager - ` +
+                'this is likely due to configuring settings from a removed xlet instance.'
+            );
+            return;
+        }
         Main.settingsManager.uuids[uuid][instance_id].remoteUpdate(key, payload);
     }
 
--- a/js/ui/cinnamonDBus.js
+++ b/js/ui/cinnamonDBus.js
@@ -391,10 +391,7 @@ var CinnamonDBus = class {
             return;
         }
         if (!Main.settingsManager.uuids[uuid][instance_id]) {
-            global.logWarning(
-                `[CinnamonDBus] [${uuid}] Unable to find instance '${instance_id}' from SettingsManager - ` +
-                'this is likely due to configuring settings from a removed xlet instance.'
-            );
+            global.logWarning(`[CinnamonDBus] [${uuid}] Unable to find instance '${instance_id}' from SettingsManager`);
             return;
         }
         Main.settingsManager.uuids[uuid][instance_id].remoteUpdate(key, payload);
--- a/js/ui/notificationDaemon.js
+++ b/js/ui/notificationDaemon.js
@@ -268,7 +268,13 @@ NotificationDaemon.prototype = {
         let ndata = this._expireNotifications[0];
 
         if (ndata) {
-            ndata.notification.destroy(MessageTray.NotificationDestroyedReason.EXPIRED);
+            if (ndata.notification) {
+                ndata.notification.destroy(MessageTray.NotificationDestroyedReason.EXPIRED);
+            } else {
+                // notification object not yet created (async PID lookup still pending)
+                this._expireNotifications.shift();
+                this._restartExpire();
+            }
         }
 
         this._expireTimer = 0;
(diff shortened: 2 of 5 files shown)


js/ui/notificationDaemon.js
```diff
@@ -173,6 +173,9 @@ NotificationDaemon.prototype = {
                 case Urgency.CRITICAL:
                     stockIcon = 'xsi-dialog-error-symbolic';
                     break;
+                default:
```
Cobinja: Unnecessary. The type Urgency only has 3 possible values, all of which are already handled.

js/ui/cinnamonDBus.js
```diff
@@ -386,6 +390,13 @@ var CinnamonDBus = class {
             );
             return;
         }
+        if (!Main.settingsManager.uuids[uuid][instance_id]) {
+            global.logWarning(
+                `[CinnamonDBus] [${uuid}] Unable to find instance '${instance_id}' from SettingsManager - ` +
+                'this is likely due to configuring settings from a removed xlet instance.'
```
mtwebster: .... uuid] Unable to find instance 'instance_id' from SettingsManager' is enough, keep the message short.

mtwebster:
I'm ok with everything but expired notification changes - 

You're removing the ndata from the expire queue, but it's still (theoretically) in flight with `GetConnectionUnixProcessIDRemote` - now once that completes this notification may never expire. Why not consider just restarting the timer?

It's all very unlikely in any event - this can only happen if a notification sets a custom timeout of under 1 second, *and* the dbus lookup takes longer than that (*extremely* unlikely for a local call like that).

Considering the complexity handling this case would add, I would just leave this code alone, unless you have a real-world scenario where this occurs.

The remainder of the PR looks fine (other than the overly-long log warning).

KAMI911:
Thanks for the review! I've pushed two follow-up commits:

- Reverted the `_expireNotification` change and left the original code as-is, per your point about the entry becoming orphaned in the expire queue once the async PID lookup resolves.
- Shortened the `updateSetting` log warning to a single line.

For reference, in case it's ever worth revisiting: instead of shifting the pending entry out of `_expireNotifications` on timer fire, it could stay in the queue and the timer could just retry shortly until `ndata.notification` exists, e.g.:

```js
_expireNotification: function() {
    let ndata = this._expireNotifications[0];

    if (ndata) {
        if (ndata.notification) {
            ndata.notification.destroy(MessageTray.NotificationDestroyedReason.EXPIRED);
        } else {
            // Notification actor not created yet (async PID lookup still
            // pending) -- keep it in the queue and retry shortly instead of
            // dropping it, so it still expires once _notifyForSource runs.
            this._expireTimer = Mainloop.timeout_add_seconds(1, Lang.bind(this, this._expireNotification));
            return false;
        }
    }

    this._expireTimer = 0;
    return false;
},
```

That keeps the entry trackable so it still expires (and still emits `NotificationClosed`) once the notification is eventually created, without needing to touch anything else. That said, I agree it's not worth the added complexity for a race this unlikely, so I'll leave it out unless it actually comes up in practice.

PR #13648, opened 2026-03-19
Remove previous track actors from coverBox

heapframe:
Fix bug where previous tracks render under current track.

This bug is visible when you play a couple tracks on something like soundcloud (with thumbnails), open the sound applet, then when closing it, the previous tracks thumbnail would be visible when fading out the current tracks thumbnail on the close animation


mtwebster:
Hi, why are actors accumulating this way? That's the bug we need to fix, not just clean up after it.

heapframe:
Ah, it was _showCover silently failing when it thinks a track has no cover art (probably another issue then, as i only found this issue by using tracks with cover art), causing them to pile up because it never swapped it out in coverBox after setting St.Icon

PR #13658, opened 2026-03-22
Add PopupDialog widget - draggable, re-focusable dialog.

mtwebster:
- baseDialog.js: adds a base class for shared code between modal
and popup dialog.
- popupDialog.js: Draggable wrapper for Dialog.Dialog, to use in
place of ModalDialog.
- The 'popup-dialog' css class can be used to help differeniation.
In the default theme, there's a faint box shadow that uses the
accent color.

ref:
https://github.com/orgs/linuxmint/discussions/774

### DIFF
--- a/data/theme/cinnamon-sass/widgets/_dialogs.scss
+++ b/data/theme/cinnamon-sass/widgets/_dialogs.scss
@@ -194,3 +194,10 @@
     }
   }
 }
+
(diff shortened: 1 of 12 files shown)


mtwebster:
merged - 81eba1c670c1e4ff1b2747d704247f7aa246ec81, 7efb8a0beb36a8736dc3e720f8268a62ad0970c4

PR #13667, opened 2026-03-25
cinnamon-settings: make extension toggles directly clickable

jleaders:
Removed the +/- buttons and just use clickable checkboxes
image

## Summary
- replace the passive enabled indicator in the Extensions list with an inline checkbox
- wire the checkbox directly to the existing enable/disable logic so extensions can be toggled from the row itself
- hide the old bottom enable/disable toolbar buttons on the Extensions page so the primary action is easier to discover

## Test plan
- [x] Overlay the modified `cinnamon-settings` Python files onto a local Cinnamon install
- [x] Open `cinnamon-settings extensions --tab installed`
- [x] Verify the bottom `+` / `-` buttons are hidden for extensions
- [x] Verify clicking the row checkbox enables and disables an extension
- [x] Verify the configure button sensitivity still tracks the enabled state

Made with [Cursor](https://cursor.com)

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/ExtensionCore.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/ExtensionCore.py
@@ -160,10 +160,12 @@ def sanitize_html(string):
 
 
 class ManageSpicesRow(Gtk.ListBoxRow):
-    def __init__(self, extension_type, metadata, size_groups):
+    def __init__(self, extension_type, metadata, size_groups, use_inline_toggle=False, toggle_callback=None):
         super().__init__()
         self.extension_type = extension_type
         self.metadata = metadata
+        self.use_inline_toggle = use_inline_toggle
+        self.toggle_callback = toggle_callback
 
         self.status_ids = {}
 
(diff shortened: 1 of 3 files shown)


anaximeno:
I personally prefer the +/- buttons here.

And even in this case, it doesn't look good in terms of UI imo, probably could be done better with a plus button besides the gear icon, perhaps, which would look disabled once the extension is enabled.

It also adds specific complexity to a generic spice settings module.

fredcw:
I like it. An extension is either enabled or disabled unlike the applets and desklets where you can add more than one, so the distinction makes sense. I think it's more intuitive.

anaximeno:
> I like it. An extension is either enabled or disabled unlike the applets and desklets where you can add more than one, so the distinction makes sense. I think it's more intuitive.

Remember you can restrict applets to one instance too if you want (I think desklets as well), while this could contribute to a specific situation I'm not sure for the generic case.

jleaders:
@anaximeno 
I'm a little bit confused why you would prefer a simple checkbox to be checked by a separate plus or minus button in a different location? I'm not joking when I say it actually took me a good fifteen minutes the first time using Cinnamon Spice Extensions to figure out how the heck to enable an extension, because, for starters, a plus button means to "add" something, not to enable something. I had already installed the extension and it was already added and my user intent was not to add a new spice extension, but to enable an existing one, so why would I push a plus button? Well I didn't ... and for fifteen minutes I sat there trying all sorts of different things and even Googling how to enable the extension and get it working. For a while I thought the extension was broken and and had a bug. Nope. It turns out to check a box I have to:
 1) select the row 
 2) move my cursor away from the box to the bottom of the UI 
 3) hit plus button
 
 Now you just hit the checkbox... like all checkboxes.

JosephMcc:
To keep things simple. Right now it is consistent with the other spices pages. This change makes it so it isn't. IMHO, it should stay like it is.

pilatomic:
jleaders : one info you might be missing, is that while extensions can only be added once, applets and desklets can be added several times, to create multiple instances, so for those, checkboxes can not be used.

jleaders:
Okay, I understand the context now. I opened the desklets page, which I had never used before, and you're right that the plus button can be hit multiple times to add more than one to the desktop. I see how the goal is to keep it consistent across multiple user interface designs. 

To be honest, I'm a little bit on the fence on the right call. You can't add multiple for the extensions, so the plus/minus just gets confusing in this context.  It's a consistency versus usability trade off.

Okay, your call. If you want to close this PR I'm fine with it.

jleaders:
One alternative idea I came up with is maybe merge both, keep the checkboxes functional, but make the plus or minus buttons also check or uncheck. So we have consistency but also usability. Let me know, and I can update this PR to match that.

PR #13674, opened 2026-03-29
printers@cinnamon.org: Rework layout, use cupsd dbus notifications for printer and job info.

mtwebster:
We've long used the cupsd notifier to tell us when to run shell commands that collect job and printer data, but this data has been provided by these notifications all along, and it much more relable than parsing shell output (which can change unexpectedly with new versions or on different distributions).

- Stop using shell commands to gather job and printer information. They're now used only to cancel or prioritize jobs, and to trigger a cupsd notification to populate the applet initially.
- Organize jobs under each printer when multiple printers are available.
- Provide access to individual printer properties.
- Move the 'Printers' menu item to the context menu, except when no printers have been discovered.
- Use job/printer state-reasons for reporting abnormal conditions such as no-paper. Icon changes accordingly.
- Cancel-all avoids permission issues for some users by cancelling jobs individually (credit @hans-fritz-pommes)
- Cancel prompts use native Cinnamon dialogs.

image


hans-fritz-pommes:
Nice!

PR #13677, opened 2026-03-30
Change session type from Application to XSession

FedGuy699:
This fixes an issue with DisplayLink monitors not working on Cinnamon, as EVDI will not initialize them if it is not an XSession. This is also the proper type anyway. https://codesearch.debian.net/search?q=Type%3DXSession


mtwebster:
Please update cinnamon2d.desktop.in also.

PR #13684, opened 2026-04-01
Update cs_general.py

### LINKED ISSUE #10287: cinnamon general settings minimum memory limit of 1024MB too high for many systems
System information in attached file.
[inxi.log](https://github.com/linuxmint/cinnamon/files/7040137/inxi.log)

I have an older system with 8GB of memory.  When started cinnamon is using about 70MB, so a 1024MB limit is far too high for this system.
[cinnamon_monitor.zip](https://github.com/linuxmint/cinnamon/files/7040162/cinnamon_monitor.zip)

Until this lower limit is relaxed I will need to continue using the script in the attached zip file to keep cinnamon memory usage under control.

Feel free to leave the default where it is, but please reduce the minimum limit to 256MB.

mixmastamyk:
Lower minimum allowable memory limit in settings GUI.  Does not change default.

Resolves #10287

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_general.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_general.py
@@ -44,7 +44,7 @@ class Module:
                 switch = GSettingsSwitch(_("Restart Cinnamon when it uses too much memory"), "org.cinnamon.launcher", "memory-limit-enabled")
                 settings.add_row(switch)
 
-                spin = GSettingsSpinButton(_("Memory limit"), "org.cinnamon.launcher", "memory-limit", _("MB"), 1024, 36000, 1, 100)
+                spin = GSettingsSpinButton(_("Memory limit"), "org.cinnamon.launcher", "memory-limit", _("MB"), 256, 36000, 1, 100)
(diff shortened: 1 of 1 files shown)


mtwebster:
256 is far too low to avoid constantly having restarts for the unsuspecting user.

This isn't intended to be some sort active of memory management feature, but an escape hatch if the user has installed some poorly-behaved applet. 1024 is a reasonable, safe minimum, and it can already easily be adjusted lower if desired:

```
gsettings set org.cinnamon.launcher memory-limit 256
```
or with dconf-editor:
image

mixmastamyk:
Hi, this is a minimum allowable value for the interface.  It is not a recommendation or even a default.  It simply lets you drop the number conveniently, if you choose.  People will decide what is appropriate on their own.

If you won't give people a real choice, there's no point for this GUI setting to exist.  In that case it would be preferable to remove it rather than provide the illusion of choice.

(Many machines, especially older ones cannot afford 1 or 2 GB to run a desktop, that's a significant fraction of an 8GB machine which are common in the market today, Apple just debuted a new one.  W95 ran in 8MB comfortably in comparison.  Only on perhaps a 32GB machine would this amount of RAM be reasonable to waste.)  Cheers,

Issue #13686, opened 2026-04-02
Power Applet: Battery icon shows charging but text label says "Not charging"


ADMarian:
### Distribution

Mint 22.3 - Cinnamon 64-bit

### Package version

6.6.7

### Graphics hardware in use

NVIDIA Corporation TU106M [GeForce RTX 2060 Mobile] | Intel Corporation CoffeeLake-H GT2 [UHD Graphics 630]

### Frequency

Always

### Bug description

Hi,

When the laptop is plugged into AC power and the battery is low (icon showing the orange level in the applet), the battery icon correctly updates to the "charging" state (bolt icon), but the text label in the panel or the tooltip incorrectly displays "Not charging". Eventually it will start working as expected

Output of `upower -i $(upower -e | grep 'BAT')`

native-path:          BAT0
  vendor:               ASUSTeK
  model:                ASUS Battery
  power supply:         yes
  updated:              Thu 02 Apr 2026 01:40:54 PM EEST (1 seconds ago)
  has history:          yes
  has statistics:       yes
  battery
    present:             yes
    rechargeable:        yes
    state:               pending-charge
    warning-level:       none
    energy:              13,419 Wh
    energy-empty:        0 Wh
    energy-full:         46,278 Wh
    energy-full-design:  66 Wh
    energy-rate:         0 W
    voltage:             15,677 V
    charge-cycles:       N/A
    percentage:          28%
    capacity:            70,1182%
    technology:          lithium-ion
    icon-name:          'battery-low-charging-symbolic'

Below a screenshot of the issue:

<img width="347" height="310" alt="Image" src="https://github.com/user-attachments/assets/d2221dc7-d692-4c2f-8b7f-37c4961aa3fb" />

### Steps to reproduce

Connect the laptop to a power source, observe the icon change to the charging state, hover over the icon or check the panel label to see the "Not charging" text.

### Expected behavior

Connect the laptop to a power source, observe the icon change to the charging state, hover over the icon or check the panel label to see the "Charging" text.

Konstantin-The-Goose:
i think i fixed it i am waiting for the pull request to be reviewed

insomniacno1:
This is basically the same issue as mine, and my laptop is connected to charger as you can see in Power settings.

21.3 Cinnamon 64bit

<img width="124" height="28" alt="Image" src="https://github.com/user-attachments/assets/b998d388-0de3-4e98-a6fe-618e58c6a66f" />
[...]

PR #13687, opened 2026-04-03
Fix battery status logic for CHARGING and PENDING_CHARGE states

Konstantin-The-Goose:
Fixed issue #13686 

Previously, PENDING_CHARGE was unreachable due to duplicated condition checks. This separates the logic so both states are handled correctly (i hope).


leigh123linux:
The commit title sucks, please fix.
Add Fixed issue #13686 to the commit message instead.

PR #13699, opened 2026-04-09
menu@cinnamon.org: Fix keyboard layout switching when menu is open

good-vi:
When the menu applet is open, the search entry captures all key events, preventing keyboard layout switching via Caps Lock (grp:caps_toggle) or other XKB layout switch keys. The ISO_Next_Group/ISO_Prev_Group keysyms generated by XKB are consumed by the search field instead of triggering the layout switch.

Handle these keysyms explicitly in _onMenuKeyPress by calling the input source manager's switcher directly and stopping event propagation.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1804,6 +1804,16 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
 
     _onMenuKeyPress(actor, event) {
         let symbol = event.get_key_symbol();
+
+        // Allow layout switching (Caps_Lock / ISO_Next_Group) when menu is open
+        if (symbol === Clutter.KEY_ISO_Next_Group ||
+            symbol === Clutter.KEY_ISO_Prev_Group ||
+            symbol === Clutter.KEY_Caps_Lock) {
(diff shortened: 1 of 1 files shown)


files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
```diff
@@ -1804,6 +1804,16 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
 
     _onMenuKeyPress(actor, event) {
         let symbol = event.get_key_symbol();
+
+        // Allow layout switching (Caps_Lock / ISO_Next_Group) when menu is open
+        if (symbol === Clutter.KEY_ISO_Next_Group ||
+            symbol === Clutter.KEY_ISO_Prev_Group ||
+            symbol === Clutter.KEY_Caps_Lock) {
```
anaximeno: Would it make sense for Caps Lock as well? Since it has a concrete function when used in an entry field to pull all text in caps.

mtwebster:
This isn't the correct place for this.

fixed for normal keybindings for any entries (not just this applet) in:
https://github.com/linuxmint/cinnamon/commit/58e1c478d71e78f9902fddf7b031aad4f97f3f8a
xkb next-group fixed in:
https://github.com/linuxmint/cinnamon/commit/1898e911f1ba45bf6a2f8d901507c9a64a16ca64

PR #13700, opened 2026-04-09
expo/overview: Refactor to Use newer/gobject class styles, easing.

mtwebster:
- Stop using Lang.bind().
- Use actor easing instead of tweening.
- Use actor/widget subclasses instead of this.actor/GenericContainers.
- Use connectObject/disconnectObject for signal handling.
- Use GLib instead of Mainloop.
- Allow early cancel of both modes.
- Cleanup rest of codebase's access to overview and expo
- Use object vfuncs instead of signal handlers where appropriate.

### DIFF
--- a/js/ui/expoThumbnail.js
+++ b/js/ui/expoThumbnail.js
@@ -1,19 +1,17 @@
 // -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
 
 const Clutter = imports.gi.Clutter;
-const Lang = imports.lang;
-const Mainloop = imports.mainloop;
+const GLib = imports.gi.GLib;
+const GObject = imports.gi.GObject;
 const Meta = imports.gi.Meta;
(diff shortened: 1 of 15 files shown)


js/ui/expoThumbnail.js
```diff
@@ -1084,21 +1115,20 @@ ExpoThumbnailsBox.prototype = {
...
+        this.add_child(this.button);
+
+        this.button.connect('enter-event', () => { this.button.show(); });
+        this.button.connect('leave-event', () => { this.button.hide(); });
+        this.button.connect('clicked', () => { this.lastHovered.remove(); this.button.hide(); });
         this.button.hide();
-                
-        this.actor.connect('scroll-event', this.onScrollEvent);
+
+        this.connect('scroll-event', this.onScrollEvent.bind(this));
```
JosephMcc: Since all this stuff is now objects, do you need to do this? Can't you just use all the parents vfuncs instead of the signal connections?

PR #13704, opened 2026-04-11
notifications@cinnamon.org - Avoid redundant update_list calls during clear all

jsharpe1:
## Problem

When clearing all notifications, each `notification.destroy()` triggers the destroy signal handler, which calls `update_list()` individually. With N notifications, this results in N+1 calls to `update_list()`, each recalculating urgency, updating icons/labels, reordering children, and requesting a relayout. When the notification count gets high, there is a noticeable hang when it clears all notifications. 50 notifications took 6.5 seconds to clear (see screenshot below).

## Fix

Added a `_clearing` flag that causes the destroy handler to skip `update_list()` during bulk clear. The single `update_list()` call at the end of `_clear_all()` handles final state. Wrapped in try/finally to guarantee the flag resets if a destroy call throws.                                                                                  

## Testing

Tested on Cinnamon 6.6.7 by triggering 50 notifications and clearing all. Verified via Looking Glass:

** Before: **
image

** After: **
image

Also verified:
  - Clear all updates tray icon and count correctly
  - Individual notification dismiss still updates the list

### DIFF
--- a/files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js
@@ -243,14 +243,15 @@ class CinnamonNotificationsApplet extends Applet.TextIconApplet {
     }
 
     _clear_all() {
-        let count = this.notifications.length;
-        if (count > 0) {
-            for (let i = count-1; i >=0; i--) {
-                this._notificationbin.remove_actor(this.notifications[i].actor);
-                this.notifications[i].destroy(NotificationDestroyedReason.DISMISSED);
-            }
-        }
+        let toDestroy = this.notifications.slice();
+
+        // Clear tracking state first so destroy handlers skip redundant work.
         this.notifications = [];
+        this._notificationbin.remove_all_children();
+
+        for (let notification of toDestroy) {
+            notification.destroy(NotificationDestroyedReason.DISMISSED);
+        }
         this.update_list();
     }
 


fredcw:
what computer are you using? I have a 4 core ryzen3 and clearing 50 notifications takes about 150ms. There's a max of 20 notifications per app as well.

jsharpe1:
@fredcw This is on my laptop - CPU is 11th Gen Intel(R) Core(TM) i7-11370H @ 3.30GHz.

Out of curiosity, I ran the same test (generating 50 notifications and clearing all) on my desktop, and did _not_ have the same performance issue. That's running a 13th Gen Intel(R) Core(TM) i5-13600KF.

fredcw:
It might be worth investigating this further. Are there any other performance issues on your laptop?

jsharpe1:
Not really on the CPU side. Memory has been the bottleneck more than anything. 

This is what my CPU usage looks like under normal usage: 
image

I did manage to capture the CPU graph when I first hit this issue:
image

enodari:
I was preparing a PR to fix this exact issue, good to know there was already one.

I can attest that the way I use Cinnamon, having a lot of apps that triggers notifications (mainly chat programs but also some custom notifications I have to keep track of things to do) I tend to read them but not dismiss them.

This make it so I regularly find myself with 100+ notifications and on my hardware (AMD Ryzen 7, 32GB RAM) clearing them takes some seconds (just tested: 5s more or less).

Not a big problem per se, but the whole UI freezes and it bothers me a bit, since it's a relatively simple fix (IMHO).

My fix would have been just this:

```
    _clear_all() {
        let toDestroy = this.notifications.slice();
        this.notifications = [];
        this._notificationbin.remove_all_children();
        for (let notification of toDestroy) {
            notification.destroy(NotificationDestroyedReason.DISMISSED);
        }
        this.update_list();
    }
```

But I'm no Cinnamon expert. With this change the clearing is immediate.

fredcw:
@enodari Yes, your method is better as it's not as complicated.

@jsharpe1 Do you want to update this PR to @enodari 's method or should I create a new one?

I just tested again with 40 notifications and it took about 2 seconds and about 38ms with this speedup. I'm not sure why it takes longer sometimes and not other times but in any case, this is still a 10x - 100x speedup so well worth it I think.

jsharpe1:
@enodari Love it, much cleaner, thanks for the suggestion

@fredcw I'll update the PR in a minute

PR #13714, opened 2026-04-16
sound settings: Split overamplication - out/in.

SyberHexen:
disappeared at some point possibly in 520b4db1f0af12c598534efe91f211786a7b2443

### DIFF
--- a/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/sound@cinnamon.org/applet.js
@@ -37,7 +37,8 @@ const VOLUME_ADJUSTMENT_STEP = 0.05; /* Volume adjustment step in % */
 const ICON_SIZE = 28;
 
 const CINNAMON_DESKTOP_SOUNDS = "org.cinnamon.desktop.sound";
-const OVERAMPLIFICATION_KEY = "allow-amplified-volume";
+const OVERAMPLIFICATION_OUT_KEY = "allow-amplified-out-volume";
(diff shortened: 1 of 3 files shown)


anaximeno:
The switch for Overamplification in sound settings is only in the Output section, so from there it doesn't make sense to apply to both Output and Input.

SyberHexen:
I agree this could be done better, but I didn't have much luck splitting the logic tied to the single overamplicfication_key to each tab. 
Alternatively the overamplification toggle could be nested under sounds tab instead to look more in-place?
## edit
Ideas:
1) Bring back the settings tab (or something similar like extras?) to have a single overamplification toggle controlling both output/input.
2) Split amplification per input / output having a toggle on each page (requires splitting the key 'allow-amplified-volume' in [cinnamon-desktop](https://github.com/linuxmint/cinnamon-desktop/blob/master/schemas/org.cinnamon.desktop.sound.gschema.xml.in#L29) and results in duplicated code.

SyberHexen:
1) Bring back settings tab for single toggle idea [here](https://github.com/SyberHexen/cinnamon/commit/2ffdebc66a57c6e8b6a4c8cb00b52e6fde831b4b)
2) This request splitting amplification per output/input page (which i'd rather squash into a single commit) assuming the key 'allow-amplified-volume' in [cinnamon-desktop](https://github.com/linuxmint/cinnamon-desktop/blob/master/schemas/org.cinnamon.desktop.sound.gschema.xml.in#L29) is also split as 'allow-amplified-out-volume' & 'allow-amplified-in-volume' keys.

While (1) still feels a bit out of place, (2) has sane placement and logic for users but is a larger change with likely unneeded code duplication.

PR #13725, opened 2026-04-22
Modified sound@cinnamon.org to autoexpand "Output device" category.

SpeeQz1:
The Linux Mint `sound@cinnamon.org` applet doesn't display the system's output devices by default when a user right clicks on the applet to open the context menu.

This change modifies it so that the category `Output device` is already expanded and doesn't hide what audio devices the user's system has which should make their discovery easier.


claudiux:
This should be an option (which already exists in the Enhanced Sound applet).

SpeeQz1:
@claudiux Implemented toggleable options, for both `Applications` and `Output device`.

`Output device` is expanded by default and `Applications` is not.

image

PR #13736, opened 2026-04-26
screenShield: Redesign the UI widgets

JosephMcc:
Use a transparent and semi-transparent style instead of the using the default style of other cinnamon dialogs. No real functionality changes here.

### DIFF
--- a/data/theme/cinnamon-sass/widgets/_screensaver.scss
+++ b/data/theme/cinnamon-sass/widgets/_screensaver.scss
@@ -7,128 +7,114 @@
     background-color: black;
   }
 
+  .icon-button {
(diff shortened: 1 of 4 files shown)


anaximeno:
Could there be a confirm button as well? (could be similar to the old screensaver or whatever button format/position is best in the new one)
Curiously, I tend to use the button in the screensaver more than Enter for confirming. Also, for example, in the password dialog, while you can confirm with the enter key, you still have the confirm button there.

JosephMcc:
> Could there be a confirm button as well?

I don't plan on adding one at the moment. Don't know if this will be merged or what changes will be made afterward if it is.

PR #13742, opened 2026-04-30
Fix battery status logic for CHARGING and PENDING_CHARGE states

Konstantin-The-Goose:
Fixed issue #13686

Previously, PENDING_CHARGE was unreachable due to duplicated condition checks. This separates the logic so both states are handled correctly (i hope).

### DIFF
--- a/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
@@ -448,7 +448,7 @@ class CinnamonPowerApplet extends Applet.TextIconApplet {
         else if (state == UPDeviceState.FULLY_CHARGED) {
             status = _("Fully charged");
         }
-        else if (state == UPDeviceState.FULLY_CHARGED) {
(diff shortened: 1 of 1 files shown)


fredcw:
Is this a typo, you've got two else if conditions the same.

Eliezermga (approved):

Konstantin-The-Goose:
> Is this a typo, you've got two else if conditions the same.

OH MY GOD yes this is a typo im so sorry

mtwebster:
So, this has been tested?  I'm suspicious now - the original basically did nothing, but it was submitted... now it's fixed.

Sorry for having to ask these types of questions, but I didn't pick this timeline we're in.

Konstantin-The-Goose:
> So, this has been tested? I'm suspicious now - the original basically did nothing, but it was submitted... now it's fixed.
> 
> Sorry for having to ask these types of questions, but I didn't pick this timeline we're in.

no sorry i havent tested it. i think it fixes the issue i mentioned in the pr, but i may be wrong

Issue #12994, opened 2025-07-22
Firefox 141 does not honor Cinnamon's dark mode on Mint 22.1


NeatNit:
# Workaround

## For Mint 22.1:

```
dconf write /org/gnome/desktop/interface/color-scheme "'prefer-dark'"
```

source: https://bugzilla.mozilla.org/show_bug.cgi?id=1978639#c24

If you need to automatically switch between light and dark mode, see this more advanced workaround: https://github.com/linuxmint/cinnamon/issues/12994#issuecomment-3190398183

## For Mint 22.2:
This issue should be fixed. If you still have this problem, please see this comment: https://github.com/linuxmint/cinnamon/issues/12994#issuecomment-3249853445 and the comments following it untill this one: https://github.com/linuxmint/cinnamon/issues/12994#issuecomment-3252938783

# Update: fix is coming (now released for Mint 22.2)
A fix is in the pipeline for Linux Mint.
See: https://github.com/linuxmint/cinnamon/issues/12994#issuecomment-3137278378

# Original bug report

### Distribution

Mint 22.1

### Package version

6.4.8

### Graphics hardware in use

AMD Radeon RX 9070 XT

### Frequency

Always

### Bug description

Full details: https://bugzilla.mozilla.org/show_bug.cgi?id=1978639

Last few comments are most relevant, specifically [comment #28](https://bugzilla.mozilla.org/show_bug.cgi?id=1978639#c28):

> Fairly sure now the issue is your distro is missing https://github.com/flatpak/xdg-desktop-portal/pull/1538 and related changes (not sure what needs to happen where to get all those fixes into linux mint).
> 
> So basically before that change (https://github.com/flatpak/xdg-desktop-portal/commit/49ee1c0d535608f6a5d0a47a1cd510958dc5e34f in particular) you could get different (conflicting) settings from two different portal implementations.
> 
> It might be that cinnamon wants to set the /org/gnome/desktop/interface/color-scheme regardless tho, since that'd trivially "fix" this issue?

### Steps to reproduce

1. Upgrade Firefox to version 141.0+linuxmint1+xia using Update Manager
2. Set Firefox's color theme to "System theme — auto"
3. Set Cinnamon to a dark theme
4. Close all windows of Firefox
5. Launch Firefox

### Expected behavior

Firefox's UI should be in dark mode

### Additional information

[...]

PR #13751, opened 2026-05-05
cs_themes: Set value of xsettings on gtk theme selected

ArqamQazi:
# Problem
Related to #12994, see my [comment](https://github.com/linuxmint/cinnamon/issues/12994#issuecomment-4321136962).

When the GTK theme is changed from advanced option it is not reflected on apps like firefox/libadwaita/flatpack apps.
I have to manually set prefer dark or light in the setting.
 
[theme-not-working-properly.webm](https://github.com/user-attachments/assets/6e2c2fb6-28a1-42ae-b7e2-b7c97fe9e15d)

I modified `_on_gtk_theme_selected` to set the xsettings value, the value of `org.x.apps.portal color-scheme` is decided and set based on theme's name.
After applying changes:

[cinnamon-2026-05-03T151933+0530.webm](https://github.com/user-attachments/assets/8e525b74-6549-42c2-a008-1ebd013017c8)

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/modules/cs_themes.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_themes.py
@@ -985,6 +985,17 @@ class Module:
         try:
             self.settings.set_string("gtk-theme", theme)
             self.set_button_chooser_text(self.theme_chooser, theme)
+            theme_lower = theme.lower()
+
+            if "darker" in theme_lower:
+                new_pref = 0
+            elif "dark" in theme_lower:
+                new_pref = 1
+            else:
+                new_pref = 2
+
+            self.xsettings.set_enum("color-scheme", new_pref)
+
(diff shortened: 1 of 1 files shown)


anaximeno:
I'm not sure if this is the best way to implement this. Because won't work well if a theme is a dark theme but just happens to not have "Dark" in the name (on the theme download section you can find some). Or as I see it goes to `2` in case neither darker or dark are found in the name, which I assume is prefer-light? In the case of a dark theme which just happens to not have "Dark" in the name it would switch to the prefer-light instead if that's the case.

mtwebster:
We set this in the 'simplified' view for the user already.

In 'advanced' mode where you can set individual toolkit themes, there's already a way to set this in the 'Settings' tab ("Dark mode"). There doesn't need to be any additional magic here, just use the controls already available, or use the 'Simplified' view for themes that support that.

PR #13758, opened 2026-05-10
menu: restore places order after search

### LINKED ISSUE #13757: Menu "Places" items order becomes reversed after using search
### Distribution

Fedora 43

### Package version

6.4.12

### Graphics hardware in use

NVIDIA NVS 5200M

### Frequency

Always

### Bug description

After using the search field in the Cinnamon menu, the order of items in the "Places" section becomes reversed.
The issue persists until Cinnamon is restarted.

### Steps to reproduce

1. Open Cinnamon menu
2. Go to the "Places" section and note the normal order
3. Type anything into the menu search field
4. Clear the search
5. Open "Places" again

### Expected behavior

The original order should remain unchanged after searching.

### Additional information

$ rpm -q cinnamon
cinnamon-6.4.12-3.fc43.x86_64

The same issue is also present in Fedora 41 and 42.

Workaround.
Restarting Cinnamon restores the correct order.
cinnamon --replace
or restarting the session.

Valley-15:
Fixes #13757

The menu already restores the original order for application, favorite, and recent items via `_resetSortOrder()`, but `placesBox` items were not restored.

This patch restores the original order of `_placesButtons` after search state changes by resetting their child indices in `placesBox`.

I was not able to test this directly in a Cinnamon desktop environment, but the issue appears to be caused by places ordering not being reset when leaving search mode.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -2866,6 +2866,10 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
         for (let i = 0; i < this._recentButtons.length; i++) {
             this.applicationsBox.set_child_at_index(this._recentButtons[i].actor, pos++);
         }
+
+        for (let i = 0; i < this._placesButtons.length; i++) {
+            this.placesBox.set_child_at_index(this._placesButtons[i].actor, i);
+        }
     }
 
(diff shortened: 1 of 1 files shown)


mtwebster:
Hi, the issue doesn't reproduce in current git master, it was probably fixed incidentally by the work done in 0cf07bf9a - Places are no longer just another category, they live in the sidebar now as of Cinnamon 6.6.

PR #13762, opened 2026-05-14
fix scrolling in notifications applet by setting a maximum height

### LINKED ISSUE #13761: Cannot scroll throught the notifications list as the length of the list grows indefinitely (gets cut off) and the scrollbar is never shown
### Distribution

Mint 22.3

### Package version

6.6.7

### Graphics hardware in use

_No response_

### Frequency

Always

### Bug description

When the number of notifications is large, I cannot scroll through the list of notifications. As a result, notifications are not accessible.

See the screenshot below:

Image

(there are a lot more notifications, but as the list length grows indefinitely, the scrollbar is never shown)

### Steps to reproduce

Create a lot of notifications. This can be done with a command like the following (creates a dummy hello notifications 20 times):
```
for i in {1..20}; do notify-send "Hello" "Notification $i"; done
```
Do not close the notification when it appears. Instead, let them close on their own (after a few seconds), and let them accumulate in the notification tray.

### Expected behavior

The scrollbar in the notifications list should be activated.

### Additional information

I initially reported a similar error in #13760, and then was able to pinpoint the specific error by discussing with developers.

This is fixed by setting a maximum height to the notifications list. I will send a PR soon.

archisman-panigrahi:
This will Fix #13761 by setting a maximum height for the notification applet, beyond which, the scrollbar will appear. Previously, the height was growing beyond the screen size, and the scrollbar never appeared.

### DIFF
--- a/files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/notifications@cinnamon.org/applet.js
@@ -109,6 +109,7 @@ class CinnamonNotificationsApplet extends Applet.TextIconApplet {
         this.scrollview.add_actor(this._notificationbin);
         this.scrollview.set_policy(St.PolicyType.NEVER, St.PolicyType.AUTOMATIC);
         this.scrollview.set_clip_to_allocation(true);
+        this.scrollview.set_style('max-height: 600px;');
 
         let vscroll = this.scrollview.get_vscroll_bar();
         vscroll.connect('scroll-start', Lang.bind(this, function() {


archisman-panigrahi:
Now the scrollbar is shown **after** applying the fix:

Untitled

Compare to what happened **before** applying the fix:

image

claudiux:
I'm afraid this solution won't work with the next version of Cinnamon.
The CSS property `max-height` is deprecated.

archisman-panigrahi:
> The CSS property `max-height` is deprecated.

What is the alternative?

claudiux:
Using `height`, after calculating the required height. I'm running into this problem with one of my applets.

anaximeno:
> Using `height`, after calculating the required height. I'm running into this problem with one of my applets.

> I'm afraid this solution won't work with the next version of Cinnamon. The CSS property `max-height` is deprecated.

Are you sure it has been deprecated? I can't find any references to that into the cinnamon code, also it seems to still be used in the codebase. I tested this with Cinnamon 6.7 on the LM 23 alfa and seems to be working as well.

I ask this because I'm planning to use `max-height` as well in this pr https://github.com/linuxmint/cinnamon/pull/13640 (I initially thought this PR was similar to mine, but this is directed towards the notification applet menu, while I'm handling the notification pop-up).

JosephMcc:
> I'm afraid this solution won't work with the next version of Cinnamon.
The CSS property max-height is deprecated.

Where?

claudiux:
Sorry, I confused `max-height` with `height: auto`.

archisman-panigrahi:
The build fails due to some sort of cjs dependency lookup, which is not related to this PR

PR #13767, opened 2026-05-18
network applet: optional VPN list cap with overflow submenu (#13562)

### LINKED ISSUE #13562: network@cinnamon.org applet missing "More" on VPN Connections in LM 22.3 & LMDE 7
### Distribution

Linux Mint 22.3 (and LMDE 7 on another computer)

### Package version

Cinnamon 6.6.4 - 6.6.7

### Graphics hardware in use

Intel Corporation UHD Graphics 630 (530 on LMDE 7 computer)

### Frequency

Always

### Bug description

The Cinnamon network@cinnamon.org applet does not have "More" list expansion when there are more than 5 stored VPN Connections. So the full list of stored VPN Connections is shown. 

There is also no scrollbar on the shown full list. And if there are numerous stored VPN Connections (I usually have from 18-36 stored VPN connections) the long list pushes the applet's top above the screen. So Wired (possibly some Wireless) is pushed out-of-view above the screen viewing area, rather than stopping at top and bringing up a scrollbar.

### Steps to reproduce

Need more than 5 listed VPN Connections to trigger "More", and I saw the same issue on each of several computers when I upgraded from LM 22.2 to 22.3 and when LMDE went to Cinnamon 6.6.4. 

I rolled back all except one each with LM 22.3 and LMDE 7 (with Cinnamon 6.6.7 now) to monitor the issue.

### Expected behavior

Return of "More" VPN Connection list expansion when there are more than 5 saved VPN Connections, with scrollbar when list is long enough to push top of applet to top off viewable screen. This has always worked for me over several Linux Mint versions on the same computers, until now.

The "More" function seems to work fine on Wireless Connections when more than 5 are seen, as it had in the past.

### Additional information

To get around this, I'm using the Cinnamon 6.4.13 network@cinnamon.org applet under ~/.local/share/cinnamon/applets, but the "More" triangles (>, v) can be a little flaky in their orientation. I'd rather use the current package version with XSI icons.

araujorm:
Fixes #13562.

NMDeviceVPN._createSection lists every configured VPN inline regardless of count, unlike the wired, wireless, mobile broadband and WireGuard sections, which cap at NUM_VISIBLE_NETWORKS (5). With many VPN connections the menu exceeds the monitor height and the items below (Network Settings, Network Connections) become unreachable.

The patch adds a per-instance cap to NMDeviceVPN, exposed through two new settings:

- "Limit VPN list" (switch, default off): gates the behaviour.
- "Visible VPN connections before \"More\"" (spinbutton, default 5, range 1-50, shown via dependency only when the switch is on): cap value.

When the switch is off the cap is Infinity, so the popup looks exactly like an unpatched Cinnamon 6.6.x. When on, items beyond the cap move into a "More" submenu, matching the existing behaviour of the other sections.

Tested on Fedora 44 (cinnamon-6.6.7-5.fc44) with 27 configured VPN connections.

### DIFF
--- a/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
@@ -978,6 +978,7 @@ NMDeviceVPN.prototype = {
         this.category = NMConnectionCategory.VPN;
 
         this._activeConnections = [];
+        this._maxVisible = Infinity;
 
         NMDevice.prototype._init.call(this, client, null, []);
     },
@@ -999,6 +1000,15 @@ NMDeviceVPN.prototype = {
 
         this._createSection();
         this.emit('active-connections-changed');
+    },
+
+    setMaxVisible: function(max) {
+        max = (max && max > 0) ? max : Infinity;
+        if (this._maxVisible === max)
+            return;
+        this._maxVisible = max;
+        this._clearSection();
+        this._createSection();
     },
 
     deactivate: function() {
@@ -1038,10 +1048,12 @@ NMDeviceVPN.prototype = {
     },
 
     _createSection: function() {
+        let pos = 0;
         for (let obj of this._connections) {
+            let isActive = this._activeConnections.some(ac => ac.connection === obj.connection);
             if (!obj.item) {
                 obj.item = new PopupMenu.PopupMenuItem(obj.name);
-                this._updateConnectionItemView(obj.item, obj.connection);
+                this._updateConnectionItemView(obj.item, obj.connection, isActive);
                 obj.item.connect('activate', Lang.bind(this, function() {
                     let activeConnection = this._activeConnections.find(ac => ac.connection === obj.connection);
                     if (activeConnection) {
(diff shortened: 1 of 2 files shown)


araujorm:
> ## Best-practices scanner
> 
> This is a regex-based check for API usage that can pose security, performance or maintainability issues, or that may already be provided by Cinnamon. Most findings are advisory and do not automatically disqualify a pull request.
> ### This check is not perfect and will not replace a normal review.
> 
> Found 2 potential issue(s):
> ## ⚠️ WARNING
> ### ⚠️ lang_bind
> 
> **[files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js:1839](https://github.com/linuxmint/cinnamon/pull/13767/files#diff-0e0c1aecc4f7537ff81c1eaab783adb635c264d1a590e043da8e94326bbfaed7R1839)**
> 
> ```
> this.settings.bind("cap-vpn-list", "_capVpnList", Lang.bind(this, this._onVpnCapChanged));
> ```
> 
> `Lang.bind()` is deprecated. Use arrow functions (`() => {}`) or `Function.prototype.bind()` instead.
> 
> **[files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js:1840](https://github.com/linuxmint/cinnamon/pull/13767/files#diff-0e0c1aecc4f7537ff81c1eaab783adb635c264d1a590e043da8e94326bbfaed7R1840)**
> 
> ```
> this.settings.bind("num-visible-vpns", "_numVisibleVpns", Lang.bind(this, this._onVpnCapChanged));
> ```
> 
> `Lang.bind()` is deprecated. Use arrow functions (`() => {}`) or `Function.prototype.bind()` instead.
> 
> _Automated pattern check._

Fixed in [latest push.](https://github.com/linuxmint/cinnamon/commit/cf5536050d2966df375bf1f713680c5a16daadde)

PR #13782, opened 2026-05-27
Add some options to cinnamon-screenshot

Cobinja:
This adds two option to cinnamon-screenshot:

- autosave-to-file
- autosave-to-clipboard

These options do what they're called, without opening the interactive window. Both options can be used at the same time.


Cobinja:
force push for fixed commit message

PR #13789, opened 2026-05-29
Fix three keyboard-related issues (#13787)

### LINKED ISSUE #13787: Some hotkeys work incorrectly: ctrl+shift, super+space, ctrl+a.
### Distribution

Linux Mint 22.3 Cinnamon x64

### Package version

Cinnamon 6.6.7

### Graphics hardware in use

_No response_

### Frequency

Always

### Bug description

I'm considering different linux distors to choose to attempt to switch from windows. So I've tried to setup hotkeys ctrl+shift for switching languages, since I've got used to those the most. And what I've noticed is that whenever I would press them at the same time, they don't work properly. Like they might sometimes, but many times they don't. So while testing it I've noticed that if you press ctrl first and then shift - it always works. If you press shift first, then while holding it ctrl - then it never works. And, of course, if you try to press them at the same time, seems like it depends on which system detects pressed first.
This is a small detail, but very inconvenient. I tried how does super+space work and I've noticed that it's better: when pressed at the same time - works perfectly. But with a noticable delay if you press space first, hold it and then press super - it never works. Even if you do it very fast but in that order - never works.
So I suggest to fix those hotkeys in a way that it doesn't matter which key of the combination is pressed first, because current behavior is very misleading.
I've specified the version, but I think that I've been trying ctrl+shift before and the same behavior was already there like few years ago. Maybe 6-8 years ago.

Also switching languaguages closes start menu.
And Ctrl+A doesn't work in the start menu input field. Only shift+home works to select all the text.

On one hand, it sounds like 3 different bugs, but on the other hand, it all has to do with hotkeys. So I'm not sure if I should report them separately.

### Steps to reproduce

1. Setup 2 languages in your system for keyboard layout.
*Note: I have flags displayed for languages in the tray, but I don't think it matters. Just saying just in case. It seems to be more distinguishable for me that way.
2. Setup hotkeys `ctrl+left shift` to go to next language.
3. Setup hotkeys `ctrl+right shift` to go to previous language.
4. Open start menu by pressing a super key on your physical keyboard, or click the start menu button in the bottom left corner.
5. Start typing some text so that you have some input in the start menu search bar.
6. Change your mind on the language you're typing the text.
7. ❗ **Press ctrl+a to select all the wrong text you've just typed. Notice that ctrl+a doesn't work here.**
8. Delete the text in the start menu search input.
9. Press super+space to change the language.
10. ❗ **Notice that the language have changed, but at the same time, because you've pressed the super button, the start menu has closed, even though it was not a single button pressed, but a part of hotkeys super+space.** Seems like it's because start menu opens on super button release, but start menu closes on super button key down event.
11. Try switching languages with hotkeys ctrl+shift, but press ctrl first, hold it, press shift, release shift, then release ctrl. Repeat few times. Notice that it works every time.
12. ❗ **Repeat the same, but this time press shift first, then ctrl. Repeat few times. Notice that it never works.**
13. Try pressing ctrl+shift at the same time multiple times. Notice that sometimes it works, sometimes it doesn't. Most likely because of the same reason: depending on which button is processed by OS first.

### Expected behavior

1. For hotkeys ctrl+shift order of pressing shouldn't matter. Like in math: the sum does not change when the variables order changes.
2. When typing in the start menu search input, ctrl+a hotkeys should selec all the input text in the search field.
3. When pressing super+space with open start menu, it shouldn't close. Instead it should close on super button release IF no other buttons are pressed.

### Additional information

I'm pretty sure that this is an old bug. I'm sure that I've noticed ctrl+shift years ago. No, reboot doesn't help. I've been noticing it on a fresh install of a system. Including live usb usage.

H09-0:
## Summary

Fix #13787 - three keyboard and shortcut issues in the modal event handler
and menu search field.

---

### Issue 1: Ctrl+Shift order-dependent for layout switching

**Root cause:** In `_stageEventHandler` (`js/ui/main.js`), when the user
presses two modifier keys (e.g., Ctrl+Shift) and the keybinding is stored
as `<Control>Shift`, only "Ctrl first, then Shift" triggers the binding.
"Shift first, then Ctrl" fails because `get_keybinding_action(ctrl_keycode,
Shift_mask)` checks for a binding where Ctrl is the trigger key, but the
binding has Shift as the trigger key.

**Fix:** Track the previous modifier keycode (`_lastModifierKeyCode`) when a
modifier press doesn't match a binding. When the next modifier is pressed,
try the reverse combination: `get_keybinding_action(prev_keycode,
modifierState | current_mask)`. This correctly matches `<Control>Shift`
when Shift is pressed first.

### Issue 2: Super+Space closes the Start Menu

**Root cause:** When Super (the overlay key) is pressed, `_stageEventHandler`
stores its action in `_modifierOnlyAction`, deferred to key-release. If the
user then presses another modifier key that doesn't match a binding,
`_modifierOnlyAction` was never cleared, causing the overlay toggle to fire
on Super release - closing the menu while the user intended Super as part
of a combo (Super+Space for layout switching).

**Fix:** Clear `_modifierOnlyAction` in the modifier key press handler when
the press doesn't match any binding. Also clear it when a non-modifier key
press arrives. This ensures a subsequent modifier-only action only fires if
no other key was pressed between press and release.

### Issue 3: Ctrl+A does not select all text in search field

**Root cause:** The menu applet's `_onMenuKeyPress` handler connects to the
search entry `clutter_text`'s `key-press-event`. ClutterText does not
implement Ctrl+A (select all) by default in the version used by Cinnamon.
The event propagated but no built-in handler acted on it.

**Fix:** Add `Clutter.KEY_A` / `Clutter.KEY_a` cases in the `_onMenuKeyPress`
switch statement. When ctrlKey is held, call
`this.searchEntryText.set_selection(0, textLen)` to select all text.

---

## Files changed

- `js/ui/main.js` (+51, -5): Fix Issues 1 and 2 in `_stageEventHandler`
- `files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js` (+8, -0): Fix Issue 3 in `_onMenuKeyPress`

## Testing

1. **Issue 1:** Set Ctrl+Shift as layout switching shortcut. With the menu
   open, press Shift first then Ctrl (both orders). Layout should switch
   regardless of order.
2. **Issue 2:** Open the Start Menu (by mouse click). Press Super+Space
   (or whatever the layout switch combo is). The menu should stay open
   and the layout should switch. Solo Super press/release should still
   toggle the menu.
3. **Issue 3:** Open the Start Menu, type text in the search field, press
   Ctrl+A. All text should be selected.

### DIFF
--- a/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js
@@ -1937,6 +1937,14 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
                     return Clutter.EVENT_STOP;
                 }
                 break;
+            case Clutter.KEY_A:
+            case Clutter.KEY_a:
+                if (ctrlKey) {
+                    let textLen = this.searchEntryText.get_text().length;
+                    this.searchEntryText.set_selection(0, textLen);
+                    return Clutter.EVENT_STOP;
+                }
+                break;
             case Clutter.KEY_Menu:
                 if (this._activeContainer === this.applicationsBox) {
                     this.toggleContextMenu(active._delegate);
--- a/js/ui/main.js
+++ b/js/ui/main.js
@@ -1372,7 +1372,7 @@ function _stageEventHandler(actor, event) {
             // modifier mask.
             if (action <= 0 && _lastModifierKeyCode > 0) {
                 let currentMask = _modifierMaskForKeySymbol(event.get_key_symbol());
-                action = global.display.get_keybinding_action(_lastModifierKeyCode, modifierState | currentMask);
+                action = global.display.get_keybinding_action(_lastModifierKeyCode, currentMask);
                 if (action > 0) {
                     let entry = keybindingManager.getBindingById(action);
                     if (!_shouldFilterKeybinding(entry)) {
@@ -1437,6 +1437,7 @@ function _findModal(actor) {
 
 function _completeModalSetup(actor, mode, onDismiss) {
     _modifierOnlyAction = 0;
+    _lastModifierKeyCode = 0;
 
     if (modalCount == 0)
         Meta.disable_unredirect_for_display(global.display);
--- a/js/ui/main.js
+++ b/js/ui/main.js
@@ -1306,12 +1306,32 @@ function _shouldFilterKeybinding(entry) {
  * to give the multi-key binding a chance to activate on key-press.
  */
 let _modifierOnlyAction = 0;
+let _lastModifierKeyCode = 0;
 
 function _isModifierKeyval(symbol) {
     return symbol === Clutter.KEY_Super_L   || symbol === Clutter.KEY_Super_R   ||
            symbol === Clutter.KEY_Control_L || symbol === Clutter.KEY_Control_R ||
            symbol === Clutter.KEY_Alt_L     || symbol === Clutter.KEY_Alt_R     ||
            symbol === Clutter.KEY_Shift_L   || symbol === Clutter.KEY_Shift_R;
+}
+
+function _modifierMaskForKeySymbol(symbol) {
+    switch (symbol) {
+        case Clutter.KEY_Control_L:
+        case Clutter.KEY_Control_R:
+            return Clutter.ModifierType.CONTROL_MASK;
+        case Clutter.KEY_Shift_L:
+        case Clutter.KEY_Shift_R:
+            return Clutter.ModifierType.SHIFT_MASK;
+        case Clutter.KEY_Alt_L:
+        case Clutter.KEY_Alt_R:
+            return Clutter.ModifierType.MOD1_MASK;
+        case Clutter.KEY_Super_L:
+        case Clutter.KEY_Super_R:
+            return Clutter.ModifierType.MOD4_MASK;
+        default:
+            return 0;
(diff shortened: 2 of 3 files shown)


H09-0:
## Edge-case review completed

I reviewed the 5 concerns raised during review. Two real bugs were found and fixed in commit 12717ad63:

1. **Reverse-match modifier state was wrong**: The old code passed `modifierState | currentMask` as the modifier state for the reverse lookup. Since `modifierState` already contains the held modifier's mask (whose keycode is `_lastModifierKeyCode`), this created a self-referential mismatch for bindings like `<Control>Shift`. Fixed to pass only `currentMask`.

2. **Stale `_lastModifierKeyCode` across modal sessions**: Added `_lastModifierKeyCode = 0` to `_completeModalSetup()` to prevent stale keycodes from causing false reverse matches when a new modal session starts.

All 5 originally flagged concerns were confirmed as already correct: cleanup on release/non-modifier keys, `_modifierOnlyAction` clearing on successful combos, three-modifier graceful degradation, left/right modifier distinction, and Ctrl+A ctrlKey-only strictness.

mtwebster:
Pretty sure all of this is fixed in git master already.  Ordered-keybinding behavior is deliberate. If order independent behavior is required, XKB Options for switching layouts now works in both x11 and Wayland sessions.

PR #13797, opened 2026-05-31
Fix issue #12641- invisible panel on second display

### LINKED ISSUE #12467: Monitor panels forgotten between connections
### Distribution

22

### Package version

6.2.9

### Graphics hardware in use

_No response_

### Frequency

Quite often

### Bug description

When I connect to a monitor (I have one at work and one at home) I like to have a panel on the monitor as well as on my laptop screen to keep the window list on the same monitor as the windows.  I add a panel and it works fine.  But then, much of time time, when I come back another day and plug the monitor in again, the panel I added is gone.  Previously, this hasn't been the case, and each monitor's configuration has been remembered.

Also, when I add the new panel, the icons of the windows are huge and are consequently cropped in the panel.  But if I edit the panel settings and change the panel height, it fixes it.  The panel height starts at 25; I move the slider to 26 and then back to 25 and the icons' sizes are correct.

### Steps to reproduce

As far as I can tell, this happens between monitors, so if I'm using the laptop in the office at work, and then come home and plug it in, that's when it tends to happen -- when I've been using a different monitor.  Or vice versa: home to work.  I use USB-C at home and either USB-C or HDMI at work (to the same monitor) if that makes a difference.

### Expected behavior

Cinnamon remembers the monitor and displays the panel as previously configured.

### Additional information

_No response_

### LINKED ISSUE #12641: Panel on the secondary screen becomes invisible
### Distribution

Fedora release 41 (Forty One)

### Package version

6.4.4

### Graphics hardware in use

Intel Corporation Alder Lake-P GT2, NVIDIA Corporation GA104M 

### Frequency

Always

### Bug description

Panel on the secondary screen does not come back when the screen is connected.

### Steps to reproduce

- Connect a laptop to a an external screen configure for "join displays"
- Create a panel on the bottom of the external screen
- put system in hibernate
- wake from hibernate without the secondary screen
- hibernate again
- wake with second screen attached - the panel will not be visible.
- In panel settings, the "next panel" button shows the props of the missing panel

### Expected behavior

Panel should appear when the screen is connected

### Additional information

_No response_

canklot:
If a secondary monitor is connected after cinnamon starts, the panels on the secondary monitor is not displayed. Calling the fullPanelLoad inside onMonitorsChanged fixes the issue.

fixes #12641 and fixes #12467

### DIFF
--- a/js/ui/panel.js
+++ b/js/ui/panel.js
@@ -942,6 +942,7 @@ var PanelManager = GObject.registerClass({
     }
 
     _onMonitorsChanged() {
+        this._fullPanelLoad();
         const oldCount = this.monitorCount;
         this.monitorCount = global.display.get_n_monitors();
 


canklot:
Here is the screen recording of the problem: https://youtu.be/TmS35RU9rtM
Here is the screen recording after applying the fix: https://youtu.be/-XPwOQYgMd4

canklot:
Hey guys, can you check the PR. Its just one line. I have been using the modified version for a month without any issues. @mtwebster @clefebvre

mtwebster:
Pretty sure this is no longer necessary as a result of:
https://github.com/linuxmint/cinnamon/commit/e3cbc04410b9f4964fec9aed421a337bd443ffd1

PR #13873, opened 2026-07-15
workspace-switcher: Add preview size setting

Shirinveg:
## What changed

Adds a size slider to the built-in workspace switcher applet.

The slider allows workspace previews to be scaled from 25% to 100% of the panel size.

## Why

Workspace previews can occupy excessive panel space, especially when several workspaces are enabled.

Previously, users had no applet setting to reduce their size.

## Implementation

The preview container, drawing area, window representations and application icons are scaled together. This avoids clipping and alignment issues.

The default value is 100%, so existing users keep the current appearance.

## Tested

* Cinnamon 6.6.7
* Horizontal panel
* Visual workspace representation
* Preview sizes from 25% to 100%


anaximeno:
Changing the theme would maybe be a way to change this. Also not all styles of workspace representation would make sense to have that setting.

PR #13811, opened 2026-06-12
Improve debian packaging

Fantu:
merge some improments from Debian packaging

PR #13878, opened 2026-07-17
keyboard: fall back to tecla for layout preview, hide menu actions for missing tools

Fantu:
### Background

`gkbd-keyboard-display` comes from **libgnomekbd**, which is unmaintained and being removed from distributions: GNOME replaced it with [tecla](https://gitlab.gnome.org/GNOME/tecla) in GNOME 45, and Debian is in the process of removing libgnomekbd entirely. It is also an X11-only GTK3 program, so the keyboard layout preview was already unavailable under Wayland.

### 1. Layout preview: fall back to tecla

The layout preview button (in the "Add layout" dialog) and the "Test layout" button were already hidden when `gkbd-keyboard-display` was missing — the feature was simply lost. This PR makes them fall back to `tecla`, hiding the buttons only when *neither* program is available. `gkbd-keyboard-display` stays preferred when installed, since tecla only renders a standard pc105 geometry (this addresses the `TODO` comment removed from `InputSources.py`). Because tecla is a GTK4 program it also works under Wayland, restoring the preview there.

tecla is invoked with a single `layout+variant` argument. `+` is accepted as the layout/variant separator by **every** tecla version (verified against the tecla 46 and 50 sources, `tecla_model_new_from_layout_name`); space and tab were only added in tecla 48, so `+` is used unconditionally for backward compatibility (Ubuntu 24.04 / Mint 22 ship tecla 46).

### 2. Applet: hide actions when the target program is missing

The keyboard applet added its **"Show Keyboard Layout"** menu entry unconditionally: with no preview program installed, clicking it just popped an *"Execution of 'gkbd-keyboard-display' failed"* error notification. It now uses the same lookup/fallback as cinnamon-settings and the entry is hidden when no preview program is available.

A second commit applies the same guard to **"Show Character Table"** (`gucharmap`), which has the identical failure mode when gucharmap is not installed.

### Possible future improvement

tecla ≥ 47 accepts a `--parent-handle` option to attach the preview window to a parent as a modal dialog. This could be used from the two cinnamon-settings dialogs (it doesn't apply to the applet, which has no parent GTK window), but it requires exporting the GTK window handle on Wayland/X11 and has no `gkbd-keyboard-display` equivalent, so it's left out of this PR to keep the fallback simple.

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/AddKeyboardLayout.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/AddKeyboardLayout.py
@@ -1,7 +1,6 @@
 #!/usr/bin/python3
 
 import gettext
-import os
 import subprocess
 
 import cairo
@@ -12,6 +11,7 @@ gi.require_version('IBus', '1.0')
 gi.require_version('Pango', '1.0')
 from gi.repository import GLib, Gio, Gtk, GObject, CinnamonDesktop, IBus, Pango
 
+from bin import util
 from bin.SettingsWidgets import Keybinding
 from xapp.SettingsWidgets import SettingsPage
 from xapp.GSettingsWidgets import PXGSettingsBackend, GSettingsSwitch
@@ -21,11 +21,28 @@ MAX_LAYOUTS_PER_GROUP = 4
 INPUT_SOURCE_SETTINGS = "org.cinnamon.desktop.input-sources"
 SHOW_ALL_SOURCES_KEY="show-all-sources"
 
-def make_gkbd_keyboard_args(layout, variant):
-    if variant:
-        return ["gkbd-keyboard-display", "-l", f"{layout}\t{variant}"]
+def get_layout_preview_program():
+    # gkbd-keyboard-display (libgnomekbd) is X11-only, so on Wayland only tecla
+    # is usable. On X11 keep gkbd-keyboard-display preferred where still
+    # available (tecla only shows a standard pc105 geometry).
+    if util.get_session_type() == "wayland":
+        candidates = ("tecla",)
     else:
-        return ["gkbd-keyboard-display", "-l", layout]
+        candidates = ("gkbd-keyboard-display", "tecla")
+    for program in candidates:
+        if GLib.find_program_in_path(program):
+            return program
+    return None
+
+def make_layout_preview_args(layout, variant):
+    program = get_layout_preview_program()
+    if program == "tecla":
+        # '+' is the layout/variant separator accepted by every tecla version
+        # (space and tab only since tecla 48)
+        return [program, f"{layout}+{variant}" if variant else layout]
+    if variant:
+        return [program, "-l", f"{layout}\t{variant}"]
+    return [program, "-l", layout]
 
 def make_ibus_display_name(engine):
     name = engine.get_longname()
@@ -99,7 +116,7 @@ class AddKeyboardLayoutDialog():
 
         self.response_id = None
 
-        if not GLib.find_program_in_path("gkbd-keyboard-display"):
+        if get_layout_preview_program() is None:
             self.preview_button.set_visible(False)
 
         self._ibus = IBus.Bus.new_async()
@@ -167,7 +184,7 @@ class AddKeyboardLayoutDialog():
         display_name = self.layouts_sort_store.get_value(iter, LAYOUT_DISPLAY_NAME_COLUMN)
         layout_layout = self.layouts_sort_store.get_value(iter, LAYOUT_LAYOUT_COLUMN)
         layout_variant = self.layouts_sort_store.get_value(iter, LAYOUT_VARIANT_COLUMN)
-        args = make_gkbd_keyboard_args(layout_layout, layout_variant)
+        args = make_layout_preview_args(layout_layout, layout_variant)
         subprocess.Popen(args)
 
     def _on_all_sources_changed(self, button, data=None):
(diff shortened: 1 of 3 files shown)


Fantu:
lmde7 test seems fails for unrelated issue

Fantu:
Pushed an additional commit that improves Wayland support: the layout preview now only considers `tecla` there.

`gkbd-keyboard-display` is X11-only — in a Wayland session it just aborts with:

```
** (gkbd-keyboard-display:12569): CRITICAL **: Error initializing GTK: Unable to open display:
```

so even where libgnomekbd is still installed, the preview was broken on Wayland.

With this change:

- **Wayland**: `tecla` is used when installed; the preview/test buttons and the applet menu entry are hidden otherwise. `gkbd-keyboard-display` is never spawned.
- **X11**: behaviour unchanged from the previous commits — `gkbd-keyboard-display` is still preferred (accurate keyboard geometries), with `tecla` as fallback.

Session detection uses `Meta.is_wayland_compositor()` in the applet and `XDG_SESSION_TYPE` in cinnamon-settings.

Fantu:
note: about d/control changes need to choose if only add tecla to deps (will be used only in wayland session for now) and keep gkbd-capplet as libgnomekbd still present in actual stable version of ubuntu and remove it only for debian 14; or if remove gkbd-capplet now and use already tecla also on xorg session

files/usr/share/cinnamon/cinnamon-settings/bin/AddKeyboardLayout.py
```diff
@@ -21,11 +21,28 @@
...
 SHOW_ALL_SOURCES_KEY="show-all-sources"
 
-def make_gkbd_keyboard_args(layout, variant):
-    if variant:
-        return ["gkbd-keyboard-display", "-l", f"{layout}\t{variant}"]
+def get_layout_preview_program():
+    # gkbd-keyboard-display (libgnomekbd) is X11-only, so on Wayland only tecla
+    # is usable. On X11 keep gkbd-keyboard-display preferred where still
+    # available (tecla only shows a standard pc105 geometry).
+    if os.environ.get("XDG_SESSION_TYPE") == "wayland":
```
mtwebster: Maybe `util.get_session_type()` here (see `files/usr/share/cinnamon/cinnamon-settings/bin/util.py`) for consistency.

Fantu:
thanks for the review, change done
"import os" removed even though it wasn't added by the PR. It seems correct, I don't see any uses in the file.
I haven't tested it after the change. Now it's late and I'm tired, but I'll test again in the next few days if necessary.

mtwebster:
We can definitely remove gir1.2-gkbd-3.0 now.

Maybe we could do `gkbd-capplet | tecla` for depends.

Fantu:
> Maybe we could do `gkbd-capplet | tecla` for depends.

in versions with libgnomekbd still present (debian 13, ubuntu 24.04 and ubuntu 26.04) it would only install gkbd-capplet so would be not good for wayland session. (libgnomekbd will be removed from debian 14 and ubuntu 26.10 or 27.04)

Fantu:
about gir1.2-gkbd-3.0 I already removed in [Debian packages from 6.6](https://salsa.debian.org/cinnamon-team/cinnamon/-/commit/c0fb7297e6960834332ffe0b3d3b8b9e912f31c7)
from https://github.com/linuxmint/cinnamon/pull/13811 I have not applied several things that I had done in Debian, if not applicable or uncertain, gir1.2-gkbd-3.0 seems to me to be sure that it is not used, I do not remember why I excluded it, if I have time I will do a comparison again and see about making another PR with the merging of other useful parts from Debian

mtwebster:
Tecla is really underwhelming. I don't like having to change 'tabs'  to see additional key levels - the nice thing about gkbd-keyboard-display was all layers are visible at once - on some complicated layouts that was bad, but tecla is much too far in the opposite direction.

I'm going to see how much trouble it is to make our own layout viewer that will be session-type independent, so I'd rather hold off on doing anything with this yet.

Is merging this or fixing the dependencies blocking you or anything on Debian right now?  Or any other PRs I haven't gotten to? (I've was out of town for over a month).

Fantu:
I already applied this PR to debian unstable/testing because I had to remove the last libgnomekbd package.
the commit that uses tecla was made while still leaving the possibility of using gkbd-keyboard-display as primary in xorg if present, since at the moment in the debian/ubuntu versions used by mint and the next one is still there for xorg sessions it would not change but it would bring an advantage for wayland which is without that part in wayland and it seems good to use this for now and then replace it if you want to create a better alternative. What do you think?

Maybe we could also try to propose an improvement in Tecla and see if they respond or do you think it's basically not good?

about other PR if you want to take a look to https://github.com/linuxmint/cinnamon/pull/13898
also this already applied in debian unstable/testing

mtwebster:
I doubt they'll be receptive. The last few times I attempted to engage it was pointless.

After recent events I don't want a new Gnome dependency, either.

Quick effort using python-xkbcommon and Gtk3. Still needs work but I think it'll be good enough:

image

mtwebster:
Done:

https://github.com/linuxmint/cinnamon/commit/467c8d3e03405304442f2961ed1dbe2faaa29481

Fantu:
seems missed the part related to commit https://github.com/linuxmint/cinnamon/pull/13878/changes/f74370b0c6ae9c6c4e4df7673d0318b40e04a788
It's a minor thing but I think it's better to handle the case where gucharmap is not installed

PR #13880, opened 2026-07-18
issue-13868

### LINKED ISSUE #13868: Applet/Extension/Desklet/Theme manager crashes with TypeError: int() argument must be ... not 'NoneType' when server response has no content-length header
### Distribution

Mint 22.3

### Package version

6.6.7

### Graphics hardware in use

_No response_

### Frequency

Always

### Bug description

I'm very new to Linux. The following report was written by Claude. 

Started getting the following message with Applets, Desklets, Actions and Themes when trying to refresh the cache (Download). 

Description:
The Applets/Desklets/Extensions/Themes download manager intermittently fails with a generic "An error occurred while trying to access the server" dialog. Running cinnamon-settings applets from a terminal and adding a traceback.print_exc() in the relevant except block in Spices.py reveals the actual cause:

  File "/usr/share/cinnamon/cinnamon-settings/bin/Spices.py", line 395, in _download
    self._url_retrieve(url, outfd, self._update_progress, binary)
  File "/usr/share/cinnamon/cinnamon-settings/bin/Spices.py", line 435, in _url_retrieve
    raise e
  File "/usr/share/cinnamon/cinnamon-settings/bin/Spices.py", line 424, in _url_retrieve
    totalSize = int(response.headers.get('content-length'))
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'

Root cause:
_url_retrieve() in Spices.py assumes every response from cinnamon-spices.linuxmint.com includes a content-length header:
response = requests.get(url, proxies=proxy_info, stream=True, timeout=15)
assert response.ok
totalSize = int(response.headers.get('content-length'))

This isn't guaranteed by HTTP (e.g. chunked transfer-encoding responses omit it), and in my case it happens intermittently, seemingly under the parallel/multi-threaded download load the manager generates. curl -sD - against the same JSON endpoint immediately afterward shows a normal response with content-length present, and the site sits behind Sucuri/Cloudproxy, which suggests the bot-protection/WAF layer is occasionally returning a different (challenge or rate-limited) response to the burst of concurrent non-browser requests, without a content-length header — and the client crashes instead of handling it gracefully.

### Steps to reproduce

Hard to reproduce reliably on demand since it depends on server/WAF behavior under load, but it recurs regularly when opening the Applets/Desklets/Extensions/Themes download tabs.

### Expected behavior

It should refresh without crashing.

### Additional information

Suggested fix:
Guard against a missing header instead of crashing:
pythoncontent_length = response.headers.get('content-length')
totalSize = int(content_length) if content_length is not None else -1
(and have the progress-reporting callback tolerate totalSize == -1 as "unknown size" rather than trying to compute a percentage from it.)
I've applied this locally and confirmed it resolves the crash for me.

androidbod:
- content length extraction fails if for some reasone the content-length header is missing from the remote response
- fix it by means of a none type fallback to length of -1
- this is the recommended patch from the ticket comments - i just took the time to build the patch pr for it
- ai: gemini helped but mostly with github process stuff - the development i did myself
- tested locally by manually adjusting the locally installed copy of Spices.py
- fixes #13868

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
+++ b/files/usr/share/cinnamon/cinnamon-settings/bin/Spices.py
@@ -420,7 +420,8 @@ class Spice_Harvester(GObject.Object):
             response = requests.get(url, proxies=proxy_info, stream=True, timeout=15)
             assert response.ok
 
-            totalSize = int(response.headers.get('content-length'))
+            content_length = response.headers.get('content-length')
+            totalSize = int(content_length) if content_length is not None else -1
 
             for data in response.iter_content(chunk_size=blockSize):
                 count += 1


claudiux:
This only concerns Cinnamon 6.6.

claudiux:
Ping @mtwebster

blakkd:
Alternative approach I used locally: use `None` as the fallback and add a guard in `_update_progress`:

```
# Line 423
cl = response.headers.get('content-length')
totalSize = int(cl) if cl else None
```

```
# Line 331
fraction = count * blockSize / float((totalSize / blockSize + 1) * blockSize) if totalSize else 0
```

The `-1` fallback works, but `None` is more semantically correct for a missing header and makes the guard in `_update_progress` self-documenting. Sharing in case it's useful.

androidbod:
okay - multiple (conflicting) solutions - so what do we do? who can make that decision?
i just want the problem gone with a regular cinnamon release :-)

androidbod:
> Alternative approach I used locally: use `None` as the fallback and add a guard in `_update_progress`:
> 
> ```
> # Line 423
> cl = response.headers.get('content-length')
> totalSize = int(cl) if cl else None
> ```
> 
> ```
> # Line 331
> fraction = count * blockSize / float((totalSize / blockSize + 1) * blockSize) if totalSize else 0
> ```
> 
> The `-1` fallback works, but `None` is more semantically correct for a missing header and makes the guard in `_update_progress` self-documenting. Sharing in case it's useful.

Doesn't `response.headers.get('content-length')` already return a `None` object as the reported issue says it does (the error message contains something with `None`)

The guardrail in `_update_progress` alone should solve that issue then if i'm not mistaken - but i barely speak python and might be missing something.

blakkd:
Yes it now does

```
$ /usr/bin/python3 -c "
import requests
r = requests.get('https://cinnamon-spices.linuxmint.com/json/applets.json', stream=True, timeout=15)
result = r.headers.get('content-length')
print('Value:', repr(result))
"
Value: None
```

I think the commit you proposed is actually sufficient, let alone if a huge refactor has occurred since then...

androidbod:
pull request can be closed as the issue was resolved with the release of cinnamon 6.6.9

PR #13899, opened 2026-07-25
Warn when a device is unplugged while still mounted

Fantu:
Related to https://github.com/orgs/linuxmint/discussions/1565 ("Please make it more clear that properly ejecting storage mediums is very important on Linux").

When a user physically pulls out a drive without ejecting it, nothing happens: no sound, no message, no hint that anything went wrong. The removal itself can't be prevented or delayed, but it can be reported afterwards, so that the user learns what the consequence of that action is.

This adds a notification for that case only. Clean removals — eject, safe removal, or a plain unmount — stay silent.

## How the detection works

GVfs makes the two cases distinguishable through the order in which it emits its removal signals. In `update_all()` (`monitor/udisks2/gvfsudisks2volumemonitor.c`), everything that disappeared in the same update is announced as:

```
drive-disconnected  ->  volume-removed  ->  mount-removed
```

So a device that is unplugged while mounted still has its mounts recorded when `drive-disconnected` arrives, while a device that was unmounted or ejected first had its `mount-removed` delivered by an earlier update, leaving nothing behind.

The mounts are tracked in a small cache keyed on the mount point. Two details forced that choice, and both are worth knowing if this code is ever revisited:

- `GMount` has no volume left by the time `mount-removed` is emitted, so a cache keyed on the volume never gets cleared;
- `drive.get_volumes()` cannot be used instead: it is empty for a clean removal, and it can still list a volume after a clean unmount.

## Why a notification and not a dialog

`CinnamonUnmountNotifier` already reports unmount progress through the message tray, while the dialogs added in 9bdd28c5 are used where the user asked for something and has to answer (password, busy device). This warning belongs to the first group: nobody asked for it and there is nothing to answer, so a modal dialog would only steal focus from whatever the user is doing.

It is rate limited to one warning per minute per device, because a failing connector can produce repeated connect/disconnect cycles.

## Testing

Signal order and liveness were traced with a GVolumeMonitor + UDisks2 job monitor on GVfs 1.54, with a USB stick, in three scenarios:

| scenario | result |
|---|---|
| eject, and safe removal from the applet | no warning |
| unplugged while mounted | warning |
| `udisksctl unmount` then unplugged | no warning |

The last one is the interesting negative case, since it doesn't go through the GIO API at all.

The feature was then verified end to end on a Mint 23 VM (Ubuntu 26.04 + the `alfa` repo, cinnamon 6.7.4): unplugging a mounted stick shows the notification in a real session.

## Known limitations

- The rate limit window (60s) is reasoned, not measured against an actually faulty connector.
- With several volumes mounted on the same drive a single warning is shown, by construction; not tested on a real multi-partition device.
- Unmounts performed outside GVfs' knowledge can't be seen, as stated in the original discussion.

## Two unrelated fixes included

They are separate commits and can be dropped or taken independently:

- `cinnamonMountOperation.js` used `xapp-media-removable`, which doesn't exist in the icon set Cinnamon uses since icon names were switched to XSI in 890f6f7f9 — the unmount notification has been falling back to a missing icon. `xsi-media-removable` is the direct equivalent.
- The removable drives applet used `xsi-drive-removable-media`, which draws a plain rounded rectangle that reads as a floppy or an optical drive at panel size. `xsi-media-removable` draws a USB stick and is immediately recognizable.

`js/ui/automountManager.js` was also added to `POTFILES.in`, which never listed it: its strings — including the existing ones — were not being extracted for translation.

warning-device-removed-unsafely

### DIFF
--- a/files/usr/share/cinnamon/applets/removable-drives@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/removable-drives@cinnamon.org/applet.js
@@ -41,7 +41,7 @@ class CinnamonRemovableDrivesApplet extends Applet.IconApplet {
     constructor(orientation, panel_height, instance_id) {
         super(orientation, panel_height, instance_id);
 
-        this.set_applet_icon_symbolic_name("xsi-drive-removable-media");
+        this.set_applet_icon_symbolic_name("xsi-media-removable");
         this.set_applet_tooltip(_("Removable drives"));
 
         global.settings.connect('changed::' + PANEL_EDIT_MODE_KEY, Lang.bind(this, this._onPanelEditModeChanged));
--- a/js/testing/testUnsafeRemovalWarning.js
+++ b/js/testing/testUnsafeRemovalWarning.js
@@ -1,0 +1,12 @@
+// Test helper for the unsafe removal warning.
+// Call from Looking Glass:
+//
+//   imports.testing.testUnsafeRemovalWarning.show()
+//   imports.testing.testUnsafeRemovalWarning.show("My USB Stick")
+//   imports.testing.testUnsafeRemovalWarning.show(null)  // unnamed device
+
+const AutomountManager = imports.ui.automountManager;
+
+function show(driveName = "JetFlash Transcend 4GB") {
+    AutomountManager.notifyUnsafeRemoval(driveName);
+}
--- a/js/ui/automountManager.js
+++ b/js/ui/automountManager.js
@@ -1,7 +1,7 @@
 // -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
 /* exported Component */
 
-const { Gio, GLib } = imports.gi;
+const { Gio, GLib, St } = imports.gi;
 const Params = imports.misc.params;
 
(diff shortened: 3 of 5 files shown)


mtwebster:
Hi, I've got a branch already I've been working on this for Cinnamon and Nemo, I'll make sure I'm covering this stuff also

mtwebster:
I'm not sure about a dialog, though, after the fact. Honestly, if they can't be bothered to read a notification warning them not to do something, they deserve what they get.

Fantu:
Some users do it on purpose and you could be right (even if it's not a good thing to have a vindictive attitude), while other users are really inexperienced or have "too little logical skills" and I think it's useful to help them, for example there's already the applet icon which is not easily intuitive and I suppose the USB stick icon would be better, then that warning in case of removal without unmounting I think helps (at least for those who don't insist on doing it anyway)

runningvisor:
I love the notification! It explains it very well. I feel like we should also still [attempt to prevent data loss](https://github.com/orgs/linuxmint/discussions/1565#discussioncomment-17818358) though. What do you think @mtwebster?

Issue #12550, opened 2024-12-03
UI: Network Manager: Swithcing between VPN connections does not update the UI


Tridy:
### Distribution

LMDE 6 Faye

### Package version

6.2.9

### Graphics hardware in use

NVIDIA Corporation GA104 [GeForce RTX 3070]

### Frequency

Always

### Bug description

In Newtork Manger when switching between VPN A and VPN B, the list of the VPNs does not get automatically updated with the currenly active VPN, and it takes one more click to update it.

### Steps to reproduce

Have 2 VPN Connections imported,
1. click on VPN A to connect (the successful connection notification comes up)
2. click on VPN B to connect (the successful connection notification comes up)
3. now VPN B is connected but the dot showing the current connection is still pointing to VPN A

additional:

4. clicking on the "VPN Connections" with the slider refreshes the list and shows the proper connection being active instead of disconnecting all VPNs.

### Expected behavior

the dot showing the currently selected VPN should move to the newly connected VPN

![vpns](https://github.com/user-attachments/assets/9cb4f68d-d3c4-48e0-ac39-e1f6e95cb87e)

MatzFan:
Seems another dupe of #7093. On LMDE simply removing the applet was a good workaround for me. The fallback Gnome NM applet GUI can toggle any number of VPN's on and off and switch between them without issue.

gallogabor:
Just to clarify the real problem for OP:

When you click on the second VPN connection, and it connects, the notifies you about the successful connection, that not means it disconnected automatically the first VPN you connected. So after clicking the second VPN, both VPN connections are active at the same time.
The real problem is that the network@cinnamon panel applet cannot show properly multiple active VPN connections to the user. The network@cinnamon panel applet always shows the first connected VPN, if multiple are active. And when you click the on-off switch, you disconnect the oldest connection, then the second oldest, etc.
When you click on the VPN on-off switch (which shows the first VPN's name), that disconnects the first (oldest) connected VPN, and that is why the applet then displaying the second connected VPN's name (because that will be the only active connection).
This network@cinnamon panel applet problem exists from about 8-10 years ago (I found this exact problem in 2016-2017 forum- and Reddit posts) and no solution appeared since.
[...]

PR #13911, opened 2026-07-30
network applet: six correctness fixes, incl. multiple active WireGuard connections

### LINKED ISSUE #12178: network@cinnamon.org applet can only show a single active WireGuard connection
### Distribution

Mint 21.3

### Package version

6.0.4

### Graphics hardware in use

_No response_

### Frequency

Always

### Bug description

Since a recent version the Network applet shows WireGuard connections, which is great!

Unfortunately it can only show one active connection but there could be multiple.

### Steps to reproduce

Create at least two WireGuard networks (with different subnets) and connect to both. The Network applet will only show one active connection.

### Expected behavior

The Network applet should show checkboxes instead of the "radio" items to toggle the Networks independently.

### Additional information

_No response_

IvanTheGeek:
Six independent correctness fixes in `network@cinnamon.org`, one per commit so they can be taken or dropped individually. No new UI.

### The bugs

**1. `updateAccessPoints()` never loops** — `for (let i = 0; i < accessPoints; i++)` compares the index to the *array* rather than its length, so the body never runs. `_accessPoints` is left empty and the `notify::strength` handlers are never reconnected, silently stopping signal-strength updates for that network.

**2. `splice(pos)` truncates the connection list** — `Array.splice(pos)` with no delete count removes everything from `pos` to the end, so removing one connection dropped every connection after it from `_connections`.

**3. The VPN-over-wireless icon never resolves** — `set_applet_icon_symbolic_name()` passes the name through unchanged and sets `St.IconType.SYMBOLIC`; St appends `-symbolic` itself at lookup time. The name built here already carried the suffix, so it asked for `xsi-network-wireless-signal-<level>-secure-symbolic-symbolic`, which does not exist. The panel icon renders as nothing while a VPN is active over wireless. Verified against the shipped icon set:

```
FOUND    xsi-network-wireless-signal-good-secure-symbolic
MISSING  xsi-network-wireless-signal-good-secure-symbolic-symbolic
```

The same commit guards `active_access_point`, which is null on a wireless device that is still activating.

**4. `_correctStateForTunnel()` returns `undefined`** — it falls through without a return when no tun/wireguard connection is active. Only reached from the currently-disabled connectivity check, but wrong either way.

**5. An already-connected VPN draws no dot when its row is first built** — `_updateConnectionItemView()` takes `active` as its third argument, but the first-build call omits it, so `active` is `undefined`. The row corrects itself on the next redraw, which is why it reads as intermittent. This is likely what #12550 is seeing.

**6. Only one WireGuard connection can be active in the UI** (#12178) — `NMDeviceWIREGUARD` kept a single `_activeConnection` slot, but every WireGuard connection is handed the *same* pseudo-device and `_syncActiveConnections()` calls `setActiveConnection()` once per connection, so each activation overwrites the previous one. With two tunnels up the menu shows one as connected and draws the other as if it were down; the section switch tears down only one and then flips straight back on.

This last one mirrors what #12930 did for `NMDeviceVPN` — keep a list, assign it once after the sync loop rather than per connection, and deactivate all of them. The VPN section got that treatment; the WireGuard section was left behind.

Rows become a switch each rather than a pick-one list, since WireGuard connections are independent of one another. That is the behaviour asked for in #12178 ("checkboxes instead of the radio items to toggle the Networks independently") and re-confirmed there in February.

### Testing

Honest about what was and wasn't done:

- Built and `node --check`ed against `master`.
- Smoke-tested by running this branch's `applet.js` on **Cinnamon 6.6.9**, which differs from master only by the four lines of #13847. With two WireGuard tunnels up simultaneously the applet reported both, each with its own working switch:
  ```
  wg active list: [wg-devvps, wg-lab-inet]
  rows: wg-devvps=ON  wg-lab-inet=ON  wg-lab=off  wg-home=off
  ```
- **Not** run on master itself — I don't have a master build here.
- Bugs 1-5 are reasoned from the code and verified by inspection; 1, 2 and 4 are single-expression fixes.

Closes #12178

---

**Disclosure:** I worked on this with Claude (Anthropic's Claude Code). It did the bulk of the code reading, diagnosis and patch authoring, and drafted this write-up; I reviewed the changes and ran them on my own machine before filing. The commits carry a `Co-Authored-By: Claude` trailer to the same effect. Flagging it so you can weight the review accordingly.

### DIFF
--- a/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
@@ -2604,6 +2604,8 @@ CinnamonNetworkApplet.prototype = {
         if (this._activeConnections.some(con => con.get_connection_type() === "wireguard")) {
             return NM.ConnectivityState.FULL;
         }
+
+        return state;
     },
 
     _proxyConnectivityCheckCallback(new_state) {
--- a/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
@@ -2532,10 +2532,9 @@ CinnamonNetworkApplet.prototype = {
                     let iconName = 'xsi-network-vpn';
                     if (mc._section == NMConnectionCategory.WIRELESS) {
                         const dev = mc._primaryDevice;
-                        if (dev) {
-                            const ap = dev.device.active_access_point;
-                            iconName = 'xsi-network-wireless-signal-' + signalToIcon(ap.strength) + '-secure-symbolic';
-                        }
+                        const ap = dev ? dev.device.active_access_point : null;
+                        if (ap)
+                            iconName = 'xsi-network-wireless-signal-' + signalToIcon(ap.strength) + '-secure';
                     }
                     this._setIcon(iconName);
                     this.set_applet_tooltip(_("Connected to the VPN"));
--- a/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
@@ -168,7 +168,7 @@ NMNetworkMenuItem.prototype = {
         accessPoints = sortAccessPoints(accessPoints);
         this.bestAP = accessPoints[0];
         this._accessPoints = [ ];
-        for (let i = 0; i < accessPoints; i++) {
+        for (let i = 0; i < accessPoints.length; i++) {
             let ap = accessPoints[i];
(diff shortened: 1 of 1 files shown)


IvanTheGeek:
The best-practices scanner findings above are addressed — all of the flagged `Lang.bind()` uses were in code this PR adds, and they are now arrow functions (which capture `this` the same way). No new `Lang.bind()` is introduced; the count in this file actually goes *down* versus master.

The scanner comment is stale rather than wrong: the workflow re-run is sitting in `action_required`, since runs from a first-time contributor need maintainer approval. It should clear once someone approves the run.

Issue #13912, opened 2026-07-30
network applet: several small correctness bugs (dead loop, splice truncation, unresolvable icon name, missing return, WPA3 unclassified)


IvanTheGeek:
### Distribution

Linux Mint 22.3 (Zena)

### Package version

cinnamon 6.6.9 — all of these are also present in `master` at 22e0e5c

### Graphics hardware in use

Not relevant

### Frequency

Always

### Bug description

Five small correctness bugs in `files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js`, found while investigating #12178. Grouping them in one issue rather than filing five, since each is a couple of lines. Four have a fix in #13911; the fifth is listed here for the record.

---

**1. `updateAccessPoints()` never executes its loop**

```js
for (let i = 0; i < accessPoints; i++) {
```

The index is compared to the *array*, not `accessPoints.length`, so the condition is `number < object` → `false` on the first iteration. `_accessPoints` is left empty and the `notify::strength` handlers are never reconnected, so signal-strength updates silently stop for that network after its access point set changes.

**2. `splice(pos)` in `_connectionRemoved()` truncates the list**

```js
this._connections.splice(pos);
```

With no delete count, `Array.splice` removes everything from `pos` to the end. Removing one connection drops every connection after it from `_connections`.

**3. The VPN-over-wireless panel icon resolves to nothing**

```js
iconName = 'xsi-network-wireless-signal-' + signalToIcon(ap.strength) + '-secure-symbolic';
```

`set_applet_icon_symbolic_name()` passes the name through unchanged and sets `St.IconType.SYMBOLIC`; **St appends `-symbolic` itself** during lookup. So this asks for `...-secure-symbolic-symbolic`. Checked against the shipped icons:

```
FOUND    xsi-network-wireless-signal-good-secure-symbolic
MISSING  xsi-network-wireless-signal-good-secure-symbolic-symbolic
```

Result: while a VPN is active over wireless, the panel icon renders as nothing at all. (`xapp-symbolic-icons` provides these, so it is not theme-dependent.)

The same block also dereferences `dev.device.active_access_point` without a null check; that is null on a wireless device that is still activating.

**4. `_correctStateForTunnel()` falls through returning `undefined`**

It returns `NM.ConnectivityState.FULL` for `tun` and `wireguard`, but has no `return` on the path where neither matches, so it yields `undefined` instead of the state it was passed. Currently only reached from the commented-out connectivity check, but wrong regardless.

[...]

PR #13913, opened 2026-07-30
network applet: unified tunnel section, per-tunnel switches, band/encryption info, and settings (proposal — take any part or none)

IvanTheGeek:
**This is an opinionated proposal, offered so you can take any part of it or none.** I am not expecting it to be merged as-is, and I won't be offended if the answer is "no thanks" or "just that one bit". #13911 is the minimal, uncontroversial subset and stands alone — if only that lands, this has still been worthwhile.

Cut from `master`, so it independently contains equivalents of the #13911 fixes. If #13911 merges first I'll happily rebase this down to just the additions.

---

## Tunnels: one section, one code path

`NMDeviceWIREGUARD` is deleted and WireGuard connections are routed into the VPN category instead, giving **one** "VPN Connections" section with each row naming its kind (WireGuard / OpenVPN / PPTP / …).

The argument for merging rather than fixing both: the collapse in #12178 existed *because* there were two classes. `NMDeviceVPN` had already been made plural by #12930; the WireGuard copy still had a single slot. Two code paths for the same job drifted apart once and can do so again. Deleting one is a net removal of ~55 lines plus special-casing in five places.

The kind label answers the obvious objection — that folding WireGuard into "VPN Connections" hides what kind of tunnel each is.

## No section-wide switch

A switch per tunnel, and the section title loses its switch. Every other master switch in this applet is one of two things:

| Section | What its title switch does |
|---|---|
| Wireless | `client.wireless_enabled` — radio power / rfkill |
| WWAN | `wwan_enabled` — radio power |
| Wired, one NIC | that device's own switch (the title *is* the device) |
| Wired, two or more NICs | **nothing — the switch is removed** |
| VPN / WireGuard | "disconnect all" — the odd one out |

Both VPN and WireGuard construct `NMWiredSectionTitleMenuItem`, the *single-device* title item, and point it at a collection. There is also no `wireguard_enabled` in NM for it to reflect. The wired section's own answer for "several members" is to drop the switch, so that is what this does.

## Panel icon

- **A tunnel adds a separate indicator** next to the link glyph instead of replacing it with a padlock. The `-secure` wifi variants already mean "this access point is encrypted" in the menu (`NMNetworkMenuItem._getIcon`), so using the same glyph for "a VPN is up" gave one symbol two meanings. `nm-applet` ships two distinct glyphs (`nm-secure-lock-symbolic`, `nm-vpn-active-lock-symbolic`) for exactly this reason.
- **The base glyph comes from the physical connection**, not `_mainConnection`, so a full-tunnel connection that owns the default route no longer hides the link it runs over.
- **The panel padlock now reflects access-point encryption**, matching what the glyph means in the menu. Previously an open hotspot and a WPA2 network looked identical in the panel, which seems worth surfacing on a laptop that roams.
- A **count badge** once more than one tunnel is up, and the tooltip names them.

## Wireless rows

- **Radio band, including 6 GHz**, with the band actually associated to in brackets: `2.4·[5]·6 GHz`. NM ships `utils_wifi_2ghz_freqs()` / `utils_wifi_5ghz_freqs()` and nothing for 6 GHz, so the classification is done locally.
- **Optional channel width and encryption type.** `_getApSecurityType()` only tests PSK and 802.1X, so WPA3-SAE and OWE fall through it as `undefined` (see #13912 item 5); `apSecurityLabel()` classifies from the raw flags and distinguishes personal from enterprise.
- **The band/strength text never ellipsizes** — a long SSID gives way instead, since a clipped name still identifies the network while a clipped signal reading does not. The SSID column is capped so a single very long name cannot drag the whole menu wider.
- **Signal strength is watched on the active access point**, redrawn only when it crosses a `signalToIcon()` bucket. Measured here: `notify::strength` fires ~8/min on the active AP but crossed a bucket 0 times in 45s stationary, so this is cheaper than the existing 10s poll, not dearer. Watching *all* APs would have been ~280 events/min with 35 in range — hence active-only.

## Settings

`settings-schema.json` gains four checkboxes: radio band, signal percentage, channel width, encryption type. **Defaults reproduce the current look** (band and percentage on, the other two off), so nothing changes for anyone who never opens it. The padlock is deliberately not switchable — it answers "is this network open?".

## Testing

Runtime-tested, not just compiled. Installed as a user override on **Cinnamon 6.6.9** (which differs from master only by the four lines of #13847) and exercised against real hardware and real access points:

- Two WireGuard tunnels up simultaneously → both listed, both switches working, count badge showing `2`.
- A NetworkManager OpenVPN profile alongside them → labelled `OpenVPN`, appearing without an applet reload.
- Security labels against live APs: `WPA2`, `WPA2/3` (a transition-mode AP that `nmcli` reports as "WPA2 WPA3"), and `Open`.
- Row alignment verified by comparing allocation boxes across every row with all four toggles on — the widest configuration.
- Settings toggles confirmed to apply live.

**Not** tested: ethernet link speed (no wired NIC on this machine) and the 6 GHz band display (the radio supports 6E but no 6 GHz APs have been in range). Both are correct by construction and unexercised. Nor has any of this run on a `master` build.

Happy to split this into separate PRs per section, drop anything, or adjust naming/defaults — whatever is easiest to review.

---

**Disclosure:** I worked on this with Claude (Anthropic's Claude Code). It did the bulk of the code reading, diagnosis and patch authoring, and drafted this write-up; I reviewed the changes and ran them on my own machine before filing. The commits carry a `Co-Authored-By: Claude` trailer to the same effect. Flagging it so you can weight the review accordingly.

### DIFF
--- a/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
@@ -3,6 +3,7 @@ const GLib = imports.gi.GLib;
 const Gio = imports.gi.Gio;
 const GObject = imports.gi.GObject;
 const Lang = imports.lang;
+const Pango = imports.gi.Pango;
 const NM = imports.gi.NM;
 const Signals = imports.signals;
 const St = imports.gi.St;
@@ -75,6 +76,20 @@ function signalToIcon(value) {
     return 'none';
 }
 
+/* '-secure' when the access point is encrypted, '' when it is open.
+   _getApSecurityType() lives on NMDeviceWireless and caches its answer on the
+   access point, so pass the device wrapper when there is one. */
+function apSecuritySuffix(dev, ap) {
+    if (!ap)
+        return '';
+
+    let type = (dev && dev._getApSecurityType) ? dev._getApSecurityType(ap)
+                                               : ap._secType;
+
+    return (type != NMAccessPointSecurity.UNKNOWN &&
+            type != NMAccessPointSecurity.NONE) ? '-secure' : '';
+}
+
 // shared between NMNetworkMenuItem and NMDeviceWireless
 function sortAccessPoints(accessPoints) {
     return accessPoints.sort(function (one, two) {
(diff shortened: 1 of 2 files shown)


IvanTheGeek:
The best-practices scanner findings above are addressed — all of the flagged `Lang.bind()` uses were in code this PR adds, and they are now arrow functions (which capture `this` the same way). No new `Lang.bind()` is introduced; the count in this file actually goes *down* versus master.

The scanner comment is stale rather than wrong: the workflow re-run is sitting in `action_required`, since runs from a first-time contributor need maintainer approval. It should clear once someone approves the run.

PR #13924, opened 2026-08-08
network@cinnamon.org: don't show the offline icon when online through an unmanaged device

Fantu:
The network applet shows the offline icon, and a "No connection" tooltip, whenever `_mainConnection` is null. That also happens on a machine that is perfectly online, when the only connectivity comes from a device NetworkManager does not manage — typically an interface configured in `/etc/network/interfaces`, with the ifupdown plugin set to `managed=false`, which is the default on Debian. There is no active connection for the applet to describe, so it concludes the machine is offline.

NetworkManager still reports `NM_STATE_CONNECTED_GLOBAL` in that situation, so this uses it as the fallback: when there is no main connection but the global state is `CONNECTED_GLOBAL`, show the generic wired icon and the `Connected to the network` tooltip that `_updateIcon()` already uses for active connections it cannot classify. Everything else is unchanged, and when the machine really is offline the global state is not `CONNECTED_GLOBAL`, so the offline icon is still shown.

Both the icons and both the strings are already used elsewhere in the same function, so there is nothing new to translate.

### Testing

Debian sid VM, cinnamon 6.6.9, NetworkManager 1.58.0, `enp1s0` configured in `/etc/network/interfaces` with `[ifupdown] managed=false` (Debian default):

```
$ nmcli device
DEVICE  TYPE      STATE                   CONNECTION
lo      loopback  connected (externally)  lo
enp1s0  ethernet  unmanaged               --

$ gdbus call --system --dest org.freedesktop.NetworkManager \
    --object-path /org/freedesktop/NetworkManager \
    --method org.freedesktop.DBus.Properties.Get \
    org.freedesktop.NetworkManager State
(<uint32 70>,)     # NM_STATE_CONNECTED_GLOBAL
```

The only active connection is the loopback one, which `_syncActiveConnections()` skips because it has no `_section`, so `_mainConnection` ends up null. Inspecting the live applet before the patch:

```
icon=xsi-network-offline  tooltip="No connection"  _mainConnection=null  client.state=70
```

and after it:

```
icon=xsi-network-wired    tooltip="Connected to the network"  _mainConnection=null  client.state=70
```

Bringing the interface down gives `state=20` and the offline icon is shown again, as expected.

| before | after |
| --- | --- |
| cinnamon-network-unmanaged-before | cinnamon-network-unmanaged-after |

This has been reported on the Debian side since 2013, https://bugs.debian.org/699773 — GNOME Shell used to carry a Debian-specific patch doing exactly this, which is what that bug asked us to port.

### DIFF
--- a/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
@@ -2409,7 +2409,7 @@ CinnamonNetworkApplet.prototype = {
                     // /etc/network/interfaces): there is no active connection
                     // to describe, but we are not offline either.
                     this._setIcon('xsi-network-wired');
-                    this.set_applet_tooltip(_("Connected to the network"));
+                    this.set_applet_tooltip(_("Connected through an unmanaged device"));
                 } else {
                     this._setIcon('xsi-network-offline');
                     this.set_applet_tooltip(_("No connection"));
--- a/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/network@cinnamon.org/applet.js
@@ -2403,8 +2403,17 @@ CinnamonNetworkApplet.prototype = {
             let mc = this._mainConnection;
 
             if (!mc) {
-                this._setIcon('xsi-network-offline');
-                this.set_applet_tooltip(_("No connection"));
+                if (this._client.state == NM.State.CONNECTED_GLOBAL) {
+                    // We are online through a device NetworkManager does not
+                    // manage (for instance one configured in
+                    // /etc/network/interfaces): there is no active connection
+                    // to describe, but we are not offline either.
+                    this._setIcon('xsi-network-wired');
+                    this.set_applet_tooltip(_("Connected to the network"));
+                } else {
+                    this._setIcon('xsi-network-offline');
+                    this.set_applet_tooltip(_("No connection"));
+                }
             } else if (mc.state == NM.ActiveConnectionState.ACTIVATING) {
                 new_delay = FAST_PERIODIC_UPDATE_FREQUENCY_SECONDS;
                 switch (mc._section) {


Fantu:
This has been reported by some users over the years, and I've also seen it a lot with Debian test VMs and the Debian unstable test PC.
It may seem irrelevant, but for example, a few years ago, when my network cable would easily come loose from my test PC, it would have been useful to immediately see, using the applet (and the notification), if the cable had become slightly loose (and therefore disconnected).
@mtwebster what do you think of this?

leigh123linux:
I'm not keen on this change, it would be better if we had a un-managed icon for this state.

Fantu:
A different icon would be better, but I don't know how it could be made to be easily recognizable and intuitive enough (It seems difficult to me with these small and minimal style icons). Any ideas?

Fantu:
Following up on the icon question I asked above: I went through the whole `xapp-symbolic-icons` set looking for something reusable, and I did not find anything good.

The nearest one is `xsi-network-wired-no-route`, the wired icon with a question mark. Even leaving aside that it is already the wired limited-connectivity icon (`applet.js:2470`), so reusing it would make two different states look identical, I do not think it works on its own terms: a question mark on the network icon reads as "something is wrong with your connection", which is the opposite of what we want to say here — the connection is fine, it is just not NetworkManager's. Everything else in the set is either an active state we would be misrepresenting, or an error/disconnected mark that would be plainly wrong.

So a dedicated icon means new artwork in xapp-symbolic-icons, and I am still not sure what the metaphor would be at that size and in that minimal style — which was my worry in the first place. If anyone has an idea for how "connected, but not managed by us" should look, I am happy to be pointed at it.

Meanwhile the current behaviour is not just imprecise, it is wrong: a machine that is fully online shows the offline icon and "No connection". That is what the Debian report has been about since 2013.

So I have updated the PR to at least make the state distinguishable, with a tooltip of its own instead of the generic one:

```js
this._setIcon('xsi-network-wired');
this.set_applet_tooltip(_("Connected through an unmanaged device"));
```

That is one new string to translate.

There is also a variant with **no new strings at all**, reusing two msgids this applet already has — `_("Connected to the network")` and `_("unmanaged")` (`applet.js:513`) — giving "Connected to the network (unmanaged)":

```js
this.set_applet_tooltip(_("Connected to the network") + " (" + _("unmanaged") + ")");
```

I did not pick it because building a sentence out of fragments, with the parentheses hardcoded around it, is awkward for translators and reads worse. But if you would rather avoid the new string, say so and I will switch to it.

If an unmanaged icon does get drawn later, swapping it in here is a one-line change.

Unrelated to this PR, but I ran into it while looking at the icons: `limited_conn` is hardcoded to `false` at `applet.js:2439`, with the real connectivity check commented out on the line above. That makes every limited-connectivity branch in `_updateIcon()` unreachable, so `xsi-network-wired-no-route` and the other `-no-route` icons are never shown. Is that a deliberate temporary state, or a leftover? I can open a separate issue if it is worth tracking.

PR #13934, opened 2026-08-15
Created a folder for organizing files in C

igotlogedoutwaaa-star:
to organize!
in the placeholder.txt, you could also list each file and give it a purpose for other devs to use.


leigh123linux:
Don't post crap!

PR #13948, opened 2026-08-23
screenshot: guard against double _ungrab in area/color/window pickers

rumours86:
**The bug.** A second BUTTON_RELEASE (or BUTTON_PRESS for `SelectWindow`) arriving within the 100ms fade-out animation schedules a second `_ungrab()`, so `Main.popModal()` runs twice for the same actor. The second call takes the `incorrect pop` error path in `main.js`, which first tears down the stage input mode and the X grab and only then throws — the exception aborts `_ungrab()` before `this._group.destroy()`, so the fullscreen selection overlay stays on screen: clicks land on the overlay and the panel appears frozen until the shell is restarted.

**Reproduce.** Take an area screenshot and click once more right after releasing the selection, e.g.:
```
gdbus call -e -d org.gnome.Shell.Screenshot -o /org/gnome/Shell/Screenshot -m org.gnome.Shell.Screenshot.SelectArea &
sleep 1.2
xdotool mousemove 400 400 mousedown 1 mousemove 700 600 mouseup 1 click 1
```
Watch `~/.xsession-errors` for `JS ERROR: Error: incorrect pop` from `screenshot.js` — one affected user had six occurrences in two days of normal use.

**The fix.** Finish the selection on the first release only, ignore extra input during the fade-out, and make `_ungrab()` idempotent in all three pickers (`SelectArea`, `PickColor`, `SelectWindow`).

Verified on Cinnamon 6.6.9 (Mint 22.3): normal area selection still returns the same geometry; the reproducer above no longer throws (3 runs), overlay is destroyed properly.

### DIFF
--- a/js/ui/screenshot.js
+++ b/js/ui/screenshot.js
@@ -405,6 +405,18 @@ class SelectArea {
     }
 
     _onButtonRelease(actor, event) {
+        // The selection is finished on the first release. Ignore any further
+        // press/release pairs arriving during the 100ms fade-out - each of
+        // them used to overwrite the result and schedule another _ungrab(),
+        // and the second popModal() would take the 'incorrect pop' error path
+        // after tearing down the X grab, aborting _ungrab() before
+        // this._group.destroy(): the fullscreen overlay stayed on screen and
+        // the panel appeared frozen.
+        if (!this.active || this._finishing)
+            return Clutter.EVENT_PROPAGATE;
+
+        this._finishing = true;
+
         this._result = this._getGeometry();
         this._group.ease({
             opacity: 0,
@@ -419,6 +431,12 @@ class SelectArea {
     }
 
     _ungrab() {
+        // One-shot guard, NOT this.active: Escape legitimately calls
+        // _ungrab() before any selection started (active is still false).
+        if (this._ungrabbed)
+            return;
+        this._ungrabbed = true;
+
         this.active = false;
 
         if (this.stage_event_id > 0) {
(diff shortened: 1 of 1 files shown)


rumours86:
Amended: the `_ungrab()` guard is now a one-shot `this._ungrabbed` flag instead of checking `this.active` — Escape legitimately calls `_ungrab()` before any selection has started (while `active` is still false), and the previous version of this PR would have blocked that path, leaving the overlay stuck. Verified both paths live: Escape before selecting and a normal selection both tear the overlay down cleanly; the double-release reproducer still passes.

rumours86:
Now verified against git master (built muffin + cinnamon 6.7.5, run nested): on unpatched master a second `_ungrab()` throws `incorrect pop` (reproduced live via the DBus Eval path); with this PR applied the double-ungrab and the Escape-before-selection paths both complete silently and `Main.modalCount` stays balanced.

rumours86:
I retested this against freshly built current master: Cinnamon `61c1b3d1cb1d83e263507fd59aede221bce9abf8` and Muffin `963a573eff827d740ba8ce9592f3b2873aa0b952` (6.7.8). These tests use actual XTEST input and public screenshot D-Bus calls, not direct `_ungrab()` calls.

Two real regressions were reproduced:

1. An additional move/click during the area's fade replaces the first 200×140 rectangle with 300×240 on unpatched master. This PR preserves the first rectangle.
2. Escape followed by an already-queued mouse press/release can reactivate the picker before its idle destruction and leave a captured-event handler behind. The next picker then hits disposed-widget/`incorrect pop` errors. The initial PR still allowed that press and left the next selection hanging.

Commit `b4913b2` fixes the remaining race by rejecting presses after `_ungrabbed` is set, in all three picker classes. I also removed the earlier overly specific animation explanation from the code comments: repeated `ease()` calls replace the existing transition, so the native event trace is a better explanation of the observed failure.

All 11 focused native checks and all 25 ordinary/input-race regression checks now pass: ordinary selections, preserving the first area/window result, and queued Escape followed by a new selection in area/color/window modes. They check real outputs, balanced modal state, actor destruction, subsequent GTK input and no new shell errors. Testing is X11-only; Wayland was not tested.

Pattern Check and CodeFactor pass for `b4913b2`. The [new Build run](https://github.com/linuxmint/cinnamon/actions/runs/35883294100) requires maintainer approval; the earlier failed notification was an approval expiry with zero jobs started.

Could a maintainer approve that Build run and review the updated patch?

mtwebster:
The actual fix is ok i think, but can you update the PR description and squash the commits with an updated message? Thanks

rumours86:
Done — squashed into a single commit with a full message covering both input sequences, and rewrote the PR description to match. The diff is byte-for-byte identical to what you reviewed (only the history changed); the branch is still mergeable against master.

PR #13949, opened 2026-08-23
power applet: debounce device menu rebuilds

rumours86:
**The bug.** `_devicesChanged()` is connected to the csd-power proxy's `g-properties-changed` and rebuilds the whole device menu — destroying and recreating every `DeviceItem` — on each signal. With a flapping AC adapter (loose plug / failing cable) csd-power emits a storm of signals: measured **126 in 3 minutes** against ~18 at idle. The flood of destroyed/recreated actors drives GJS into high-frequency incremental GC; while it runs, GJS blocks JS callbacks from C — including the shell's own paint vfunc (`UiActor.vfunc_paint` in `layout.js`, which paints *everything*). Result: every time the charger blinks, the entire screen freezes or goes black for seconds (`Attempting to run a JS callback during garbage collection ... The offending callback was paint(), a vfunc` at exactly the frame rate in `.xsession-errors`), and blocked destroy callbacks leave zombie notification banners floating over all windows.

**Measurements** (Cinnamon 6.6.9, 3000x2000 eDP): deliberately flapping the charger produced 200 unpainted frames over 9s with the stock applet; removing the applet from the panel dropped it to 29; with this debounce the blackouts are gone entirely (user-confirmed over many plug/unplug cycles).

**The fix.** Coalesce bursts into one rebuild per 500ms: the first signal is handled immediately (icon still reacts instantly), the rest of a burst is folded into a single trailing rebuild.

### DIFF
--- a/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/power@cinnamon.org/applet.js
@@ -579,6 +579,33 @@ class CinnamonPowerApplet extends Applet.TextIconApplet {
     }
 
     _devicesChanged() {
+        // Coalesce bursts of change notifications into one rebuild per 500ms.
+        // csd-power emits g-properties-changed for every UPower device update;
+        // a flapping AC adapter (loose plug) produces a storm of them (measured
+        // 126 signals in 3 minutes), and rebuilding the device menu - destroying
+        // and recreating every DeviceItem - on each signal floods GJS with
+        // garbage.  The resulting high-frequency GC blocks JS callbacks,
+        // including the shell's own paint path, so the whole screen goes black
+        // for seconds at a time (and pending actor destroys leak, leaving stuck
+        // notification banners).  The first signal is handled immediately, the
+        // rest of the burst is folded into one trailing rebuild.
+        if (this._rebuildGate === undefined)
+            this._rebuildGate = 0;
+        if (this._rebuildGate > 0) {
+            this._rebuildDirty = true;
+            return;
+        }
+        this._rebuildDirty = false;
+        this._rebuildGate = GLib.timeout_add(GLib.PRIORITY_DEFAULT, 500, () => {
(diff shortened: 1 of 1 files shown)


rumours86:
Now verified against git master (built muffin + cinnamon 6.7.5, run nested, master power applet loaded on the panel): 10 rapid `_devicesChanged()` calls result in exactly **2** full menu rebuilds — one immediate, one trailing with the final state — instead of 10. The applet loads and functions normally with the patch.

rumours86:
Retested against freshly built current master: Cinnamon `61c1b3d1cb1d83e263507fd59aede221bce9abf8` and Muffin `963a573eff827d740ba8ce9592f3b2873aa0b952` (both 6.7.8), plus current cinnamon-desktop and cinnamon-settings-daemon. This used the real applet and St actors in an isolated X11 session. A controlled service on the private session bus supplied power-device signals and delayed replies; it was not a physical charger test.

I compared unpatched master and this PR in separate native sessions, both at UI scale 1. The same burst of 40 real D-Bus property-change signals caused 40 GetPrimaryDevice/GetDevices pairs, menu rebuilds and panel updates on master, versus 2 of each with the PR. Both actual panel labels reached the final 71%. Request/reply counts were measured at the real Gio proxy and agreed with the service-side counts; the test did not assume a particular rebuild count. Sustained updates also reached the final 82%.

I added `efd6d99` to cancel the pending timeout and ignore notifications/in-flight replies after removal. Three native lifecycle tests failed with the original PR: removing with a pending trailing refresh, removing before GetPrimaryDevice returned, and removing before GetDevices returned. All seven power tests now pass, including repeated removal/re-addition and final-state checks.

The Build notification was an expired approval request, not a compiler/test failure: the old run started zero jobs. Could a maintainer approve [the new Build run](https://github.com/linuxmint/cinnamon/actions/runs/35876126262)? Pattern Check and CodeFactor already pass. The separate dependency failures on current master's own Build run remain outside this applet change.

PR #13950, opened 2026-08-24
keyboard applet: never ellipsize the layout code in the panel

rumours86:
When the panel runs out of space St shrinks the keyboard applet's label and renders the two-letter layout code ("En", "Ru") as "…". The ellipsis is as wide as the code itself, so nothing is saved — the current layout just becomes invisible, which defeats the applet's purpose.

The recently added `min-width: 2.5em` stabilizes the width in the normal case, but under real panel pressure St can still allocate below the natural size and Pango ellipsizes. This change additionally disables ellipsizing on the label (`Pango.EllipsizeMode.NONE`), so the code stays readable however tight the panel gets.

Spotted by a user with a full window list: the applet collapsed to "…" while barely changing width.

### DIFF
--- a/files/usr/share/cinnamon/applets/keyboard@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/keyboard@cinnamon.org/applet.js
@@ -1,5 +1,6 @@
 const Applet = imports.ui.applet;
 const St = imports.gi.St;
+const Pango = imports.gi.Pango;
 const Main = imports.ui.main;
 const PopupMenu = imports.ui.popupMenu;
 const Util = imports.misc.util;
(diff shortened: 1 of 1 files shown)


rumours86:
Now verified against git master (built muffin + cinnamon 6.7.5, run nested at ui scale 2): reproduced on unpatched master — an `applet-label` St.Label with the stock `min-width: 2.5em` squeezed to a 30px allocation still reports `layout.is_ellipsized() == true` (the code renders as "…"); with `ellipsize = Pango.EllipsizeMode.NONE` the same setup reports `is_ellipsized() == false` and the code stays readable.

rumours86:
I revisited the reproduction on freshly built current master (Cinnamon `61c1b3d`, Muffin `963a573`, 6.7.8), using the actual keyboard applet on a horizontal panel, two real XKB sources and UI scales 1 and 2.

I could not reproduce ellipsizing in that setup. The label retains its existing `min-width: 2.5em` allocation (30 px / 60 px), while excessive panel pressure pushes the applet off-screen. Switching Pango END/NONE does not fix that overflow.

The earlier August test forced a standalone label to a chosen allocation. That demonstrates the Pango property, but is not sufficient evidence that the real master panel produces that allocation. The original installed 6.6.9/local keyboard modifications also differed from current master, so that observation should not be treated as a master reproducer.

I do not currently have a supported current-master reproduction and useful visible before/after result for this PR.

Closing this until there is a supported current-master reproducer and a useful visible before/after result.

PR #13951, opened 2026-08-24
keyboard: fix flag icons being half-size and squashed on HiDPI

rumours86:
With "show flags" enabled the keyboard applet's flag icons render **squashed (2x too wide) and at half the intended size** on HiDPI.

Two related causes:

1. `St.TextureCache.load_image_from_file_async()` returns an actor **pre-scaled by the ui scale**. `SubscriptableFlagIcon` puts it into its `St.Bin` without correcting the size, so the bin clamps only the height: with ui scale 2 a 320×240 (4:3) `iso-flag-png` file renders at twice its proper width — e.g. **64×24** for a requested height of 24. Fixed by scaling the returned actor back to the logical size (a no-op at ui scale 1).

2. The applet requests the flag at the logical icon size, so on HiDPI the image is half the intended size. Fixed by multiplying by `global.ui_scale` at both call sites (the menu path used to have this multiplication and lost it at some point).

Verified live on a 3000×2000 laptop panel (ui scale 2): the panel flag went from a squashed 64×24 to a correct 4:3 image at the intended size; layouts menu flags likewise. At ui scale 1 both changes are no-ops.

### DIFF
--- a/files/usr/share/cinnamon/applets/keyboard@cinnamon.org/applet.js
+++ b/files/usr/share/cinnamon/applets/keyboard@cinnamon.org/applet.js
@@ -178,7 +178,7 @@ class CinnamonKeyboardApplet extends Applet.Applet {
             let actor = null;
 
             if (source.type === 'ibus' || this._inputSourcesManager.showFlags) {
-                actor = this._inputSourcesManager.createFlagIcon(source, POPUP_MENU_ICON_STYLE_CLASS, 22);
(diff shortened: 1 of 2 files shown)


rumours86:
Closing: **not reproducible on git master**. I built current master (muffin + cinnamon 6.7.5) and ran it nested with `MUFFIN_DEBUG_DUMMY_MONITOR_SCALES=2`: the texture cache returns the image at the correct aspect (`load_image_from_file_async(-1, 48)` → 64×48), and the panel flag renders correctly sized at 64×48 with the stock applet. Both halves of this fix address a 6.6-only behavior, so per the no-backports policy there is nothing to merge here.

Issue #13866, opened 2026-07-14
Feature Request: Add 'Rename User' option to Users & Groups settings


sporteka2:
## Problem

There is no way to rename a user account from the Cinnamon GUI. Renaming a user currently requires complex CLI steps that are error-prone.

## Suggested Solution

Add a "Rename..." button in **Users and Groups** (`cinnamon-settings-users`) that:

1. Validates the new username
2. Runs `usermod`/`groupmod` behind the scenes
3. Offers to rename and migrate the home directory
4. Checks for and warns about running processes, open sessions, Docker, etc.
5. Updates `/etc/subuid`, `/etc/subgid`, crontabs, systemd user units automatically

## Why it matters

- Switching display managers can break username-based session detection
- Users want to simplify their username
- Current CLI approach (usermod + groupmod + manual fixes) is fragile and intimidating for non-technical users

PR #13953, opened 2026-08-24
feat(cinnamon-settings-users): add user account rename support

farhan11166:
## Summary

Adds a `Rename...` action to Cinnamon Users and Groups.

The rename workflow:

- Validates the new username
- Checks for duplicate usernames
- Warns about running processes and open sessions
- Renames the login with `usermod`
- Migrates the home directory
- Renames a same-named private group
- Updates `/etc/subuid` and `/etc/subgid`
- Migrates the user crontab
- Refreshes the users and groups lists
- Displays errors when system commands fail

## Testing

- Python syntax check passed
- GTK UI XML validation passed
- `git diff --check` passed
- Runtime testing on a real Cinnamon installation is pending(APolgies will try that in a weeks time my systems can take load of an addtional distro tried on vm but was lagging will do asap if possible anyone could check would be glad shall msotly work as i didnt meddle with other features jsut added an addtional one). 

## Known limitations

- Docker configuration is not automatically migrated
- Systemd user-unit configuration is not automatically rewritten
- Arbitrary references to the old username in configuration files are not changed
- Full runtime testing is still required with a disposable user account

This resolves majority of #13866 except the limitations mentioned above which could be implemented iteratively once this is merged

### DIFF
--- a/files/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py
+++ b/files/usr/share/cinnamon/cinnamon-settings-users/cinnamon-settings-users.py
@@ -7,6 +7,8 @@ import gettext
 import shutil
 import re
 import subprocess
+import tempfile
+import stat
 from random import randint
 from setproctitle import setproctitle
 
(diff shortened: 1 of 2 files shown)


farhan11166:
@mtwebster could you go through the changes .Also if my pull req is lacking in certain sectors  please let me know.Thank you.

PR #14001, opened 2026-09-15
appsys: Fix UTF-8 handling when capitalizing executable names

MiQ71:
## Description

`capitalize()` advances the input string by one byte after uppercasing
the first character using `name + 1`.

This produces invalid UTF-8 when the executable basename starts with a
multi-byte UTF-8 character.

For example, when two applications have the same display name and one
of their executable basenames is `ätest`, the app deduplication code
tries to generate a unique name such as:

    UTF8 Cinnamon Test (Ätest)

Since `ä` is encoded as two bytes in UTF-8, `name + 1` points to the
second byte of the character. The resulting unique name therefore
contains invalid UTF-8.

When the menu applet sorts the resulting application names, this can
cause SpiderMonkey to throw:

    malformed UTF-8 character sequence

and prevent `menu@cinnamon.org` from loading.

Use `g_utf8_next_char()` instead, so the pointer advances past the
complete first UTF-8 character.

## Reproducer

Two desktop entries were created with the same `Name` and different
executables:

    Name=UTF8 Cinnamon Test
    Exec=/tmp/ätest

and

    Name=UTF8 Cinnamon Test
    Exec=/tmp/normaltest

Both executable targets were valid and executable.

With Cinnamon 6.6.9 before this change, restarting Cinnamon caused the
menu applet to fail with:

    malformed UTF-8 character sequence at offset 22

The stack trace led through `appSort()`, `getApps()`, `_refreshApps()`
and `CinnamonMenuApplet`.

## Testing

The change was tested by rebuilding the Linux Mint 22.3
`cinnamon 6.6.9+zena` package with this patch and repeating the same
reproducer.

After the change:

- the menu applet loads normally
- no malformed UTF-8 exception occurs
- both applications are displayed correctly as:
  - `UTF8 Cinnamon Test (Ätest)`
  - `UTF8 Cinnamon Test (Normaltest)`

ASCII executable names continue to behave as before.


src/cinnamon-app-system.c
mtwebster: We should check that the name is NULL here and bail early if so. Replace this line with maybe...

```
  const gchar *exec;
  exec = _cinnamon_app_get_executable (app);

  if (exec == NULL)
    {
      DEBUG_RENAMING ("      Skipping rename of app at %s - it has no executable\n",
                      cinnamon_app_get_id (app));

      return;
    }
```

`g_utf8_next_char` has no error-checking, and will crash if it encounters NULL.

mtwebster:
Hi, this is a good catch, but we need to ensure g_utf8_next_char never encounters NULL.

Having no `Exec` entry is valid (though not recommended) for apps that are marked `DBusActivatable`.

https://specifications.freedesktop.org/desktop-entry/latest/recognized-keys.html

MiQ71:
Good point, thanks! I've added the NULL check before calling `g_path_get_basename()`.

I also tested this with a valid `DBusActivatable=true` desktop entry without an `Exec` key. `Gio.DesktopAppInfo` loads it successfully and `get_executable()` returns `NULL`. With a second application using the same display name to trigger the rename path, Cinnamon starts normally and the menu remains functional.

I also re-tested the original UTF-8 reproducer with both changes applied; the menu loads correctly and the duplicate entries are displayed as `UTF8 Cinnamon Test (Ätest)` and `UTF8 Cinnamon Test (Normaltest)`.

